PR #4, opened 2018-08-28
Add initial controller logic

sttts:


sttts:
@mfojtik rebased. ptal.

mfojtik:
@sttts the basic skeleton looks good, can you replace the SSCS in types/vars across the code?

pkg/operator/operator.go
```diff
@@ -0,0 +1,271 @@
...
+	operatorConfigClient operatorconfigclientv1alpha1.KubeapiserverV1alpha1Interface,
+	appsv1Client appsclientv1.AppsV1Interface,
+	corev1Client coreclientv1.CoreV1Interface,
+	rbacv1Client rbacclientv1.RbacV1Interface,
+) *KubeApiserverOperator {
+	c := &KubeApiserverOperator{
+		operatorConfigClient: operatorConfigClient,
+		appsv1Client:         appsv1Client,
+		corev1Client:         corev1Client,
+		rbacv1Client:         rbacv1Client,
```
mfojtik: we probably don't need rbac here (appsv1 and corev1 seems ok for workloads/etc)...

pkg/operator/operator.go
```diff
@@ -0,0 +1,271 @@
...
+			currentActualVerion = &ver
+		}
+	}
+	desiredVersion, err := semver.Parse(operatorConfig.Spec.Version)
+	if err != nil {
+		// TODO report failing status, we may actually attempt to do this in the "normal" error handling
+		return err
+	}
+
+	v311_00_to_unknown := versioning.NewRangeOrDie("3.11.0", "3.12.0")
```
mfojtik: i guess this will be 3.11.0, 4.0.0 ?

pkg/operator/operator.go
```diff
@@ -0,0 +1,271 @@
...
+	switch {
+	case v311_00_to_unknown.BetweenOrEmpty(currentActualVerion) && v311_00_to_unknown.Between(&desiredVersion):
+		var versionAvailability operatorsv1alpha1.VersionAvailablity
+		operatorConfig.Status.TaskSummary = "sync-[3.11.0,3.12.0)"
+		operatorConfig.Status.TargetAvailability = nil
+		// TODO: implement: versionAvailability, errors = sync_v311_00_to_latest(c, operatorConfig, operatorConfig.Status.CurrentAvailability)
+		operatorConfig.Status.CurrentAvailability = &versionAvailability
+
+	default:
+		operatorConfig.Status.TaskSummary = "unrecognized"
```
mfojtik: perhaps provide more details here

pkg/operator/operator.go
```diff
@@ -0,0 +1,271 @@
...
+
+	operatorsv1alpha1 "github.com/openshift/api/operator/v1alpha1"
+	operatorconfigclientv1alpha1 "github.com/openshift/cluster-kube-apiserver-operator/pkg/generated/clientset/versioned/typed/kubeapiserver/v1alpha1"
+	operatorconfiginformerv1alpha1 "github.com/openshift/cluster-kube-apiserver-operator/pkg/generated/informers/externalversions/kubeapiserver/v1alpha1"
+	"github.com/openshift/library-go/pkg/operator/v1alpha1helpers"
+	"github.com/openshift/library-go/pkg/operator/versioning"
+)
+
+const (
+	targetNamespaceName = "openshift-kube-apiserver"
```
mfojtik: @deads2k i believe we are going for a component per namespace?

deads2k: > @deads2k i believe we are going for a component per namespace?

We are.  Let's start here and see if a minor update to the checkpointer can let us run them outside of kube-system.

@sttts what do you think about a poststarthook that creates a set of "special" namespaces so that the static pods can be mirrored in a namespace?

deads2k:
I think this is a good block to build on top of

/lgtm

PR #7, opened 2018-08-30
Initial manifests

deads2k:
Builds on https://github.com/openshift/cluster-kube-apiserver-operator/pull/6

```bash
// assuming oc cluster up
make build-images
oc apply -f manifests/install/operator.yaml
# run it again so the CRD is likely to be accepted
oc apply -f manifests/install/operator.yaml
// wait
```

The operator starts up and produces enough resources to create a pod that crashes on bad/missing config.  I'm going to start the API pulls for config.


manifests/v3.11.0/kube-apiserver/defaultconfig.yaml
```diff
@@ -0,0 +1,6 @@
+# TODO this will change to KubeAPISererConfig
```
ericavonb: s/Serer/Server/

manifests/v3.11.0/kube-apiserver/deployment.yaml
```diff
@@ -0,0 +1,47 @@
...
+spec:
+  strategy:
+    type: RollingUpdate
+  selector:
+    matchLabels:
+      app: openshift-kube-apiserver
+      apiserver: "true"
+  template:
+    metadata:
+      name: openshift-kube-apiserver
```
mfojtik: should we add:

```
checkpointer.alpha.coreos.com/checkpoint: "true"
```

annotation here (it is used to tell the checkpointer to snapshot the pods)

deads2k:
lgtm'd via slack

PR #8, opened 2018-09-10
update to use structured config

deads2k:
merge after https://github.com/openshift/origin/pull/20897

/assign @sttts


deads2k:
@sttts @mfojtik This now includes all my changes required to run in cluster up.

pkg/operator/operator.go
```diff
@@ -30,7 +30,7 @@ import (
 )
 
 const (
-	targetNamespaceName = "openshift-kube-apiserver"
+	targetNamespaceName = "openshift-kube-apiserverwq"
```
derekwaynecarr: why this namespace?

sttts: vi user

PR #9, opened 2018-09-11
Exposed config

deads2k:
This exposes information the openshift apiserver operator needs.  Technically it works, but it errors when merging configuration because a sparse config doesn't technically require an apiversion and kind.  This produces an interesting/awful hot loop.

The bug is preexisting, so this is still strictly an improvement.

/assign @sttts @mfojtik

### DIFF
--- a/manifests/v3.11.0/kube-apiserver/public-info-role.yaml
+++ b/manifests/v3.11.0/kube-apiserver/public-info-role.yaml
@@ -1,0 +1,16 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  namespace: openshift-kube-apiserver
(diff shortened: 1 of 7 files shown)


manifests/v3.11.0/kube-apiserver/public-info-role.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
```
sttts: can you comment about the purpose? Maybe even put a comment into the yaml here.

sttts: It's for the operator status object, right?

Is the role name agreed upon somewhere or is it our buest guess?

PR #16, opened 2018-09-14
Stop hotlooping due to unnecessary status update

sttts:


deads2k:
I think we should still wire in a real rate limiter at some point.

PR #20, opened 2018-09-18
render: move bootkube daemonset back to kube-system

sttts:
This is necessary because bootkube-start requires this.


sttts:
@mfojtik ptal

Issue #18, opened 2018-09-17
building images destroys IDE data


deads2k:
```
[deads@deads-02 cluster-openshift-apiserver-operator]$ make build-images 
hack/build-images.sh
[openshift/origin-cluster-openshift-apiserver-operator] --> FROM openshift/origin-release:golang-1.10 as 0
[openshift/origin-cluster-openshift-apiserver-operator] --> COPY . /go/src/github.com/openshift/cluster-openshift-apiserver-operator
[openshift/origin-cluster-openshift-apiserver-operator] --> RUN cd /go/src/github.com/openshift/cluster-openshift-apiserver-operator && go build ./cmd/cluster-openshift-apiserver-operator
[openshift/origin-cluster-openshift-apiserver-operator] --> FROM centos:7 as 1
[openshift/origin-cluster-openshift-apiserver-operator] --> COPY --from=0 /go/src/github.com/openshift/cluster-openshift-apiserver-operator/cluster-openshift-apiserver-operator /usr/bin/cluster-openshift-apiserver-operator
[openshift/origin-cluster-openshift-apiserver-operator] --> Committing changes to openshift/origin-cluster-openshift-apiserver-operator:d78ed24 ...
[openshift/origin-cluster-openshift-apiserver-operator] --> Tagged as openshift/origin-cluster-openshift-apiserver-operator:latest
[openshift/origin-cluster-openshift-apiserver-operator] --> Done
[openshift/origin-cluster-openshift-apiserver-operator] Removing .idea/
[openshift/origin-cluster-openshift-apiserver-operator] Removing _output/

```

@mfojtik @sttts any idea what would case this?

mfojtik:
the prefix looks like imagebuilder, but I was not able to find anything inside the imagebuilder source code that can cause this

PR #21, opened 2018-09-18
remove git clean when building images

mfojtik:
Fixes https://github.com/openshift/cluster-kube-apiserver-operator/issues/18

### DIFF
--- a/hack/lib/build/images.sh
+++ b/hack/lib/build/images.sh
@@ -83,8 +83,6 @@ function os::build::image::internal::generic() {
 		fi
 	fi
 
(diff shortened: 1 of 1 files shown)


deads2k:
/assign @sttts 

he probably knows what this does

hack/lib/build/images.sh
```diff
@@ -83,8 +83,6 @@ function os::build::image::internal::generic() {
 		fi
 	fi
 
-	# ensure the temporary contents are cleaned up
-	git clean -fdx "${directory}"
```
tnozicka: +1, scripts should not do that

PR #23, opened 2018-09-19
render: generic asset rendering

sttts:
No need to change go code when templates or assets change.


pkg/assets/template.go
```diff
@@ -0,0 +1,39 @@
+package assets
```
mfojtik: looks like a library-go candidate

mfojtik: pkg/assets there

pkg/assets/asset.go
```diff
@@ -1,81 +1,74 @@
 package assets
 
 import (
-	"bytes"
```
mfojtik: this is wrong, we have library-go/assets now

PR #24, opened 2018-09-19
render: switch to kube-core file layout

sttts:
Goal: be a drop-in for kube-core in the installer.


sttts:
/hold

until smoke test with cluster-up is green

manifests/bootkube/manifests/kube-apiserver-daemonset.yaml
```diff
@@ -24,6 +24,7 @@ spec:
         k8s-app: "kube-apiserver"
         openshift.io/control-plane: "true"
         openshift.io/component: "api"
+        checkpointer.alpha.coreos.com/checkpoint: "true"
```
mfojtik: do we want to checkpoint these? i thought we want to checkpoint only operator created resources

/cc @deads2k

mfojtik:
/hold

@sttts can you bump library-go and replace the template?

one question about checkpointer otherwise LGTM

manifests/bootkube/manifests/kube-apiserver-daemonset.yaml
sttts: Makes sense. removing.

sttts:
/hold cancel

Moved to asset lib to library-go and remove the checkpointing annotation.

PR #26, opened 2018-09-19
WIP: render: towards replacing kube-core, with temporary manifests

sttts:
This will unblock our work from other teams' work and from finishing all of our own operators.


deads2k:
still WIP?  rebase eliminates some of teh diff?

PR #31, opened 2018-09-25
render: wire through etcd server URLs

sttts:


cmd/cluster-kube-apiserver-operator/render/render.go
```diff
@@ -190,3 +199,21 @@ func (r *renderOpts) configFromDefaultsPlusOverride(tlsOverride string) ([]byte,
 
 	return yml, nil
 }
+
+func readFileTemplate(fname string, data interface{}) ([]byte, error) {
```
deads2k: didn't you make a library-go function for this?

deads2k:
one question that may deserve later cleanup

/lgtm

cmd/cluster-kube-apiserver-operator/render/render.go
sttts: Don't think that makes sense. The whole render command plumbing belongs into a library-go package though.

PR #34, opened 2018-09-28
update manifest filenames to conform to expected CVO update payload format

juanvallejo:
update manifest filenames to conform to expected CVO update payload format

cc @smarterclayton @soltysh @rajatchopra


smarterclayton:
Separate your component name parts with dashes as described before.  `_cluster-kube-apiserver-operator_`

mfojtik:
@smarterclayton @juanvallejo we also need to update the Dockerfile to have the required label, correct?

PR #42, opened 2018-10-02
Add basic yaml linting test

soltysh:
@deads2k @mfojtik wdyt?


soltysh:
@mfojtik || @deads2k I hope this is good as a first step for validating yamls. If you're ok with it I'll copy this pattern to the rest of operators we have. From there I'll start building more thorough checks for these.

PR #45, opened 2018-10-02
tolerate any taint

deads2k:
/assign @abhinavdahiya @mfojtik


abhinavdahiya:
`/lgtm`

The tolerate all is from https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/#concepts

PR #49, opened 2018-10-03
remove extra namespace manifest

mfojtik:


sttts:
Why? Just because it is created in the render phase already?

deads2k:
> Why? Just because it is created in the render phase already?

we don't use this one.  There is an ns.yaml which we use.

PR #51, opened 2018-10-04
switch etcd ca from root-ca to etcd-client-ca

abhinavdahiya:
etcd server certs in opnsihft/installer are signed by etcd-client-ca
and not the root-ca

### DIFF
--- a/bindata/bootkube/config/bootstrap-config-overrides.yaml
+++ b/bindata/bootkube/config/bootstrap-config-overrides.yaml
@@ -20,7 +20,7 @@ servingInfo:
   clientCA: /etc/kubernetes/secrets/kube-ca.crt # origin 3.11: ca.crt
   keyFile: /etc/kubernetes/secrets/apiserver.key # origin 3.11: master.server.key
 storageConfig:
-  ca: /etc/kubernetes/secrets/root-ca.crt # origin 3.11: ca.crt
+  ca: /etc/kubernetes/secrets/etcd-client-ca.crt # origin 3.11: ca.crt
   certFile: /etc/kubernetes/secrets/etcd-client.crt # origin 3.11: master.etcd-client.crt
   keyFile: /etc/kubernetes/secrets/etcd-client.key # origin 3.11: master.etcd-client.key
   urls: {{range .EtcdServerURLs}}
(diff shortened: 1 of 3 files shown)


sttts:
> etcd server certs in opnsihft/installer are signed by etcd-client-ca

How odd.

sttts:
Does this mean every etcd client key owner can play man-in-the-middle for other etcd clients?

abhinavdahiya:
> Does this mean every etcd client key owner can play man-in-the-middle for other etcd clients?

The ca is unfortunately named like that. But root-ca signs etcd-client-ca and we use etcd-client-ca to sign all peer,server and etcd client certs.

deads2k:
> > Does this mean every etcd client key owner can play man-in-the-middle for other etcd clients?
> 
> The ca is unfortunately named like that. But root-ca signs etcd-client-ca and we use etcd-client-ca to sign all peer,server and etcd client certs.

That's a problem. By signing them that way, @sttts concern is true.  You distribute client certs widely and you don't want them to validate against the same ca that signs the server unless you tightly control all names for the clients. Even then, given how we use it, I wouldn't recommend it.

abhinavdahiya:
And etcd server certs don't bundle the etcd-client-ca, therefore we can't use root-ca for trust yet.

deads2k:
> Also wouldn't a client trying to man-in-the-middle attack other clients; need to have a server certificate to do that successfully?

the etcd server can set `--trusted-ca-file` to the client-ca.  It can separately set the `--peer-trusted-ca-file` right?

abhinavdahiya:
https://github.com/openshift/machine-config-operator/blob/master/templates/_base/master/units/etcd-member.yaml#L66-L95

deads2k:
@ericavonb @enj care to weigh in here?  They're named and used correctly in the operator and kube-apiserver usage, but the signing pattern for etcd seems weird.

mrogers950:
@sttts @deads2k @abhinavdahiya if the server certs had only the TLS Web Server EKU and the client certs had only the TLS Web Client EKU then we don't need separate client and server CAs to prevent misuse of the client certs in that way. But I see the etcd client cert has both client and server EKUs, plus DNS SANs so the misuse seems possible. I seem to recall this being an unfortunate necessity by etcd, do you know is this still the case?

sttts:
> I seem to recall this being an unfortunate necessity by etcd, do you know is this still the case?

@mrogers950 I see `-peer-trusted-ca-file` and `-trusted-ca-file` in etcd. For the former, I believe you are right: that one is used for client and server peer authn. But `-trusted-ca-file` is only used by etcd clients, i.e. one-way.

abhinavdahiya:
/close
https://github.com/openshift/cluster-kube-apiserver-operator/pull/56 made the ca for etcd customizable.

PR #52, opened 2018-10-05
Make sure the namespace is created first

mfojtik:


deads2k:
ordering sucks and this is still luck

/lgtm

PR #53, opened 2018-10-05
add etcd service endpoints

juanvallejo:
cc @deads2k 

Work in progress...

### DIFF
--- a/bindata/bootkube/manifests/etcd-service-endpoints.yaml
+++ b/bindata/bootkube/manifests/etcd-service-endpoints.yaml
@@ -1,14 +1,12 @@
 apiVersion: v1
-kind: Service
+kind: Endpoints
 metadata:
   name: etcd
   namespace: kube-system
-spec:
+subsets:
+- addresses:
+  - ip: {{ .MasterIPAddress }}
   ports:
   - name: etcd
     port: 2379
-    protocol: TCP
-  - name: legacy-etcd
-    port: 4001
-    targetPort: 2379
     protocol: TCP
--- a/cmd/cluster-kube-apiserver-operator/render/render.go
+++ b/cmd/cluster-kube-apiserver-operator/render/render.go
@@ -32,6 +32,7 @@ type manifestOpts struct {
 	secretsHostPath       string
 	lockHostPath          string
 	etcdServerURLs        []string
+	masterIpAddress       string
 }
 
 // renderOpts holds values to drive the render command.
@@ -69,6 +70,9 @@ func NewRenderCommand() *cobra.Command {
 	cmd.Flags().StringVar(&renderOpts.manifest.configFileName, "manifest-config-file-name", "kube-apiserver-config.yaml", "The config file name inside the manifest-config-host-path.")
 	cmd.Flags().StringVar(&renderOpts.manifest.cloudProviderHostPath, "manifest-cloud-provider-host-path", "/etc/kubernetes/cloud", "A host path mounted into the apiserver pods to hold cloud provider configuration.")
 	cmd.Flags().StringArrayVar(&renderOpts.manifest.etcdServerURLs, "manifest-etcd-server-urls", []string{"https://127.0.0.1:2379"}, "The etcd server URL, comma separated.")
+	cmd.Flags().StringVar(&renderOpts.manifest.masterIpAddress, "temporary-never-use-in-production-create-etcd-endpoints", "", "The IP address of the master0 host")
+	cmd.Flags().MarkDeprecated("temporary-never-use-in-production-create-etcd-endpoints", "This flag is meant to provide a temporary workaround for obtaining the IP address of the first master node running an etcd service")
+	cmd.Flags().MarkHidden("temporary-never-use-in-production-create-etcd-endpoints")
 
 	cmd.Flags().StringVar(&renderOpts.assetOutputDir, "asset-output-dir", "", "Output path for rendered manifests.")
 	cmd.Flags().StringVar(&renderOpts.assetInputDir, "asset-input-dir", "", "A path to directory with certificates and secrets.")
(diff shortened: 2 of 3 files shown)


juanvallejo:
@deads2k Thanks, done. Opened PR against installer setting the new param during the render step: https://github.com/openshift/installer/pull/426

bindata/bootkube/manifests/etcd-service-endpoints.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Endpoints
+metadata:
+  name: etcd
+  namespace: kube-system
+subsets:
+- addresses:
+  - ip: {{ .MasterIPAddress }}
```
wking: Does this have to be an IP address?  Can't we use DNS or something less brittle?

wking: For example, [we're already pushing etcd endpoints][1] which the templates can consume.  Is there a reason you can't use them?  (I don't actually understand what this PR is trying to accomplish, so take my feedback with large grains of salt ;).

[1]: https://github.com/openshift/installer/blob/669ce4419d2ce1d79a8aacbe8a29c39dfeff7f5c/pkg/asset/ignition/bootstrap/bootstrap.go#L143-L146

cmd/cluster-kube-apiserver-operator/render/render.go
```diff
@@ -69,6 +70,7 @@ func NewRenderCommand() *cobra.Command {
 	cmd.Flags().StringVar(&renderOpts.manifest.configFileName, "manifest-config-file-name", "kube-apiserver-config.yaml", "The config file name inside the manifest-config-host-path.")
 	cmd.Flags().StringVar(&renderOpts.manifest.cloudProviderHostPath, "manifest-cloud-provider-host-path", "/etc/kubernetes/cloud", "A host path mounted into the apiserver pods to hold cloud provider configuration.")
 	cmd.Flags().StringArrayVar(&renderOpts.manifest.etcdServerURLs, "manifest-etcd-server-urls", []string{"https://127.0.0.1:2379"}, "The etcd server URL, comma separated.")
+	cmd.Flags().StringVar(&renderOpts.manifest.masterIpAddress, "manifest-master-ip-address", "192.168.124.11", "The IP address of the master0 host")
```
mfojtik: If this is used only by etcd service, I will make it specific to etcd, like:

```
--manifest-etcd-service-ip
```

Also I would not hard-code the IP address here, but rather in installer?

mfojtik: also maybe mark it as hidden or deprecated

bindata/bootkube/manifests/etcd-service-endpoints.yaml
mfojtik: @wking it means you use the DNS provided by the cloud provider instead of in-cluster DNS, right?

mfojtik:
@deads2k @juanvallejo @sttts why do we need in-cluster etcd service and our endpoints? Can't we just use the external DNS names that installer create for us? There is an option for that in render command `manifest-etcd-server-urls` that we can just set like this: https://github.com/openshift/installer/pull/420/files#diff-9b4608714b6a03f9757416e09352587aR64

sttts:
/hold

We should not enter the etcd operator business. These IPs must be maintained later-on, e.g. when masters are replaced. Either we get an etcd (endpoint) operator which does that, or we should only point to static DNS names.

deads2k:
> /hold
> 
> We should not enter the etcd operator business. These IPs must be maintained later-on, e.g. when masters are replaced. Either we get an etcd (endpoint) operator which does that, or we should only point to static DNS names.

There is https://jira.coreos.com/browse/CORS-859 .  We need something to unstick the openshift operators.

deads2k:
@abhinavdahiya @aaronlevy To allay @sttts concerns, you both acknowledge that this doesn't mean that the master team is picking up the etcd service and endpoint management.  We're only doing this to unstick the openshift apis until you make a more permanent solution.

aaronlevy:
I'm not sure how this is any different than the list of etcd servers you have to maintain for the apiserver.

bindata/bootkube/manifests/etcd-service-endpoints.yaml
aaronlevy: Correct, it would not be using in-cluster DNS. And also agree with @wking  - we should use the same list of etcd endpoints that we already have available

abhinavdahiya:
The apiservers should be using the list of static dns names for etcd members as @mfojtik pointed out. the installer already has the list of static dns names for each etcd member.

deads2k:
> I'm not sure how this is any different than the list of etcd servers you have to maintain for the apiserver.

Using a service, we do not have to manage the list of individual endpoints.



> The apiservers should be using the list of static dns names for etcd members as @mfojtik pointed out. the installer already has the list of static dns names for each etcd member.

Where would the etcd operator place those DNS names for us to observe?  We've been using a public configmap like [this](https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/bindata/v3.11.0/kube-apiserver/public-info.yaml).  Would you create one for us to use instead of this?

deads2k:
/close

Based on slack conversation with @aaronlevy @abhinavdahiya and @spahl , the list of urls should be rendered as a configmap in the openshift/installer (never in the kube-apiserver-operator).

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  namespace: openshift-etcd
  name: public-info
data:
  # urls is a json array containing the etcd URLs that a clusterNetwork pod should use to access etcd.
  urls: "["https://the-dns-name-here:2379", "https://the-otherdns-or-ip-name-here:2379"]"
```

smarterclayton:
So there are two parts here:

1. whether components use a service or not to access stuff
2. what the data format for that config is

I don't have a strong opinion on 1 in the near term.

The problem with configmap for 2 is that we already have a structured schema object which handles these things.  Endpoints have all of the correct data for this already, including hostname and ip and nodeName.  Regardless of whether you force components to bypass the service in 1 (because we aren't signing the certs? / other reasons) creating a configmap that is reproducing endpoints seems wrong.

If we can make a case that there is a nonzero probability that we will ever use a service for etcd, it's probably more correct for us to represent this particular config in the api object designed for it.

PR #54, opened 2018-10-08
bump to latest library-go

deads2k:
Does the bump, a rename, and collapsing onto helper functions.

/assign @mfojtik @sttts


deads2k:
same change as the rest. tagging.

PR #56, opened 2018-10-09
render: make etcd-serving-ca customizable

sttts:
Replacement of https://github.com/openshift/cluster-kube-apiserver-operator/pull/51 to unblock us and to move the issue into the installer (where it can be resolved one way or the other).


deads2k:
@sttts @mfojtik I'm still seeing issues.  See https://github.com/openshift/installer/issues/440

sttts:
> I'm still seeing issues.

We have not merged the use of this flag into the installer yet.

PR #57, opened 2018-10-09
render: use custom etcd-serving-ca also in bootstrap config overrides

sttts:


mfojtik:
/lgtm

this made the API server working in bootstrap

PR #58, opened 2018-10-09
read etcd endpoint information

deads2k:
react to changes in format in https://github.com/openshift/installer/pull/431

/assign @mfojtik

### DIFF
--- a/pkg/operator/observe_config.go
+++ b/pkg/operator/observe_config.go
@@ -12,6 +12,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/diff"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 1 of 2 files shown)


pkg/operator/observe_config.go
```diff
@@ -64,27 +65,30 @@ func (c ConfigObserver) sync() error {
...
+	etcdURLs := []string{}
 	etcdEndpoints, err := c.kubeClient.CoreV1().Endpoints(etcdNamespaceName).Get("etcd", metav1.GetOptions{})
 	if err != nil && !errors.IsNotFound(err) {
 		return err
 	}
 	if etcdEndpoints != nil {
 		for _, subset := range etcdEndpoints.Subsets {
 			for _, address := range subset.Addresses {
-				etcdURLs = append(etcdURLs, "https://"+address.Hostname+":2379")
+				etcdURLs = append(etcdURLs, "https://"+address.Hostname+"."+etcdEndpoints.Annotations["alpha.installer.openshift.io/dns-suffix"]+":2379")
```
sttts: same comment as for the other repo about the dot

deads2k: > same comment as for the other repo about the dot

I think we fail in general.  If this lasts longer than a few weeks we're going to have some pretty significant issues in general.

PR #61, opened 2018-10-10
remove etcd service

deads2k:
We no longer need to create this.

/assign @sttts @mfojtik


mfojtik:
/lgtm

This is logically not belonging to this operator anyway.

PR #63, opened 2018-10-11
manifests: provide kube-system configs and secrets

mfojtik:
Copy the secrets into `kube-system` namespace for the bootkube daemonset and change the daemonset to use them instead of relying on hostPaths, which does not exists on the master nodes.

/cc @deads2k 
/cc @sttts

### DIFF
--- a/bindata/bootkube/manifests/kube-system-secret-aggregator-client.yaml
+++ b/bindata/bootkube/manifests/kube-system-secret-aggregator-client.yaml
@@ -1,0 +1,14 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: aggregator-client
+  namespace: kube-system
+  labels:
+    tier: "control-plane"
+    k8s-app: "kube-apiserver"
+    openshift.io/control-plane: "true"
+    openshift.io/component: "api"
+type: SecretTypeTLS
+data:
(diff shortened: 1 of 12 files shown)


mfojtik:
do no merge yet, pending manual end-2-end test

mfojtik:
@sttts i have to rewire the post-bootstrap config file and update all paths there

bindata/bootkube/manifests/kube-system-secret-aggregator-client.yaml
```diff
@@ -0,0 +1,14 @@
...
+  name: aggregator-client
+  namespace: kube-system
+  labels:
+    tier: "control-plane"
+    k8s-app: "kube-apiserver"
+    openshift.io/control-plane: "true"
+    openshift.io/component: "api"
+type: SecretTypeTLS
+data:
+  tls.crt: {{ .Assets | load "apiserver-proxy.crt" | base64 }}
```
sttts: didn't we have the `ident 4` before?

sttts: ic, it's a secret... nvmd.

mfojtik:
Smoke tested, this is working as expected, no errors and the kube-apiserver daemonset come up.

PR #64, opened 2018-10-11
Makefile: add origin-release target

sttts:
`DOCKER_ORG=sttts make origin-release`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
 GOFLAGS :=
+DOCKER_ORG ?= $(USER)
 
 all build:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,4 +1,5 @@
 GOFLAGS :=
+DOCKER_ORG ?= $(USER)
```
deads2k: I feel like I should hardcode a conversion from `deads` to `deads2k` :)

deads2k:
/lgtm

@sttts mind setting up the other repos?
@ironcladlou our solution

Makefile
soltysh: Sign me up for it too ;)

PR #67, opened 2018-10-12
Allow to specify image pull policy via operator API

mfojtik:
To demonstrate https://github.com/openshift/api/pull/109

/cc @deads2k 
/hold

Will unhold this when the API is merged and I will bump.


mfojtik:
@smarterclayton I can see a lot of:

```
error: .status.conditions accessor error: Failure is of the type string, expected map[string]interface{}
timeout waiting for router to be available
```

in e2e-aws...

/cc @ironcladlou

deads2k:
little surprised to see cobra move, but ok. We may need to reswizzle something there later.

/lgtm

PR #69, opened 2018-10-12
render: pluralize --config-overrides-file

sttts:


sttts:
Blocked by https://github.com/openshift/installer/pull/457.

mfojtik:
@sttts https://github.com/openshift/installer/pull/457 this won't merge until this merge... i think we have a dependency loop here. To break the loop, we should keep the singular flag in addition to plural, to allow the current installer to pass and to merge this PR. Right know, installer will call the non-existing flag and fail?

sttts:
@mfojtik readded the old flag, as you did for the controller manager

PR #73, opened 2018-10-16
rename conflicting file

deads2k:
when two render steps create files with the same name, only one wins during a copy and we got the "wrong" one.


deads2k:
@mfojtik @sttts why didn't I have a bindata update?

deads2k:
@sjenning thanks for the issue.  This and the other should fix it.

deads2k:
lgtm'd via slack.

wking:
Comparing this with openshift/cluster-kube-controller-manager-operator#50, it looks like they used to have the same name and same content:

```console
$ (cd cluster-kube-controller-manager-operator && git show origin/pr/50 -- bindata/bootkube/manifests/configmap-client-ca.yaml)
commit fde6d6a092ed70c44b26e37d5c29fe36413f904c
Author: David Eads <deads@redhat.com>
Date:   Tue Oct 16 13:57:11 2018 -0400

    rename conflicting file

diff --git a/bindata/bootkube/manifests/configmap-client-ca.yaml b/bindata/bootkube/manifests/configmap-client-ca.yaml
deleted file mode 100644
index e0044cc..0000000
--- a/bindata/bootkube/manifests/configmap-client-ca.yaml
+++ /dev/null
@@ -1,9 +0,0 @@
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: client-ca
-  namespace: {{ .Namespace }}
-data:
-  ca-bundle.crt: |
-    {{ .Assets | load "kube-ca.crt" | indent 4 }}
-
$ (cd cluster-kube-apiserver-operator && git show origin/pr/73 -- bindata/bootkube/manifests/configmap-client-ca.yaml)
commit 2f0de2bdab59ddaa22051498df31dab62949b8e6
Author: David Eads <deads@redhat.com>
Date:   Tue Oct 16 13:56:55 2018 -0400

    rename conflicting file

diff --git a/bindata/bootkube/manifests/configmap-client-ca.yaml b/bindata/bootkube/manifests/configmap-client-ca.yaml
deleted file mode 100644
index e0044cc..0000000
--- a/bindata/bootkube/manifests/configmap-client-ca.yaml
+++ /dev/null
@@ -1,9 +0,0 @@
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: client-ca
-  namespace: {{ .Namespace }}
-data:
-  ca-bundle.crt: |
-    {{ .Assets | load "kube-ca.crt" | indent 4 }}
-
```

Does the filename still matter?  Or are these somehow getting set up without [per-project prefixes][1]?  I'm not particularly familiar with how the filename affects the pushed asset, so feel free to ignore if this is a noob Kubernetes issue ;).

[1]: https://github.com/openshift/cluster-version-operator/blame/fe673cb712fa5e27001488fc088ac91bb553353d/docs/dev/operators.md#L50-L55

sjenning:
The filename does matter here.  The CVO just copies all these manifests from all the operators into a directory (single namespaces for all rendered files).  Operator can overwrite one another if they have the same file name, as was the case here vs the `cluster-kube-controller-manager-operator`.

wking:
> The CVO just copies all these manifests from all the operators into a directory (single namespaces for all rendered files).

Without [per-project prefixes][1]?  Maybe the CVO should borrow that logic from the release-image generator?

> Operator can overwrite one another if they have the same file name, as was the case here vs the `cluster-kube-controller-manager-operator`.

But was the file content actually different between these two operators?  The templates were the same, and the only variable is `.Namespace`.

[1]: https://github.com/openshift/cluster-version-operator/blame/fe673cb712fa5e27001488fc088ac91bb553353d/docs/dev/operators.md#L50-L55

sttts:
This has nothing to do with CVO, but it due to the simplicity of the bootkube script: https://github.com/openshift/installer/blob/master/pkg/asset/ignition/bootstrap/content/bootkube.go#L112

sjenning:
yes, my bad.  these are the bootkube resources not the release resources :man_facepalming:

PR #82, opened 2018-10-22
add simple installer

deads2k:
Adds a simple installer command that takes content prefixes and a deployment ID, then fluffs them all up on disk and smashes the pod content into place.

/assign @sttts

### DIFF
--- a/pkg/cmd/installer/cmd.go
+++ b/pkg/cmd/installer/cmd.go
@@ -1,0 +1,193 @@
+package installer
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+
+	"github.com/golang/glog"
+	"github.com/spf13/cobra"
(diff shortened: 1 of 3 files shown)


deads2k:
> 0700 ? At least for the secrets.

TODO, we'll start here.

pkg/cmd/installer/cmd.go
```diff
@@ -0,0 +1,193 @@
...
+
+	cmd := &cobra.Command{
+		Use:   "installer",
+		Short: "Install static pod and related resources",
+		Run: func(cmd *cobra.Command, args []string) {
+			if err := o.Validate(); err != nil {
+				glog.Fatal(err)
+			}
+
+			if err := o.Run(); err != nil {
```
soltysh: I think you're missing the `Complete` invocation.

PR #84, opened 2018-10-22
add new control loops

deads2k:
```
	// meet our control loops.  Each has a specific job and they operator independently so that each is very simply to write and test
	// 1. prereqs - this creates configmaps and secrets to be copied for static pods.  It writes to a single target for each resource
	//    (no spin numbers on the individual secrets).  This (and other content) are targets for the deploymentContent loop.
	// 2. deploymentContent - this watches multiple resources for "latest" input that has changed from the most current deploymentID.
	//    When a change is found, it creates a new deployment by copying resources and adding the deploymentID suffix to the names
	//    to make a theoretically immutable set of deployment data.  It then bumps the latestDeploymentID and starts watching again.
	// 3. installer - this watches the latestDeploymentID and the list of kubeletStatus (alpha-sorted list).  When a latestDeploymentID
	//    appears that doesn't match the current latest for first kubeletStatus and the first kubeletStatus isn't already transitioning,
	//    it kicks off an installer pod.  If the next kubeletStatus doesn't match the immediate prior one, it kicks off that transition.

```

/assign @sttts 

I plan to add one more control loop, but @sanchezl needed an example.


deads2k:
@sttts I've got one more control loop to write, but lets start here.  You can see where I'm headed and this should give you enough inputs to create that installer pod.

pkg/apis/kubeapiserver/v1alpha1/types.go
```diff
@@ -43,6 +43,25 @@ type KubeApiserverOperatorConfigSpec struct {
...
 	operatorsv1alpha1api.OperatorStatus `json:",inline" protobuf:"bytes,1,opt,name=operatorStatus"`
+
+	// latestDeploymentID is the deploymentID of the most recent deployment
+	LatestDeploymentID int32 `json:"latestDeploymentID"`
+
+	TargetKubeletStates []KubeletState `json:"kubeletStates"`
+}
+
+type KubeletState struct {
+	KubeletName string `json:"kubeletName"`
```
sttts: nodeName ?

pkg/apis/kubeapiserver/v1alpha1/types.go
```diff
@@ -43,6 +43,25 @@ type KubeApiserverOperatorConfigSpec struct {
...
+
+	// currentDeploymentID is the ID of the most recently successful deployment
+	CurrentDeploymentID int32 `json:"currentDeploymentID"`
+	// targetDeploymentID is the ID of the deployment we're trying to apply
+	TargetDeploymentID int32 `json:"targetDeploymentID"`
+	// lastFailedDeploymentID is the ID of the deployment we tried and failed to deploy.
+	LastFailedDeploymentID int32 `json:"lastFailedDeploymentID"`
+
+	// errors is a list of the errors we've had
+	Errors []string `json:"errors"`
```
sttts: "we've had" means? This must be more specific.

pkg/operator/deployment_content_v311_00.go
```diff
@@ -0,0 +1,143 @@
...
+	operatorConfigOriginal := operatorConfig.DeepCopy()
+	latestDeploymentID := operatorConfig.Status.LatestDeploymentID
+	isLatestDeploymentCurrent, reason := isLatestDeploymentCurrent(c.kubeClient, latestDeploymentID)
+
+	// check to make sure that the latestDeploymentID has the exact content we expect.  No mutation here, so we start creating the next Deployment only when it is required
+	if isLatestDeploymentCurrent {
+		return false, nil
+	}
+
+	glog.Infof("new deployment triggered by %q", reason)
```
sttts: also print the ID, makes debugging easier.

pkg/operator/deployment_content_v311_00.go
```diff
@@ -0,0 +1,143 @@
...
+	"deployment-kube-apiserver-config",
+	"aggregator-client-ca",
+	"client-ca",
+	"etcd-serving-ca",
+	"kubelet-serving-ca",
+	"sa-token-signing-certs",
+}
+
+// directCopySecrets is a list of secrets that are directly copied for the current values.  A different actor/controller
+//// modifies these.
```
sttts: nit: slashes

sttts:
"prereqs" is not a good name. How about:
- targetConfigReconciler
- deploymentController
- installerController

pkg/operator/prereqs_v311_00.go
```diff
@@ -0,0 +1,145 @@
...
+		})
+		if !reflect.DeepEqual(operatorConfigOriginal, operatorConfig) {
+			_, updateError := c.operatorConfigClient.KubeApiserverOperatorConfigs().UpdateStatus(operatorConfig)
+			return true, updateError
+		}
+		return true, nil
+	}
+
+	v1alpha1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, operatorv1alpha1.OperatorCondition{
+		Type:   "PrereqCreationFailing",
```
sttts: "prereq" is confusing as a name. See other comment. "TargetConfigCreationFailed" would be much clearer.

deads2k:
All comments accepted.  I'm out this morning though.  You mind fixing it up?

deads2k:
comments fixed.  Merging to move our branch forward.

PR #85, opened 2018-10-23
provide config observation operator status 

sanchezl:
Updates the Failed condition after syncing the observed config.
All needed configs are required for this operator, so previously observed values are retained when a configuration key is missing or unreadable.
Failed condition is triggered (status=true) when a problem is detected, and cleared(status=false) if reading the config is successful.

### DIFF
--- a/pkg/operator/observe_config.go
+++ b/pkg/operator/observe_config.go
@@ -4,17 +4,20 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"reflect"
+	"strings"
 	"time"
 
 	"github.com/ghodss/yaml"
 	"github.com/golang/glog"
+	"github.com/imdario/mergo"
 
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/diff"
+	"k8s.io/apimachinery/pkg/util/rand"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	kubeinformers "k8s.io/client-go/informers"
@@ -24,11 +27,19 @@ import (
 	"k8s.io/client-go/util/flowcontrol"
 	"k8s.io/client-go/util/workqueue"
 
+	"github.com/openshift/api/operator/v1alpha1"
 	operatorconfigclientv1alpha1 "github.com/openshift/cluster-kube-apiserver-operator/pkg/generated/clientset/versioned/typed/kubeapiserver/v1alpha1"
 	kubeapiserveroperatorinformers "github.com/openshift/cluster-kube-apiserver-operator/pkg/generated/informers/externalversions"
+	"github.com/openshift/library-go/pkg/operator/v1alpha1helpers"
 )
 
-type observeConfigFunc func(kubernetes.Interface, *rest.Config, map[string]interface{}) (map[string]interface{}, error)
+const configObservationErrorConditionReason = "ConfigObservationError"
+
+// observeConfigFunc observes configuration and returns the observedConfig. This function should not return an
+// observedConfig that would cause the service being managed by the operator to crash. For example, if a required
+// configuration key cannot be observed, consider reusing the configuration key's previous value. Errors that occur
+// while attempting to generate the observedConfig should be returned in the errs slice.
+type observeConfigFunc func(kubeClient kubernetes.Interface, clientConfig *rest.Config, currentConfig map[string]interface{}) (observedConfig map[string]interface{}, errs []error)
 
 type ConfigObserver struct {
 	operatorConfigClient operatorconfigclientv1alpha1.KubeapiserverV1alpha1Interface
@@ -39,8 +50,8 @@ type ConfigObserver struct {
 	queue workqueue.RateLimitingInterface
 
 	rateLimiter flowcontrol.RateLimiter
-	// observers are used to build the observed configuration. They are called in
-	// order and are expected to mutate the config based on cluster state.
+	// observers are called in an undefined order and their results are merged to
+	// determine the observed configuration.
 	observers []observeConfigFunc
 }
 
@@ -58,8 +69,8 @@ func NewConfigObserver(
 
 		rateLimiter: flowcontrol.NewTokenBucketRateLimiter(0.05 /*3 per minute*/, 4),
 		observers: []observeConfigFunc{
-			observeEtcdEndpoints,
-			observeClusterConfig,
+			observeStorageURLs,
+			observeRestrictedCIDRs,
 		},
 	}
 
@@ -74,16 +85,6 @@ func NewConfigObserver(
 // must be information that is logically "owned" by another component.
 func (c ConfigObserver) sync() error {
 
-	observedConfig := map[string]interface{}{}
-	var err error
-
-	for _, observer := range c.observers {
-		observedConfig, err = observer(c.kubeClient, &rest.Config{}, observedConfig)
-		if err != nil {
-			return err
-		}
-	}
-
 	operatorConfig, err := c.operatorConfigClient.KubeAPIServerOperatorConfigs().Get("instance", metav1.GetOptions{})
 	if err != nil {
 		return err
@@ -92,63 +93,134 @@ func (c ConfigObserver) sync() error {
 	// don't worry about errors
 	currentConfig := map[string]interface{}{}
 	json.NewDecoder(bytes.NewBuffer(operatorConfig.Spec.ObservedConfig.Raw)).Decode(&currentConfig)
-	if reflect.DeepEqual(currentConfig, observedConfig) {
-		return nil
+
+	var errs []error
+	var observedConfigs []map[string]interface{}
+	for _, i := range rand.Perm(len(c.observers)) {
+		var currErrs []error
+		observedConfig, currErrs := c.observers[i](c.kubeClient, &rest.Config{}, currentConfig)
+		observedConfigs = append(observedConfigs, observedConfig)
+		errs = append(errs, currErrs...)
 	}
 
-	glog.Infof("writing updated observedConfig: %v", diff.ObjectDiff(operatorConfig.Spec.ObservedConfig.Object, observedConfig))
-	operatorConfig.Spec.ObservedConfig = runtime.RawExtension{Object: &unstructured.Unstructured{Object: observedConfig}}
-	if _, err := c.operatorConfigClient.KubeAPIServerOperatorConfigs().Update(operatorConfig); err != nil {
-		return err
+	mergedObservedConfig := map[string]interface{}{}
+	for _, observedConfig := range observedConfigs {
+		mergo.Merge(&mergedObservedConfig, observedConfig)
+	}
+
+	if !equality.Semantic.DeepEqual(currentConfig, mergedObservedConfig) {
+		glog.Infof("writing updated observedConfig: %v", diff.ObjectDiff(operatorConfig.Spec.ObservedConfig.Object, mergedObservedConfig))
+		operatorConfig.Spec.ObservedConfig = runtime.RawExtension{Object: &unstructured.Unstructured{Object: mergedObservedConfig}}
+		operatorConfig, err = c.operatorConfigClient.KubeAPIServerOperatorConfigs().Update(operatorConfig)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("kubeapiserveroperatorconfigs/instance: error writing updated observed config: %v", err))
+		}
+	}
+
+	status := operatorConfig.Status.DeepCopy()
+	if len(errs) > 0 {
+		var messages []string
+		for _, currentError := range errs {
+			messages = append(messages, currentError.Error())
+		}
+		v1alpha1helpers.SetOperatorCondition(&status.Conditions, v1alpha1.OperatorCondition{
+			Type:    v1alpha1.OperatorStatusTypeFailing,
+			Status:  v1alpha1.ConditionTrue,
+			Reason:  configObservationErrorConditionReason,
+			Message: strings.Join(messages, "\n"),
+		})
+	} else {
+		condition := v1alpha1helpers.FindOperatorCondition(status.Conditions, v1alpha1.OperatorStatusTypeFailing)
+		if condition != nil && condition.Status != v1alpha1.ConditionFalse && condition.Reason == configObservationErrorConditionReason {
+			condition.Status = v1alpha1.ConditionFalse
+			condition.Reason = ""
+			condition.Message = ""
+		}
+	}
+
+	if !equality.Semantic.DeepEqual(operatorConfig.Status, status) {
+		operatorConfig.Status = *status
+		_, err = c.operatorConfigClient.KubeAPIServerOperatorConfigs().UpdateStatus(operatorConfig)
+		if err != nil {
+			return err
+		}
 	}
 
 	return nil
 }
 
-// observeEtcdEndpoints reads the etcd endpoints from the endpoints object and then manually pull out the hostnames to
-// get the etcd urls for our config. Setting them observed config causes the normal reconciliation loop to run
-func observeEtcdEndpoints(kubeClient kubernetes.Interface, clientConfig *rest.Config, observedConfig map[string]interface{}) (map[string]interface{}, error) {
-	etcdURLs := []string{}
+// observeStorageURLs observes the storage config URLs. If there is a problem observing the current storage config URLs,
+// then the previously observed storage config URLs will be re-used.
+func observeStorageURLs(kubeClient kubernetes.Interface, clientConfig *rest.Config, currentConfig map[string]interface{}) (observedConfig map[string]interface{}, errs []error) {
+	observedConfig = map[string]interface{}{}
+	storageConfigURLsPath := []string{"storageConfig", "urls"}
+	if currentEtcdURLs, found, _ := unstructured.NestedStringSlice(currentConfig, storageConfigURLsPath...); found {
+		unstructured.SetNestedStringSlice(observedConfig, currentEtcdURLs, storageConfigURLsPath...)
+	}
+
+	var etcdURLs []string
 	etcdEndpoints, err := kubeClient.CoreV1().Endpoints(etcdNamespaceName).Get("etcd", metav1.GetOptions{})
 	if errors.IsNotFound(err) {
-		return observedConfig, nil
+		errs = append(errs, fmt.Errorf("endpoints/etcd.kube-system: not found"))
+		return
 	}
 	if err != nil {
-		return observedConfig, err
+		errs = append(errs, err)
+		return
 	}
-	for _, subset := range etcdEndpoints.Subsets {
-		for _, address := range subset.Addresses {
-			etcdURLs = append(etcdURLs, "https://"+address.Hostname+"."+etcdEndpoints.Annotations["alpha.installer.openshift.io/dns-suffix"]+":2379")
+	dnsSuffix := etcdEndpoints.Annotations["alpha.installer.openshift.io/dns-suffix"]
+	if len(dnsSuffix) == 0 {
+		errs = append(errs, fmt.Errorf("endpoints/etcd.kube-system: alpha.installer.openshift.io/dns-suffix annotation not found"))
+		return
+	}
+	for subsetIndex, subset := range etcdEndpoints.Subsets {
+		for addressIndex, address := range subset.Addresses {
+			if address.Hostname == "" {
+				errs = append(errs, fmt.Errorf("endpoints/etcd.kube-system: subsets[%v]addresses[%v].hostname not found", subsetIndex, addressIndex))
+				continue
+			}
+			etcdURLs = append(etcdURLs, "https://"+address.Hostname+"."+dnsSuffix+":2379")
 		}
 	}
-	if len(etcdURLs) > 0 {
-		unstructured.SetNestedStringSlice(observedConfig, etcdURLs, "storageConfig", "urls")
-	} else {
-		glog.Warningf("no etcd endpoints found")
+
+	if len(etcdURLs) == 0 {
+		errs = append(errs, fmt.Errorf("endpoints/etcd.kube-system: no etcd endpoint addresses found"))
 	}
-	return observedConfig, nil
+	if len(errs) > 0 {
+		return
+	}
+	unstructured.SetNestedStringSlice(observedConfig, etcdURLs, storageConfigURLsPath...)
+	return
 }
 
-// observeClusterConfig observes CIDRs from cluster-config-v1 in order to populate list of restrictedCIDRs
-func observeClusterConfig(kubeClient kubernetes.Interface, clientConfig *rest.Config, observedConfig map[string]interface{}) (map[string]interface{}, error) {
+// observeRestrictedCIDRs observes list of restrictedCIDRs.
+func observeRestrictedCIDRs(kubeClient kubernetes.Interface, clientConfig *rest.Config, currentConfig map[string]interface{}) (observedConfig map[string]interface{}, errs []error) {
+	observedConfig = map[string]interface{}{}
+	restrictedCIDRsPath := []string{"admissionPluginConfig", "openshift.io/RestrictedEndpointsAdmission", "configuration", "restrictedCIDRs"}
+	//if currentRestrictedCIDRs, found, _ := unstructured.NestedStringSlice(currentConfig, restrictedCIDRsPath...); found {
+	//	unstructured.SetNestedStringSlice(observedConfig, currentRestrictedCIDRs, restrictedCIDRsPath...)
+	//}
+
 	clusterConfig, err := kubeClient.CoreV1().ConfigMaps("kube-system").Get("cluster-config-v1", metav1.GetOptions{})
 	if errors.IsNotFound(err) {
-		glog.Warningf("cluster-config-v1 not found in the kube-system namespace")
-		return observedConfig, nil
+		errs = append(errs, fmt.Errorf("configmap/cluster-config-v1.kube-system: not found"))
+		return
 	}
 	if err != nil {
-		return observedConfig, err
+		errs = append(errs, err)
+		return
 	}
 
 	installConfigYaml, ok := clusterConfig.Data["install-config"]
 	if !ok {
-		return observedConfig, nil
+		errs = append(errs, fmt.Errorf("configmap/cluster-config-v1.kube-system: install-config not found"))
+		return
 	}
 	installConfig := map[string]interface{}{}
 	err = yaml.Unmarshal([]byte(installConfigYaml), &installConfig)
 	if err != nil {
-		glog.Warningf("Unable to parse install-config: %s", err)
-		return observedConfig, nil
+		errs = append(errs, err)
+		return
 	}
 
 	// extract needed values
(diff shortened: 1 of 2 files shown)


pkg/operator/observe_config.go
```diff
@@ -92,12 +84,42 @@ func (c ConfigObserver) sync() error {
...
-		return nil
+		glog.Infof("writing updated observedConfig: %v", diff.ObjectDiff(operatorConfig.Spec.ObservedConfig.Object, observedConfig))
+		operatorConfig.Spec.ObservedConfig = runtime.RawExtension{Object: &unstructured.Unstructured{Object: observedConfig}}
+	}
+	if len(errors) > 0 {
+		var message bytes.Buffer
+		var i int
+		for i, err = range errors {
+			if i > 0 {
+				message.WriteString("\n")
```
deads2k: readable above all else until you prove this is bad in a profile..  make a slice of strings from the errors, then `strings.Join`

pkg/operator/observe_config.go
```diff
@@ -92,12 +84,42 @@ func (c ConfigObserver) sync() error {
...
+			}
+			message.WriteString(err.Error())
+		}
+		v1alpha1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, v1alpha1.OperatorCondition{
+			Type:    "Failing",
+			Status:  v1alpha1.ConditionTrue,
+			Reason:  "SynchronizationError",
```
deads2k: Start a file with our special reasons.  Also this name is bad.  `ConfigObservationError`

pkg/operator/observe_config.go
```diff
@@ -92,12 +84,42 @@ func (c ConfigObserver) sync() error {
...
+			Message: message.String(),
+		})
+	} else {
+		v1alpha1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, v1alpha1.OperatorCondition{
```
deads2k: This code is wrong.  You should only be changing this status if the previous value was your own synchronization error.

pkg/operator/observe_config.go
```diff
@@ -92,12 +84,42 @@ func (c ConfigObserver) sync() error {
...
+			Type:    "Failing",
+			Status:  v1alpha1.ConditionFalse,
+			Reason:  "SynchronizationSuccess",
+			Message: "TargetConfigurationObservedSuccessfully",
+		})
 	}
-
-	glog.Infof("writing updated observedConfig: %v", diff.ObjectDiff(operatorConfig.Spec.ObservedConfig.Object, observedConfig))
-	operatorConfig.Spec.ObservedConfig = runtime.RawExtension{Object: &unstructured.Unstructured{Object: observedConfig}}
 	if _, err := c.operatorConfigClient.KubeApiserverOperatorConfigs().Update(operatorConfig); err != nil {
```
deads2k: do a check to see if the status is equal before updating.  `.DeepCopy` at the beginning.

pkg/operator/observe_config.go
```diff
@@ -92,12 +84,42 @@ func (c ConfigObserver) sync() error {
 	// don't worry about errors
 	currentConfig := map[string]interface{}{}
 	json.NewDecoder(bytes.NewBuffer(operatorConfig.Spec.ObservedConfig.Raw)).Decode(&currentConfig)
+	observedConfig := map[string]interface{}{}
+	json.NewDecoder(bytes.NewBuffer(operatorConfig.Spec.ObservedConfig.Raw)).Decode(&observedConfig)
+
+	var errors []error
+	for _, observer := range c.observers {
+		var e []error
```
deads2k: `currErrors`

pkg/operator/observe_config.go
```diff
@@ -107,14 +129,16 @@ func (c ConfigObserver) sync() error {
...
 // get the etcd urls for our config. Setting them observed config causes the normal reconciliation loop to run
-func observeEtcdEndpoints(kubeClient kubernetes.Interface, clientConfig *rest.Config, observedConfig map[string]interface{}) (map[string]interface{}, error) {
+// REQUIRED EndPoints etcd.kube-system
+// REQUIRED etcd.kube-system/subsets[], len > 0
+func observeEtcdEndpoints(kubeClient kubernetes.Interface, clientConfig *rest.Config, observedConfig map[string]interface{}) (map[string]interface{}, []error) {
 	etcdURLs := []string{}
 	etcdEndpoints, err := kubeClient.CoreV1().Endpoints(etcdNamespaceName).Get("etcd", metav1.GetOptions{})
 	if errors.IsNotFound(err) {
-		return observedConfig, nil
+		return observedConfig, []error{fmt.Errorf("Endpoints/etcd.kube-system not found")}
```
deads2k: This doesn't appear to match your proposal.  Aren't you supposed to be returning the map with the old value set?

pkg/operator/observe_config.go
```diff
@@ -24,11 +24,13 @@ import (
...
 	"k8s.io/client-go/util/workqueue"
 
+	"github.com/openshift/api/operator/v1alpha1"
 	operatorconfigclientv1alpha1 "github.com/openshift/cluster-kube-apiserver-operator/pkg/generated/clientset/versioned/typed/kubeapiserver/v1alpha1"
 	kubeapiserveroperatorinformers "github.com/openshift/cluster-kube-apiserver-operator/pkg/generated/informers/externalversions"
+	"github.com/openshift/library-go/pkg/operator/v1alpha1helpers"
 )
 
-type observeConfigFunc func(kubernetes.Interface, *rest.Config, map[string]interface{}) (map[string]interface{}, error)
+type observeConfigFunc func(kubernetes.Interface, *rest.Config, map[string]interface{}) (map[string]interface{}, []error)
```
deads2k: name the args.  I think that map is `previousObservedConfig`

deads2k: > name the args. I think that map is `previousObservedConfig`

Oh, it isn't.  You need to add that one to build this method correctly.

pkg/operator/observe_config.go
```diff
@@ -124,31 +148,33 @@ func observeEtcdEndpoints(kubeClient kubernetes.Interface, clientConfig *rest.Co
...
 	} else {
-		glog.Warningf("no etcd endpoints found")
+		return observedConfig, []error{fmt.Errorf("no etcd endpoints found")}
 	}
 	return observedConfig, nil
 }
 
 // observeClusterConfig observes CIDRs from cluster-config-v1 in order to populate list of restrictedCIDRs
-func observeClusterConfig(kubeClient kubernetes.Interface, clientConfig *rest.Config, observedConfig map[string]interface{}) (map[string]interface{}, error) {
+// REQUIRED ConfigMap cluster-config-v1.kube-system
```
deads2k: Comments like this are prone to being out of date.

deads2k:
@juanvallejo  ptal

@sanchezl We have at least three other teams waiting to add to our config observation loop.  Any added before this pattern is applied to all repos you have to migrate (stick).  Any added after, they have to conform (carrot).  :)

pkg/operator/observe_config.go
sanchezl: A file in source? Or a shared document?

pkg/operator/observe_config.go
sanchezl: Not for the errors.

pkg/operator/observe_config.go
sanchezl: The observed config being composed was initialized with the previously observed config, so no need to pass the previously observed config separately.

pkg/operator/observe_config.go
```diff
@@ -74,30 +77,46 @@ func NewConfigObserver(
...
+	}
+	if len(errs) > 0 {
+		var messages []string
+		for _, currentError := range errs {
+			messages = append(messages, currentError.Error())
+		}
+		v1alpha1helpers.SetOperatorCondition(&operatorConfig.Status.Conditions, v1alpha1.OperatorCondition{
+			Type:    v1alpha1.OperatorStatusTypeFailing,
+			Status:  v1alpha1.ConditionTrue,
+			Reason:  "ConfigObservationError",
```
juanvallejo: Is this defined as a const or global var somewhere?

pkg/operator/observe_config.go
```diff
@@ -74,30 +77,46 @@ func NewConfigObserver(
...
 
-	glog.Infof("writing updated observedConfig: %v", diff.ObjectDiff(operatorConfig.Spec.ObservedConfig.Object, observedConfig))
-	operatorConfig.Spec.ObservedConfig = runtime.RawExtension{Object: &unstructured.Unstructured{Object: observedConfig}}
+	var errs []error
+	for _, observer := range c.observers {
+		var currErrs []error
+		observedConfig, currErrs = observer(c.kubeClient, &rest.Config{}, observedConfig)
+		errs = append(errs, currErrs...)
+	}
+	if !reflect.DeepEqual(currentConfig, observedConfig) {
```
juanvallejo: @deads2k worth using the apimachinery equality.Semantic.DeepEqual helper here?

pkg/operator/observe_config.go
```diff
@@ -124,31 +143,30 @@ func observeEtcdEndpoints(kubeClient kubernetes.Interface, clientConfig *rest.Co
...
 
 // observeClusterConfig observes CIDRs from cluster-config-v1 in order to populate list of restrictedCIDRs
-func observeClusterConfig(kubeClient kubernetes.Interface, clientConfig *rest.Config, observedConfig map[string]interface{}) (map[string]interface{}, error) {
+func observeClusterConfig(kubeClient kubernetes.Interface, clientConfig *rest.Config, observedConfig map[string]interface{}) (map[string]interface{}, []error) {
+
 	clusterConfig, err := kubeClient.CoreV1().ConfigMaps("kube-system").Get("cluster-config-v1", metav1.GetOptions{})
 	if errors.IsNotFound(err) {
-		glog.Warningf("cluster-config-v1 not found in the kube-system namespace")
-		return observedConfig, nil
+		return observedConfig, []error{fmt.Errorf("cluster-config-v1 not found in the kube-system namespace")}
```
juanvallejo: maybe worth including the resource type in the error message? (e.e `configmap cluster-config-v1...`)

pkg/operator/observe_config.go
sanchezl: Done

pkg/operator/observe_config.go
sanchezl: I've added this info to all messages.

pkg/operator/observe_config.go
```diff
@@ -107,48 +136,59 @@ func (c ConfigObserver) sync() error {
...
 	if errors.IsNotFound(err) {
-		return observedConfig, nil
+		return observedConfig, []error{fmt.Errorf("endpoints/etcd.kube-system: not found")}
 	}
 	if err != nil {
-		return observedConfig, err
+		return observedConfig, []error{err}
+	}
+	dnsSuffix, ok := etcdEndpoints.Annotations["alpha.installer.openshift.io/dns-suffix"]
+	if !ok || dnsSuffix == "" {
```
juanvallejo: I think it should be okay to just check for the length of the string: `if len(dnsSuffix) == 0 {`

juanvallejo:
overall, this lgtm

pkg/operator/observe_config.go
```diff
@@ -92,12 +90,43 @@ func (c ConfigObserver) sync() error {
...
+		var currErrs []error
+		observedConfig, currErrs = observer(c.kubeClient, &rest.Config{}, observedConfig)
+		errs = append(errs, currErrs...)
 	}
 
-	glog.Infof("writing updated observedConfig: %v", diff.ObjectDiff(operatorConfig.Spec.ObservedConfig.Object, observedConfig))
-	operatorConfig.Spec.ObservedConfig = runtime.RawExtension{Object: &unstructured.Unstructured{Object: observedConfig}}
+	if len(errs) > 0 {
+		var messages []string
+		for _, currentError := range errs {
```
soltysh: @deads2k any particular reason we don't want to use `k8s.io/apimachinery/pkg/util/errors.Aggregate` and get rid of that loop?

pkg/operator/observe_config.go
soltysh: Doesn't matter.

pkg/operator/observe_config_test.go
```diff
@@ -1,41 +1,400 @@
 package operator
 
 import (
+	"encoding/json"
+	"github.com/ghodss/yaml"
```
soltysh: Don't mix third party imports with builtins.

pkg/operator/observe_config.go
deads2k: > @deads2k any particular reason we don't want to use `k8s.io/apimachinery/pkg/util/errors.Aggregate` and get rid of that loop?

It doesn't give me pretty newlines.

pkg/operator/observe_config.go
```diff
@@ -92,12 +90,43 @@ func (c ConfigObserver) sync() error {
 	// don't worry about errors
 	currentConfig := map[string]interface{}{}
 	json.NewDecoder(bytes.NewBuffer(operatorConfig.Spec.ObservedConfig.Raw)).Decode(&currentConfig)
-	if reflect.DeepEqual(currentConfig, observedConfig) {
-		return nil
+	observedConfig := map[string]interface{}{}
+	json.NewDecoder(bytes.NewBuffer(operatorConfig.Spec.ObservedConfig.Raw)).Decode(&observedConfig)
+	currentStatus := operatorConfig.Status.DeepCopy()
```
deads2k: we actually want to mutate the copy, not the original.

deads2k: > we actually want to mutate the copy, not the original.

it has to do with working from caches later.

pkg/operator/observe_config.go
```diff
@@ -24,11 +25,18 @@ import (
...
+	"github.com/openshift/api/operator/v1alpha1"
 	operatorconfigclientv1alpha1 "github.com/openshift/cluster-kube-apiserver-operator/pkg/generated/clientset/versioned/typed/kubeapiserver/v1alpha1"
 	kubeapiserveroperatorinformers "github.com/openshift/cluster-kube-apiserver-operator/pkg/generated/informers/externalversions"
+	"github.com/openshift/library-go/pkg/operator/v1alpha1helpers"
 )
 
-type observeConfigFunc func(kubernetes.Interface, *rest.Config, map[string]interface{}) (map[string]interface{}, error)
+const (
+	configObservationError   = "ConfigObservationError"
+	configObservationSuccess = "ConfigObservationSuccess"
```
deads2k: no.  this will result in fighting control loops.

pkg/operator/observe_config.go
```diff
@@ -92,12 +90,43 @@ func (c ConfigObserver) sync() error {
...
+		})
+	} else {
+		condition := v1alpha1helpers.FindOperatorCondition(operatorConfig.Status.Conditions, v1alpha1.OperatorStatusTypeFailing)
+		if condition != nil && condition.Status == v1alpha1.ConditionTrue && condition.Reason == configObservationError {
+			condition.Status = v1alpha1.ConditionFalse
+			condition.Reason = configObservationSuccess
+			condition.Message = "Configuration observed successfully"
+		}
+	}
+	if reflect.DeepEqual(currentConfig, observedConfig) && reflect.DeepEqual(currentStatus, operatorConfig.Status) {
```
deads2k: if the whole thing is equal, why check status too?

pkg/operator/observe_config.go
```diff
@@ -92,12 +90,43 @@ func (c ConfigObserver) sync() error {
...
+		return nil
+	}
+	if !reflect.DeepEqual(currentConfig, observedConfig) {
```
deads2k: ok, this code is weird.  Just check the spec update, do it.  Then check the status update (from the result of the first update!) and do that.

pkg/operator/observe_config.go
deads2k: > Doesn't matter.

no unnecesary `ok`

pkg/operator/observe_config.go
```diff
@@ -107,48 +136,59 @@ func (c ConfigObserver) sync() error {
...
+		for addressIndex, address := range subset.Addresses {
+			if address.Hostname == "" {
+				errs = append(errs, fmt.Errorf("endpoints/etcd.kube-system: subsets[%v]addresses[%v].hostname not found", subsetIndex, addressIndex))
+				continue
+			}
+			etcdURLs = append(etcdURLs, "https://"+address.Hostname+"."+dnsSuffix+":2379")
 		}
 	}
+	if len(errs) > 0 {
+		return observedConfig, errs
```
deads2k: I'm still missing my "use the old config", aren't I?

pkg/operator/observe_config.go
```diff
@@ -157,20 +197,26 @@ func observeClusterConfig(kubeClient kubernetes.Interface, clientConfig *rest.Co
...
-		return observedConfig, nil
+	restrictedCIDRs, found, err := unstructured.NestedStringSlice(observedConfig, "admissionPluginConfig", "openshift.io/RestrictedEndpointsAdmission", "configuration", "restrictedCIDRs")
+	if err != nil || !found {
+		restrictedCIDRs = []string{"", ""}
 	}
-	if cidr := networking["podCIDR"]; cidr != nil {
-		restrictedCIDRs = append(restrictedCIDRs, fmt.Sprintf("%v", cidr))
+	errs := []error{}
+	podCIDR, _, err := unstructured.NestedString(installConfig, "networking", "podCIDR")
+	if err != nil {
```
deads2k: when you get to two conditions (not just if/else, but if/else if), use a switch case.

pkg/operator/observe_config.go
deads2k: > I'm still missing my "use the old config", aren't I?

This isn't strictly worse than what I had before, so I can take it as a followup.  A quick followup.

pkg/operator/observe_config.go
sanchezl: Help me understand. Are we saying a single reason, such as 'ConfigObservation' ?

pkg/operator/observe_config.go
sanchezl: Status might of changed, even if the observed config has not.

pkg/operator/observe_config.go
deads2k: > @deads2k worth using the apimachinery equality.Semantic.DeepEqual helper here?

no

pkg/operator/observe_config.go
```diff
@@ -24,11 +25,15 @@ import (
...
+	"github.com/openshift/api/operator/v1alpha1"
 	operatorconfigclientv1alpha1 "github.com/openshift/cluster-kube-apiserver-operator/pkg/generated/clientset/versioned/typed/kubeapiserver/v1alpha1"
 	kubeapiserveroperatorinformers "github.com/openshift/cluster-kube-apiserver-operator/pkg/generated/informers/externalversions"
+	"github.com/openshift/library-go/pkg/operator/v1alpha1helpers"
 )
 
-type observeConfigFunc func(kubernetes.Interface, *rest.Config, map[string]interface{}) (map[string]interface{}, error)
+const configObservationErrorConditionReason = "ConfigObservationError"
+
+type observeConfigFunc func(kubeClient kubernetes.Interface, clientConfig *rest.Config, observedConfig map[string]interface{}) (map[string]interface{}, []error)
```
deads2k: please doc this api.  I can't figure out what it is doing.

pkg/operator/observe_config.go
```diff
@@ -92,63 +87,112 @@ func (c ConfigObserver) sync() error {
 	// don't worry about errors
 	currentConfig := map[string]interface{}{}
 	json.NewDecoder(bytes.NewBuffer(operatorConfig.Spec.ObservedConfig.Raw)).Decode(&currentConfig)
-	if reflect.DeepEqual(currentConfig, observedConfig) {
-		return nil
+	observedConfig := map[string]interface{}{}
+	json.NewDecoder(bytes.NewBuffer(operatorConfig.Spec.ObservedConfig.Raw)).Decode(&observedConfig)
```
deads2k: Reading top to bottom, this just seems wrong.  You have a previous config and you have the observed config you're building up.  Using the previous value means active clearing instead of active adding.  That seems backwards, dangerous, and wrong.

pkg/operator/observe_config.go
```diff
@@ -92,63 +87,112 @@ func (c ConfigObserver) sync() error {
...
+		if condition != nil && condition.Status == v1alpha1.ConditionTrue && condition.Reason == configObservationErrorConditionReason {
+			condition.Status = v1alpha1.ConditionFalse
+			condition.Reason = configObservationErrorConditionReason
+			condition.Message = "Configuration observed successfully"
+		}
+	}
+
+	if !reflect.DeepEqual(operatorConfig.Status, status) {
+		operatorConfig.Status = *status
+		_, err = c.operatorConfigClient.KubeAPIServerOperatorConfigs().Update(operatorConfig)
```
deads2k: UpdateStatus

deads2k:
Please actually try this and cause a failure.  I would expect you to see dueling controllers.

pkg/operator/observe_config.go
sanchezl: Done

pkg/operator/observe_config.go
sanchezl: Done

pkg/operator/observe_config_test.go
sanchezl: Done

pkg/operator/observe_config.go
sanchezl: Done

pkg/operator/observe_config.go
sanchezl: Done

pkg/operator/observe_config.go
sanchezl: For this particular operator, there is nothing to clear as all config keys are required. Is there a situation where the list of required keys might of changed? e.g. during an upgrade of the kube-apiserver?

deads2k: > For this particular operator, there is nothing to clear as all config keys are required. Is there a situation where the list of required keys might of changed? e.g. during an upgrade of the kube-apiserver?

A case where you have optional keys.  A different version of the operator.  Start empty and build up, don't start full and clear.

sanchezl:
> Please actually try this and cause a failure. I would expect you to see dueling controllers.

I've tried it and it appears to be working fine.

pkg/operator/observe_config.go
```diff
@@ -24,11 +25,15 @@ import (
...
+	"github.com/openshift/api/operator/v1alpha1"
 	operatorconfigclientv1alpha1 "github.com/openshift/cluster-kube-apiserver-operator/pkg/generated/clientset/versioned/typed/kubeapiserver/v1alpha1"
 	kubeapiserveroperatorinformers "github.com/openshift/cluster-kube-apiserver-operator/pkg/generated/informers/externalversions"
+	"github.com/openshift/library-go/pkg/operator/v1alpha1helpers"
 )
 
-type observeConfigFunc func(kubernetes.Interface, *rest.Config, map[string]interface{}) (map[string]interface{}, error)
+const configObservationErrorConditionReason = "ConfigObservationError"
+
+type observeConfigFunc func(kubeClient kubernetes.Interface, clientConfig *rest.Config, currentConfig, observedConfig map[string]interface{}) []error
```
deads2k: still missing doc.

pkg/operator/observe_config.go
```diff
@@ -92,63 +87,120 @@ func (c ConfigObserver) sync() error {
 	// don't worry about errors
 	currentConfig := map[string]interface{}{}
 	json.NewDecoder(bytes.NewBuffer(operatorConfig.Spec.ObservedConfig.Raw)).Decode(&currentConfig)
-	if reflect.DeepEqual(currentConfig, observedConfig) {
-		return nil
+	observedConfig := map[string]interface{}{}
+	//json.NewDecoder(bytes.NewBuffer(operatorConfig.Spec.ObservedConfig.Raw)).Decode(&observedConfig)
```
deads2k: don't keep cruft

pkg/operator/observe_config.go
```diff
@@ -92,63 +87,120 @@ func (c ConfigObserver) sync() error {
...
+		}
+		v1alpha1helpers.SetOperatorCondition(&status.Conditions, v1alpha1.OperatorCondition{
+			Type:    v1alpha1.OperatorStatusTypeFailing,
+			Status:  v1alpha1.ConditionTrue,
+			Reason:  configObservationErrorConditionReason,
+			Message: strings.Join(messages, "\n"),
+		})
+	} else {
+		condition := v1alpha1helpers.FindOperatorCondition(status.Conditions, v1alpha1.OperatorStatusTypeFailing)
+		if condition != nil && condition.Status == v1alpha1.ConditionTrue && condition.Reason == configObservationErrorConditionReason {
```
deads2k: check if the condition isn't false instead of checking if it is true.

pkg/operator/observe_config.go
```diff
@@ -92,63 +87,120 @@ func (c ConfigObserver) sync() error {
...
+			condition.Status = v1alpha1.ConditionFalse
+			condition.Reason = ""
+			condition.Message = ""
+		}
+	}
+
+	if !reflect.DeepEqual(operatorConfig.Status, status) {
```
deads2k: equalities.Semantic.DeepEqual

pkg/operator/observe_config.go
```diff
@@ -92,63 +87,120 @@ func (c ConfigObserver) sync() error {
...
+		var currErrs []error
+		currErrs = observer(c.kubeClient, &rest.Config{}, currentConfig, observedConfig)
+		errs = append(errs, currErrs...)
 	}
 
-	glog.Infof("writing updated observedConfig: %v", diff.ObjectDiff(operatorConfig.Spec.ObservedConfig.Object, observedConfig))
-	operatorConfig.Spec.ObservedConfig = runtime.RawExtension{Object: &unstructured.Unstructured{Object: observedConfig}}
-	if _, err := c.operatorConfigClient.KubeAPIServerOperatorConfigs().Update(operatorConfig); err != nil {
-		return err
+	if !reflect.DeepEqual(currentConfig, observedConfig) {
```
deads2k: equalities.Semantic.DeepEqual

pkg/operator/observe_config.go
```diff
@@ -92,63 +87,120 @@ func (c ConfigObserver) sync() error {
...
+		return []error{fmt.Errorf("endpoints/etcd.kube-system: no etcd endpoints found")}
 	}
-	return observedConfig, nil
+	return nil
 }
 
 // observeClusterConfig observes CIDRs from cluster-config-v1 in order to populate list of restrictedCIDRs
-func observeClusterConfig(kubeClient kubernetes.Interface, clientConfig *rest.Config, observedConfig map[string]interface{}) (map[string]interface{}, error) {
+func observeClusterConfig(kubeClient kubernetes.Interface, clientConfig *rest.Config, currentConfig, observedConfig map[string]interface{}) []error {
+	restrictedCIDRsPath := []string{"admissionPluginConfig", "openshift.io/RestrictedEndpointsAdmission", "configuration", "restrictedCIDRs"}
```
deads2k: do this in the above method too

pkg/operator/observe_config.go
```diff
@@ -157,20 +209,28 @@ func observeClusterConfig(kubeClient kubernetes.Interface, clientConfig *rest.Co
 	//     networking:
 	//       podCIDR: 10.2.0.0/16
 	//       serviceCIDR: 10.3.0.0/16
-	restrictedCIDRs := []string{}
-	networking, ok := installConfig["networking"].(map[string]interface{})
-	if !ok {
-		return observedConfig, nil
+	restrictedCIDRs, found, err := unstructured.NestedStringSlice(observedConfig, restrictedCIDRsPath...)
+	if err != nil || !found {
+		restrictedCIDRs = []string{"", ""}
```
deads2k: why two empty strings instead of an empt slice?

pkg/operator/observe_config.go
```diff
@@ -157,20 +209,28 @@ func observeClusterConfig(kubeClient kubernetes.Interface, clientConfig *rest.Co
...
-		glog.Warningf("No value found for install-config/networking/podCIDR.")
+	errs := []error{}
+	podCIDR, _, err := unstructured.NestedString(installConfig, "networking", "podCIDR")
+	switch {
+	case err != nil:
+		errs = append(errs, err)
+	case podCIDR == "":
+		errs = append(errs, fmt.Errorf("configmap/cluster-config-v1.kube-system: install-config/networking/podCIDR not found"))
+	default:
+		restrictedCIDRs[0] = podCIDR
```
deads2k: append instead of relying on length above

pkg/operator/observe_config.go
```diff
@@ -4,17 +4,20 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"reflect"
+	"k8s.io/apimachinery/pkg/api/equality"
```
soltysh: Don't mix k8s and built-in deps.

PR #86, opened 2018-10-23
Add cabundle configmap observer

juanvallejo:
cc @deads2k @sttts 

Adds a ca-bundle configmap, whose ca.crt value is populated by the serving-service-cert-signer controller.  Adds an observer for the ca.crt value of the new configmap

Will add a test

### DIFF
--- a/pkg/operator/observe_config.go
+++ b/pkg/operator/observe_config.go
@@ -33,7 +33,8 @@ type observeConfigFunc func(kubernetes.Interface, *rest.Config, map[string]inter
 type ConfigObserver struct {
 	operatorConfigClient operatorconfigclientv1alpha1.KubeapiserverV1alpha1Interface
 
-	kubeClient kubernetes.Interface
+	kubeClient   kubernetes.Interface
+	clientConfig *rest.Config
 
 	// queue only ever has one item, but it has nice error handling backoff/retry semantics
 	queue workqueue.RateLimitingInterface
@@ -48,11 +49,14 @@ func NewConfigObserver(
 	operatorConfigInformers kubeapiserveroperatorinformers.SharedInformerFactory,
 	kubeInformersForKubeSystemNamespace kubeinformers.SharedInformerFactory,
 	operatorConfigClient operatorconfigclientv1alpha1.KubeapiserverV1alpha1Interface,
+	kubeInformersForOpenshiftServiceCertSignerNamespace kubeinformers.SharedInformerFactory,
 	kubeClient kubernetes.Interface,
+	clientConfig *rest.Config,
 ) *ConfigObserver {
 	c := &ConfigObserver{
 		operatorConfigClient: operatorConfigClient,
 		kubeClient:           kubeClient,
+		clientConfig:         clientConfig,
 
 		queue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "ConfigObserver"),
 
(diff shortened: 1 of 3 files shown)


pkg/operator/observe_config.go
```diff
@@ -105,6 +111,20 @@ func (c ConfigObserver) sync() error {
...
 
+func observeCABundleConfig(kubeClient kubernetes.Interface, clientConfig *rest.Config, observedConfig map[string]interface{}) (map[string]interface{}, error) {
+	caBundleConfig, err := kubeClient.CoreV1().ConfigMaps(operatorNamespaceName).Get("service-ca-bundle-configmap", metav1.GetOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		return nil, err
+	}
+	if caBundleConfig != nil {
+		// TODO(juanvallejo): reflect any issues in operator status
+		if val := caBundleConfig.Data["service-ca.crt"]; len(val) > 0 {
+			// TODO: what is the nested field path? Where do we store the service-ca info?
```
juanvallejo: @deads2k or @sttts are we adding this to a volume definition on the operator config?

0000_10_cluster-kube-apiserver-operator_02_cabundleconfigmap.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: openshift-core-operators
+  name: service-ca-bundle-configmap
+  annotations:
+    service.alpha.openshift.io/inject-cabundle: "true"
```
deads2k: clever, but don't do this.  @openshift/sig-security and @mrogers950  in particular should be able to tell you which configmap in their operator namespace provides *just* the ca bundle for their signer.  For compatibility, this one will be injected with their signer *and* the signer for the old way.  We want to keep them separate.

mrogers950: @juanvallejo  it's the `signing-cabundle` configmap in `openshift-service-cert-signer`. The data item is `cabundle.crt`and contains the PEM bundle.

pkg/operator/observe_config.go
```diff
@@ -48,24 +49,29 @@ func NewConfigObserver(
...
 			observeEtcdEndpoints,
 			observeClusterConfig,
+			observeCABundleConfig,
 		},
 	}
 
 	operatorConfigInformers.Kubeapiserver().V1alpha1().KubeApiserverOperatorConfigs().Informer().AddEventHandler(c.eventHandler())
 	kubeInformersForKubeSystemNamespace.Core().V1().Endpoints().Informer().AddEventHandler(c.eventHandler())
 	kubeInformersForKubeSystemNamespace.Core().V1().ConfigMaps().Informer().AddEventHandler(c.eventHandler())
+	kubeInformersForOpenshiftServiceCertSignerNamespace.Core().V1().Namespaces().Informer().AddEventHandler(c.eventHandler())
```
deads2k: namespace looks wrong.  You're looking for a configmap change.

pkg/operator/observe_config.go
```diff
@@ -105,6 +111,19 @@ func (c ConfigObserver) sync() error {
 	return nil
 }
 
+func observeCABundleConfig(kubeClient kubernetes.Interface, clientConfig *rest.Config, observedConfig map[string]interface{}) (map[string]interface{}, error) {
+	caBundleConfig, err := kubeClient.CoreV1().ConfigMaps(serviceCertSignerNamespaceName).Get("signing-cabundle", metav1.GetOptions{})
+	if err != nil && !errors.IsNotFound(err) {
```
deads2k: NotFound gets separate handling before err.

pkg/operator/observe_config.go
deads2k: > @deads2k or @sttts are we adding this to a volume definition on the operator config? Not sure of the path here

You actually need to do two things: 
1) do a `SyncConfigMap` call in the main config to copy the CA locally
2) watch for that change and if you have a value, set this value for a path.  You need to create a new bit in our binary's config (kubecontorlplane group)

deads2k:
I think this is actually needed in kube-controller-manager.

Once we get the information available in our namespace, I think the task belongs to @openshift/sig-security 

@derekwaynecarr fyi, more config wiring for a team's feature.

/close

juanvallejo:
Opened against kube-controller-manager repo: https://github.com/openshift/cluster-kube-controller-manager-operator/pull/61

PR #88, opened 2018-10-24
fix case of the kubeapiserveroperatorconfig type

deads2k:
`APIServer` instead of `Apiserver`.

/assign @sttts


sttts:
I always liked Apiserver. But fine with me, we have APIServer elsewhere. So let's be consistent.

lgtm if it is green.

PR #91, opened 2018-10-25
render: temporary --disable-phase-2 flag

sttts:
Depends on https://github.com/openshift/library-go/pull/58.


deads2k:
I see no reason to hold this.

deads2k:
> I see no reason to hold this.

wait, I remember now.

PR #94, opened 2018-10-26
make the deployment control loop generic

deads2k:
This is needed before start promoting to library-go for re-use in controller manager


deads2k:
now with more bug fixes.

deads2k:
@sttts Once we get this in, I'll move the bits to library-go and pull them down into the kube-controller-manager.

sttts:
This is really hard to review. Github shows removal and creation of tons of files. Any way to split code moves and code changes?

pkg/apis/kubeapiserver/v1alpha1/types.go
```diff
@@ -1,6 +1,7 @@
 package v1alpha1
 
 import (
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/staticpodcontroller"
```
soltysh: Down to openshift imports :stuck_out_tongue_closed_eyes:

pkg/operator/staticpod_interface.go
```diff
@@ -0,0 +1,51 @@
...
+	instance, err := c.informers.Kubeapiserver().V1alpha1().KubeAPIServerOperatorConfigs().Lister().Get("instance")
+	if err != nil {
+		return nil, nil, "", err
+	}
+
+	return &instance.Spec.OperatorSpec, &instance.Status.StaticPodConfigStatus, instance.ResourceVersion, nil
+}
+
+func (c *staticPodOperatorClient) UpdateStatus(resourceVersion string, status *staticpodcontroller.StaticPodConfigStatus) (*staticpodcontroller.StaticPodConfigStatus, error) {
+	original, err := c.informers.Kubeapiserver().V1alpha1().KubeAPIServerOperatorConfigs().Lister().Get("instance")
```
soltysh: You're swallowing the error, why not using `CurrentStatus` here also?

pkg/operator/staticpodcontroller/installer_controller.go
```diff
@@ -0,0 +1,348 @@
...
+		ret.TargetDeploymentID = 0
+
+		errors := []string{}
+		for _, containerStatus := range installerPod.Status.ContainerStatuses {
+			if containerStatus.State.Terminated != nil && len(containerStatus.State.Terminated.Message) > 0 {
+				errors = append(errors, fmt.Sprintf("%q: %s", containerStatus.Name, containerStatus.State.Terminated.Message))
+			}
+		}
+		if len(errors) == 0 {
+			errors = append(errors, "no detailed termination message, see `oc get -n %q pods/%q -oyaml`", installerPod.Namespace, installerPod.Name)
```
soltysh: Maybe also suggest seeing events?

pkg/operator/staticpodcontroller/installer_controller.go
```diff
@@ -0,0 +1,348 @@
...
+}
+
+func getInstallerPodName(deploymentID int32, nodeName string) string {
+	return fmt.Sprintf("installer-%d-%s", deploymentID, nodeName)
+}
+
+// createInstallerPod creates the installer pod with the secrets required to
+func (c *InstallerController) createInstallerPod(currKubeletState *KubeletState, operatorSpec *operatorv1alpha1.OperatorSpec, deploymentID int32) (*KubeletState, error) {
+	required := resourceread.ReadPodV1OrDie(v311_00_assets.MustAsset("v3.11.0/kube-apiserver/installer-pod.yaml"))
+	switch corev1.PullPolicy(operatorSpec.ImagePullPolicy) {
```
soltysh: Why we care about this validation here? Or if we do validate why not other fields too?

pkg/operator/staticpodcontroller/interfaces.go
```diff
@@ -0,0 +1,37 @@
...
+	"k8s.io/client-go/tools/cache"
+)
+
+type OperatorClient interface {
+	Informer() cache.SharedIndexInformer
+	Get() (*operatorv1alpha1.OperatorSpec, *StaticPodConfigStatus, string, error)
+	UpdateStatus(string, *StaticPodConfigStatus) (*StaticPodConfigStatus, error)
+}
+
+// TODO move to openshift/api I think
```
soltysh: Why? IIUC this is only for kube-apiserver, is it not?

pkg/operator/staticpod_interface.go
deads2k: > You're swallowing the error, why not using `CurrentStatus` here also?

you need a full object of some kind to build the update payload until we switch to a dynamic client.

pkg/operator/staticpodcontroller/installer_controller.go
deads2k: > Why we care about this validation here? Or if we do validate why not other fields too?

Shrugs. @mfojtik's seems reasonable.

pkg/operator/staticpodcontroller/interfaces.go
deads2k: > Why? IIUC this is only for kube-apiserver, is it not?

apiserver, controller, scheduler, possibly etcd.

deads2k:
pull updated on latest, generic, library-go

sttts:
Now this is much better to read.

/lgtm

Needs rebase.

PR #98, opened 2018-10-31
render: remove phase 2 by default

sttts:
This turns `--disable-phase-2` into a no-op, to be removed when the installer stops using that flag.


cmd/cluster-kube-apiserver-operator/render/render.go
```diff
@@ -73,9 +70,10 @@ func (r *renderOpts) AddFlags(fs *pflag.FlagSet) {
 	fs.StringArrayVar(&r.etcdServerURLs, "manifest-etcd-server-urls", r.etcdServerURLs, "The etcd server URL, comma separated.")
 	fs.StringVar(&r.etcdServingCA, "manifest-etcd-serving-ca", r.etcdServingCA, "The etcd serving CA.")
 
+	// TODO: remove when the installer has stopped using it
 	fs.BoolVar(&r.disablePhase2, "disable-phase-2", r.disablePhase2, "Disable rendering of the phase 2 daemonset and dependencies.")
 	fs.MarkHidden("disable-phase-2")
```
soltysh: You can remove that, `MarkDeprecated` does that.

PR #101, opened 2018-10-31
configure kube apiserver w/ internalregistryhostname published by registry

bparees:


bparees:
@deads2k @juanvallejo ptal, goes with https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/26 and https://github.com/openshift/cluster-openshift-apiserver-operator/pull/41

pkg/operator/observe_config.go
```diff
@@ -187,6 +195,28 @@ func observeClusterConfig(kubeClient kubernetes.Interface, clientConfig *rest.Co
 	return observedConfig, nil
 }
 
+// observeInternalRegistryHostname reads the internal registry hostname from the cluster configuration as provided by
+// the registry operator.
+func observeInternalRegistryHostname(kubeClient kubernetes.Interface, clientConfig *rest.Config, observedConfig map[string]interface{}) (map[string]interface{}, error) {
+	imageConfigClient, err := configv1client.NewForConfig(clientConfig)
+	if err != nil {
+		return nil, err
```
deads2k: @sanchezl your status pull missed, so you'll need to tidy up this one too.

pkg/operator/observe_config.go
```diff
@@ -187,6 +195,28 @@ func observeClusterConfig(kubeClient kubernetes.Interface, clientConfig *rest.Co
 	return observedConfig, nil
 }
 
+// observeInternalRegistryHostname reads the internal registry hostname from the cluster configuration as provided by
+// the registry operator.
+func observeInternalRegistryHostname(kubeClient kubernetes.Interface, clientConfig *rest.Config, observedConfig map[string]interface{}) (map[string]interface{}, error) {
```
deads2k: needs unit test

deads2k:
Just looking for a test.

bparees:
@deads2k ok this is all cleaned up, tests added, and ready for re-review. Note that what I did here will need to be carried over into @juanvallejo's openshift-apiserver operator PR also, assuming he's going to add a unit test to it.

bparees:
@deads2k i'd like to see this landed today so we can copy the patterns over to https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/26 and https://github.com/openshift/cluster-openshift-apiserver-operator/pull/41  as well as the work @adambkaplan is doing for buildconfig.

pkg/operator/observe_config.go
```diff
@@ -19,16 +19,21 @@ import (
...
 	operatorconfigclientv1alpha1 "github.com/openshift/cluster-kube-apiserver-operator/pkg/generated/clientset/versioned/typed/kubeapiserver/v1alpha1"
 	kubeapiserveroperatorinformers "github.com/openshift/cluster-kube-apiserver-operator/pkg/generated/informers/externalversions"
 )
 
-type observeConfigFunc func(kubernetes.Interface, *rest.Config, map[string]interface{}) (map[string]interface{}, error)
+type Listers struct {
+	imageConfigLister configlistersv1.ImageLister
+}
+
+type observeConfigFunc func(kubernetes.Interface, Listers, map[string]interface{}) (map[string]interface{}, error)
```
deads2k: should be able to remove the `kubernetes.Interface` entirely.

pkg/operator/observe_config.go
```diff
@@ -19,16 +19,21 @@ import (
...
 	"k8s.io/client-go/util/workqueue"
 
+	configinformers "github.com/openshift/client-go/config/informers/externalversions"
+	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
 	operatorconfigclientv1alpha1 "github.com/openshift/cluster-kube-apiserver-operator/pkg/generated/clientset/versioned/typed/kubeapiserver/v1alpha1"
 	kubeapiserveroperatorinformers "github.com/openshift/cluster-kube-apiserver-operator/pkg/generated/informers/externalversions"
 )
 
-type observeConfigFunc func(kubernetes.Interface, *rest.Config, map[string]interface{}) (map[string]interface{}, error)
+type Listers struct {
```
deads2k: needs the particular kube listers

pkg/operator/observe_config_test.go
```diff
@@ -39,3 +42,48 @@ func TestObserveClusterConfig(t *testing.T) {
 		t.Error(restrictedCIDRs[1])
 	}
 }
+
+type FakeImageConfigLister struct {
```
deads2k: Creative.  I owed you 
```go
indexer := cache.NewIndexer(controller.KeyFunc, cache.Indexers{})
for _, obj := range test.startingClusterRoles {
	indexer.Add(obj)
}
rbaclisters.NewClusterRoleLister(indexer)
```

So I'll take it, but I think this construction is easier

pkg/operator/observe_config.go
```diff
@@ -18,54 +18,68 @@ import (
...
+			imageConfigLister: configInformer.Config().V1().Images().Lister(),
+			endpointLister:    kubeInformersForKubeSystemNamespace.Core().V1().Endpoints().Lister(),
+			configmapLister:   kubeInformersForKubeSystemNamespace.Core().V1().ConfigMaps().Lister(),
 		},
 	}
 
 	operatorConfigInformers.Kubeapiserver().V1alpha1().KubeAPIServerOperatorConfigs().Informer().AddEventHandler(c.eventHandler())
 	kubeInformersForKubeSystemNamespace.Core().V1().Endpoints().Informer().AddEventHandler(c.eventHandler())
 	kubeInformersForKubeSystemNamespace.Core().V1().ConfigMaps().Informer().AddEventHandler(c.eventHandler())
+	configInformer.Config().V1().Images().Informer().AddEventHandler(c.eventHandler())
```
deads2k: here's a neat side effect.  If you don't wait for your informers to sync once, you can race and destroy your config.

deads2k:
Just the bit about waiting for sync and this lgtm.

@juanvallejo @sanchezl don't forget about the wait to sync.

pkg/operator/starter.go
```diff
@@ -94,6 +102,13 @@ func RunOperator(clientConfig *rest.Config, stopCh <-chan struct{}) error {
 	kubeInformersClusterScoped.Start(stopCh)
 	kubeInformersForOpenshiftKubeAPIServerNamespace.Start(stopCh)
 	kubeInformersForKubeSystemNamespace.Start(stopCh)
+	configInformers.Start(stopCh)
+
+	cache.WaitForCacheSync(stopCh, operatorConfigInformers.Kubeapiserver().V1alpha1().KubeAPIServerOperatorConfigs().Informer().HasSynced,
```
deads2k: this block goes here https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/pkg/operator/observe_config.go#L196

juanvallejo:
lgtm pending @deads2k comments

bparees:
@deads2k cache syncs added.

bparees:
@deads2k bump

juanvallejo:
@bparees just a heads up, at least in my PR, waiting for the config image cache to sync was blocking forever: https://github.com/openshift/cluster-openshift-apiserver-operator/pull/41/files#diff-a32feedccc2c285e54b07ce4f24afbf4R198

bparees:
> @bparees just a heads up, at least in my PR, waiting for the config image cache to sync was blocking forever

i'm going to assume that's because the imageconfig resource doesn't exist when this operator is coming up... @deads2k ?

deads2k:
/lgtm

@sanchezl I accidentally merged these in the wrong order.  Can you pick up the rebase for Ben since it was your changes?

bparees:
@deads2k we also need to resolve the "imageconfig never syncs" issue that @juanvallejo noted.

juanvallejo:
@bparees 
> i'm going to assume that's because the imageconfig resource doesn't exist when this operator is coming up

I am also not seeing the "cluster" image.config.openshift.io resource come up on my local install. I don't believe this is causing CI failures for me, but it might explain why I am seeing the cache sync wait forever

bparees:
> I am also not seeing the "cluster" image.config.openshift.io resource come up on my local install.

you won't see it until the registry successfully deploys which, by default, it does not today because there's no storage configuration.

juanvallejo:
@bparees 
> you won't see it until the registry successfully deploys which, by default, it does not today because there's no storage configuration.

@deads2k would it make sense to disable that cache sync wait for now?

bparees:
> @deads2k would it make sense to disable that cache sync wait for now?

since the CRD should be defined, i don't see why the cache should be blocked, even if there is no instance of the CRD.

sanchezl:
> @sanchezl please confirm you are picking up and rebasing this. This PR is now on course to be blocking demos for milestone 3.

@bparees  confirmed

PR #103, opened 2018-11-01
Use common build machinery

tnozicka:
/cc @deads2k 

@sttts can you confirm this retains the special functionality you use around `origin-release`? should be the same just the image build (name) is controlled by these 3 vars https://github.com/openshift/library-go/blob/255848f5975dc710dcc4434670f5b00e00e266fa/alpha-build-machinery/make/targets/openshift/images.mk#L1-L3


deads2k:
I defer to @sttts 

/assign @sttts

PR #104, opened 2018-11-02
 add metrics

deads2k:
This pulls in the new library-go which includes metrics enablement.

/assign @sttts


sttts:
Sgtm. What is missing to make it non-wip?

deads2k:
> Sgtm. What is missing to make it non-wip?

Just the real bump. Ready now.

Issue #83, opened 2018-10-22
initialize restrictedCIDRs during the bootstrapping phase


sanchezl:
Find out whether we have access to the cluster config during the bootstrapping phase executed at https://github.com/openshift/installer/blob/master/pkg/asset/ignition/bootstrap/content/bootkube.go#L31 .

It's a text/template. So maybe we have the cluster config in one form or another around to be plugged in, or even as file on the bootstrap node.

_Originally posted by @sttts in https://github.com/openshift/cluster-kube-apiserver-operator/pull/74#issuecomment-431285403_

sanchezl:
@damemi @sttts

PR #105, opened 2018-11-05
[WIP] Render RestrictedCIDRs during bootstrapping phase

damemi:
WIP - depends on https://github.com/openshift/library-go/pull/77 to add helpers to library-go

Reads `cluster-config.yaml` cluster configmap file to parse out the `PodCIDR` and `ServiceCIDR` and writes them to the config overrides for phase 1

Fixes https://github.com/openshift/cluster-kube-apiserver-operator/issues/83


damemi:
cc @sttts

bindata/bootkube/config/bootstrap-config-overrides.yaml
```diff
@@ -24,4 +24,11 @@ storageConfig:
   certFile: /etc/kubernetes/secrets/etcd-client.crt # origin 3.11: master.etcd-client.crt
   keyFile: /etc/kubernetes/secrets/etcd-client.key # origin 3.11: master.etcd-client.key
   urls: {{range .EtcdServerURLs}}
-  - {{.}}{{end}}
\ No newline at end of file
+  - {{.}}{{end}}
+admissionPluginConfig:
+  openshift.io/RestrictedEndpointsAdmission:
+    configuration:
```
sttts: is this completely missing in the defaultconfig.yaml? We cannot merge those sub-objects, only replace them.

cmd/cluster-kube-apiserver-operator/render/render.go
```diff
@@ -69,6 +72,7 @@ func (r *renderOpts) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&r.lockHostPath, "manifest-lock-host-path", r.lockHostPath, "A host path mounted into the apiserver pods to hold lock.")
 	fs.StringArrayVar(&r.etcdServerURLs, "manifest-etcd-server-urls", r.etcdServerURLs, "The etcd server URL, comma separated.")
 	fs.StringVar(&r.etcdServingCA, "manifest-etcd-serving-ca", r.etcdServingCA, "The etcd serving CA.")
+	fs.StringVar(&r.configFile, "config-file", r.configFile, "ClusterConfig ConfigMap file.")
```
sttts: how about `--cluster-config-file` ?

sttts: May this be empty? If not, let's validate the value.

cmd/cluster-kube-apiserver-operator/render/render.go
```diff
@@ -121,14 +125,27 @@ type TemplateData struct {
...
+	RestrictedCIDRs []string
 }
 
 // Run contains the logic of the render command.
 func (r *renderOpts) Run() error {
+	installConfigFileData, err := ioutil.ReadFile(r.configFile)
+	if err != nil {
+		return err
+	}
+	restrictedCIDRs, err := discoverRestrictedCIDRs(v1alpha1helpers.InstallConfigFromFile(installConfigFileData))
```
sttts: would be simpler to just parse the file and then parse the ClusterConfig Go struct into discoverRestrictedCIDRs. Wdyt? Functional is nice, but does not really make it simpler here.

damemi: I agree with removing the functionality, but because we still have to parse out the InstallConfig from the data in the config map I'd like to do that step as well and pass the InstallConfig struct directly instead (see changes to https://github.com/openshift/library-go/pull/77)

deads2k:
@sttts @mfojtik @smarterclayton do we actually need this to be consistent? Can we instead claim that we trust that actors using the bootstrap node enough that we don't need a complete admission chain?

deads2k:
/hold

We aren't going to pick up an installer dependency in this repo.

Also, since we created the issue, the bootstrap node has been removed from the cluster. The need for this is less clear.

sttts:
> @sttts @mfojtik @smarterclayton do we actually need this to be consistent? Can we instead claim that we trust that actors using the bootstrap node enough that we don't need a complete admission chain?

For the apiserver we could decide that a complete admission chain is not needed. But the controller-manager has the same settings, but they influence the logic of the controllers:

```
  cluster-cidr:
  - "10.2.0.0/16"
  service-cluster-ip-range:
  - "10.3.0.0/16"
```

smarterclayton:
InstallerConfig will probably end up in openshift/api

damemi:
@squeed could you shed some light on the importance of importing these CIDRs?

squeed:
I filed issue #106 proposing a much simpler solution: the installer creates a `Cluster.cluster.k8s.io/v1alpha1` that includes all ServiceCIDRs and ClusterCIDRs. Does that fix the nasty dependency chain?

squeed:
I can't speak as to in which order the AdmissionController has to be configured - that's your wheelhouse. But it is important from a security perspective that, eventually, it is configured.

sttts:
> But it is important from a security perspective that, eventually, it is configured.

@squeed  "eventually" after bootstrapping everything will be set up. This PR is about the bootstrap phase. If I understand correctly, for admission we can live with missing CIDRs, but for the controller manager it's not only cosmetics, but service IPs will be wrong if the CIDRs do not match the network setup. Can you confirm?

squeed:
@sttts good question! (I didn't realize this PR would also have that effect.. is that part of the WIP?) 

I'm not sure. Certainly the controller manager needs to be started with the correct `--service-cluster-ip-range` in the steady state, but that should not be necessary until we actually try and create some Services without a ServiceIP. (I'm not sure about this, so we should test). Are we restarting the controller-manager during the pivot to the running cluster?

Either way, the ideal state would be for the operator to read the ServiceCIDR from the Cluster object, and render that to the controller-manager config. Can we do that as part of the bootstrap flow?

PR #107, opened 2018-11-06
fix printing version

mfojtik:
1) Before the `--version` flag was not defined and we were not setting the `-ldflag` during the go build. I think it only makes sense to set it when building the images so I updated the Dockerfile. Now:

```
$ docker run --rm  openshift/origin-cluster-kube-apiserver-operator:latest --version
cluster-kube-apiserver-operator version v0.0.0-alpha.0-188-gbf4d1be
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,9 @@
 #
 FROM openshift/origin-release:golang-1.10
 COPY . /go/src/github.com/openshift/cluster-kube-apiserver-operator
-RUN cd /go/src/github.com/openshift/cluster-kube-apiserver-operator && go build ./cmd/cluster-kube-apiserver-operator
+WORKDIR /go/src/github.com/openshift/cluster-kube-apiserver-operator
+ENV GO_PACKAGE github.com/openshift/cluster-kube-apiserver-operator
+RUN go build -ldflags "-X $GO_PACKAGE/pkg/version.versionFromGit=$(git describe --long --tags --abbrev=7 --match 'v[0-9]*')" ./cmd/cluster-kube-apiserver-operator
 
 FROM centos:7
 RUN mkdir -p /usr/share/bootkube/manifests
--- a/cmd/cluster-kube-apiserver-operator/main.go
+++ b/cmd/cluster-kube-apiserver-operator/main.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/openshift/cluster-kube-apiserver-operator/cmd/cluster-kube-apiserver-operator/render"
 	"github.com/openshift/cluster-kube-apiserver-operator/pkg/cmd/operator"
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/version"
 	"github.com/openshift/library-go/pkg/operator/staticpod/installerpod"
 )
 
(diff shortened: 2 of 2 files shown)


cmd/cluster-kube-apiserver-operator/main.go
```diff
@@ -1,32 +1,30 @@
...
 	"github.com/openshift/cluster-kube-apiserver-operator/pkg/cmd/operator"
 	"github.com/openshift/library-go/pkg/operator/staticpod/installerpod"
+	"github.com/openshift/library-go/pkg/serviceability"
 )
 
 func main() {
 	rand.Seed(time.Now().UTC().UnixNano())
 
-	pflag.CommandLine.SetNormalizeFunc(utilflag.WordSepNormalizeFunc)
-	pflag.CommandLine.AddGoFlagSet(goflag.CommandLine)
```
mfojtik: @deads2k i think you copied this from `kubectl` but for some reason it brings all test flags into this command... I searched trough imports and haven't found any offending package that does that so far. 

Not sure whether we want to risk that in future some of the vendored packages pollute the go flag and we get "extra" flags set?

/cc @soltysh

Dockerfile
```diff
@@ -11,3 +13,4 @@ COPY --from=0 /go/src/github.com/openshift/cluster-kube-apiserver-operator/clust
 
 COPY manifests /manifests
 LABEL io.openshift.release.operator true
+ENTRYPOINT ["/usr/bin/cluster-kube-apiserver-operator"]
```
mfojtik: @smarterclayton will this break the CVO or CI in any way?

I think it is much nicer for anyone to just run the image and get the operator binary executed automatically than guess the binary (`cluster-kube-apiserver-operator`) name. WDYT?

Dockerfile
```diff
@@ -2,7 +2,9 @@
 #
 FROM openshift/origin-release:golang-1.10
 COPY . /go/src/github.com/openshift/cluster-kube-apiserver-operator
-RUN cd /go/src/github.com/openshift/cluster-kube-apiserver-operator && go build ./cmd/cluster-kube-apiserver-operator
+WORKDIR /go/src/github.com/openshift/cluster-kube-apiserver-operator
+ENV GO_PACKAGE github.com/openshift/cluster-kube-apiserver-operator
+RUN go build -ldflags "-X $GO_PACKAGE/pkg/version.versionFromGit=$(git describe --long --tags --abbrev=7 --match 'v[0-9]*')" ./cmd/cluster-kube-apiserver-operator
```
mfojtik: @sttts @deads2k i guess we only care about having the version set to actual version in the image we build?

cmd/cluster-kube-apiserver-operator/main.go
```diff
@@ -1,32 +1,30 @@
...
-
 	logs.InitLogs()
 	defer logs.FlushLogs()
 
+	defer serviceability.BehaviorOnPanic(os.Getenv("OPENSHIFT_ON_PANIC"), version.Get())()
```
mfojtik: @deads2k since the operator is long running process, I guess we want to define the panic behavior and set the profile?

Dockerfile
sttts: Instead of?

Dockerfile
sttts: Who runs the operator manually?

Dockerfile
mfojtik: instead of having the version set in make build

PR #111, opened 2018-11-07
observe imagePolicyConfig/internalRegistryHostname

sanchezl:
Rebase of pull request https://github.com/openshift/cluster-kube-apiserver-operator/pull/101.
TODO: add some unit tests.

### DIFF
--- a/pkg/operator/observe_config.go
+++ b/pkg/operator/observe_config.go
@@ -21,13 +21,14 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	kubeinformers "k8s.io/client-go/informers"
-	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/rest"
+	corelistersv1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/flowcontrol"
 	"k8s.io/client-go/util/workqueue"
 
 	"github.com/openshift/api/operator/v1alpha1"
+	configinformers "github.com/openshift/client-go/config/informers/externalversions"
+	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
 	operatorconfigclientv1alpha1 "github.com/openshift/cluster-kube-apiserver-operator/pkg/generated/clientset/versioned/typed/kubeapiserver/v1alpha1"
 	kubeapiserveroperatorinformers "github.com/openshift/cluster-kube-apiserver-operator/pkg/generated/informers/externalversions"
 	"github.com/openshift/library-go/pkg/operator/v1alpha1helpers"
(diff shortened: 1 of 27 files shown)


juanvallejo:
@sanchezl you might want to bump deps here to pick up [this fix to openshift/api](https://github.com/openshift/api/pull/123)

bparees:
@sanchezl you're going to need to include a dependency bump.

bparees:
@juanvallejo jinx

pkg/operator/observe_config.go
```diff
@@ -257,13 +278,44 @@ func observeRestrictedCIDRs(kubeClient kubernetes.Interface, clientConfig *rest.
...
+// observeInternalRegistryHostname reads the internal registry hostname from the cluster configuration as provided by
+// the registry operator.
+func observeInternalRegistryHostname(listers Listers, currentConfig map[string]interface{}) (observedConfig map[string]interface{}, errs []error) {
+	observedConfig = map[string]interface{}{}
+	internalRegistryHostnamePath := []string{"imagePolicyConfig", "internalRegistryHostname"}
+	currentInternalRegistryHostname, _, _ := unstructured.NestedStringSlice(currentConfig, internalRegistryHostnamePath...)
+
+	configImage, err := listers.imageConfigLister.Get("cluster")
+	if errors.IsNotFound(err) {
+		errs = append(errs, fmt.Errorf("image.config.openshift.io/cluster.kube-system: not found"))
```
bparees: cluster.kubesystem?  imageconfig is a cluster scoped resource.

bparees: also why are we returning an error here?  I thought we just wanted to log this.

bparees: also comments on each of these decisions would be helpful for the next person.. (why we are treating "not found" and "error retrieving" differently, for example)

sanchezl: @bparees , @deads2k : please review these scenarios

if the **image.config.openshift.io/cluster** resource does not exist, just log a message, do not set **imagePolicyConfig.internalRegistryHostName** and do not return an error.
```gherkin
# The internal registry is not expected to be setup by the openshift-installer.
Scenario: New cluster startup.
  Given image.config.openshift.io/cluster resource does not exist
  And imagePolicyConfig.internalRegistryHostName was not previously observed
  When an attempt is made to observe imagePolicyConfig.internalRegistryHostName
  Then imagePolicyConfig.internalRegistryHostName will not be set
  And no error will be returned
  And the Failing condition will not be set
```

but what if its not first startup and a previous **image.config.openshift.io/cluster** resource was deleted?
```gherkin
Scenario: Resource image.config.openshift.io/cluster deleted
  Given image.config.openshift.io/cluster resource does not exist
  And imagePolicyConfig.internalRegistryHostName was previously observed
  When an attempt is made to observe imagePolicyConfig.internalRegistryHostName
  Then imagePolicyConfig.internalRegistryHostName will not be set
  And no error will be returned
  And the Failing condition will not be set
```

if an error (other than NotFound) occurs getting the **image.config.openshift.io/cluster** resource, return the previously observed **imagePolicyConfig.internalRegistryHostName** if any.
```gherkin
Scenario: Error getting resource image.config.openshift.io/cluster (previously observed).
  Given imagePolicyConfig.internalRegistryHostName was previously observed
  When an attempt is made to observe imagePolicyConfig.internalRegistryHostName
  And an error occurs getting the image.config.openshift.io/cluster resource
  Then imagePolicyConfig.internalRegistryHostName will be set to the previously observed value
  And error will be returned
  And the Failing condition will be set

Scenario: Error getting resource image.config.openshift.io/cluster (not previously observed).
  Given imagePolicyConfig.internalRegistryHostName was not previously observed
  When an attempt is made to observe imagePolicyConfig.internalRegistryHostName
  And an error occurs getting the image.config.openshift.io/cluster resource
  Then imagePolicyConfig.internalRegistryHostName will not be set
  And error will be returned
  And the Failing condition will be set
```

If **status.internalRegistryHostName** is not set, then do not set **imagePolicyConfig.internalRegistryHostName**, even if there was a previous value for it.
```gherkin
Scenario: status.internalRegistryHostName not set in image.config.openshift.io/cluster
  Given image.config.openshift.io/cluster resource exists
  And status.internalRegistryHostName not set
  When an attempt is made to observe imagePolicyConfig.internalRegistryHostName
  Then imagePolicyConfig.internalRegistryHostName will not be set
  And no error will be returned
  And the Failing condition will not be set
```

bparees: > if the image.config.openshift.io/cluster resource does not exist, just log a message, do not set imagePolicyConfig.internalRegistryHostName and do not return an error.

sounds right.

> but what if its not first startup and a previous image.config.openshift.io/cluster resource was deleted?

don't care, still don't set it.  (this means the previous setting, if any, will be cleared from the apiserver config, right?).

> if an error (other than NotFound) occurs getting the image.config.openshift.io/cluster resource, return the previously observed imagePolicyConfig.internalRegistryHostName if any.

right.

> If status.internalRegistryHostName is not set, then do not set imagePolicyConfig.internalRegistryHostName, even if there was a previous value for it.

right.

mfojtik:
@sanchezl can you split all changes in vendor/ into a separate commit? (call it `bump(*):`)

pkg/operator/observe_config.go
sanchezl: Ok.

PR #112, opened 2018-11-08
staticpods: wait for static pod to be running

sttts:


bindata/v3.11.0/kube-apiserver/pod.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
   name: openshift-kube-apiserver
   labels:
     app: openshift-kube-apiserver
+    deployment-id: "DEPLOYMENT_ID"
```
deads2k: good

pkg/operator/starter.go
```diff
@@ -104,6 +112,37 @@ func RunOperator(clientConfig *rest.Config, stopCh <-chan struct{}) error {
 	return fmt.Errorf("stopped")
 }
 
+func StaticPodIsRunningFunc(podlister corev1listers.PodLister, ns string) installer.StaticPodInstallStateFunc {
+	return func(installerPod *corev1.Pod, deploymentID int32, nodeName string) (installer.StaticPodInstallState, error) {
+		switch installerPod.Status.Phase {
+		case corev1.PodSucceeded:
+			// installer succeeded, wait for running static pod
+			pod, err := podlister.Pods(ns).Get("openshift-kube-apiserver-" + nodeName)
```
deads2k: They are likely to all be the same but for name and namespace, right?  We already pass the namespace, we could pass a name prefix.

pkg/operator/starter.go
```diff
@@ -104,6 +112,37 @@ func RunOperator(clientConfig *rest.Config, stopCh <-chan struct{}) error {
...
+			if err != nil {
+				if errors.IsNotFound(err) {
+					return installer.StaticPodInstallStateRunning, nil
```
deads2k: This reads more like Pending to me.  The installer is done, but we haven't received any status yet.

pkg/operator/starter.go
```diff
@@ -104,6 +112,37 @@ func RunOperator(clientConfig *rest.Config, stopCh <-chan struct{}) error {
...
+				}
+				return installer.StaticPodInstallStateRunning, err
+			}
+			if pod.Labels[deploymentIDLabel] != strconv.Itoa(int(deploymentID)) {
+				return installer.StaticPodInstallStateRunning, nil
```
deads2k: yeah, pending

pkg/operator/starter.go
deads2k: nit: `staticPod` or `targetPod` or `operandPod` would help me reading below

pkg/operator/starter.go
```diff
@@ -104,6 +112,37 @@ func RunOperator(clientConfig *rest.Config, stopCh <-chan struct{}) error {
...
+			switch pod.Status.Phase {
+			case corev1.PodRunning, corev1.PodSucceeded:
+				return installer.StaticPodInstallStateSuccess, nil
+			case corev1.PodFailed:
+				return installer.StaticPodInstallStateFailed, nil
+			}
+
+			return installer.StaticPodInstallStateRunning, nil
+		case corev1.PodFailed:
+			return installer.StaticPodInstallStateFailed, nil
```
deads2k: I wouldn't expect this case to do that

pkg/operator/starter.go
```diff
@@ -104,6 +112,37 @@ func RunOperator(clientConfig *rest.Config, stopCh <-chan struct{}) error {
 	return fmt.Errorf("stopped")
 }
 
+func StaticPodIsRunningFunc(podlister corev1listers.PodLister, ns string) installer.StaticPodInstallStateFunc {
+	return func(installerPod *corev1.Pod, deploymentID int32, nodeName string) (installer.StaticPodInstallState, error) {
+		switch installerPod.Status.Phase {
```
deads2k: It seems weird to pass the installerPod here.  I'm not seeing how that affects whether a particular deployment's status

vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -47,6 +47,36 @@ type InstallerController struct {
...
+	// StaticPodInstallStateRunning means that the static pod installation of a node is running.
+	StaticPodInstallStateRunning = StaticPodInstallState(iota)
+	// StaticPodInstallStateSuccess means that the static pod installation of a node has succeeded.
+	StaticPodInstallStateSuccess
+	// StaticPodInstallStateSuccess means that the static pod installation of a node has failed.
+	StaticPodInstallStateFailed
+)
+
+// StaticPodInstallStateFunc derives the status of the installer.
+type StaticPodInstallStateFunc func(installerPod *corev1.Pod, deploymentID int32, nodeName string) (StaticPodInstallState, error)
```
deads2k: Given a nodeName, we could return the current deployment and it's state, right?  That seems generally useful to building levels and would also satisfy the need here.  You're waiting for it to return the "correct" deploymentID and the running state.

vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -167,13 +198,17 @@ func (c *InstallerController) newNodeStateForInstallInProgress(currNodeState *op
 	if err != nil {
 		return nil, err
 	}
-	switch installerPod.Status.Phase {
-	case corev1.PodSucceeded:
```
deads2k: Can we nest below here based on the deploymentID and state returned from the `StaticPodInstallStateFunc`?  I think we probably want different status (or at least log messages) about each of the 6 interesting states.

pkg/operator/starter.go
```diff
@@ -104,6 +112,37 @@ func RunOperator(clientConfig *rest.Config, stopCh <-chan struct{}) error {
 	return fmt.Errorf("stopped")
 }
 
+func StaticPodIsRunningFunc(podlister corev1listers.PodLister, ns string) installer.StaticPodInstallStateFunc {
```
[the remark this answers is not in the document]
sttts: A pity go does not use [currying](https://en.wikipedia.org/wiki/Currying.)

bindata/v3.11.0/kube-apiserver/pod.yaml
```diff
@@ -32,7 +32,7 @@ spec:
       httpGet:
         scheme: HTTPS
         port: 6443
-        path: healthz/ready
+        path: healthz
```
deads2k: pull this commit into a another PR and merge it.

vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -47,16 +49,66 @@ type InstallerController struct {
...
+	StaticPodStateReady
+	// StaticPodStateFailed means that the static pod installation of a node has failed.
+	StaticPodStateFailed
+)
+
+const deploymentIDLabel = "deployment-id"
+
+// StaticPodStateFunc derives the status of the installed static pod. It returns the state, an array of error strings
+// in case of StaticPodStateFailed and an error in case the state couldn't be checked.
+type StaticPodStateFunc func(deploymentID int32, nodeName string) (StaticPodState, []string, error)
```
deads2k: seems like deploymentID can be an output,not an input.  There is only one possible per node.

vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -83,18 +140,34 @@ func NewInstallerController(
 func (c *InstallerController) createInstallerController(operatorSpec *operatorv1alpha1.OperatorSpec, originalOperatorStatus *operatorv1alpha1.StaticPodOperatorStatus, resourceVersion string) (bool, error) {
```
deads2k: wow, I chose a really bad name

vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -112,7 +185,14 @@ func (c *InstallerController) createInstallerController(operatorSpec *operatorv1
 			} else {
 				glog.V(2).Infof("%q is in transition to %d, but has not made progress", currNodeState.NodeName, currNodeState.TargetDeploymentGeneration)
 			}
-			break
+
+			if !pendingNewDeployment {
+				break
+			}
+
+			// fall through to start the next deployment
```
deads2k: Falling through means you may update status again.  This means that you need to have the new operatorstatus as returned by `UpdateStatus`

vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -83,18 +140,34 @@ func NewInstallerController(
...
 		if i > 0 {
-			prevNodeState = &operatorStatus.NodeStatuses[i-1]
+			prevNodeState = &operatorStatus.NodeStatuses[prev]
 		}
 
 		// if we are in a transition, check to see if our installer pod completed
 		if currNodeState.TargetDeploymentGeneration > currNodeState.CurrentDeploymentGeneration {
-			// TODO check to see if our installer pod completed.  Success or failure there indicates whether we should be failed.
-			newCurrNodeState, err := c.newNodeStateForInstallInProgress(currNodeState)
+			pendingNewDeployment := operatorStatus.LatestAvailableDeploymentGeneration > currNodeState.TargetDeploymentGeneration
```
deads2k: This change bothers me.  Can you detect this a line earlier, avoid this branch, and just enter the "normal" new deployment flow?

bindata/v3.11.0/kube-apiserver/pod.yaml
sttts: https://github.com/openshift/cluster-kube-apiserver-operator/pull/118

vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/installer/installer_controller.go
sttts: We have to know which one is the target. There can be an old static pod. Hence, deployment id must be an input. Will rename it.

sttts: `targetDeploymentID`

vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/installer/installer_controller.go
sttts: Tried that. We still want to wait for the installer to finish, otherwise they might be racing.

vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/installer/installer_controller.go
sttts: Compare the updated logic:
- the "target > current" if clause logic is as before
- no fallthrough anymore, but a break
- on resync the first node is found that is not ready to its reported current deployment id. It's used as the new start of the pending deployment id.

=> the "operator dies" TODO is resolved as the operator can pick up an unready static pod at any time.
=> the new deployment flow is uniform again.

vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -47,16 +49,66 @@ type InstallerController struct {
 
 	// installerPodImageFn returns the image name for the installer pod
 	installerPodImageFn func() string
+
+	// installStateFunc derives the static pod install state of a node
```
mfojtik: nit: comment

vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -47,16 +49,66 @@ type InstallerController struct {
...
+const (
+	// staticPodStatePending means that the installed static pod is not up yet.
+	StaticPodStatePending = StaticPodState(iota)
+	// StaticPodStateReady means that the installed static pod is ready.
+	StaticPodStateReady
+	// StaticPodStateFailed means that the static pod installation of a node has failed.
+	StaticPodStateFailed
+)
+
+const deploymentIDLabel = "deployment-id"
```
mfojtik: nit: did we ended up calling this "revision" ? `deployment-revision` ?

vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -47,16 +49,66 @@ type InstallerController struct {
...
+
+// StaticPodStateFunc derives the status of the installed static pod. It returns the state, an array of error strings
+// in case of StaticPodStateFailed and an error in case the state couldn't be checked.
+type StaticPodStateFunc func(targetDeploymentID int32, nodeName string) (StaticPodState, []string, error)
```
mfojtik: why we need this nested function stuff?

vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -47,16 +49,66 @@ type InstallerController struct {
...
+
+// NewDefaultStaticPodStateFunc returns a static pod state func which checks the pods readiness via the pod condition.
+func NewDefaultStaticPodStateFunc(podlister corev1listers.PodLister, ns string, staticPodNamePrefix string) StaticPodStateFunc {
+	return func(deploymentID int32, nodeName string) (StaticPodState, []string, error) {
+		pod, err := podlister.Pods(ns).Get(staticPodNamePrefix + "-" + nodeName)
```
mfojtik: nit: convert `staticPodNamePrefix + "-" + nodeName` to a `getStaticPodName` function

vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -47,16 +49,66 @@ type InstallerController struct {
...
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				return StaticPodStatePending, nil, nil
+			}
+			return StaticPodStatePending, nil, err
+		}
+		if pod.Labels[deploymentIDLabel] != strconv.Itoa(int(deploymentID)) {
```
mfojtik: if we get a pod which does not have the matching revision (deployment id) what does it mean for this controller? how can that happen? do we want to ignore that pod and wait for the deployment ID passed to this function?

vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -83,18 +140,51 @@ func NewInstallerController(
 func (c *InstallerController) createInstallerController(operatorSpec *operatorv1alpha1.OperatorSpec, originalOperatorStatus *operatorv1alpha1.StaticPodOperatorStatus, resourceVersion string) (bool, error) {
 	operatorStatus := originalOperatorStatus.DeepCopy()
 
+	if len(operatorStatus.NodeStatuses) == 0 {
+		return false, nil
+	}
+
+	// find upgrading node as this will be the first to start new deployment (to minimize number of down nodes)
```
mfojtik: @sttts i would probably put this logic into separate function? like `determineNodeToStart()` or something... I think this logic is very important for the operator?

vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/installer/installer_controller.go
sttts: why nested?

vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/installer/installer_controller.go
sttts: It happens all the time during deployment. Until the kubelet has updated the static pod, the controller sees the old deployment id.

sttts: This `deploymentID` should be named `targetDeploymentID`. Renamed that now. Then it's clearer I think.

vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/installer/installer_controller.go
deads2k: > why we need this nested function stuff?

I think he means the closure.  Same distaste as me I suspect :)

sttts: I am a fan of closures. Luckily Golang has them.

vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/installer/installer_controller.go
sttts: done

vendor/github.com/openshift/library-go/pkg/operator/staticpod/controller/installer/installer_controller.go
sttts: still seeing deployment-id in the code

sttts:
Replaced by some other library-go bump.

PR #114, opened 2018-11-11
Add a RHEL7 dockerfile and standarize format

smarterclayton:


smarterclayton:
the new install is apparently unstable right now, which is why this is
failing.  commented in slack.

PR #116, opened 2018-11-13
handle case of missing image.config.openshift.io type definition

deads2k:
When the entire CRD is missing, we can still synchronize most of the config and simply use the last known value for the image config itself.  We need to observe most of the config before this loop starts.

/assign @sanchezl @juanvallejo @sttts 

@sanchezl can you manually verify that if no CRD is present when this process starts, adding the CRD later and setting a value is properly observed?  If not, please add a unit test showing it breaking and find a way to fix it.

@bparees note the structural change

### DIFF
--- a/pkg/operator/observe_config.go
+++ b/pkg/operator/observe_config.go
@@ -40,13 +40,15 @@ type Listers struct {
 	imageConfigLister configlistersv1.ImageLister
 	endpointsLister   corelistersv1.EndpointsLister
 	configmapLister   corelistersv1.ConfigMapLister
+
+	imageConfigSynced cache.InformerSynced
 }
 
 // observeConfigFunc observes configuration and returns the observedConfig. This function should not return an
 // observedConfig that would cause the service being managed by the operator to crash. For example, if a required
 // configuration key cannot be observed, consider reusing the configuration key's previous value. Errors that occur
 // while attempting to generate the observedConfig should be returned in the errs slice.
-type observeConfigFunc func(listers Listers, currentConfig map[string]interface{}) (observedConfig map[string]interface{}, errs []error)
+type observeConfigFunc func(listers Listers, existingConfig map[string]interface{}) (observedConfig map[string]interface{}, errs []error)
 
 type ConfigObserver struct {
 	operatorConfigClient operatorconfigclientv1alpha1.KubeapiserverV1alpha1Interface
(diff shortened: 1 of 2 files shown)


pkg/operator/observe_config.go
```diff
@@ -280,28 +280,34 @@ func observeRestrictedCIDRs(listers Listers, currentConfig map[string]interface{
 
 // observeInternalRegistryHostname reads the internal registry hostname from the cluster configuration as provided by
 // the registry operator.
-func observeInternalRegistryHostname(listers Listers, currentConfig map[string]interface{}) (observedConfig map[string]interface{}, errs []error) {
-	observedConfig = map[string]interface{}{}
+func observeInternalRegistryHostname(listers Listers, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
```
deads2k: @sanchezl we don't use named return values.  Please update the others.

pkg/operator/observe_config.go
```diff
@@ -280,28 +280,34 @@ func observeRestrictedCIDRs(listers Listers, currentConfig map[string]interface{
...
+	errs := []error{}
+	prevObservedConfig := map[string]interface{}{}
+
 	internalRegistryHostnamePath := []string{"imagePolicyConfig", "internalRegistryHostname"}
-	currentInternalRegistryHostname, _, _ := unstructured.NestedStringSlice(currentConfig, internalRegistryHostnamePath...)
+	if currentInternalRegistryHostname, _, _ := unstructured.NestedString(existingConfig, internalRegistryHostnamePath...); len(currentInternalRegistryHostname) > 0 {
+		unstructured.SetNestedField(prevObservedConfig, currentInternalRegistryHostname, internalRegistryHostnamePath...)
```
deads2k: @sanchezl this structure makes it a little easier to follow the logic of "return the previous value"

pkg/operator/observe_config.go
```diff
@@ -89,14 +88,15 @@ func NewConfigObserver(
 			imageConfigLister: configInformer.Config().V1().Images().Lister(),
 			endpointsLister:   kubeInformersForKubeSystemNamespace.Core().V1().Endpoints().Lister(),
 			configmapLister:   kubeInformersForKubeSystemNamespace.Core().V1().ConfigMaps().Lister(),
+			imageConfigSynced: configInformer.Config().V1().Images().Informer().HasSynced,
+		},
+		cachesSynced: []cache.InformerSynced{
```
deads2k: notice that the optional types aren't listed here.

bparees:
lgtm.

pkg/operator/observe_config.go
sanchezl: :+1:

PR #119, opened 2018-11-15
Force config change when spec.ForceRedeploymentReason changes

sttts:
Or did I miss any other existing mechanism to redeploy?


deads2k:
> Or did I miss any other existing mechanism to redeploy?

https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/pkg/operator/target_config_reconciler_v311_00.go#L125

deads2k:
linked the mechanism above.  I think eventually it ends up on the anchor from @damemi 

/close

PR #121, opened 2018-11-15
clear failing status previously set during config reconcile

sanchezl:
Resolves JIRA Issue [MSTR-219](https://jira.coreos.com/browse/MSTR-219).

### DIFF
--- a/pkg/operator/target_config_reconciler.go
+++ b/pkg/operator/target_config_reconciler.go
@@ -24,9 +24,10 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


deads2k:
What do you think about having every `<ControlLoop>Failing` condition aggregated into the `Failing` condition during our update loop instead of trying to avoid conflicts wtih each other.

deads2k:
make https://github.com/openshift/library-go/blob/master/pkg/operator/status/controller.go handle unioning multiple *Failing conditions.

PR #124, opened 2018-11-15
generic config observer

deads2k:
This is an attempt to generify the logic in the config observer so we can stop copy/pasting everywhere.

/assign @sanchezl @sttts 

/hold
holding for test

I think the tests will be relatively portable, but don't merge until I have them.


deads2k:
/hold cancel

tests added

@sanchezl lets close this off.  then promote to library-go, then fix the others.

PR #125, opened 2018-11-16
metrics: add svc for metrics and wire serving certs to deployment

mfojtik:
@deads2k @sttts we have the secret volume that we don't mount into container... this is fixing this in addition of metrics service that should be used to serve the operator metrics.

pre-requirement: https://github.com/openshift/library-go/pull/88
/hold

### DIFF
--- a/manifests/0000_10_cluster-kube-apiserver-operator_06_deployment.yaml
+++ b/manifests/0000_10_cluster-kube-apiserver-operator_06_deployment.yaml
@@ -28,6 +28,8 @@ spec:
         volumeMounts:
         - mountPath: /var/run/configmaps/config
           name: config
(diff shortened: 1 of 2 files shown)


manifests/0000_10_cluster-kube-apiserver-operator_06_deployment.yaml
```diff
@@ -28,6 +28,8 @@ spec:
         volumeMounts:
         - mountPath: /var/run/configmaps/config
           name: config
+        - mountPath: /var/run/secrets/serving-certs
+          name: serving-cert
```
deads2k: remove the default modes below

deads2k:
just a nit.

lgtm otherwise

mfojtik:
@deads2k nit fixed.

PR #131, opened 2018-11-20
Use openshift-kube-apiserve as container name for easier debugging

sttts:
"crictl ps -a" does not show pod names. So during disaster recovery full names for the containers is helpful.


derekwaynecarr:
‘crictl pods’ should show names, but this change is still good for me.

sttts:
> ‘crictl pods’ should show names

good to know :+1:

PR #136, opened 2018-11-21
fix serving certs path to match file observer

mfojtik:


mfojtik:
self-lgtm, just matching the path with file observer.

deads2k:
flakes, oh joy.  @soltysh get ready for more upstreams.

/retest

PR #137, opened 2018-11-21
silence image.config.openshift.io not found messages

mfojtik:


pkg/operator/configobservation/images/observe_images.go
```diff
@@ -30,14 +30,16 @@ func ObserveInternalRegistryHostname(genericListers configobserver.Listers, exis
...
+		glog.V(4).Infof("image.config.openshift.io/cluster: not found")
 		return observedConfig, errs
 	}
 	if err != nil {
+		glog.Warningf("error getting image.config.openshift.io/cluster: %v", err)
 		return prevObservedConfig, errs
 	}
 	internalRegistryHostName := configImage.Status.InternalRegistryHostname
 	if len(internalRegistryHostName) > 0 {
+		glog.V(4).Infof("setting internal registry hostname to: %q", internalRegistryHostName)
```
sttts: shouldn't this be differential? We don't want an info (even verbosity 4) if nothing changed.

mfojtik: yeah, I think we want info when things change... I can compare current value (we should do that anyway)

pkg/operator/configobservation/images/observe_images.go
```diff
@@ -30,14 +31,18 @@ func ObserveInternalRegistryHostname(genericListers configobserver.Listers, exis
 	observedConfig := map[string]interface{}{}
 	configImage, err := listers.ImageConfigLister.Get("cluster")
 	if errors.IsNotFound(err) {
-		glog.Warningf("image.config.openshift.io/cluster: not found")
+		glog.V(4).Infof("image.config.openshift.io/cluster: not found")
 		return observedConfig, errs
 	}
 	if err != nil {
+		glog.Warningf("error getting image.config.openshift.io/cluster: %v", err)
```
deads2k: errors here are reported in status, why log them?

deads2k: oh, I see.  missing the append.  That's what should be fixed there.

pkg/operator/configobservation/images/observe_images.go
```diff
@@ -14,11 +14,12 @@ import (
 // the registry operator.
 func ObserveInternalRegistryHostname(genericListers configobserver.Listers, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
 	listers := genericListers.(configobservation.Listers)
-	errs := []error{}
+	var errs []error
 	prevObservedConfig := map[string]interface{}{}
 
 	internalRegistryHostnamePath := []string{"imagePolicyConfig", "internalRegistryHostname"}
-	if currentInternalRegistryHostname, _, _ := unstructured.NestedString(existingConfig, internalRegistryHostnamePath...); len(currentInternalRegistryHostname) > 0 {
+	currentInternalRegistryHostname, _, _ := unstructured.NestedString(existingConfig, internalRegistryHostnamePath...)
```
deads2k: let's be consistent.  I think this previous (my fault on the original name)

pkg/operator/configobservation/images/observe_images.go
```diff
@@ -30,14 +31,18 @@ func ObserveInternalRegistryHostname(genericListers configobserver.Listers, exis
...
 		return observedConfig, errs
 	}
 	if err != nil {
+		glog.Warningf("error getting image.config.openshift.io/cluster: %v", err)
 		return prevObservedConfig, errs
 	}
 	internalRegistryHostName := configImage.Status.InternalRegistryHostname
 	if len(internalRegistryHostName) > 0 {
+		if currentInternalRegistryHostname != internalRegistryHostName {
+			glog.V(4).Infof("setting internal registry hostname to: %q", internalRegistryHostName)
```
deads2k: I expect this to be done in the library-go method.  Perhaps you want a better diff here: https://github.com/openshift/library-go/blob/master/pkg/operator/configobserver/config_observer_controller.go#L107 ?

deads2k:
This code exists in two or three repos

bparees:
> @bparees @deads2k I can see this message even after >10m of cluster startup. Is that expected? Or we observing wrong field?

you'll see it until you have a valid imageconfig object.  If nothing else, you'll get one of those when the registry comes up.  I doubt you're standing up a registry because you don't care and (today) one doesn't come up automatically.

PR #145, opened 2018-11-28
Update event recorder and wire events into resource apply

mfojtik:


mfojtik:
Added https://github.com/openshift/library-go/pull/118 and https://github.com/openshift/library-go/pull/118 to see what the output will be.

/hold

(until i get real bump)

Issue #143, opened 2018-11-26
servicesSubnet is hardcoded to 10.3.0.0


vrutkovs:
`servicesSubnet` is set to 10.3.0.0 in master's kube API server - https://github.com/openshift/cluster-kube-apiserver-operator/blob/4f197523d36c70e06e29fc758452e9638fd20d53/bindata/v3.11.0/kube-apiserver/defaultconfig.yaml#L89
Installer has `serviceCIDR` set to `172.30.0.0/16` so most namespaces have `Cluster IP 172.30.220.6 is not within the service CIDR 10.3.0.0/16; please recreate service` event

wking:
The installer default changed in openshift/installer@c6c9d832e94d0c677f3d8e6170e4306dd00df23a (openshift/installer#712).  Ideally operators should be pulling this value from [the network config][1].

[1]: https://github.com/openshift/installer/blob/v0.4.0/pkg/asset/manifests/network.go#L107-L118

wking:
Example of pulling thihis from the network config:

```console
$ oc get -o jsonpath="{.spec.serviceNetwork}{'\n'}" networkconfig default
172.30.0.0/16
```

smarterclayton:
Operators are not manually setting this.  Services created by the CVO are getting assigned the 172.x block, and then it looks like we change what the value is.  The CVO service for instance has no hardcoded IP and is getting a 172.

vrutkovs:
This is breaking openshift API server on GCP setup:
```
E1127 13:27:12.158406       1 controller.go:111] loading OpenAPI spec for "v1.image.openshift.io" failed with: failed to retrieve openAPI spec, http error: ResponseCode: 503, Body: service unavailable

E1127 13:27:50.036693       1 available_controller.go:327] v1.template.openshift.io failed with: Get https://172.30.222.191:443: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```
```
# oc get svc -n openshift-apiserver
NAME      TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)   AGE
api       ClusterIP   172.30.222.191   <none>        443/TCP   47m
```
```
# oc get events -n openshift-apiserver
1m          43m          16        api.156afbe83ec48845               Service                                            Warning   ClusterIPOutOfRange      ipallocator-repair-controller       Cluster IP 172.30.222.191 is not within the service CIDR 10.3.0.0/16; please recreate service
[...]

PR #146, opened 2018-11-28
network-observer: set serviceSubnet in apiserver config

sttts:
Fixes https://github.com/openshift/installer/pull/712


wking:
Fixes #143 too.

The subject here has a "netowrk" -> "network" typo.

PR #147, opened 2018-11-29
operator: wait for spec.observedConfig to be filled before installing

sttts:
Installing an initial apiserver with broken, incomplete might fail the pivot from bootstrap to operator mode. We should therefore wait until the config has been observed once before installing any static pod to the masters.

This might reduce flakes.

### DIFF
--- a/pkg/operator/target_config_reconciler.go
+++ b/pkg/operator/target_config_reconciler.go
@@ -91,6 +91,12 @@ func (c TargetConfigReconciler) sync() error {
 		return nil
 	}
 
+	// block until config is obvserved
+	if len(operatorConfig.Spec.ObservedConfig.Raw) == 0 {
+		glog.Info("Waiting for observed configuration to be available")
+		return nil
+	}
+
 	requeue, err := createTargetConfigReconciler_v311_00_to_latest(c, c.eventRecorder, operatorConfig)
 	if requeue && err == nil {
 		return fmt.Errorf("synthetic requeue request")


mfojtik:
/lgtm

https://github.com/openshift/cluster-kube-apiserver-operator/pull/145 is longer in queue.

Lets see which one will land first ;-)

sttts:
> Lets see which one will land first ;-)

🍿

deads2k:
I'm not a fan of this.  I think we should instead make sure that when kube-apiservers gracefully shutdown, that they remove themselves from endpoints.

/hold

abhinavdahiya:
Not sure if this is valid info here, but the bootstrap node doesn't run any k8s networking so i would assume nothing can talk to the bootstrap apiserver on service ip.

_disregard if not useful :innocent:_

deads2k:
> That's not enough. E.g. we deploy a wrong service subnet and we have services which are broken. We cannot just deploy a default config and accept a minute of wrong behaviour which does not repair itself.

ok. Requiring certain fields to be set seems valid, but just requiring non-empty isn't enough.

You can release this one, but it will need to be more selective moving forward.

sttts:
Rebased. Reapplying label.

sttts:
Removing hold. We can and should iterate on this, this is fine. But perfect is the enemy of good (quote @deads2k), and this breaks clusters.

deads2k:
@sttts this should be an event along with a status indicating we aren't healthy and it should be looking for specific stanzas.

PR #149, opened 2018-11-29
make static pods node critical priority

mfojtik:
/cc @aveshagarwal 
/cc @deads2k

### DIFF
--- a/bindata/v3.11.0/kube-apiserver/installer-pod.yaml
+++ b/bindata/v3.11.0/kube-apiserver/installer-pod.yaml
@@ -7,6 +7,7 @@ metadata:
     app: installer
 spec:
   serviceAccountName: installer-sa
+  priorityClassName: system-node-critical
   containers:
   - name: apiserver
     image: ${IMAGE}
--- a/bindata/v3.11.0/kube-apiserver/pod.yaml
+++ b/bindata/v3.11.0/kube-apiserver/pod.yaml
@@ -36,6 +36,9 @@ spec:
       initialDelaySeconds: 10
       timeoutSeconds: 10
   hostNetwork: true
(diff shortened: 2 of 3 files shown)


aveshagarwal:
/lgtm

@mfojtik need to assign priority class to the operator too.

deads2k:
This change seems weird for a static pod.  Why are we doing it?  The scheduler isn't involved in deciding whether and how to place or evict this.  Seems like the scheduler should be aware of static pods and behave as expected.

bindata/v3.11.0/kube-apiserver/installer-pod.yaml
```diff
@@ -7,6 +7,7 @@ metadata:
     app: installer
 spec:
   serviceAccountName: installer-sa
+  priorityClassName: system-node-critical
```
deads2k: This one seems ok. The static pod change looks wrong.

aveshagarwal: The priority to static pod is given to avoid eviction by kubelet in resource crunch scenarios:

https://github.com/kubernetes/kubernetes/blob/a3ccea9d8743f2ff82e41b6c2af6dc2c41dc7b10/pkg/kubelet/eviction/eviction_manager.go#L543-#L549

https://github.com/kubernetes/kubernetes/blob/25cbd1c753fbed15e47c6331ddf043cc00111792/pkg/kubelet/types/pod_update.go#L148

https://github.com/kubernetes/kubernetes/blob/25cbd1c753fbed15e47c6331ddf043cc00111792/pkg/kubelet/types/pod_update.go#L194

deads2k:
If we need to it seems like resource request could be useful too.

/hold cancel

PR #156, opened 2018-12-04
config observers: add OWNER files with config owners

sttts:


pkg/operator/configobservation/etcd/OWNERS
```diff
@@ -0,0 +1,6 @@
+reviewers:
+  - abhinavdahiya
+  - wking
```
wking: hmm, why am I here?  I'm not even sure what this directory is about...

wking: Filed #674 removing myself.  I guess we can talk about whether I should be a reviewer/approver in that PR.

PR #161, opened 2018-12-05
add memory request to kube-apiserver and operator pods

sjenning:
One in a series of PRs to set memory requests for all control plane and high memory use components.

We are encountering master instability (https://github.com/openshift/installer/pull/408 https://github.com/openshift/installer/pull/785) as the number of components increase because many components run `BestEffort` giving the scheduler no information on pod resource usage.

On a steady state empty cluster, `kube-apiserver` uses 980Mi.

{pod_name="openshift-kube-apiserver-dev-master-0"} | 979714048

Setting request to `1Gi`.

This component uses the most memory and CPU of any component.  Possible target for future optimization.

@deads2k @sttts  @derekwaynecarr


sttts:
Better than no requests at all. We can increase if we see need.

/lgtm
/approve

PR #165, opened 2018-12-07
bindata: set target-ram-mb to 1Gi in default config

sjenning:
In https://github.com/openshift/cluster-kube-apiserver-operator/pull/161, the memory request was set to `1Gi`.  However, the apiserver cache size scales with the amount of RAM on the node by default.  Through experimentation, is seems to be 30% of RAM on the node, but I can't find the code that actually does this.

Either way, there is a flag `target-ram-mb` to manually control it.  In order for the memory request for the kube-apiserver to be valid for all deployments, we need to set this.

The upstream heuristic for watch cache sizing is 60MB/node, which is pretty generous.  The means that a 1Gi `target-ram-mb` should be valid for a ~20 node cluster according to the upstream heuristic and probably even 3-4x that in reality.  I figure this is our 95% case.

@deads2k @sttts @smarterclayton @derekwaynecarr

### DIFF
--- a/bindata/v3.11.0/kube-apiserver/defaultconfig.yaml
+++ b/bindata/v3.11.0/kube-apiserver/defaultconfig.yaml
@@ -25,6 +25,8 @@ apiServerArguments:
   - etcd3
   storage-media-type:
   - application/vnd.kubernetes.protobuf
+  target-ram-mb:
+  - "1024"
 auditConfig:
   auditFilePath: ""
   enabled: false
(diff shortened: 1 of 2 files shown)


derekwaynecarr:
is it possible to scale as a factor of downward api resource limits?

derekwaynecarr:
i mean could we make apiserver itself understand the downward api to apply a default?

either way, @deads2k and i discussed this, and i think its fine to move this to global config for control plane.  in a future world, if left unset, ideally we find a way to auto tune this tunable.

derekwaynecarr:
i am fine with this change now, we should plumb it from global config in a follow-on.

will defer to @deads2k for review.

PR #166, opened 2018-12-07
fix status

deads2k:

PR #167, opened 2018-12-11
Revert "add clusteroperator resource to CVO payload"

deads2k:
Reverts openshift/cluster-kube-apiserver-operator#163

A different option.

@smarterclayton


deads2k:
@abhinavdahiya @wking

deads2k:
@crawford an option

abhinavdahiya:
So the PR that is fixing this problem is not working yet https://github.com/openshift/cluster-kube-apiserver-operator/pull/166#issuecomment-446295991

Also a lot of people are failing to install due to this one example: https://github.com/openshift/installer/issues/857

So lets go ahead with revert this and add this back in #166 . @deads2k :innocent: 

/lgtm

PR #171, opened 2018-12-17
add service monitor to CVO payload

deads2k:
Adds the service monitor to the CVO payload in a position that should come last.  These work locally for me

/assign @mfojtik


manifests/0000_90_cluster-kube-apiserver-operator_01_prometheusrole.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  # TODO this should be a clusterrole
```
deads2k: @openshift/sig-auth @openshift/sig-monitoring can you two confirm.  This looks like a cluster role that needs to be bound locally.

manifests/0000_90_cluster-kube-apiserver-operator_03_servicemonitor.yaml
```diff
@@ -0,0 +1,27 @@
...
+    interval: 30s
+    metricRelabelings:
+    - action: drop
+      regex: etcd_(debugging|disk|request|server).*
+      sourceLabels:
+      - __name__
+    port: https
+    scheme: https
+    tlsConfig:
+      # TODO where do you mount the service-ca.crt?
```
deads2k: @openshift/sig-monitoring to monitor our operators securely, you need to have access to the service-ca bundle from @openshift/sig-auth

deads2k:
```
level=error msg="Error: Error applying plan:"
level=error
level=error msg="1 error occurred:"
level=error msg="\t* module.vpc.aws_vpc_endpoint.s3: 1 error occurred:"
level=error msg="\t* aws_vpc_endpoint.s3: Error creating VPC Endpoint: VpcEndpointLimitExceeded: The maximum number of VPC endpoints has been reached."
level=error msg="\tstatus code: 400, request id: 36689b63-fed3-4aa3-a6f0-c6c1cb78d5bb"
level=error
```

wasn't me!

/test e2e-aws

manifests/0000_90_cluster-kube-apiserver-operator_03_servicemonitor.yaml
s-urbaniak: @deads2k just rechecked, we are already mounting the service-ca bundle into prometheus, example config is here: https://github.com/openshift/cluster-monitoring-operator/blob/master/assets/kube-state-metrics/service-monitor.yaml#L16-L18

manifests/0000_90_cluster-kube-apiserver-operator_01_prometheusrole.yaml
s-urbaniak: hmm, I might be missing something, but I don't see the necessity for a binding here, cc @squat @metalmatze

PR #177, opened 2019-01-02
add additional certificate management

deads2k:
Adds cert rotation for the other certificates that the kube-apiserver needs to manage.

/assign @mfojtik


pkg/operator/starter.go
```diff
@@ -5,7 +5,8 @@ import (
 	"os"
 	"time"
 
-	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/aggregatorproxyrotation"
+	"github.com/openshift/library-go/pkg/operator/certrotation"
```
mfojtik: nit: imports

mfojtik:
one nit, LGTM otherwise

pkg/operator/starter.go
deads2k: fixed

PR #179, opened 2019-01-03
Add revision status configmap to bindata

damemi:
Adds a status configmap to the operator to be used by the revision controller. This needs to exist already or the revision controller will return an error when it tries to copy it.

Depended on by https://github.com/openshift/library-go/pull/157


sttts:
Could and shouldn't this be a CRD?

damemi:
@deads2k PTAL along with https://github.com/openshift/library-go/pull/157 and https://github.com/openshift/library-go/pull/159 to build on this

deads2k:
inline what you need since it's basically empty

/close

PR #181, opened 2019-01-04
add generic, configurable config sync controller

deads2k:
This needs unit tests, but this is where my head is at.

I hit an unfortunate operator bug related to required secrets, so I'm going to sort that out next.

@mrogers950 @mfojtik

### DIFF
--- a/pkg/operator/resourcesynccontroller/interfaces.go
+++ b/pkg/operator/resourcesynccontroller/interfaces.go
@@ -1,0 +1,19 @@
+package resourcesynccontroller
+
+// ResourceLocation describes coordinates for a resource to be synced
+type ResourceLocation struct {
+	Namespace string
+	Name      string
+}
+
+var emptyResourceLocation = ResourceLocation{}
+
+// ResourceSyncer allows changes to syncing rules by this controller
+type ResourceSyncer interface {
+	// SyncConfigMap indicates that a configmap should be copied from the source to the destination.  It will also
+	// mirror a deletion from the source.  If the source is a zero object the destination will be deleted.
+	SyncConfigMap(destination, source ResourceLocation) error
+	// SyncSecret indicates that a secret should be copied from the source to the destination.  It will also
+	// mirror a deletion from the source.  If the source is a zero object the destination will be deleted.
+	SyncSecret(destination, source ResourceLocation) error
+}
--- a/pkg/operator/resourcesynccontroller/resourcesync_controller.go
+++ b/pkg/operator/resourcesynccontroller/resourcesync_controller.go
@@ -1,0 +1,237 @@
+package resourcesynccontroller
+
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/golang/glog"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	"github.com/openshift/library-go/pkg/operator/staticpod/controller/common"
+)
+
+const (
+	operatorStatusResourceSyncControllerFailing = "ResourceSyncControllerFailing"
+	controllerWorkQueueKey                      = "key"
+)
+
+// ResourceSyncController is a controller that will copy source configmaps and secrets to their destinations.
+// It will also mirror deletions by deleting destinations.
+type ResourceSyncController struct {
+	// syncRuleLock is used to ensure we avoid races on changes to syncing rules
+	syncRuleLock sync.RWMutex
+	// configMapSyncRules is a map from destination location to source location
+	configMapSyncRules map[ResourceLocation]ResourceLocation
+	// secretSyncRules is a map from destination location to source location
+	secretSyncRules map[ResourceLocation]ResourceLocation
+
+	// knownNamespaces is the list of namespaces we are watching.
+	knownNamespaces sets.String
+
+	preRunCachesSynced []cache.InformerSynced
+
+	// queue only ever has one item, but it has nice error handling backoff/retry semantics
+	queue workqueue.RateLimitingInterface
+
+	kubeClient           kubernetes.Interface
+	operatorConfigClient common.OperatorClient
+	eventRecorder        events.Recorder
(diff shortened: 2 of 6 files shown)


pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -0,0 +1,221 @@
...
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	"github.com/openshift/library-go/pkg/operator/staticpod/controller/common"
+)
+
+const (
+	operatorStatusResourceSyncControllerFailing = "ResourceSyncControllerFailing"
+	controllerWorkQueueKey                      = "key"
+)
+
+type ResourceSyncController struct {
```
sttts: go doc. What does this controller do? Please document controllers at least high level. It's very hard for new people to read otherwise.

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -0,0 +1,221 @@
...
+	queue workqueue.RateLimitingInterface
+
+	kubeClient           kubernetes.Interface
+	operatorConfigClient common.OperatorClient
+	eventRecorder        events.Recorder
+}
+
+var _ ResourceSyncer = &ResourceSyncController{}
+
+// NewResourceSyncController creates a new backing resource controller.
```
sttts: update comment

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -0,0 +1,221 @@
...
+func NewResourceSyncController(
+	operatorConfigClient common.OperatorClient,
+	// kubeInformersForNamespaces is a map of namespaces to their informers
```
sttts: no need for this comment. The var name says it all.

pkg/operator/resourcesynccontroller/interfaces.go
```diff
@@ -0,0 +1,13 @@
+package resourcesynccontroller
+
+// ResourceLocation describes coordinates for a resource to be synced
+type ResourceLocation struct {
+	Namespace string
+	Name      string
+}
+
+// ResourceSyncer allows changes to syncing rules by this controller
```
sttts: Why not SyncConfigMap and SyncSecret?

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -0,0 +1,221 @@
...
+
+const (
+	operatorStatusResourceSyncControllerFailing = "ResourceSyncControllerFailing"
+	controllerWorkQueueKey                      = "key"
+)
+
+type ResourceSyncController struct {
+	// syncRuleLock is used to ensure we avoid races on changes to syncing rules
+	syncRuleLock sync.RWMutex
+
```
sttts: remove the empty line as the lock belongs to the maps below.

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -0,0 +1,221 @@
...
+		// TODO: Should we try to actively remove the resources created by this controller here?
+		return nil
+	}
+
+	c.syncRuleLock.RLock()
+	defer c.syncRuleLock.RUnlock()
+
+	errors := []error{}
+
+	for destination, source := range c.configMapSyncRules {
```
sttts: `from, to` would make it easier to read.

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -0,0 +1,221 @@
...
+		_, _, err := resourceapply.SyncSecret(c.kubeClient.CoreV1(), c.eventRecorder, source.Namespace, source.Name, destination.Namespace, destination.Name)
+		if err != nil {
+			errors = append(errors, err)
+		}
+	}
+
+	if len(errors) > 0 {
+		message := ""
+		for _, err := range errors {
+			message = message + err.Error() + "\n"
```
sttts: didn't we have a func for that formatting?

pkg/operator/starter.go
```diff
@@ -5,6 +5,8 @@ import (
 	"os"
 	"time"
 
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/resourcesynccontroller"
+
```
sttts: order

pkg/operator/resourcesynccontroller/resourcesync_controller.go
deads2k: > `from, to` would make it easier to read.

golang `copy` method uses dest,src

sttts: also fine. even d, s.

deads2k:
comments addressed, unit tests added

pkg/operator/resourcesynccontroller/interfaces.go
```diff
@@ -0,0 +1,19 @@
...
+	Namespace string
+	Name      string
+}
+
+var emptyResourceLocation = ResourceLocation{}
+
+// ResourceSyncer allows changes to syncing rules by this controller
+type ResourceSyncer interface {
+	// SyncConfigMap indicates that a configmap should be copied from the source to the destination.  It will also
+	// mirror a deletion from the source.  If the source, is a zero object the destination will not be synced.
```
sttts: If the source is a zero object, ...

Same below. Actually, the logic is just looking at the namespace, isn't it?

deads2k: > f the source is a zero object, ...
> 
> Same below. Actually, the logic is just looking at the namespace, isn't it?

fixed

pkg/operator/starter.go
```diff
@@ -71,6 +74,30 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
...
+			machineSpecifiedGlobalConfigNamespace: kubeInformersForMachineSpecifiedGlobalConfigNamespace,
+			operatorNamespace:                     kubeInformersForOpenshiftKubeAPIServerOperatorNamespace,
+			targetNamespaceName:                   kubeInformersForOpenshiftKubeAPIServerNamespace,
+			"kube-system":                         kubeInformersForKubeSystemNamespace,
+		},
+		kubeClient,
+		ctx.EventRecorder,
+	)
+	if err := resourceSyncController.SyncConfigMap(
+		resourcesynccontroller.ResourceLocation{Namespace: etcdNamespaceName, Name: "etcd-serving-ca"},
```
sttts: order is dest, src. We should standardize on one. Library-go's sync uses src, dest.

sttts:
https://github.com/openshift/cluster-kube-apiserver-operator/pull/181#pullrequestreview-189864409 looks wrong.

Otherwise, lgtm.

pkg/operator/starter.go
deads2k: > order is dest, src. We should standardize on one. Library-go's sync uses src, dest.

https://golang.org/pkg/builtin/#copy

soltysh:
@deads2k is this intended to land in library-go at some point?

PR #183, opened 2019-01-07
start logical composition of the client ca bundle

deads2k:
Our client-ca bundle is logically composed of several cabundles maintained by different actors.  We need to add a cluster config option for the cluster-admin to specify one too.

/assign @sttts

### DIFF
--- a/bindata/bootkube/manifests/configmap-initial-kube-apiserver-client-ca.yaml
+++ b/bindata/bootkube/manifests/configmap-initial-kube-apiserver-client-ca.yaml
@@ -1,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: initial-client-ca
+  namespace: openshift-config
+data:
+  ca-bundle.crt: |
+    {{ .Assets | load "kube-ca.crt" | indent 4 }}
+
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -104,10 +104,31 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 	); err != nil {
 		return err
 	}
+	if err := resourceSyncController.SyncConfigMap(
+		resourcesynccontroller.ResourceLocation{Namespace: operatorNamespace, Name: "initial-client-ca"},
+		resourcesynccontroller.ResourceLocation{Namespace: userSpecifiedGlobalConfigNamespace, Name: "initial-client-ca"},
+	); err != nil {
+		return err
+	}
+	// this ca bundle contains certs used to sign CSRs (kubelet serving and client certificates)
+	if err := resourceSyncController.SyncConfigMap(
+		resourcesynccontroller.ResourceLocation{Namespace: operatorNamespace, Name: "csr-controller-ca"},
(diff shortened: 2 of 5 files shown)


bindata/bootkube/manifests/configmap-initial-kube-apiserver-client-ca.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: initial-client-ca
+  namespace: openshift-config
+data:
+  ca-bundle.crt: |
+    {{ .Assets | load "kube-ca.crt" | indent 4 }}
```
deads2k: this gives us as a starting point the one used by the installer to set everything up.

pkg/operator/starter.go
```diff
@@ -98,10 +98,31 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 	); err != nil {
 		return err
 	}
+	if err := resourceSyncController.SyncConfigMap(
+		resourcesynccontroller.ResourceLocation{Namespace: operatorNamespace, Name: "initial-client-ca"},
```
deads2k: Collapsing into a single namespace here means less plumbing for wiring together different informers from different namespaces to trigger.

pkg/operator/starter.go
```diff
@@ -98,10 +98,31 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
...
+	// this ca bundle contains certs used to sign CSRs (kubelet serving and client certificates)
+	if err := resourceSyncController.SyncConfigMap(
+		resourcesynccontroller.ResourceLocation{Namespace: operatorNamespace, Name: "csr-controller-ca"},
+		resourcesynccontroller.ResourceLocation{Namespace: machineSpecifiedGlobalConfigNamespace, Name: "csr-controller-ca"},
+	); err != nil {
+		return err
+	}
+	// this ca bundle contains certs used by the kube-apiserver to verify client certs
+	if err := resourceSyncController.SyncConfigMap(
+		resourcesynccontroller.ResourceLocation{Namespace: machineSpecifiedGlobalConfigNamespace, Name: "kube-apiserver-client-ca"},
```
deads2k: This is an outbound sync that we will consume in the openshift-apiserver operator so that we never flap on different static pods restarting with different certs.  We will always use what this operator says is good.

pkg/operator/target_config_reconciler_v311_00.go
```diff
@@ -101,3 +108,62 @@ func managePod_v311_00_to_latest(client coreclientv1.ConfigMapsGetter, recorder
 	configMap.Data["version"] = version.Get().String()
 	return resourceapply.ApplyConfigMap(client, recorder, configMap)
 }
+
+func manageClientCABundle(lister corev1listers.ConfigMapLister, client coreclientv1.ConfigMapsGetter, recorder events.Recorder) (*corev1.ConfigMap, bool, error) {
```
deads2k: I expect this method to become generic, but we'll start it here for now.

mfojtik:
For the record, it failed on:

```
openshift-tests [Conformance][templates] templateinstance readiness test should report ready soon after all annotated objects are ready [Suite:openshift/conformance/parallel/minimal]
```

deads2k:
approved via slack

PR #185, opened 2019-01-09
Add observer for the oauth metadata config

mrogers950:
@openshift/sig-auth @deads2k


pkg/operator/configobservation/auth/observe_metadata.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		errs = append(errs, err)
+	}
+	if len(currentMetadataFilePath) > 0 {
+		if err := unstructured.SetNestedField(prevObservedConfig, currentMetadataFilePath, topLevelMetadataFilePath...); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	if !listers.AuthConfigSynced() {
```
mfojtik: move this up

pkg/operator/configobservation/auth/observe_metadata.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	metadataConfigMap, err := listers.AuthConfigmapLister.ConfigMaps(targetNamespaceName).Get(oauthMetadataConfigMap.Name)
+	if err != nil {
+		recorder.Eventf("ObserveAuthMetadataConfigMap", "Failed to get oauth metadata configMap %s/%s: %v", targetNamespaceName, oauthMetadataConfigMap.Name, err)
+		return observedConfig, errs
+	}
+
+	if _, ok := metadataConfigMap.Data[oauthMetadataConfigMap.Key]; ok {
+		path := fmt.Sprintf("/etc/kubernetes/static-pod-resources/configmaps/%s/%s", oauthMetadataConfigMap.Name, oauthMetadataConfigMap.Key)
```
mfojtik: nit: filepath.Join

mrogers950:
Updated with some changes in the general design. I added in the ability to call the resourceSyncController SyncConfigMap in the observer. This is so we can update the sync source dynamically based on the authentication config's spec or status fields. If set, spec.OAuthMetadata is used for the source. Otherwise use status.IntegratedOAuthMetadata if set and the OAuth type is integrated. Always call SyncConfigMap to keep the source updated and set the well-known oauthMetadataFile path.

mrogers950:
Updated again after some in-person feedback from @enj . @deads2k @mfojtik PTAL

pkg/operator/configobservation/auth/auth_metadata.go
```diff
@@ -0,0 +1,112 @@
...
+const (
+	targetNamespaceName   = "openshift-kube-apiserver"
+	oauthMetadataFilePath = "/etc/kubernetes/static-pod-resources/configmaps/oauth-metadata/oauthMetadata"
+	configNamespace       = "openshift-config"
+	managedNamespace      = "openshift-config-managed"
+)
+
+// ObserveAuthMetadata fills in authConfig.OauthMetadataFile with the path for a configMap referenced by the authentication
+// config.
+func ObserveAuthMetadata(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
```
enj: The `ObserveAuthMetadata` func looks fine to me.  I will defer the specifics of the rest of this PR to @openshift/sig-master (the signature for the observe func does look like it is getting out of hand).

pkg/operator/configobservation/auth/auth_metadata.go
```diff
@@ -0,0 +1,112 @@
...
+	listers := genericListers.(configobservation.ResourceSyncListers)
+	errs := []error{}
+	prevObservedConfig := map[string]interface{}{}
+
+	topLevelMetadataFilePath := []string{"authConfig", "oauthMetadataFile"}
+	currentMetadataFilePath, _, err := unstructured.NestedString(existingConfig, topLevelMetadataFilePath...)
```
soltysh: Use `SetNestedStringSlice` instead

pkg/operator/configobservation/auth/auth_metadata.go
```diff
@@ -0,0 +1,112 @@
...
+	if err != nil {
+		errs = append(errs, err)
+	}
+	if len(currentMetadataFilePath) > 0 {
+		if err := unstructured.SetNestedField(prevObservedConfig, currentMetadataFilePath, topLevelMetadataFilePath...); err != nil {
```
soltysh: `SetNestedStringSlice`

pkg/operator/configobservation/auth/auth_metadata.go
deads2k: > The `ObserveAuthMetadata` func looks fine to me. I will defer the specifics of the rest of this PR to @openshift/sig-master (the signature for the observe func does look like it is getting out of hand).

Please be specific.

You need listers, you need an event recorder, you need the existingconfig.  Using a struct while we settle what this looks like in 4.0 will allow code to keep "working" while not understanding semantic differences in behavior. 
 Comments like this without a viable alternative proposed are simply noise

deads2k: > The `ObserveAuthMetadata` func looks fine to me. I will defer the specifics of the rest of this PR to @openshift/sig-master (the signature for the observe func does look like it is getting out of hand).

No.  @openshift/sig-auth is the owner of this function.  If it causes rippling failures the master team will disable it, but the ownership belongs to the team providing the config.

pkg/operator/configobservation/auth/auth_metadata.go
deads2k: > Use `SetNestedStringSlice` instead

that seems odd.  It's a straight string field, not a slice, right?

mrogers950: yep, just a string field so using SetNestedStringSlice does not seem right to me.

pkg/operator/configobservation/auth/auth_metadata.go
```diff
@@ -0,0 +1,112 @@
...
+		return prevObservedConfig, errs
+	}
+
+	// Unsets oauthMetadataFile if we had an empty source.
+	if len(sourceConfigMap) == 0 {
+		return observedConfig, errs
+	}
+
+	// Set oauthMetadataFile.
+	if err := unstructured.SetNestedField(observedConfig, oauthMetadataFilePath, topLevelMetadataFilePath...); err != nil {
```
deads2k: `SetNestedSTring`?

deads2k: nm

pkg/operator/configobservation/interfaces.go
```diff
@@ -5,14 +5,22 @@ import (
 	"k8s.io/client-go/tools/cache"
 
 	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
+	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
 )
 
+type ResourceSyncListers struct {
+	SyncController *resourcesynccontroller.ResourceSyncController
```
deads2k: promote into library-go using interface `ResourceSyncer`

pkg/operator/starter.go
```diff
@@ -91,6 +84,15 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		kubeClient,
 		ctx.EventRecorder,
 	)
+	configObserver := configobservercontroller.NewConfigObserver(
```
deads2k: move below where they are all added.

pkg/operator/starter.go
```diff
@@ -332,6 +334,7 @@ var deploymentConfigMaps = []string{
 	"etcd-serving-ca",
 	"kubelet-serving-ca",
 	"sa-token-signing-certs",
+	"oauth-metadata",
```
deads2k: this will cause kube-apiserver to never rollout until the CR is present because you're requiring it.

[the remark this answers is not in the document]
mrogers950: Yeah, I didn't realize these would be required configMaps. oauth-metadata is our only sync destination so we just need it available to the static pods when the sync happens.

pkg/operator/configobservation/auth/auth_metadata.go
soltysh: yeah, sorry was looking at two similar stuff at once

mrogers950:
This is all caught up with https://github.com/openshift/library-go/pull/177 now.

pkg/operator/configobservation/auth/auth_metadata.go
```diff
@@ -0,0 +1,112 @@
...
+
+	var (
+		sourceNamespace string
+		sourceConfigMap string
+		statusConfigMap string
+	)
+
+	specConfigMap := authConfig.Spec.OAuthMetadata.Name
+
+	switch {
```
deads2k: need a default case for glog.V(2) if nothing else

PR #186, opened 2019-01-10
Sync kube-system/extension-apiserver-authentication directly to shorten race

sttts:
This is a 80% fix for the race between kube-apiserver and aggregated apiservers using the new aggregator proxy cert and using the new request header CA.


mfojtik:
let make the e2e test pass multiple times to see how much we improved.

PR #188, opened 2019-01-10
images: fix image observer event reporting

mfojtik:
`[]string` and `[]interface{}` are not the same, thus we can't use `reflect.DeepEqual()`. This change gives us string slice, so the check that report events when value changed actually works (instead of spamming the server right now).

/cc @bparees 
/cc @sttts

### DIFF
--- a/pkg/operator/configobservation/images/observe_images.go
+++ b/pkg/operator/configobservation/images/observe_images.go
@@ -10,6 +10,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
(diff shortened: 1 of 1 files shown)


bparees:
@mfojtik @sttts this did not fix anything:

56m         2h           184       openshift-kube-apiserver-operator.1579cd657e7056fb   Deployment               Normal    ObserveExternalRegistryHostnameChanged   openshift-cluster-kube-apiserver-operator   External registry hostname changed to [foo.bar.com]


15m         6h           483       openshift-kube-apiserver-operator.1579c2d8a898eaff   Deployment               Normal    ObserveAllowedRegistriesForImport        openshift-cluster-kube-apiserver-operator   Allowed registries for import changed to []

i'm going to submit a PR ripping this event recording logic out entirely.

PR #189, opened 2019-01-10
installer: QoS class guaranteed

sttts:
We saw OOM killing the installers frequently. Successful runs showed around 25MB of resident memory. 100MB gives the installer some space to breath.


deads2k:
This isn't the pod you're looking for.  See https://github.com/openshift/library-go/blob/master/pkg/operator/staticpod/controller/installer/manifests/installer-pod.yaml

Update this pull to delete the file.

sttts:
@deads2k updated. ptal.

PR #194, opened 2019-01-11
rename fewer resources

deads2k:
The big move is constant failing.  This is a smaller set that will hopefully merge and helps move the ball.


pkg/operator/starter.go
```diff
@@ -153,8 +153,14 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		ctx.EventRecorder,
 	)
 	clusterOperatorStatus := status.NewClusterOperatorStatusController(
-		"openshift-kube-apiserver-operator",
-		[]configv1.ObjectReference{},
+		"kube-apiserver",
+		[]configv1.ObjectReference{
```
deads2k: these are needed for debugging

PR #195, opened 2019-01-12
add related object

deads2k:
this is scoped down to just adding the related objects so that `openshift-must-gather` will work.


deads2k:
we installed ok.  I don't want to force it yet.  osin operator is pretty unhappy.
/retest

PR #197, opened 2019-01-14
disable endpoint reconciliation on bootstrap master

deads2k:
specify a set of apiServerArguments to override the endpoint-reconciler-type


deads2k:
So this worked, but it exposed a neat thing.  the kas-operator uses the service network to communicate with kube-apiservers in case the local one dies.  Without registering endpoints, we don't have a kube-apiserver.  Until we find a real problem with registering the bootstrap one, we'll leave it.

@sttts final outcome

sttts:
I thought this was intentional, especially during bootstrap time there is no local apiserver.

PR #198, opened 2019-01-14
remove broken event recording logic

bparees:
this is spinning events and as the owner of this config, i'm not interested in the event, so i'm removing it rather than investing further time debugging why the comparison is failing.

### DIFF
--- a/pkg/operator/configobservation/images/observe_images.go
+++ b/pkg/operator/configobservation/images/observe_images.go
@@ -3,14 +3,12 @@ package images
 import (
 	"bytes"
 	"encoding/json"
-	"reflect"
 
 	"github.com/golang/glog"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
-	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/library-go/pkg/operator/configobserver"
 	"github.com/openshift/library-go/pkg/operator/events"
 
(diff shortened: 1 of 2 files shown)


bparees:
@adambkaplan ptal

bparees:
> I suspect DeepEqual requires the arrays to be exactly identical.

Yeah, I suspected the same (some sort of ordering issue) except that @sjenning saw this happening even with an empty array:

`Allowed registries for import changed to []`

so it's something more subtle than that.

mfojtik:
@bparees i'm not happy this was removed. I think this event is useful for admin that want to know the registry external/internal hostname was updated (admin or a script can wait for that event to come before allowing clients to pull the the images via external route/etc.). I think we should fix the logic here by converting the field into sorted array and comparing.

pkg/operator/configobservation/images/observe_images.go
```diff
@@ -169,15 +163,6 @@ func ObserveAllowedRegistriesForImport(genericListers configobserver.Listers, re
 			return prevObservedConfig, append(errs, err)
 		}
 	}
-
-	existingLocations := []configv1.RegistryLocation{}
-	for _, location := range existingAllowedRegistries {
-		existingLocations = append(existingLocations, location.(configv1.RegistryLocation))
-	}
-	if !reflect.DeepEqual(existingLocations, configImage.Spec.AllowedRegistriesForImport) {
```
mfojtik: @bparees why was this ripped off? afaik this part worked fine?

pkg/operator/configobservation/images/observe_images.go
```diff
@@ -115,10 +113,6 @@ func ObserveExternalRegistryHostnames(genericListers configobserver.Listers, rec
 		}
 	}
 
-	if !reflect.DeepEqual(existingHostnames, externalRegistryHostnames) {
```
sttts: about the empty case, we don't identify empty and nil here. Maybe we should.

pkg/operator/configobservation/images/observe_images.go
bparees: It did not work fine:
https://github.com/openshift/cluster-kube-apiserver-operator/pull/188#issuecomment-454184145

483 events from this, when the array was empty.

PR #200, opened 2019-01-15
images observer: fix empty slice case and simplify

sttts:


mfojtik:
LGTM

@bparees this should stop the spamming and leave the events there for debugging. PTAL

bparees:
Given the issues we've had with this so far, I think it needs a test that confirms that after observing some array of values, a subsequent observation of that same array(possibly re-ordered) does not emit a new event.

ie given an existing config with [a,b], observing [b,a] should not emit an event.

sttts:
@bparees added idempotency tests.

bparees:
> I conclude that order is significant. Is this right?

yes.  The order of the arrays in spec and status is significant, as is the order of the merged array that is ultimately provided to the apiserver.  The first entry is used as the actual public docker registry hostname for imagestreams (since we can only have one value for that).

bparees:
/lgtm

thank you @sttts

PR #201, opened 2019-01-15
Change OperatorSpec to StaticPodOperatorSpec

damemi:
Makes use of the change in https://github.com/openshift/api/pull/171 to add revision limit fields to operator spec


mfojtik:
/lgtm
/hold
(unhold when the e2e is green)

damemi:
/hold cancel
e2e green

PR #203, opened 2019-01-16
set cpu request on kube-apiserver and operator

sjenning:
@deads2k @smarterclayton @derekwaynecarr


smarterclayton:
/approve

As discussed in slack, we are setting requests proportional to what would be necessary on a 2 core master, with 300m for etcd, 300m for both apiservers, 300m for all 3 major controllers, and then leave the remainder for operators.

lgtm but leave to team to review

/retest

PR #204, opened 2019-01-17
explicitly list metrics port as first containerPort

juanvallejo:
cc @deads2k


mfojtik:
the change looks fine to me

PR #206, opened 2019-01-17
refactor controllers into separate packages

deads2k:
This is a code shuffle to make the main start easier to read.


deads2k:
there is no functional change here.

PR #207, opened 2019-01-17
 allow configuring log level of operand

deads2k:
adds a commit on top of https://github.com/openshift/cluster-kube-apiserver-operator/pull/206 to configure glog


pkg/operator/targetconfigcontroller/target_config_reconciler.go
```diff
@@ -208,7 +208,18 @@ func managePod(client coreclientv1.ConfigMapsGetter, recorder events.Recorder, o
 	if len(imagePullSpec) > 0 {
 		required.Spec.Containers[0].Image = imagePullSpec
 	}
-	required.Spec.Containers[0].Args = append(required.Spec.Containers[0].Args, fmt.Sprintf("-v=%d", 2))
+	switch operatorConfig.Spec.LogLevel {
```
soltysh: @deads2k how about making this translation from loglevel to an actual `-v` number be part of library-go? Although this is documented in the API I'm pretty sure it might get lost in translation.

PR #208, opened 2019-01-18
tighten rotation intervals

deads2k:
This tightens our rotation interval so we can get reports of broken clusters if we don't work right.

@sttts I've picked up your scheduler fix.


deads2k:
bumped to get the better rotation code, minor updates in the bump.  tagging based on previous slack conversation

deads2k:
@openshift/sig-networking https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_cluster-kube-apiserver-operator/208/pull-ci-openshift-cluster-kube-apiserver-operator-master-e2e-aws/769 looks like a networking problem.  I see `apiservice/v1.apps.openshift.io: not available: no response from https://172.30.83.61:443: Get https://172.30.83.61:443: dial tcp 172.30.83.61:443: connect: no route to host`.  "no route to host" isn't about missing endpoints.

/retest

PR #212, opened 2019-01-21
 move to openshift/api types

deads2k:
switches us to use the new API type

@tnozicka is there anything I have to do the makefile?


Makefile
```diff
@@ -1,11 +1,6 @@
...
 .PHONY: all
 
-# Codegen module needs setting these required variables
-CODEGEN_OUTPUT_PACKAGE :=github.com/openshift/cluster-kube-apiserver-operator/pkg/generated
-CODEGEN_API_PACKAGE :=github.com/openshift/cluster-kube-apiserver-operator/pkg/apis
-CODEGEN_GROUPS_VERSION :=kubeapiserver:v1alpha1
-
 # Include the library makefile
 include $(addprefix ./vendor/github.com/openshift/library-go/alpha-build-machinery/make/, \
 	operator.mk \
```
tnozicka: @deads2k this is what includes the codegen package - those variables above are those that the package requires. at the beginning codegen seemed to be something used in all of our operator so it was included by default but we should revisit that choice once our 4.0 repos are setup based on usage.

to get rid of it you need to choose the targets manually like in https://github.com/openshift/cluster-config-operator/blob/929d69fff0ed1f31bc43276b582e957bbafd41b0/Makefile#L5-L10

deads2k:
retagging

deads2k:
```
                        "lastTransitionTime": "2019-01-23T17:38:19Z",
                        "message": "error syncing: timed out waiting for the condition during syncRequiredMachineConfigPools: error pool master is not ready. status: (total: 3, updated: 3, unavailable: 1)",
                        "status": "True",
                        "type": "Failing"
```

/retest

deads2k:
found the issue.  once https://github.com/openshift/cluster-authentication-operator/pull/48 merges, this should pass

PR #213, opened 2019-01-21
stop permafailing status

deads2k:
`TargetConfigController` was setting a status in response to being unable to set a status and created a situation of permafailing status. In addition to being a dumb thing to do (why try to set the thing you couldn't set), this stuck us in the CVO.

/assign @mfojtik @sanchezl


deads2k:
closing for https://github.com/openshift/cluster-kube-apiserver-operator/pull/215

PR #218, opened 2019-01-23
make the target namespace removable

deads2k:
We should be able to delete every namespace except the `openshift-config` namespace and recover.  This clearly separates the initially created certificates from those that are managed.

/assign @mfojtik @sttts

### DIFF
--- a/bindata/bootkube/manifests/configmap-aggregator-client-ca.yaml
+++ b/bindata/bootkube/manifests/configmap-aggregator-client-ca.yaml
@@ -1,8 +1,0 @@
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: aggregator-client-ca
-  namespace: openshift-config-managed
-data:
-  ca-bundle.crt: |
-    {{ .Assets | load "aggregator-ca.crt" | indent 4 }}
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -153,6 +153,10 @@ func createTargetConfig(c TargetConfigController, recorder events.Recorder, oper
 	if err != nil {
 		errors = append(errors, fmt.Errorf("%q: %v", "configmap/aggregator-client-ca", err))
 	}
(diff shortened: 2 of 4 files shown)


bindata/bootkube/manifests/configmap-aggregator-client-ca.yaml
```diff
@@ -1,8 +0,0 @@
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: aggregator-client-ca
```
deads2k: we don't need to bootstrap this.  It should be done via the operator.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -234,6 +238,24 @@ func manageClientCABundle(lister corev1listers.ConfigMapLister, client coreclien
...
+func manageKubeletServingCABundle(lister corev1listers.ConfigMapLister, client coreclientv1.ConfigMapsGetter, recorder events.Recorder) (*corev1.ConfigMap, bool, error) {
+	requiredConfigMap, err := resourcesynccontroller.CombineCABundleConfigMaps(
+		resourcesynccontroller.ResourceLocation{Namespace: operatorclient.TargetNamespace, Name: "kubelet-serving-ca"},
+		lister,
+		nil, // TODO remove this
+		nil, // TODO remove this
+		// this is from the installer and contains the value they think we should have
+		resourcesynccontroller.ResourceLocation{Namespace: operatorclient.GlobalUserSpecifiedConfigNamespace, Name: "initial-kubelet-serving-ca"},
+		// this is from kube-controller-manager and indicates the ca-bundle.crt to verify the signed kubelet serving certs
+		resourcesynccontroller.ResourceLocation{Namespace: operatorclient.GlobalMachineSpecifiedConfigNamespace, Name: "csr-controller-ca"},
```
sttts: who creates and rotates that?

sttts: and what happens if the configmap does not exist? Is it just ignored?

deads2k: > and what happens if the configmap does not exist? Is it just ignored?

yes, ignored.

deads2k:
prometheus is the only consistent failure

PR #220, opened 2019-01-25
Enable --enable-aggregator-routing by default

sttts:
Use endpoints IPs directly instead of service IPs, taking some of the networking logic out of the (stability) equation.


squeed:
:+1:

squeed:
Further context: even if networking were long-run 100% reliable, kube-proxy still has non-zero convergence time. In times of apiserver disruption, this naturally can take longer, which would exacerbate any problems. This change takes one more component out of the critical path.

PR #221, opened 2019-01-25
bump(*): set failing status for failing installer pod

damemi:
Includes the change from https://github.com/openshift/library-go/pull/207


soltysh:
There's a newer bump picking up additional changes that we'll go with https://github.com/openshift/cluster-kube-apiserver-operator/pull/225
/close

PR #222, opened 2019-01-26
Trust cluster-ingress-operator's CA certificate

Miciah:
This PR is related to [NE-139](https://jira.coreos.com/browse/NE-139).

This PR depends on https://github.com/openshift/cluster-ingress-operator/pull/109.

* `pkg/operator/operatorclient/interfaces.go` (`IngressOperatorNamespace`): New constant for cluster-ingress-operator's namespace.
* `pkg/operator/resourcesynccontroller/resourcesynccontroller.go` (`NewResourceSyncController`): Sync cluster-ingress-operator's router-ca configmap.
* `pkg/operator/starter.go` (`RunOperator`): Watch cluster-ingress-operator's namespace.  Add the namespace to related resources in the clusteroperator.
* `pkg/operator/targetconfigcontroller/targetconfigcontroller.go`
(`NewTargetConfigController`): React to changes to configmaps in cluster-ingress-operator's namespace.
(`manageClientCABundle`): Incorporate cluster-ingress-operator's CA certificate into the trust bundle.

### DIFF
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -84,6 +84,7 @@ func NewTargetConfigController(
 	kubeInformersForNamespaces.InformersFor(operatorclient.GlobalUserSpecifiedConfigNamespace).Core().V1().ConfigMaps().Informer().AddEventHandler(c.eventHandler())
 	kubeInformersForNamespaces.InformersFor(operatorclient.GlobalMachineSpecifiedConfigNamespace).Core().V1().ConfigMaps().Informer().AddEventHandler(c.eventHandler())
 	kubeInformersForNamespaces.InformersFor(operatorclient.OperatorNamespace).Core().V1().ConfigMaps().Informer().AddEventHandler(c.eventHandler())
+	kubeInformersForNamespaces.InformersFor(operatorclient.IngressOperatorNamespace).Core().V1().ConfigMaps().Informer().AddEventHandler(c.eventHandler())
 	kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Core().V1().ConfigMaps().Informer().AddEventHandler(c.eventHandler())
 
 	return c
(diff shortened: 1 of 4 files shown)


pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -230,6 +231,8 @@ func manageClientCABundle(lister corev1listers.ConfigMapLister, client coreclien
 		resourcesynccontroller.ResourceLocation{Namespace: operatorclient.GlobalMachineSpecifiedConfigNamespace, Name: "csr-controller-ca"},
 		// this bundle is what this operator uses to mint new client certs it directly manages
 		resourcesynccontroller.ResourceLocation{Namespace: operatorclient.OperatorNamespace, Name: "managed-kube-apiserver-client-ca-bundle"},
+		// this bundle is what cluster-ingress-operator uses to mint default certificates for routers
+		resourcesynccontroller.ResourceLocation{Namespace: operatorclient.IngressOperatorNamespace, Name: "router-ca"},
```
sttts: this means that we trust any client cert signed by the ingress CA (and groups and users claimed inside it). Is this really what we want?

mrogers950:
@sttts @Miciah Sorry for the confusion, this actually needs to go into cluster-kube-controller-manager-operator in order to wire into the service account CA bundle. 
https://github.com/openshift/cluster-kube-controller-manager-operator/blob/83d4f615d39b3131ad7fea65288633880daae5ca/pkg/operator/targetconfigcontroller/targetconfigcontroller.go#L205
One thing to note is that including the router CA into the bundle should be optional in some way, particularly when the router CA is user supplied (or the user opts out of having a default router CA generated).

Miciah:
> Sorry for the confusion, this actually needs to go into cluster-kube-controller-manager-operator in order to wire into the service account CA bundle.
> https://github.com/openshift/cluster-kube-controller-manager-operator/blob/83d4f615d39b3131ad7fea65288633880daae5ca/pkg/operator/targetconfigcontroller/targetconfigcontroller.go#L205

@mrogers950, thanks for the pointer! Will this suffice to get OpenShift Console to trust the router CA?

> One thing to note is that including the router CA into the bundle should be optional in some way, particularly when the router CA is user supplied (or the user opts out of having a default router CA generated).

To keep things a little simpler, cluster-ingress-operator always generates the CA and default certificates, even if the administrator configures it to use a custom certificate.  Since cluster-ingress-operator always creates the configmap, is it acceptable if cluster-kube-controller-manager-operator always trusts it (even if it's superfluous in the custom-router-certificate scenario)?

enj:
> is it acceptable if cluster-kube-controller-manager-operator always trusts it (even if it's superfluous in the custom-router-certificate scenario)?

No, that just leaves open an attack vector which we are actively trying to mitigate.

Miciah:
> No, that just leaves open an attack vector which we are actively trying to mitigate.

If I understand correctly, the only way to achieve this without horrible hacks in cluster-kube-controller-manager-operator or changes to library-go would be for cluster-ingress-operator to create or delete the configmap conditionally as clusteringresses are created, updated, and deleted.  I'll bring this up with my team.

Miciah:
Superseded by https://github.com/openshift/cluster-kube-controller-manager-operator/pull/145.

PR #225, opened 2019-01-28
bump(*): library-go to get priority class on installer pods

mfojtik:
Bump to get the priority class on installer pods.


glide.lock
```diff
@@ -242,7 +242,7 @@ imports:
   - config/informers/externalversions/internalinterfaces
   - config/listers/config/v1
 - name: github.com/openshift/library-go
-  version: 22c06b1b6dd66dc827b452b77b4cc48cd7820e91
+  version: 5730ef49f037925f9566a64af72327e8575620c5
```
soltysh: That doesn't look correct, I can't find that commit.

soltysh: Wrong PR.

PR #228, opened 2019-01-28
Disable local-storage features

gnufied:
Disable local-storage feature as per  JIRA- STOR-183


gnufied:
As expected local-storage tests are failing but https://github.com/openshift/origin/pull/21881 should disable them.

deads2k:
Given that the bootstrap control plane shouldn't put bad data in, you probably need to update https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/bindata/bootkube/config/bootstrap-config-overrides.yaml too

sttts:
@deads2k the defaultconfig is also the basis for the bootstrap control plane. So no need for an override.

PR #234, opened 2019-01-31
[DO NOT MERGE] - temporarily log pods/status to debug phase transition issues

tnozicka:
/hold


sjenning:
I noticed the need for this just today realizing the audit level wasn't set to record the whole resource :+1:

tnozicka:
yeah, the answer I got was to debug it in a PR only, not to merge it even temporarily. Feel free to take it over to another PR if you need to debug the pod transition issues.

/close

PR #236, opened 2019-02-01
tighten kube-apiserver names

deads2k:
1. kubeapiserveroperatorconfig is now called cluster
2. clusteroperator is now kube-apiserver
3. resources in our namespace no longer have the openshift prefix

### DIFF
--- a/manifests/0000_10_kube-apiserver-operator_02_service.yaml
+++ b/manifests/0000_10_kube-apiserver-operator_02_service.yaml
@@ -2,9 +2,9 @@ apiVersion: v1
 kind: Service
 metadata:
   annotations:
-    service.alpha.openshift.io/serving-cert-secret-name: openshift-kube-apiserver-operator-serving-cert
+    service.alpha.openshift.io/serving-cert-secret-name: kube-apiserver-operator-serving-cert
   labels:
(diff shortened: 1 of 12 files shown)


manifests/0000_10_kube-apiserver-operator_02_service.yaml
```diff
@@ -2,9 +2,9 @@ apiVersion: v1
 kind: Service
 metadata:
   annotations:
-    service.alpha.openshift.io/serving-cert-secret-name: openshift-kube-apiserver-operator-serving-cert
+    service.alpha.openshift.io/serving-cert-secret-name: kube-apiserver-operator-serving-cert
   labels:
-    app: openshift-kube-apiserver-operator
+    app: kube-apiserver-operator
```
sttts: when will we rename the repos?

deads2k: > when will we rename the repos?

Never.  Clayton prefers `cluster-` in the binary and I don't care enough to face the image pain

soltysh: Yes, please!

PR #241, opened 2019-02-06
remove cruft and depend on cluster-config-operator

deads2k:
cluster-config-operator creates our CRDs so they should always be present and we can wait.

/assign @mfojtik


deads2k:
flaky tests

/retest

pkg/operator/configobservation/images/observe_images.go
```diff
@@ -35,11 +34,6 @@ func ObserveInternalRegistryHostname(genericListers configobserver.Listers, reco
 		}
 	}
 
-	if !listers.ImageConfigSynced() {
```
mfojtik: this will make our logs nice again ;-)

sttts: why don't we care anymore?

PR #244, opened 2019-02-08
move all irreplaceable content to openshift-config

deads2k:
The installer provides a set of certificates that we cannot replace, cannot remove, and cannot rotate.  These are the only irreplaceable content in our operator and target namespaces. This pull moves that content into `openshift-config` since we cannot manage it.  In theory, this will make it possible to delete the entire operator and target namespace and recover from it.

/assign @sttts @mfojtik


deads2k:
discussed in slack last week.

/retest

PR #245, opened 2019-02-08
update admission configuration for groupification

deads2k:
Admission plugins and their configuration types have been groupified.  We tolerate both at the moment to make this transition, but once we merge this pull we will remove the old configuration.

@openshift/sig-networking fyi
/assign @sttts @mfojtik @soltysh


squeed:
We already use the group `network.openshift.io` for configuring various aspects of the network controller, mostly related to IP allocation and security. I see that this isn't a registered type, so it's not that big a deal, but do you think it will be confusing?

PR #238, opened 2019-02-05
make it easy to specify multiple SA token verification certs

deads2k:
SA token verification takes a list of signing public keys to verify the SA tokens.  This separates out the initial one create by the installer and makes it possible to union more in.

I'm going to open an origin pull to make this take a directory, alpha sort, and use that.

/assign @sttts

PR #247, opened 2019-02-11
start requiring kubelet trust

deads2k:
This pulls finishes the wiring for requiring proper serving certs from kubelets

@sjenning 
/assign @sttts @mfojtik


deads2k:
kubelets are only signed by the CSR, so if we trust the CSR bundle we should be fine.

pkg/operator/resourcesynccontroller/resourcesynccontroller.go
```diff
@@ -32,6 +32,13 @@ func NewResourceSyncController(
 	); err != nil {
 		return nil, err
 	}
+	// this ca bundle contains certs used trust kubelets
```
sttts: to trust

sttts:
Only the typo, lgtm otherwise.

deads2k:
I need https://github.com/openshift/cluster-kube-apiserver-operator/pull/238 to have a conditional sync function.

PR #250, opened 2019-02-12
Switch operator to use protobuf client for non-CR resources

mfojtik:
This builds on https://github.com/openshift/library-go/pull/233 and switches the controller loops that deal with non-CR resources to client-go with protobuf enforced.

/cc @deads2k 
/cc @smarterclayton 

**NOTE:** Remove the artificial bump before merging.


soltysh:
Surprisingly fast green, do you have data before and after this change?

mfojtik:
Only network and DNS are present in clusteroperators so this must be some networking flake.

/retest

pkg/operator/starter.go
```diff
@@ -29,7 +29,7 @@ import (
 )
 
 func RunOperator(ctx *controllercmd.ControllerContext) error {
-	kubeClient, err := kubernetes.NewForConfig(ctx.KubeConfig)
+	protoKubeClient, err := kubernetes.NewForConfig(ctx.ProtoKubeConfig)
```
deads2k: don't rename this

mfojtik: fixed.

mfojtik:
Fixed client rename, LGTM per Slack from @deads2k

mfojtik:
Re-applied LGTM as the only change was a panic fix in library-go.

PR #253, opened 2019-02-13
bump(*): library-go to fix prune controller panic

damemi:
This should (hopefully) fix some of the problems we were seeing with pruning, we had an older version of library go that (probably among other things) didn't catch a panic when there were 0 IDs to prune


deads2k:
out of date

PR #254, opened 2019-02-13
 handle multiple sources of sa token cert verifiers

deads2k:
I think my ns in https://github.com/openshift/cluster-kube-apiserver-operator/pull/238

### DIFF
--- a/pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
+++ b/pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
@@ -1,7 +1,8 @@
 package configobservercontroller
 
 import (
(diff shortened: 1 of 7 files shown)


pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -1,7 +1,8 @@
 package configobservercontroller
 
 import (
-	kubeinformers "k8s.io/client-go/informers"
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/configobservation/satokencerts"
```
mfojtik: nit: order

PR #257, opened 2019-02-14
retry configobservation updates

deads2k:


deads2k:
healthier than most runs, this will help with flakiness merging

PR #261, opened 2019-02-18
bump(*): library-go with logging changes

mfojtik:


mfojtik:
NOTE: I really need to figure out why we always bumping `google.golang.org/appengine` ... it seems to be used by library-go and client-go and I think we should pin it to kube 1.12 levels in there to prevents these?

deads2k:
/lgtm




> NOTE: I really need to figure out why we always bumping `google.golang.org/appengine` ... it seems to be used by library-go and client-go and I think we should pin it to kube 1.12 levels in there to prevents these?

icky, but ok.  YOu tried clearing the cache?

mfojtik:
@deads2k yeah, `glide cc` before every bump, yet still getting these...

PR #262, opened 2019-02-18
correctly set version information

deads2k:
/hold

holding until I get the library-go changes

This sets the version information for our component.

/assign @mfojtik


mfojtik:
/approve

LGTM (feel free to tag when the bump gets in)

PR #263, opened 2019-02-18
explain where kube-apiserver resources come from

deads2k:
This adds a command to the `cluster-kube-apiserver-operator` that explains where resources come from by building a digraph where nodes are resources and edges represent logical flow.  The roots are

1. Payload
2. Installer
3. User

Here is the state we're in today:

resources 1


deads2k:
weird wedge

/retest

PR #265, opened 2019-02-19
Add kube-scheduler group

ravisantoshgudimetla:
While working on https://github.com/openshift/cluster-kube-scheduler-operator/pull/62, I realized that, we are not populating groups and UID fields for `DefaultInfo` struct. I think we need to have the group when we are generating certs here:

https://github.com/openshift/cluster-kube-controller-manager-operator/blob/master/vendor/github.com/openshift/library-go/pkg/crypto/crypto.go#L754-#L756

/cc @deads2k @sttts

### DIFF
--- a/pkg/operator/certrotationcontroller/certrotationcontroller.go
+++ b/pkg/operator/certrotationcontroller/certrotationcontroller.go
@@ -6,6 +6,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 
+	"github.com/pborman/uuid"
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/operatorclient"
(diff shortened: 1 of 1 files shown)


pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -131,7 +132,7 @@ func NewCertRotationController(
 			Validity:          1 * 4 * time.Hour,
 			RefreshPercentage: 0.5,
 			ClientRotation: &certrotation.ClientRotation{
-				UserInfo: &user.DefaultInfo{Name: "system:kube-scheduler"},
+				UserInfo: &user.DefaultInfo{Name: "system:kube-scheduler", Groups: []string{user.KubeScheduler}, UID: uuid.NewRandom().String()},
```
deads2k: we don't actually need to set these.  Do you need a group?

ravisantoshgudimetla: TBH, no if we don't have a group(in kube-scheduler's case, we don't have) but I was trying to populate every field possible to find out if this was causing the issue in openshift/cluster-kube-scheduler-operator#62.

ravisantoshgudimetla:
Closing it as my testing locally, did not help in debugging the issue.

PR #266, opened 2019-02-19
render: merge feature gates

sttts:
Feature gates in the defaultconfig have to be overridable by user config. This means we should append those values, not replace them.

### DIFF
--- a/pkg/cmd/render/render.go
+++ b/pkg/cmd/render/render.go
@@ -1,6 +1,7 @@
 package render
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


sttts:
/hold

Requires https://github.com/openshift/library-go/pull/252.

pkg/cmd/render/render.go
```diff
@@ -154,6 +156,21 @@ func (r *renderOpts) Run() error {
...
+				if !ok {
+					v, _ := json.Marshal(dest)
+					return nil, fmt.Errorf("expected apiServerArguments.feature-gates to be a string slice, but found: %v", v)
+				}
+				source, ok := src.([]interface{})
+				if !ok {
+					v, _ := json.Marshal(source)
+					return nil, fmt.Errorf("expected apiServerArguments.feature-gates to be a string slice, but found: %v", v)
+				}
+				return append(dest, source...), nil // append, later feature gates override earlier
```
deads2k: I expected us to stop here.  Stomping will allow us to set one value when we observe configuration, but allow full flexibility if the user goes for unsupportedConfigOverrides.

PR #268, opened 2019-02-19
make config observation faster

deads2k:
/hold


deads2k:
@openshift/sig-master who's ready for this? We'd do it library-go

deads2k:
passed except for a known flake.  We could do this.

/hold

hold because it should be library-go first

mfojtik:
/retest

(all known flakes)

deads2k:
we're going for this.  We're going go to get this in a bump

PR #270, opened 2019-02-19
update to latest static pod manager

deads2k:
this should set the versions correctly.


deads2k:
known bootstrap user flake.

PR #271, opened 2019-02-19
Simplify conditions

deads2k:
proof that https://github.com/openshift/library-go/pull/258 works.

/hold


deads2k:
now a simple bump

PR #273, opened 2019-02-20
Generate CRD schemata

sttts:

PR #275, opened 2019-02-20
[WIP] Generate CRDs with kubebuilder tool

damemi:
This is in progress because of a couple PRs that it depends on, but demonstrates how we can generate CRDs with kubebuilder


damemi:
This depends on:
1. https://github.com/kubernetes-sigs/controller-tools/pull/144, to add the ability to generate from paths other than `pkg/apis` (awaiting lgtm)
2. https://github.com/openshift/api/pull/214, to add a tag to the API type which will generate the singular name

There are also some small differences that I'm not sure how to deal with. The `subresources: status` section in the original CRD seems to be replaced with a straight up `status` section, is that a problem? Seems like an artifact of the generation tool looking at all the types. Can probably exclude or modify that with a tag in kubebuilder

cc @sttts @deads2k

soltysh:
Is this alternate solution to https://github.com/openshift/cluster-kube-apiserver-operator/pull/273 ?

damemi:
this is an alternative, yes, I think we're trying to ultimately come up with something between the two

damemi:
closing for https://github.com/openshift/cluster-kube-apiserver-operator/pull/312

PR #276, opened 2019-02-20
[WIP]  prove we can adopt an installer cert

deads2k:
I don't actually want to adopt this one.  I just want to prove we *can* adopt it.

/assign @sttts


deads2k:
proof of https://github.com/openshift/library-go/pull/266

deads2k:
@sttts yeah, this appears to have worked based on secret content

mfojtik:
LGTM

PR #279, opened 2019-02-21
switch to new aggregator cert

deads2k:
switches our bootstrapping to the short lived cert from https://github.com/openshift/installer/pull/1232

/assign @mfojtik @sttts

### DIFF
--- a/bindata/bootkube/config/config-overrides.yaml
+++ b/bindata/bootkube/config/config-overrides.yaml
@@ -1,12 +1,5 @@
 apiVersion: kubecontrolplane.config.openshift.io/v1
 kind: KubeAPIServerConfig
-aggregatorConfig:
-  proxyClientInfo:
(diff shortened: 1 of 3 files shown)


bindata/bootkube/config/config-overrides.yaml
```diff
@@ -1,12 +1,5 @@
 apiVersion: kubecontrolplane.config.openshift.io/v1
 kind: KubeAPIServerConfig
-aggregatorConfig:
-  proxyClientInfo:
-    certFile: /var/run/secrets/aggregator-client/tls.crt
-    keyFile: /var/run/secrets/aggregator-client/tls.key
```
sttts: the operator delays startup until it has added these "in code"?

deads2k: > the operator delays startup until it has added these "in code"?

set here and since it is required for a revision, the installer won't start without it.

PR #282, opened 2019-02-21
adopt signers from installer for rotating serving certs

deads2k:
This doesn't wire us to serve with these new certs, but it does establish adoption of signing certs so we can rotate in the future.

/assign @sttts 

To make this actually work, we need to turn the list of hostnames into a function and we need to re-sign when it changes.


deads2k:
adopting signers as discussed.  ha proxy flake again.

PR #283, opened 2019-02-22
bump(*): library-go to pick installer fix and rolebinding fix

mfojtik:


deads2k:
replaced by https://github.com/openshift/cluster-kube-apiserver-operator/pull/287 because of conflict

PR #284, opened 2019-02-22
fix setting loglevel for the operand

mfojtik:
@deads2k this is fun: https://bugzilla.redhat.com/show_bug.cgi?id=1679898

We append to args which will append to bash but not to hypershift? I haven't trying if the loglevel setting actually work yet, have you?

### DIFF
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -225,17 +225,18 @@ func managePod(client coreclientv1.ConfigMapsGetter, recorder events.Recorder, o
 	if len(imagePullSpec) > 0 {
 		required.Spec.Containers[0].Image = imagePullSpec
 	}
+
 	switch operatorConfig.Spec.LogLevel {
 	case operatorv1.Normal:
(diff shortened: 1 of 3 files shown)


pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -225,17 +225,18 @@ func managePod(client coreclientv1.ConfigMapsGetter, recorder events.Recorder, o
 	if len(imagePullSpec) > 0 {
 		required.Spec.Containers[0].Image = imagePullSpec
 	}
+
 	switch operatorConfig.Spec.LogLevel {
 	case operatorv1.Normal:
-		required.Spec.Containers[0].Args = append(required.Spec.Containers[0].Args, fmt.Sprintf("-v=%d", 2))
+		required.Spec.Containers[0].Args[len(required.Spec.Containers[0].Args)-1] = required.Spec.Containers[0].Args[len(required.Spec.Containers[0].Args)-1] + " " + fmt.Sprintf("-v=%d", 2)
```
sttts: can we factor that logic out? looks super ugly.

deads2k:
replaced by https://github.com/openshift/cluster-kube-apiserver-operator/pull/285

PR #286, opened 2019-02-22
Dynamically set hostnames

deads2k:
proof of https://github.com/openshift/library-go/pull/270

/hold

illegal fake bump

/assign @sttts

### DIFF
--- a/pkg/operator/certrotationcontroller/dynamic_serving.go
+++ b/pkg/operator/certrotationcontroller/dynamic_serving.go
@@ -1,0 +1,40 @@
+package certrotationcontroller
+
+import (
+	"sync"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// DynamicServingRotation is a threadsafe struct to provide hostname methods for certrotation.ServingRotation info.
+// This allows us to change the hostnames and get our certs regenerated.
+type DynamicServingRotation struct {
+	lock             sync.RWMutex
+	hostnames        []string
+	hostnamesChanged chan struct{}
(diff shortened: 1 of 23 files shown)


pkg/operator/certrotationcontroller/dynamic_serving.go
```diff
@@ -0,0 +1,38 @@
+package certrotationcontroller
+
+import (
+	"sync"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+type DynamicServingRotation struct {
```
sttts: what is this? comments are missing

deads2k:
comments added

vendor/github.com/openshift/library-go/pkg/operator/certrotation/client_cert_rotation_controller.go
```diff
@@ -158,6 +160,25 @@ func (c *CertRotationController) Run(workers int, stopCh <-chan struct{}) {
 
 	}, time.Minute, stopCh)
 
+	// if we have a need to force rechecking the cert, use this channel to do it.
+	var targetRefresh <-chan struct{}
+	if refresher, ok := c.TargetRotation.CertCreator.(TargetCertRechecker); ok {
+		targetRefresh = refresher.RecheckChannel()
+	}
+	if targetRefresh != nil {
```
sttts: move this into the first if clause.

deads2k:
Spoke on slack.   We're agreed this is the correct direction.

PR #287, opened 2019-02-22
make installer pod tolerant

deads2k:


deads2k:
merged in another pull.

PR #288, opened 2019-02-22
trust the new initial kubelet client CA

deads2k:
kubelets have clients signed by this short lived cert


deads2k:
ha proxy flake.

PR #290, opened 2019-02-22
provide ca bundle for trusting the kube-apiserver

deads2k:
We need to provide a cabundle for trusting the kube-apiserver since we're minting certificates.


deads2k:
haproxy flake,

PR #291, opened 2019-02-22
 indicate cert rotation refresh in durations

deads2k:
Also eliminates duplicate generation


deads2k:
merged as part of larger pull

PR #292, opened 2019-02-22
adopt aggregator-signer

deads2k:
Instead of generating a new aggregotor signer, adopt the installer's one day cert.


deads2k:
adoption works.

PR #294, opened 2019-02-22
adopt controlplane client signer

deads2k:
There is a short lived controlplane signer we can adopt and use to avoid unifying trust.


deads2k:
cadvisor

/retest

deads2k:
cadvisor is flaking in other pulls too, the rest passed.

PR #298, opened 2019-02-24
extend trust for new csr

deads2k:


deads2k:
Impressive that adding trust causes some machine thing to fail.

deads2k:
/hold

trying https://github.com/openshift/cluster-kube-apiserver-operator/pull/299

this MCD thing is crazy....

deads2k:
we created a new configmap to get around teh MCO

PR #299, opened 2019-02-24
trust the kubelet-signer as client

deads2k:


cgwalters:
> If you require bootstrap cluster information, can you add yourself as a renderer to the installer?  if you want to consume, can you consume from the output of operators, not their installer inputs

To be clear the MCO isn't *parsing* this data (AFAIK).  It is acting as a mechanism to lay down files.  The mechanical reason this PR is failing is because the `MachineConfig` object rendered during the bootstrap phase differs from the cluster.

And that's because...this changes the CA certs dropped down at install time, but it looks like the MCO's code to derive the "in cluster" configs would need updating?

https://github.com/openshift/machine-config-operator/blob/17c4ed04dac5f787c19b749b62730e2959dbcd8e/pkg/operator/operator.go#L282

deads2k:
/hold
we just obsoleted this because the MCO was pinning on this to mean something else, but the MCO team doesn't know why they are using it.

@cgwalters  fyi

PR #302, opened 2019-02-25
bump(library-go): proof for https://github.com/openshift/library-go/pull/275

sttts:
Proves https://github.com/openshift/library-go/pull/275.


sttts:
I don't think we need this because empty ResourceVersion means to bypass cache.

PR #304, opened 2019-02-25
reject upgradeable if unsupportedConfigOverrides set

deads2k:
Adds a control loop to set an upgradeable condition to false if unsupportedConfigOverrdies are set.

I will separately add metrics since I have to lookup what kind of metric to use.


deads2k:
after editing a key

```
  - lastTransitionTime: 2019-02-25T16:39:03Z
    message: 'UnsupportedConfigOverridesUpgradeable: setting: [some]'
    reason: UnsupportedConfigOverridesUpgradeable
    status: "False"
    type: Upgradeable

```

mfojtik:
LGTM

deads2k:
bumped

PR #306, opened 2019-02-26
default the registry internal hostname to the correct expected value

bparees:


bparees:
@deads2k fyi i believe this is where the initial/wrong default was coming from.

mfojtik:
@bparees 👍 

I think this will prevent 1 restart :-)

PR #308, opened 2019-02-26
minor fitting to set versions for our operator

deads2k:
based on comments in https://github.com/openshift/cluster-image-registry-operator/pull/219, these keys should function properly with the payload we have today.


smarterclayton:
Checking, maybe I broke this.

deads2k:
@smarterclayton we look good: 
```
                "versions": [
                    {
                        "name": "operator",
                        "version": "0.0.1-2019-03-01-193053"
                    },
                    {
                        "name": "kube-apiserver",
                        "version": "1.12.4"
                    }
```

smarterclayton:
Yes that is correct.

PR #311, opened 2019-02-27
do not default the internal registry hostname to anything

bparees:


deads2k:
/lgtm

This correctly reflects the state of the system as not having a registry so that consumers of this value can distinguish between "no registry exists" and "a registry exists" state.

PR #312, opened 2019-02-27
Generate CRD schema

damemi:
This is an update from https://github.com/openshift/cluster-kube-apiserver-operator/pull/273 which uses the new fork of controller-tools checked into library-go to set `optional` as the default for fields


damemi:
cc @sttts

manifests/0000_10_kube-apiserver-operator_01_config.crd.yaml
```diff
@@ -12,3 +12,202 @@ spec:
...
+                their operands.
+              type: string
+            managementState:
+              description: managementState indicates whether and how the operator
+                should manage the component
+              type: string
+            observedConfig:
+              description: observedConfig holds a sparse config that controller has
+                observed from the cluster state.  It exists in spec because it is
+                an input to the level for the operator +nullable
```
damemi: note: this shouldn't be here, see https://github.com/openshift/kubernetes-sigs-controller-tools/pull/3#discussion_r261752842

glide.lock
```diff
@@ -827,4 +850,14 @@ imports:
   - pkg/handler
   - pkg/util
   - pkg/util/proto
+- name: sigs.k8s.io/controller-tools
+  version: 43466124052c1a1aa7d8fd33624b00bc111fb7cf
+  repo: https://github.com/openshift/kubernetes-sigs-controller-tools.git
```
soltysh: This needs to be backed by a proper `glide.yaml` entry to ensure every `make update-deps` picks that and nothing else.

soltysh:
/hold
until `glide.yaml` gets proper fix

damemi:
covered in https://github.com/openshift/cluster-kube-apiserver-operator/pull/273

PR #316, opened 2019-03-01
remove cert rotation

deads2k:
Remove cert rotation as an easier to support alternative compared to handling customization of rotation parameters by cluster admins proposed here: https://github.com/openshift/api/pull/199#discussion_r261408978

/hold

holding while the different options are considered.


deads2k:
kube-apiserver ran on the first try.  cool

deads2k:
We have a new plan that won't support configuration, so we don't need this

PR #317, opened 2019-03-01
operator: add init container to fix audit permissions

mfojtik:
Using init container we can get rid of the bash wrapper in apiserver pod and get rid of ugly string append for setting the log level (and any future extra argument we will need).

/cc @sttts 
/cc @deads2k


bindata/v3.11.0/kube-apiserver/pod.yaml
```diff
@@ -9,17 +9,22 @@ metadata:
     revision: "REVISION"
   deletionGracePeriodSeconds: 65 # a bit more than the kube-apiserver shutdown timeout of 60 sec
 spec:
+  initContainers:
```
deads2k: can static pods have init containers?

mfojtik: yes, the audit log files are in artifacts

PR #318, opened 2019-03-01
extend minimum duration

deads2k:
extend the shutdown duration to catch iptables updates.  This is long and will reduce the rollout rate, so we really want https://github.com/openshift/cluster-kube-apiserver-operator/pull/308 to help speed us back up/

### DIFF
--- a/bindata/v3.11.0/kube-apiserver/defaultconfig.yaml
+++ b/bindata/v3.11.0/kube-apiserver/defaultconfig.yaml
@@ -31,7 +31,7 @@ apiServerArguments:
   feature-gates:
   - PersistentLocalVolumes=false # disable local volumes for 4.0, owned by sig-storage/hekumar@redhat.com
   minimal-shutdown-duration:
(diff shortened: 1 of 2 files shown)


sttts:
LGTM. Live cert reload will help to reduce impact.

mfojtik:
I would rather not add 90s for each API server rollout especially when live cert reload can reduce the number of rollouts and decrease the time we are available... this seems like we are going to trade the time we gain to mitigate SDN issue.

squeed:
Interesting.

squeed:
Interesting. This was because of that issue we saw where an iptables-save failed to roll out, right @sttts? Given that we retry syncs every 30 seconds anyways, 90 seconds might be too large.

sttts:
/lgtm

Please rebase asap.

soltysh:
Closing in favor of https://github.com/openshift/cluster-kube-apiserver-operator/pull/352

PR #320, opened 2019-03-04
bring in new cert command

deads2k:
proof of https://github.com/openshift/library-go/pull/295


deads2k:
hold for upper lgtm. 
There is discussion with @mfojtik

PR #325, opened 2019-03-05
Disallow Unmanaged+Removed for management state

sttts:
Based on https://github.com/openshift/cluster-kube-apiserver-operator/pull/273.

Last commit: replaces https://github.com/openshift/cluster-kube-apiserver-operator/pull/324 in a couple of lines of yaml.


manifests/0000_10_kube-apiserver-operator_01_config.crd.yaml-merge-patch
```diff
@@ -0,0 +1,8 @@
+spec:
+  validation:
+    openAPIV3Schema:
+      properties:
+        spec:
+          properties:
+            managementState:
+              pattern: "^Managed|Removed$"
```
damemi: should probably add unmanaged and force too? https://github.com/openshift/api/blob/master/operator/v1/types.go#L29-L39

sttts: yes, will update. Was guessing here, just to prove it works.

mfojtik: no, for kube apiserver we only support Managed

damemi: @mfojtik does every operator have different supported states out of those 4? If that's the case then we won't be able to define a pattern tag in the api to generate this in the future

sttts: Compare https://github.com/openshift/cluster-kube-apiserver-operator/pull/324, which was the motivation for this PR.

PR #326, opened 2019-03-05
use dynamic reloading of certs to avoid rolling out new deployment versions

deads2k:
since https://github.com/openshift/origin/pull/22229 merged, we can reload most

We need https://github.com/openshift/origin/pull/22239 to reload the aggregator client

/hold


deads2k:
/retest

after this we just need to make the syncer suicide and switch the rest.

PR #330, opened 2019-03-07
README/Makefile: replace origin-release with oc-adm-release

sttts:
Still smoke testing.


damemi:
/lgtm

this step is the only way I could actually get a custom release image to work

PR #331, opened 2019-03-07
copy observed cert secrets and configmaps to correct namespace

sanchezl:
This should fix https://bugzilla.redhat.com/show_bug.cgi?id=1685704.


deads2k:
This breaks the user-CA, right?  https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/pkg/operator/targetconfigcontroller/targetconfigcontroller.go#L274

It's not a thing we can unit test, but it does highlight a need for the e2e operator tests.

sanchezl:
> This breaks the user-CA, right? https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/pkg/operator/targetconfigcontroller/targetconfigcontroller.go#L274
> 
> It's not a thing we can unit test, but it does highlight a need for the e2e operator tests.

Will add e2e test in another PR. Currently targetcontroller is not being notified when a new config map is created.

test/e2e/user_certs_test.go
```diff
@@ -0,0 +1,157 @@
...
+	test.FatalIfErr(t, err)
+	configClient, err := configclient.NewForConfig(kubeConfig)
+	test.FatalIfErr(t, err)
+	operatorClient, err := operatorclient.NewForConfig(kubeConfig)
+	test.FatalIfErr(t, err)
+
+	initialConfigGeneration := test.GetKubeAPIServerOperatorConfigGeneration(t, operatorClient)
+
+	t.Log("Given the kube-apiserver-operator is available.")
+	test.WaitForKubeAPIServerClusterOperatorAvailableNotProgressingNotFailing(t, configClient)
```
sttts: that's a func name :-)

test/library/library.go
```diff
@@ -0,0 +1,35 @@
...
+	"path"
+	"regexp"
+	"runtime"
+	"testing"
+	"time"
+)
+
+var (
+	waitPollInterval = time.Second
+	waitPollTimeout  = 10 * time.Minute
```
sttts: we have `ForeverTestTimeout` in `k8s.io/apimachinery/pkg/util/wait`.

test/library/library.go
```diff
@@ -0,0 +1,35 @@
...
+	"math/big"
+	"path"
+	"regexp"
+	"runtime"
+	"testing"
+	"time"
+)
+
+var (
+	waitPollInterval = time.Second
```
sttts: not sure this const makes sense. Poll intervals depend a lot on the actual thing you are polling, don't they?

test/library/library.go
```diff
@@ -0,0 +1,35 @@
...
+	waitPollTimeout  = 10 * time.Minute
+)
+
+func GenerateNameForTest(t *testing.T, prefix string) string {
```
sttts: which kind of name?

test/library/library.go
```diff
@@ -0,0 +1,35 @@
...
+	middle := regexp.MustCompile("[^a-zA-Z0-9]+").ReplaceAll([]byte(t.Name()), []byte("-"))
```
sttts: what does this do?

test/library/library.go
sanchezl: `ForeverTestTimeout` is only 30 seconds.

test/library/library.go
sanchezl: Added godoc.

test/library/library.go
sanchezl: Added godoc.

test/library/library.go
sanchezl: I guess as we add more tests we could stop using this constant if needed.

sttts: Am against using consts before we need them. I know @mfojtik disagrees :)

test/library/crypto.go
```diff
@@ -0,0 +1,87 @@
...
+		SerialNumber:          serialNumber,
+		KeyUsage:              x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign,
+		BasicConstraintsValid: true,
+		IsCA: true,
+	}
+	if parent == nil {
+		parent = template
+	}
+	var der []byte
+	if der, err = x509.CreateCertificate(rand.Reader, template, parent, result.PrivateKey.Public(), result.PrivateKey); err != nil {
```
sttts: wondering whether we want a fixture for that. We had tests before which failed because of low entropy in the kernel or something.

test/library/library.go
sttts: randomResourceName would be a good func name. says everything one needs to know.

test/library/library.go
```diff
@@ -0,0 +1,42 @@
...
+	n, err := rand.Int(rand.Reader, big.NewInt(math.MaxInt64))
+	FatalIfErr(t, err)
+	name := []byte(fmt.Sprintf("%s%s-%016x", prefix, t.Name(), n.Int64()))
+	// make the name (almost) suitable for use as a dns label
+	// only a-z, 0-9, and '-' allowed
+	name = regexp.MustCompile("[^a-zA-Z0-9]+").ReplaceAll(name, []byte("-"))
+	// collapse multiple `-`
+	name = regexp.MustCompile("-+").ReplaceAll(name, []byte("-"))
+	// ensure no `-` at beginning or end
+	return strings.Trim(string(name), "-")
```
sttts: this func looks like a perl program: everything is a regex :D

test/library/library.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// FatalIfErr marks the test as failed if there was an error.
+func FatalIfErr(t *testing.T, err error) {
```
sttts: let's use `github.com/stretchr/testify/assert`. It already has all those helpers.

test/library/crypto.go
sttts: we can watch the test. No need to do that now.

test/library/crypto.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+type CryptoMaterials struct {
+	PrivateKey  *rsa.PrivateKey
+	Certificate *x509.Certificate
+}
+
+// NewServerCertificate returns crypto materials suitable for use by a server. The hosts specified will be added as
+// subject alternate names.
+func NewServerCertificate(t *testing.T, signerCert *x509.Certificate, hosts ...string) *CryptoMaterials {
```
sttts: don't we have helpers in library-go which do the same?

sttts:
Still don't understand why all the changes are in one commit. I understand the namespace change (and the adapted tests). But what is the role of TestNamedCertificates? What does it have to do with a namespace change?

test/e2e/user_certs_test.go
```diff
@@ -0,0 +1,162 @@
...
+			_, err := configClient.APIServers().Update(apiserver)
+			test.FatalIfErr(t, err)
+			test.WaitForKubeAPIServerClusterOperatorAvailableNotProgressingNotFailing(t, configClient)
+		}
+	}()
+	apiServer, err := configClient.APIServers().Get("cluster", metav1.GetOptions{})
+	if errors.IsNotFound(err) {
+		apiServer, err = configClient.APIServers().Create(&configv1.APIServer{ObjectMeta: metav1.ObjectMeta{Name: "cluster"}})
+	}
+	test.FatalIfErr(t, err)
```
sttts: can this race? a Get, another actor which creates, and then your create => boom

test/e2e/user_certs_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+	t.Log("Given the kube-apiserver-operator is available.")
+	test.WaitForKubeAPIServerClusterOperatorAvailableNotProgressingNotFailing(t, configClient)
+
+	t.Log("And some new tls Secrets have been created in the openshift-config namespace.")
+	rootCA := test.NewCertificateAuthorityCertificate(t, nil)
+
+	type testCase struct {
+		serverName           string
+		expectedSerialNumber string
```
sttts: docs

test/e2e/user_certs_test.go
```diff
@@ -0,0 +1,162 @@
...
+	"k8s.io/client-go/util/cert"
+
+	configv1 "github.com/openshift/api/config/v1"
+	configclient "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+	operatorclient "github.com/openshift/client-go/operator/clientset/versioned/typed/operator/v1"
+
+	test "github.com/openshift/cluster-kube-apiserver-operator/test/library"
+)
+
+func TestNamedCertificates(t *testing.T) {
```
sttts: have a hard time to follow what this actually tests. A 3 liner doc would help a lot.

test/e2e/user_certs_test.go
```diff
@@ -0,0 +1,162 @@
...
+	operatorClient, err := operatorclient.NewForConfig(kubeConfig)
+	test.FatalIfErr(t, err)
+
+	initialConfigGeneration := test.GetKubeAPIServerOperatorConfigGeneration(t, operatorClient)
+
+	t.Log("Given the kube-apiserver-operator is available.")
+	test.WaitForKubeAPIServerClusterOperatorAvailableNotProgressingNotFailing(t, configClient)
+
+	t.Log("And some new tls Secrets have been created in the openshift-config namespace.")
+	rootCA := test.NewCertificateAuthorityCertificate(t, nil)
```
sttts: the log does not match: this does not create anything in the openshift-config namespace

test/e2e/user_certs_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+	type testCase struct {
+		serverName           string
+		expectedSerialNumber string
+	}
+	var testCases []testCase
+
+	var namedCertificates []configv1.APIServerNamedServingCert
```
sttts: what is this? For what will it be used.

test/e2e/user_certs_test.go
```diff
@@ -0,0 +1,162 @@
...
+			testCases = append(testCases, testCase{
+				serverName:           host,
+				expectedSerialNumber: tls.Certificate.SerialNumber.String(),
+			})
+		}
+	}
+
+	t.Log("When the cluster is configured the use of the named certificates.")
+	defer func() {
+		t.Log("Restoring global cluster configuration")
```
sttts: wondering, everything we do here in the e2e test, are there side-effect to other tests? Are there other tests running at all? <--- am just curious. Don't know the e2e context for our operators.

sanchezl:
I will split out the e2e test into a second PR and resolve comments by @sttts there.

PR #332, opened 2019-03-08
defaultconfig: bump http2-max-streams-per-connection to 2000

mfojtik:
To mitigate https://github.com/kubernetes/kubernetes/issues/74412

/cc @sttts 
/cc @deads2k 
/cc @tnozicka

### DIFF
--- a/bindata/v3.11.0/kube-apiserver/defaultconfig.yaml
+++ b/bindata/v3.11.0/kube-apiserver/defaultconfig.yaml
@@ -32,6 +32,8 @@ apiServerArguments:
   - PersistentLocalVolumes=false # disable local volumes for 4.0, owned by sig-storage/hekumar@redhat.com
   minimal-shutdown-duration:
   - 3s # give SDN some time to converge
+  http2-max-streams-per-connection:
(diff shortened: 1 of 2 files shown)


mfojtik:
@sttts @deads2k looks like this flag makes apiserver to crash loop...

deads2k:
> @sttts @deads2k looks like this flag makes apiserver to crash loop...

got a log?

bindata/v3.11.0/kube-apiserver/defaultconfig.yaml
```diff
@@ -32,6 +32,8 @@ apiServerArguments:
   - PersistentLocalVolumes=false # disable local volumes for 4.0, owned by sig-storage/hekumar@redhat.com
   minimal-shutdown-duration:
   - 3s # give SDN some time to converge
+  http2-max-streams-per-connection:
+  - 2000  # recommended is 1000, but we need to mitigate https://github.com/kubernetes/kubernetes/issues/74412
```
deads2k: "2000" to make sure it serializes as a string

mfojtik: i bet that was it! good catch

PR #336, opened 2019-03-14
targetconfig: if apiserver shutdown error occur do not set failing=true but retry

mfojtik:
Fixes this CI flake:

`Mar 14 04:37:14.404 E clusteroperator/kube-apiserver changed Failing to True: TargetConfigControllerFailing: TargetConfigControllerFailing: "v3.11.0/kube-apiserver/ns.yaml" (string): an error on the server ("apiserver is shutting down.") has prevented the request from succeeding (get namespaces openshift-kube-apiserver)`

### DIFF
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -10,7 +10,17 @@ import (
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
 
+	operatorv1 "github.com/openshift/api/operator/v1"
+	operatorv1client "github.com/openshift/client-go/operator/clientset/versioned/typed/operator/v1"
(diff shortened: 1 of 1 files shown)


pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -150,43 +150,54 @@ func (c TargetConfigController) sync() error {
...
+			continue
+		}
 		if currResult.Error != nil {
-			errors = append(errors, fmt.Errorf("%q (%T): %v", currResult.File, currResult.Type, currResult.Error))
+			allErrors = append(allErrors, fmt.Errorf("%q (%T): %v", currResult.File, currResult.Type, currResult.Error))
 		}
 	}
 
+	// If we hit the apiserver shutdown error, retry without setting failing=true condition.
+	if hasServiceUnavailableError {
```
sttts: why is this special? Why not other errors as well?

PR #338, opened 2019-03-18
Bump cert rotation to 30 days

soltysh:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1688610

/assign @deads2k @sttts 

For now:
/hold

### DIFF
--- a/pkg/operator/certrotationcontroller/certrotationcontroller.go
+++ b/pkg/operator/certrotationcontroller/certrotationcontroller.go
@@ -21,6 +21,9 @@ import (
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
+// rotationDay is one day. Set this to short values for testing, e.g. 5 minutes.
+const rotationDay = 24 * time.Hour
+
 type CertRotationController struct {
 	certRotators []*certrotation.CertRotationController
 
@@ -67,8 +70,8 @@ func NewCertRotationController(
 		certrotation.SigningRotation{
 			Namespace:     operatorclient.OperatorNamespace,
 			Name:          "aggregator-client-signer",
-			Validity:      8 * time.Hour, // to be 10 days
-			Refresh:       4 * time.Hour, // to be 4 days
+			Validity:      30 * rotationDay,
+			Refresh:       15 * rotationDay,
 			Informer:      kubeInformersForNamespaces.InformersFor(operatorclient.OperatorNamespace).Core().V1().Secrets(),
 			Lister:        kubeInformersForNamespaces.InformersFor(operatorclient.OperatorNamespace).Core().V1().Secrets().Lister(),
 			Client:        kubeClient.CoreV1(),
@@ -85,8 +88,8 @@ func NewCertRotationController(
 		certrotation.TargetRotation{
 			Namespace: operatorclient.TargetNamespace,
 			Name:      "aggregator-client",
-			Validity:  1 * 4 * time.Hour, // to be 5 days
-			Refresh:   2 * time.Hour,     // this could stay.
+			Validity:  30 * rotationDay,
+			Refresh:   15 * rotationDay,
 			CertCreator: &certrotation.ClientRotation{
 				UserInfo: &user.DefaultInfo{Name: "system:openshift-aggregator"},
 			},
@@ -107,8 +110,8 @@ func NewCertRotationController(
 		certrotation.SigningRotation{
 			Namespace:     operatorclient.OperatorNamespace,
 			Name:          "localhost-serving-signer",
-			Validity:      10 * 365 * 24 * time.Hour, // this comes from the installer
-			Refresh:       8 * 365 * 24 * time.Hour,  // this means we effectively do not rotate
+			Validity:      10 * 365 * rotationDay, // this comes from the installer
+			Refresh:       8 * 365 * rotationDay,  // this means we effectively do not rotate
 			Informer:      kubeInformersForNamespaces.InformersFor(operatorclient.OperatorNamespace).Core().V1().Secrets(),
 			Lister:        kubeInformersForNamespaces.InformersFor(operatorclient.OperatorNamespace).Core().V1().Secrets().Lister(),
 			Client:        kubeClient.CoreV1(),
@@ -125,8 +128,8 @@ func NewCertRotationController(
 		certrotation.TargetRotation{
 			Namespace: operatorclient.TargetNamespace,
 			Name:      "localhost-serving-cert-certkey",
-			Validity:  1 * 4 * time.Hour, // to be 5 days
-			Refresh:   2 * time.Hour,     // this could stay.
+			Validity:  30 * rotationDay,
+			Refresh:   15 * rotationDay,
 			CertCreator: &certrotation.ServingRotation{
 				Hostnames: func() []string { return []string{"localhost", "127.0.0.1"} },
 			},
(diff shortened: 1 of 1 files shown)


pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -309,8 +309,8 @@ func NewCertRotationController(
 		certrotation.SigningRotation{
 			Namespace:     operatorclient.OperatorNamespace,
 			Name:          "kube-control-plane-signer",
-			Validity:      8 * time.Hour, // to be 10 days
-			Refresh:       4 * time.Hour, // to be 4 days
+			Validity:      30 * 24 * time.Hour, // 30 days to be in-sync with cluster-signing-duration set for kube-cluster-manager
+			Refresh:       15 * 24 * time.Hour, // 50% of validity time to be able to respond to changes
```
sttts: what about all the other values? This does not help much, it's only changing the CA validity.

pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -309,8 +309,8 @@ func NewCertRotationController(
 		certrotation.SigningRotation{
 			Namespace:     operatorclient.OperatorNamespace,
 			Name:          "kube-control-plane-signer",
-			Validity:      8 * time.Hour, // to be 10 days
-			Refresh:       4 * time.Hour, // to be 4 days
+			Validity:      30 * 24 * time.Hour, // 30 days to be in-sync with cluster-signing-duration set for kube-cluster-manager
```
deads2k: signer should be good for 60 days

pkg/operator/certrotationcontroller/certrotationcontroller.go
deads2k: refreshed every 30 days

deads2k: > what about all the other values? This does not help much, it's only changing the CA validity.

yeah, you need to get the others

pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -327,8 +330,8 @@ func NewCertRotationController(
 		certrotation.TargetRotation{
 			Namespace: operatorclient.TargetNamespace,
 			Name:      "kube-apiserver-cert-syncer-client-cert-key",
-			Validity:  1 * 4 * time.Hour, // to be 5 days
-			Refresh:   2 * time.Hour,     // to be 1 day
+			Validity:  30 * rotationDay,
+			Refresh:   15 * rotationDay,
```
sttts: for the refresh I would have expected much shorter time, more like 1 or 2 days.

pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -21,6 +21,9 @@ import (
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
+// rotationDay is one day. Set this to short values for testing, e.g. 5 minutes.
+const rotationDay = 24 * time.Hour
```
mfojtik: can we make this var and get the value via function? so in case we want to make this much shorter (like 1 * minute) there is a way to do it (either by a hidden flag or env var.)

pkg/operator/certrotationcontroller/certrotationcontroller.go
soltysh: Why would you want to make it that short? I've spoke with @deads2k and for now this value is ok.

pkg/operator/certrotationcontroller/certrotationcontroller.go
gbraad: The constants as defined in the installer codebase was clearer in intent: https://github.com/openshift/installer/blob/0be3dfc7f6c6444d17d2edb86e2c52e9d008f9d7/pkg/asset/tls/tls.go#L21-L32

```
const (
	keySize = 2048

	// ValidityOneDay sets the validity of a cert to 24 hours.
	ValidityOneDay = time.Hour * 24

	// ValidityOneYear sets the validity of a cert to 1 year.
	ValidityOneYear = ValidityOneDay * 365

	// ValidityTenYears sets the validity of a cert to 10 years.
	ValidityTenYears = ValidityOneYear * 10
)
```

pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -67,8 +70,8 @@ func NewCertRotationController(
 		certrotation.SigningRotation{
 			Namespace:     operatorclient.OperatorNamespace,
 			Name:          "aggregator-client-signer",
-			Validity:      8 * time.Hour, // to be 10 days
-			Refresh:       4 * time.Hour, // to be 4 days
+			Validity:      30 * rotationDay,
```
gbraad: Suggested to say `ValidityOneMonth`

pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -107,8 +110,8 @@ func NewCertRotationController(
 		certrotation.SigningRotation{
 			Namespace:     operatorclient.OperatorNamespace,
 			Name:          "localhost-serving-signer",
-			Validity:      10 * 365 * 24 * time.Hour, // this comes from the installer
-			Refresh:       8 * 365 * 24 * time.Hour,  // this means we effectively do not rotate
+			Validity:      10 * 365 * rotationDay, // this comes from the installer
```
gbraad: Suggested `ValidityTenYears`

pkg/operator/certrotationcontroller/certrotationcontroller.go
sttts: I disagree. the day is the unit we might want to change for testing.

pkg/operator/certrotationcontroller/certrotationcontroller.go
sttts: same

pkg/operator/certrotationcontroller/certrotationcontroller.go
mfojtik: This is not normal users will tweak and we will discourage anyone from tweaking this except us debugging issues related with rotation where we need to decrease the interval to like 5 minutes.

I think as @sttts mentioned, we just need an unit that we can override, which is why I suggested this to not be a const but a func that can read the unit (minute, hour...) from either an env var or some hidden flag of operator binary.

mfojtik:
/lgtm

As a follow up I would like to have a way to tweak the base time unit (hour -> minute)... Will prefer a hidden flag for the operator binary, but env var will also be ok ;-)

pkg/operator/certrotationcontroller/certrotationcontroller.go
gbraad: in that case it might be handier to use a function that returns a consistent set of values

```
Validity : SigningRotationValidity(),
Refresh: SigningRotationRefresh(),
```

or similar, and have those functions handle the internal logic of time unit. This way you have to modify the values in a single location.

pkg/operator/certrotationcontroller/certrotationcontroller.go
gbraad: It was not meant to be configurable code, but cleaner code. Get the value from a function that handles the Refresh and Validity values creates consistency: https://github.com/openshift/cluster-kube-apiserver-operator/pull/338#discussion_r266485530

Seems it already got merged

PR #349, opened 2019-03-20
fix up owners files

bparees:


adambkaplan:
@bparees shouldn't someone from @openshift/sig-master own this?

adambkaplan:
/lgtm

disregard my prior comment

bparees:
@adambkaplan no, devex owns the image config observation/handling part.  (and other teams own the observation of their config objects)

PR #351, opened 2019-03-21
Remove cluster api dependency

enxebre:
Fix https://jira.coreos.com/browse/CLOUD-357
Remove dependency with cluster api cluster object:
- discoverIPs from network config
- fallback to discover from cluster api object

Also fixes an issue that was shadowing the returned err https://github.com/openshift/cluster-kube-apiserver-operator/pull/351/files#diff-5e45df1052a09901f6732e8a838541ebL206

Once we get this in we need to do the same for the controller manager.
Then we can merge https://github.com/openshift/installer/pull/1449
Then we can remove the fallback to discover from cluster api object

### DIFF
--- a/pkg/cmd/render/render.go
+++ b/pkg/cmd/render/render.go
@@ -150,9 +150,8 @@ func (r *renderOpts) Run() error {
 		if err != nil {
 			return err
 		}
-		err = discoverCIDRs(clusterConfigFileData, &renderConfig)
-		if err != nil {
-			return fmt.Errorf("unable to parse restricted CIDRs from cluster-api config %q: %v", r.clusterConfigFile, err)
+		if err = discoverCIDRs(clusterConfigFileData, &renderConfig); err != nil {
+			return fmt.Errorf("unable to parse restricted CIDRs from config %q: %v", r.clusterConfigFile, err)
 		}
 	}
 	if err := r.manifest.ApplyTo(&renderConfig.ManifestConfig); err != nil {
@@ -187,6 +186,15 @@ func mustReadTemplateFile(fname string) genericrenderoptions.Template {
 }
 
 func discoverCIDRs(clusterConfigFileData []byte, renderConfig *TemplateData) error {
+	if err := discoverCIDRsFromNetwork(clusterConfigFileData, renderConfig); err != nil {
+		if err = discoverCIDRsFromClusterAPI(clusterConfigFileData, renderConfig); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func discoverCIDRsFromNetwork(clusterConfigFileData []byte, renderConfig *TemplateData) error {
 	configJson, err := yaml.YAMLToJSON(clusterConfigFileData)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


pkg/cmd/render/render.go
```diff
@@ -140,9 +140,11 @@ func (r *renderOpts) Run() error {
 		if err != nil {
 			return err
 		}
-		err = discoverCIDRs(clusterConfigFileData, &renderConfig)
-		if err != nil {
-			return fmt.Errorf("unable to parse restricted CIDRs from cluster-api config %q: %v", r.clusterConfigFile, err)
+		if err = discoverCIDRsDeprecated(clusterConfigFileData, &renderConfig); err != nil {
```
abhinavdahiya: Wouldn't other way around make sense .. check desired if failed check deprecated?

pkg/cmd/render/fixtures/cluster-api-cluster.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: machine.openshift.io/v1beta1
+kind: Cluster
+metadata:
+  creationTimestamp: null
+  name: cluster
```
deads2k: this object may not be created here.  If the MCO needs this resource it must supply it's own render command or place it directly into the installer. The kube-apiserver-operator must not become a "this is a convenient place for other unrelated operators to place manifests".

enxebre: this is only for unit testing. The object is created by the installer. This component is consuming an object which needs to be removed see https://github.com/openshift/installer/pull/1449/files

deads2k:
/hold

The kas-operator isn't a spot for rendering resources produced and managed by other operators.  You'll need to create your own render command for the installer or convince @abhinavdahiya to let you place the code directly into the installer (and I think there's a reasonable case for this since the installer represents the information in some manner today).

sttts:
There is https://github.com/openshift/installer/blob/master/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template#L163. Isn't that a good place?

pkg/cmd/render/render.go
```diff
@@ -10,18 +10,17 @@ import (
...
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/v311_00_assets"
+	genericrender "github.com/openshift/library-go/pkg/operator/render"
+	genericrenderoptions "github.com/openshift/library-go/pkg/operator/render/options"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
-
```
sttts: we don't mix the imports. Please undo this change.

sttts:
/hold cancel
/approve

Lgtm.

@abhinavdahiya @squeed please take a look and tag if this is what you expect.

squeed:
Coming back from PTO. Quick question - is the configuration data being read the installer configuration, or the contents of `Network.config.openshift.io/v1/cluster`?

enxebre:
@squeed https://github.com/openshift/installer/blob/master/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template#L130

squeed:
@enxebre great, thanks!

PR #358, opened 2019-03-26
Fix copy operation to create proper /manifests/ dirs

soltysh:
https://github.com/openshift/cluster-kube-apiserver-operator/pull/354/commits/0bd73e26498bf2244f65ca0a9fba05ddbdc8ae4c broke manifests copying, this uses the proper format for copying dir contents.

/assign @sttts @smarterclayton


smarterclayton:
Wow, how did that merge?

soltysh:
Looks like it was problem on my end using old imagebuilder and I got additionally misguided by https://docs.docker.com/engine/reference/builder/#copy which states that dest dirs need to end with `/`.

PR #359, opened 2019-03-27
Move certrotation configmap to openshift-config namespace

soltysh:
This is to be able to tweak certrotation base from a single configmap for both kube operators. This is in-pair with https://github.com/openshift/cluster-kube-controller-manager-operator/pull/203. 

/assign @mfojtik


deads2k:
rather than twiddle this again, create the library-go function.

PR #361, opened 2019-04-01
configobservation: do not delete non-existing secrets

mfojtik:
This PR reduce the DELETE calls we make from config observation when syncing the named certificates. Currently this sync produces higher amount of API requests (all fail with 404):

`{client="cluster-kube-apiserver-operator/v0.0.0 (linux/amd64) kubernetes/$Format",code="404",contentType="application/vnd.kubernetes.protobuf",endpoint="https",job="apiserver",namespace="default",resource="secrets",scope="namespace",service="kubernetes",verb="DELETE"}	3.6703839643850538`

The logic here is to use lister and get all secrets/configs names into a string set. When syncing the resources and when we detect that target resource does not exists (is not in the set), instead of calling resource sync, which will call DELETE, we just skip it.

Fixes: 1/2 of https://bugzilla.redhat.com/show_bug.cgi?id=1694188


mfojtik:
/hold

logical bug

pkg/operator/configobservation/apiserver/observe_apiserver.go
```diff
@@ -238,7 +336,7 @@ func deleteSyncRules(names ...string) syncActionRules {
 
 // syncObservedResources copies or deletes resources, sources in GlobalUserSpecifiedConfigNamespace and destinations in OperatorNamespace namespace.
 // Errors are collected, i.e. it's not failing on first error.
-func syncObservedResources(syncResource resourceSyncFunc, syncRules syncActionRules) []error {
+func syncObservedResources(syncResource resourceSyncFunc, syncRules syncActionRules, shouldSyncResource resourceSyncCheckFunc) []error {
```
sttts: how about renaming shouldSyncResource to `syncer` and wrapping `syncResource`?

pkg/operator/configobservation/apiserver/observe_apiserver.go
```diff
@@ -220,11 +272,57 @@ func (o *apiServerObserver) observe(genericListers configobserver.Listers, recor
...
+	errs = append(errs, syncObservedResources(resourceSync, resourceSyncRules, resourceSyncCheck)...)
 
 	return observedConfig, errs
 }
 
+type resourceSyncCheckFunc = func(source, target resourcesynccontroller.ResourceLocation) bool
+
+// shouldSyncSecrets is used to check if we should initiate sync between source and target secret resource.
+// This exists to prevent deletion of target resource on every sync.
+func shouldSyncSecrets(source, destination resourcesynccontroller.ResourceLocation, lister v1.SecretLister) bool {
```
sttts: how about `syncResourceWithCache(source, destination, inCache) func(source, dest)`, i.e. dynamically?

pkg/operator/configobservation/apiserver/observe_apiserver.go
```diff
@@ -182,22 +182,74 @@ type apiServerObserver struct {
...
+}
+
+func (o *apiServerObserver) observeConfigMaps(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
+	listers := genericListers.(configobservation.Listers)
+	var errs []error
 
+	resourceSync := listers.ResourceSyncer().SyncConfigMap
 	previouslyObservedConfig, errs := extractPreviouslyObservedConfig(existingConfig, o.configPaths...)
 
+	resourceSyncCheck := func(source, target resourcesynccontroller.ResourceLocation) bool {
```
sttts: `syncResourceWithTargetCache(source, target, func(target) bool { .... })`

mfojtik: fixed.

mfojtik:
@sttts ready for next pass, now with green unit test

pkg/operator/configobservation/apiserver/observe_apiserver.go
```diff
@@ -220,11 +273,30 @@ func (o *apiServerObserver) observe(genericListers configobserver.Listers, recor
...
+	errs = append(errs, syncObservedResources(resourceSync, resourceSyncRules, resourceExists)...)
 
 	return observedConfig, errs
 }
 
+type resourceExistsFunc = func(resourcesynccontroller.ResourceLocation) bool
+
+func syncResourceWithTargetCache(source, destination resourcesynccontroller.ResourceLocation, resourceSync resourceSyncFunc, resourceExists resourceExistsFunc) error {
+	// if source name is not set, it does not exists
+	sourceExists := len(source.Name) > 0
```
sttts: `sourceExists := len(source.Name) > 0 && resourceExists(source)`

deads2k:
/hold

I was expecting a GET here https://github.com/openshift/library-go/blob/master/pkg/operator/resourcesynccontroller/resourcesync_controller.go#L152 and we assume that we use the caching getter.

pkg/operator/configobservation/apiserver/observe_apiserver.go
```diff
@@ -238,7 +310,7 @@ func deleteSyncRules(names ...string) syncActionRules {
 
 // syncObservedResources copies or deletes resources, sources in GlobalUserSpecifiedConfigNamespace and destinations in OperatorNamespace namespace.
 // Errors are collected, i.e. it's not failing on first error.
-func syncObservedResources(syncResource resourceSyncFunc, syncRules syncActionRules) []error {
+func syncObservedResources(syncResource resourceSyncFunc, syncRules syncActionRules, resourceExists resourceExistsFunc) []error {
```
sttts: would have passed the whole value of `syncResourceWithTargetCache(destination, source, syncResource, resourceExists)` as `syncResource` parameter here. Wouldn't that be simpler?

sanchezl:
Should we consider changing the [ObserverFunc](https://github.com/openshift/cluster-kube-apiserver-operator/blob/1ef690e485222e559ea0aa27388abcb2620fc372/vendor/github.com/openshift/library-go/pkg/operator/configobserver/config_observer_controller.go#L41) to return the sync rules and move the sync logic to library-go for all controllers to use?

mfojtik:
@sttts @sanchezl with library-go change, i don't think we have to do anything here. The resource sync controller will always issue cached GET before calling delete and that will prevent the DELETE requests   on non-existing resources.

PR #362, opened 2019-04-01
use the new installer certs to serve

deads2k:
switch to the specific kube-apiserver serving certs for serving the API.

### DIFF
--- a/pkg/operator/certrotationcontroller/certrotationcontroller.go
+++ b/pkg/operator/certrotationcontroller/certrotationcontroller.go
@@ -216,7 +216,7 @@ func NewCertRotationController(
 		},
 		certrotation.TargetRotation{
 			Namespace: operatorclient.TargetNamespace,
-			Name:      "loadbalancer-serving-certkey",
+			Name:      "external-loadbalancer-serving-certkey",
 			Validity:  30 * rotationDay,
 			Refresh:   15 * rotationDay,
 			CertCreator: &certrotation.ServingRotation{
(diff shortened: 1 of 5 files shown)


bindata/bootkube/config/config-overrides.yaml
```diff
@@ -6,9 +6,6 @@ kubeletClientInfo:
   keyFile: /var/run/secrets/kubelet-client/tls.key
 serviceAccountPublicKeyFiles:
 - /var/run/configmaps/sa-token-signing-certs/ca-bundle.crt
-servingInfo:
```
sttts: where are these set now? in defaultconfig.yml?

deads2k: > where are these set now? in defaultconfig.yml?

the bootstrap has another override file that already has these.

deads2k:
@abhinavdahiya @sdodson what's the ETA for bootstrap container logs and master container logs on failing installs?

abhinavdahiya:
This PR has not gone green even once. This does not look like a flake.

Can you hold and see if the actually works? Because otherwise the bot is going to keep retesting.


Also I didn't see any exception for this.

sdodson:
/hold
Clear this after we believe the tests will pass.

deads2k:
Hmmm... One node failing?

/retest

deads2k:
@derekwaynecarr this is borked, but without bootstrap logs I can't figure out why the operators are failing. 
 We need this pull (well, one nearly identical), to land the fixes for the kubelet/kube-apiserver trust.

PR #364, opened 2019-04-01
use new certs for service network and localhost

deads2k:
The load balancer has not yet been switched.

/assign @sanchezl

### DIFF
--- a/bindata/bootkube/manifests/config.apiserver.yaml
+++ b/bindata/bootkube/manifests/config.apiserver.yaml
@@ -1,0 +1,4 @@
+apiVersion: config.openshift.io/v1
+kind: APIServer
+metadata:
+  name: cluster
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -126,6 +126,7 @@ func (c TargetConfigController) sync() error {
 	}
 	requiredPaths := [][]string{
 		{"serviceAccountPublicKeyFiles"},
+		{"servingInfo", "namedCertificates"},
 	}
 	for _, requiredPath := range requiredPaths {
 		_, found, _ := unstructured.NestedFieldNoCopy(existingConfig, requiredPath...)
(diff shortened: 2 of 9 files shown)


bindata/v3.11.0/kube-apiserver/defaultconfig.yaml
```diff
@@ -96,9 +96,15 @@ servicesSubnet: 10.3.0.0/16 # ServiceCIDR
...
   bindNetwork: tcp4
+<<<<<<< HEAD
   certFile: /etc/kubernetes/static-pod-certs/secrets/serving-cert/tls.crt
   clientCA: /etc/kubernetes/static-pod-certs/configmaps/client-ca/ca-bundle.crt
   keyFile: /etc/kubernetes/static-pod-certs/secrets/serving-cert/tls.key
+=======
+  certFile: /etc/kubernetes/static-pod-resources/secrets/service-network-serving-certkey/tls.crt
+  clientCA: /etc/kubernetes/static-pod-resources/configmaps/client-ca/ca-bundle.crt
+  keyFile: /etc/kubernetes/static-pod-resources/secrets/service-network-serving-certkey/tls.key
+>>>>>>> f3d97c30... serve using SNI and rotated certificates
```
sdodson: merge conflict cruft?

deads2k:
looks like it mostly worked.  I need to update some e2e tests.

deads2k:
this moves us closer.  Starting here.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -126,6 +126,7 @@ func (c TargetConfigController) sync() error {
 	}
 	requiredPaths := [][]string{
 		{"serviceAccountPublicKeyFiles"},
+		{"servingInfo", "namedCertificates"},
```
sttts: this break upgrade. My cluster does not have this field.

bindata/bootkube/manifests/config.apiserver.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: config.openshift.io/v1
+kind: APIServer
+metadata:
+  name: cluster
```
sttts: this does not validate: spec is required

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
deads2k: > this breaks upgrade. My cluster does not have this field.

The control loop for configobservation will add it as soon as it runs.

PR #366, opened 2019-04-03
Report operand version using environment variable set in operator deployment

mfojtik:
builds on https://github.com/openshift/library-go/pull/326

PR #369, opened 2019-04-03
Report operand version using environment variable set in operator deployment

deads2k:
fixes typo from https://github.com/openshift/cluster-kube-apiserver-operator/pull/366


deads2k:
I suspect the rebase broke SNI.  If https://github.com/openshift/cluster-kube-apiserver-operator/pull/370 shows the same failure, I plan to manually merge.

deads2k:
I pinned everything to avoid unexpected interactions.  If this still fails aws-operator, I'm planning to merge anyway and @sanchezl can try to figure out what about the resource sync update broke the test.  The check itself looks correct.

PR #373, opened 2019-04-05
Set terminationGracePeriodSeconds in pod spec

sttts:
Setting deletionGracePeriodSeconds in ObjectMeta is just ignored and never had any effect.

Follow-up TODO: verify that cri-o+kubelet follow this value.

### DIFF
--- a/bindata/bootkube/bootstrap-manifests/kube-apiserver-pod.yaml
+++ b/bindata/bootkube/bootstrap-manifests/kube-apiserver-pod.yaml
@@ -45,6 +45,7 @@ spec:
         path: healthz
       initialDelaySeconds: 10
       timeoutSeconds: 10
+  terminationGracePeriodSeconds: 100 # bit more than 35s (minimal termination period) + 60s (apiserver graceful termination)
   volumes:
   - hostPath:
       path: {{ .SecretsHostPath }}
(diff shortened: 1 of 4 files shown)


bindata/bootkube/bootstrap-manifests/kube-apiserver-pod.yaml
```diff
@@ -45,6 +45,7 @@ spec:
         path: healthz
       initialDelaySeconds: 10
       timeoutSeconds: 10
+  terminationGracePeriodSeconds: 100 # bit more than 35s (minimal termination period) + 60s (apiserver graceful termination)
```
tnozicka: this should be lower by say at leas 5 seconds then

https://github.com/openshift/cluster-kube-apiserver-operator/blob/d754b48/bindata/v3.11.0/kube-apiserver/pod.yaml#L23

(or the other one higher)

tnozicka: I guess the other one could be 2* as high then, if the apiserver doesn't terminate within 100 s and is hard killed that adds 60 s by kernel TIME_WAIT so the init container should wait at least the sum of those 2 periods

sttts: The apiserver shuts down after 95s max. This is about cri-o enforcing the shutdown if the apiserver does not terminate by itself. We could add something to the timeout, I agree.

sttts: fixed

PR #376, opened 2019-04-09
config manifest: fix missing spec

sttts:
Spec is required: https://github.com/openshift/api/blob/master/config/v1/types_apiserver.go#L15


bindata/bootkube/manifests/config.apiserver.yaml
```diff
@@ -2,3 +2,4 @@ apiVersion: config.openshift.io/v1
 kind: APIServer
 metadata:
   name: cluster
+spec:
```
sttts: this wrong. I hate yaml: this translates to `spec: null`.

PR #377, opened 2019-04-09
Bump to 1.13

mfojtik:


glide.yaml
```diff
@@ -33,11 +32,11 @@ import:
 - package: github.com/apparentlymart/go-cidr
   version: v1.0.0
 - package: github.com/stretchr/testify
-  version: ~1.3.0
+  version: c679ae2cc0cb27ec3293fea7e254e47386f05d69
 
 # pinned values to land upgrade
 - package: github.com/evanphx/json-patch
-  version: 36442dbdb585210f8d5a1b45e67aa323c197d5c4
+  version: 5858425f75500d40c52783dce87d085a483ce135
```
mfojtik: build error otherwise

glide.yaml
```diff
@@ -33,11 +32,11 @@ import:
 - package: github.com/apparentlymart/go-cidr
   version: v1.0.0
 - package: github.com/stretchr/testify
-  version: ~1.3.0
+  version: c679ae2cc0cb27ec3293fea7e254e47386f05d69
```
mfojtik: glide complains that something require this SHA

glide.yaml
```diff
@@ -1,23 +1,22 @@
...
-  version: kubernetes-1.12.5
+  version: kubernetes-1.13.4
 - package: k8s.io/apiextensions-apiserver
-  repo: https://github.com/openshift/kubernetes-apiextensions-apiserver
-  version: 36be1826c980fa7ca82d91645e06e3157f159322
+  version: kubernetes-1.13.4
 - package: sigs.k8s.io/controller-tools
   repo: https://github.com/openshift/kubernetes-sigs-controller-tools
-  version: master
+  version: 43466124052c1a1aa7d8fd33624b00bc111fb7cf
```
mfojtik: glide complains about conflict

glide.yaml
```diff
@@ -1,23 +1,22 @@
...
+  version: kubernetes-1.13.4
 - package: k8s.io/kube-aggregator
-  version: kubernetes-1.12.5
+  version: kubernetes-1.13.4
 - package: k8s.io/code-generator
-  version: kubernetes-1.12.5
+  version: kubernetes-1.13.4
 - package: k8s.io/apiextensions-apiserver
-  repo: https://github.com/openshift/kubernetes-apiextensions-apiserver
-  version: 36be1826c980fa7ca82d91645e06e3157f159322
```
mfojtik: moved to upstream again?

sttts: no, due to nullable we need our fork, until @mdame's container is used for crd-schema-gen.

glide.yaml
soltysh: isn't there a specific version we can pin that will contain that commit?

mfojtik:
🤞

glide.yaml
mfojtik: we don't use this in our code, looks like transitive dep that something else pin to that SHA, so I would not be concerned?

sanchezl: That commit is in v1.2.2 and v1.3.0.

glide.yaml
```diff
@@ -1,43 +1,43 @@
...
-  version: 8ae2a9c33ba2a3d2ed5dc13d536ca935fc9625b9 # FIXME: pre-1.13-rebase SHA
+  version: master
 - package: github.com/openshift/library-go
-  version: temporary-1.12 # FIXME: This is temporary until we manage to merge 1.13 branch
+  version: master
 - package: github.com/jteeuwen/go-bindata
   version: a0ff2567cfb70903282db057e799fd826784d41d
 - package: k8s.io/gengo/args
-  version: 01a732e01d00cb9a81bb0ca050d3e6d2b947927b
+  version: master
```
deads2k: I expected this to be pinned, but I can allow this first.

glide.yaml
```diff
@@ -1,43 +1,43 @@
...
 - package: github.com/golang/glog
```
ravisantoshgudimetla: Do we still need `glog`?

PR #383, opened 2019-04-10
try plumbing the kubernetes service as an IP to SNI

deads2k:
This adds bits for bootstrap and non-bootstrap config.


deads2k:
needs a clean https://openshift-gce-devel.appspot.com/builds/origin-ci-test/logs/branch-ci-openshift-origin-master-images/ to try again

deads2k:
/assign @sanchezl @sttts 

This is one of the few stopping us from removing kube-ca altogether (so close!).  It's not pretty and it doesn't do a good job of multiple service networks, but it's definitely closer than what we have.

PR #384, opened 2019-04-10
fix the ca-bundle used by MCD

deads2k:
The kube-ca is dead.  This is used to terminate serving certs from the kube-apiserver

requires https://github.com/openshift/installer/pull/1593


sdodson:
/lgtm
/hold
Coordinating merge across repos

deads2k:
I got a good run locally with the two pulls. Prepare your keys!

PR #385, opened 2019-04-11
Make kube-apiserver audit log permissions more restrictive on start

JAORMX:
While the directory permissions are limited on starting the
openshift-kube-apiserver pod; this also initiates the audit logs
permissions to a more secure 0600.

This makes the underlying library (lumberjack) reuse these permissions
throughout the deployment, instead of keeping using the 0644 which was
the default.

Note that the file permissions in the directory were already providing
protection from accessing those files. This merely promotes good
patterns for the log files.

### DIFF
--- a/bindata/v3.11.0/kube-apiserver/pod.yaml
+++ b/bindata/v3.11.0/kube-apiserver/pod.yaml
@@ -12,7 +12,7 @@ spec:
     - name: fix-audit-permissions
       image: ${IMAGE}
       imagePullPolicy: IfNotPresent
-      command: ['/bin/sh', '-c', 'chmod 0700 /var/log/kube-apiserver']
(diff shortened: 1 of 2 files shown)


sttts:
How can we test that lumberjack keeps the permissions? Is that documented at least?

JAORMX:
@sttts It's not documented AFAIK, but here's the code where that happens https://github.com/natefinch/lumberjack/blob/v2.0/lumberjack.go#L216  basically, it reads if the file already exists and uses the permissions for that file.

JAORMX:
Note that this won't be needed once we update the lumberjack version that's being used: https://github.com/natefinch/lumberjack/pull/83

soltysh:
> Note that this won't be needed once we update the lumberjack version that's being used: [natefinch/lumberjack#83](https://github.com/natefinch/lumberjack/pull/83)

That ^ merge, I'd prefer we bump the lib rather than doing this.
/hold

JAORMX:
@soltysh that's fine with me.

PR #386, opened 2019-04-11
manifests: label certificate manifests as managed

mfojtik:
This adds label to all certificates we get from the installer in bootstrap. It allows them to show up in prometheus (so we don't have to wait for the rotated certs to rotate) and also allows to query them in the system.

### DIFF
--- a/bindata/bootkube/manifests/configmap-admin-kubeconfig-client-ca.yaml
+++ b/bindata/bootkube/manifests/configmap-admin-kubeconfig-client-ca.yaml
@@ -3,6 +3,8 @@ kind: ConfigMap
 metadata:
   name: admin-kubeconfig-client-ca
   namespace: openshift-config
+  labels:
+    "auth.openshift.io/managed-certificate-type": ca-bundle
 data:
   ca-bundle.crt: |
     {{ .Assets | load "admin-kubeconfig-ca-bundle.crt" | indent 4 }}
--- a/bindata/bootkube/manifests/configmap-csr-controller-ca.yaml
+++ b/bindata/bootkube/manifests/configmap-csr-controller-ca.yaml
@@ -3,6 +3,8 @@ kind: ConfigMap
 metadata:
   name: csr-controller-ca
   namespace: openshift-config-managed
+  labels:
+    "auth.openshift.io/managed-certificate-type": ca-bundle
 data:
   ca-bundle.crt: |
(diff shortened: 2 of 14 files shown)


bindata/bootkube/manifests/configmap-admin-kubeconfig-client-ca.yaml
```diff
@@ -3,6 +3,8 @@ kind: ConfigMap
 metadata:
   name: admin-kubeconfig-client-ca
   namespace: openshift-config
+  labels:
+    "auth.openshift.io/managed-certificate-type": ca-bundle
```
deads2k: danger!

Be really sure before you mess with this.  This data is gone forever if you change it.  @soltysh

bindata/bootkube/manifests/configmap-admin-kubeconfig-client-ca.yaml
```diff
@@ -3,6 +3,8 @@ kind: ConfigMap
 metadata:
   name: admin-kubeconfig-client-ca
   namespace: openshift-config
```
deads2k: things in this namespace cannot be replaced.

bindata/bootkube/manifests/configmap-admin-kubeconfig-client-ca.yaml
mfojtik: @deads2k i'm not going to change this, the label means the prometheus will pick this up and report the expiration for these certs.

mfojtik: for inreplacable certs, we might want some special annotation maybe "openshift.io/do-not-touch-this"

bindata/bootkube/manifests/configmap-admin-kubeconfig-client-ca.yaml
mfojtik: sure, but we can still report the hours till these expire via prometheus.

mfojtik:
/hold

re-open in 4.1

bindata/bootkube/manifests/configmap-admin-kubeconfig-client-ca.yaml
soltysh: Why do we have to manually label those? Isn't that your controller job? If these are the initial certs, do we care?

PR #396, opened 2019-04-12
use the CVO to create the operator config

deads2k:
The CVO supports create-only resources now, so we can use the CVO instead writing our own code a dozen times.

/assign @damemi


damemi:
/lgtm

farewell old friend

deads2k:
needs https://github.com/openshift/installer/pull/1561

deads2k:
/hold 

missing logs

PR #399, opened 2019-04-12
ObserveAuthMetadata: default authentication.spec.type

enj:
This prevents a state where the KAS will fail to honor metadata permanently.


pkg/operator/configobservation/auth/auth_metadata.go
```diff
@@ -39,7 +39,7 @@ func ObserveAuthMetadata(genericListers configobserver.Listers, recorder events.
 	}
 
 	observedConfig := map[string]interface{}{}
-	authConfig, err := listers.AuthConfigLister.Get("cluster")
+	authConfigNoDefaults, err := listers.AuthConfigLister.Get("cluster")
 	if errors.IsNotFound(err) {
```
deads2k: Based on some of our recent experiences, I think we should treat this like an empty object to default and continue.

PR #400, opened 2019-04-13
update the default cert to match

deads2k:
https://github.com/openshift/cluster-kube-apiserver-operator/pull/395 changed the default cert to avoid using kube-ca and still serve a working service IP address.  Somehow it merged by bot (still looking at that).  This should fix HEAD


deads2k:
hrm.

/retest

PR #402, opened 2019-04-13
stop trying to use generation without observedgeneration

deads2k:
We never set observedgeneration (oops), which means that trying to use generation is inexact at best.  I think we're rolling out and somehow not seeing the transitions.  This reduces the precision of the check, but may improve stability.


deads2k:
no one's here today.  Self lgtm to see if this helps.  We can revert the test later if necessary.

PR #403, opened 2019-04-13
disable user sni check

deads2k:
Alright, nuclear option.  SNI has to be working in order for us to terminate localhost for controller and scheduler and terminating kubelet load balancer connections.  Something is wrong with detection of completion of progressing.  This disables the naughty test to land other pulls.


deads2k:
https://github.com/openshift/cluster-kube-apiserver-operator/pull/400 passed, so this is just the flake from hell

PR #404, opened 2019-04-13
try to wait for actual success of sni rollout

deads2k:
alternative to https://github.com/openshift/cluster-kube-apiserver-operator/pull/402 and https://github.com/openshift/cluster-kube-apiserver-operator/pull/403 .  It tries until it actually works.  Hopefully this runs right and we don't have to disable.

### DIFF
--- a/test/e2e/user_certs_test.go
+++ b/test/e2e/user_certs_test.go
@@ -7,6 +7,9 @@ import (
 	"net/url"
 	"strings"
 	"testing"
+	"time"
+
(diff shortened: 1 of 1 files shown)


deads2k:
https://github.com/openshift/cluster-kube-apiserver-operator/pull/400 passed, so this is just the flake from hell

deads2k:
picked up by https://github.com/openshift/cluster-kube-apiserver-operator/pull/406

PR #405, opened 2019-04-15
add internal load balancer cert

deads2k:
Updates us to rotate teh internal load-balancer serving cert and serve with it.

/assign @mfojtik 
@sjenning


pkg/operator/certrotationcontroller/externalloadbalancer.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"k8s.io/klog"
+
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/tools/cache"
+)
+
+func (c *CertRotationController) syncExternalLoadBalancerHostnames() error {
+	hostnames := sets.NewString()
```
mfojtik: why is this set if we only use it to insert one item?

pkg/operator/certrotationcontroller/internalloadbalancer.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"k8s.io/klog"
+
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/tools/cache"
+)
+
+func (c *CertRotationController) syncInternalLoadBalancerHostnames() error {
+	hostnames := sets.NewString()
```
mfojtik: same question here

deads2k:
lgtm'd via slack

deads2k:
terraform

/retest

PR #407, opened 2019-04-15
remove resources we no longer use

deads2k:
Multiple refactors have led to confusing cruft that has a long life on our clusters.  This removes that we don't use or honor


mfojtik:
/retest

Ci flake (`release "release-latest" failed: pod release-latest was already deleted`)

PR #412, opened 2019-04-16
Increase minimum-shutdown-duration by 10s to 45s

sttts:
Alternative to https://github.com/openshift/cluster-kube-apiserver-operator/pull/411.


smarterclayton:
I think this is the one we actually want.  I'm pretty convinced before the 30s elb takeout and the 35s wait was too short, but now that we are 20 for ELB takeout I think 45s is sufficient.  Can you rebase this and let it run a few times?

deads2k:
replaced by https://github.com/openshift/cluster-kube-apiserver-operator/pull/433

/close

PR #413, opened 2019-04-16
Increase minimum-shutdown-duration by 35s to 70s

sttts:
Alternative to https://github.com/openshift/cluster-kube-apiserver-operator/pull/411


smarterclayton:
Will test and evaluate this weekend in concert with the health check interval decrease.

PR #415, opened 2019-04-17
Add audit policy ConfigMap observer

sttts:
This adds an observer of the openshift-config/audit-policy ConfigMap. It is supposed to contain a audit.k8s.io/v1beta1.Policy file. It is syntactically validated to be valid YAML, not that it can be decoded into a Policy struct (the API server will do that).

This PR does not add central configuration of other settings like audit webhooks.

### DIFF
--- a/pkg/operator/configobservation/audit/observe_audit_policy.go
+++ b/pkg/operator/configobservation/audit/observe_audit_policy.go
@@ -1,0 +1,74 @@
+package audit
+
+import (
+	gojson "encoding/json"
+	"fmt"
+
+	yaml2 "sigs.k8s.io/yaml"
+
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/configobservation"
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/operatorclient"
+)
+
+const policyConfigMapName = "audit-policy"
+
+// ObserveAuditPolicy observes the openshift-config/audit-policy ConfigMap and merges it into the kube-apiserver config.
+func ObserveAuditPolicy(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
+	listers := genericListers.(configobservation.Listers)
+	errs := []error{}
+	prevObservedConfig := map[string]interface{}{}
+
+	// copy non-empty .auditConfig.policyConfiguration from existingConfig to prevObservedConfig
+	policyPath := []string{"auditConfig", "policyConfiguration"}
+	existingPolicy, _, err := unstructured.NestedFieldNoCopy(existingConfig, policyPath...)
+	if err != nil {
+		return prevObservedConfig, append(errs, err)
+	}
+	if existingPolicy != nil {
+		if err := unstructured.SetNestedField(prevObservedConfig, existingPolicy, policyPath...); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	observedConfig := map[string]interface{}{}
+
+	// get optional openshift-config/audit-policy file, fallback to default config (i.e. observed config unset)
+	policyConfigMap, err := listers.ConfigmapLister.ConfigMaps(operatorclient.GlobalUserSpecifiedConfigNamespace).Get(policyConfigMapName)
+	switch {
+	case errors.IsNotFound(err):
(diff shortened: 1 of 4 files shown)


pkg/operator/configobservation/audit/observe_audit_policy.go
```diff
@@ -0,0 +1,74 @@
...
+
+// ObserveAuditPolicy observes the openshift-config/audit-policy ConfigMap and merges it into the kube-apiserver config.
+func ObserveAuditPolicy(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
+	listers := genericListers.(configobservation.Listers)
+	errs := []error{}
+	prevObservedConfig := map[string]interface{}{}
+
+	// copy non-empty .auditConfig.policyConfiguration from existingConfig to prevObservedConfig
+	policyPath := []string{"auditConfig", "policyConfiguration"}
+	existingPolicy, _, err := unstructured.NestedFieldNoCopy(existingConfig, policyPath...)
```
mfojtik: don't we want to check if it is non-empty?

pkg/operator/configobservation/audit/observe_audit_policy.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	if existingPolicy != nil {
+		if err := unstructured.SetNestedField(prevObservedConfig, existingPolicy, policyPath...); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	observedConfig := map[string]interface{}{}
+
+	// add initial-sa-token-signing-certs configmap mount path to saTokenCertDirs if configmap exists
```
soltysh: ???

pkg/operator/configobservation/audit/observe_audit_policy.go
```diff
@@ -0,0 +1,74 @@
...
+	policyConfigMap, err := listers.ConfigmapLister.ConfigMaps(operatorclient.GlobalUserSpecifiedConfigNamespace).Get(policyConfigMapName)
+	switch {
+	case errors.IsNotFound(err):
+		// switch back to default policy by leaving observedConfig empty
+	case err != nil:
+		// we had an error, return what we had before and exit. this really shouldn't happen
+		return prevObservedConfig, append(errs, err)
+	case len(policyConfigMap.Data["policy.yaml"]) > 0:
+		// TODO: maybe verify config here
```
soltysh: You mean you'd like to verify before applying it to server to ensure it doesn't crash the server/rejected by the server?

mfojtik:
LGTM

will defer to @soltysh to tag

pkg/operator/configobservation/audit/observe_audit_policy.go
sttts: copy&pasta

pkg/operator/configobservation/audit/observe_audit_policy.go
sttts: We ignore nil, but leave everything else. I think this is correct. If somebody writes `{}` or `{garbage: 42}` we want the API server to interpret that, not us here.

pkg/operator/configobservation/audit/observe_audit_policy.go
sttts: yes, return an early error. But that binds the operator livecycle to that of the server. Maybe not desired.

soltysh: I don't think we do that kind of validation elsewhere in the operator code, so I'd do it here until proven we need it.

pkg/operator/configobservation/audit/observe_audit_policy.go
soltysh: :stuck_out_tongue_closed_eyes:

sttts:
/hold

for smoke tests to be green.

deads2k:
/hold

I'd like to think more carefully about this before we choose to land it on a 1.13 base.  I think it is too late in the release to add this.

JAORMX:
So.... https://github.com/kubernetes/enhancements/blob/c7ab5663d31b6e3ce848d3929f33a9d6df2eebd1/keps/sig-auth/0014-dynamic-audit-configuration.md seems to be in alpha stage. What are the options to enable that feature into the deployment? Would that be an appropriate solution that could be added to the next release?

JAORMX:
@deads2k So... This would have been an easy way to support custom configurations for the audit policy. There currently doesn't seem to be an easy way to do this, but to do it via an operator like it's being suggested here. There is a proposal [1], but it seems to be still in development and under discussion [2]. There is also the possibility of enabling the webhook backend; however, that would require us to ship another process (or server) that reads the audit logs and knows how to configure them. So I still think this would be a better and easier solution. When does the development for the next release start?

[1] https://github.com/kubernetes/enhancements/blob/c7ab5663d31b6e3ce848d3929f33a9d6df2eebd1/keps/sig-auth/0014-dynamic-audit-configuration.md
[2] https://github.com/kubernetes/kubernetes/pull/71230

PR #422, opened 2019-04-18
switch to relying on etcd info in openshift-config

deads2k:
No resource we use should be located in kube-system anymore

/assign @mfojtik


deads2k:
terraform

/retest

PR #426, opened 2019-04-19
generate and maintain our own kube-apiserver-to-kubelet-signer

deads2k:
This removes the final kube-ca cert for real this time.

/hold

holding so that I can check to be sure we successfully collect pod logs


deads2k:
Logs work

/hold cancel

PR #427, opened 2019-04-22
speed up leader election

deads2k:
this helps with upgrade timing because save one minute on the lease.


deads2k:
infra

/retest

PR #428, opened 2019-04-22
fix flaky e2e test

sanchezl:
The `TestNamedCertificates` e2e test occasionally timeouts waiting for changes to take effect. This PR works around that by:
* looping on the first test until it passes, showing that at least one kube-apiserver is running with the configuration to be tested
* Then looping on each subsequent test until it passes (just in case it takes a couple of tries to hit the updated kube-apiserver).


deads2k:
well, it doesn't win for beauty, but ok.  We need the flake fix, merging

PR #430, opened 2019-04-22
update deployment for upgrade and install-gather

deads2k:
for upgrades, we need to have deployment strategy recreate on our operators in case they are ever scaled. We don't want to ever get a lease for new and later get a lease for old.

We need a distinct operator container to make install-gather happy at the moment.

@openshift/sig-master


deads2k:
/retest


I bet the installer pull swapping names broken the SNI test

deads2k:
@sanchezl have a look tomorrow

deads2k:
merging over permafailed operator test with https://github.com/openshift/cluster-kube-apiserver-operator/pull/432

PR #432, opened 2019-04-23
switch to internal hostname

deads2k:
https://github.com/openshift/installer/pull/1633 swapped names


deads2k:
only touches a test local to this repo (not shared) and we expect this to be the fix.  It flaked

PR #434, opened 2019-04-23
Wait for unrevisioned certs to be available

deads2k:
fixes https://bugzilla.redhat.com/show_bug.cgi?id=1702104


deads2k:
terraform and the sni flake.

/retest

PR #438, opened 2019-04-25
Include operator resource deletion recovery for revision controller

damemi:
Bumps to include changes from https://github.com/openshift/library-go/pull/368


damemi:
/hold cancel

this is merged in library-go now

PR #441, opened 2019-04-26
Revision hotlooping e2e test

damemi:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1699471

When operators hotloop creating new revisions, old revisions get their status stuck `InProgress` which doesn't count them toward the Failed/Succeeded revision limits, allowing revision counts to get very high. This tries to check a couple things:

1. That the revision count isn't higher than the sum of the failed/succeeded limits. This would indicate InProgress revisions (of which there shouldn't ever be more than 1)
2. That revisions aren't being rapidly created, by polling and tracking new revisions created. 30 seconds or so of no new revisions is probably good enough to see that we're not currently hotlooping
3. If 20 revisions have been created, we've got too much

`#2` and `#3` are kind of redundant, so if either one is preferred we can remove the other.

Any feedback on this approach appreciated. If this goes through we can work on adding similar tests to other operators

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -1,15 +1,22 @@
 package e2e
 
 import (
+	"fmt"
+	"strings"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 
 	configclient "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+	operatorversionedclient "github.com/openshift/client-go/operator/clientset/versioned"
+	operatorv1informers "github.com/openshift/client-go/operator/informers/externalversions"
 
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/operatorclient"
 	test "github.com/openshift/cluster-kube-apiserver-operator/test/library"
 )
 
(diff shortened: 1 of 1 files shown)


damemi:
Running this locally I'm getting a nil pointer dereference [here](https://github.com/openshift/cluster-kube-apiserver-operator/pull/441/files#diff-382f5df4828722855ac47b478acdf9f4R63), which digging into it turns out the call to `operatorClient.GetStaticPodOperatorState()` is returning back nil for the operatorSpec (error `kubeapiserver.operator.openshift.io "cluster" not found`).

If we remove the parts of the test that are trying to compare current revisions to the actual limits set in the operator, then this shouldn't be a problem because that's all I need the static pod spec for. In that case we'd just have a loop that checks we're not rapidly creating a bunch of new revisions, which imo would be sufficient. Thoughts?

damemi:
@sanchezl @deads2k any opinions on my last comment? If we can get past that bug this should pass CI hopefully

test/e2e/operator_test.go
```diff
@@ -37,3 +44,79 @@ func TestOperandImageVersion(t *testing.T) {
...
+	}
+
+	revisions, err := getRevisionStatuses(kubeClient)
+	require.NoError(t, err)
+
+	// Check if revisions are being quickly created to test for operator hotlooping
+	changes := 0
+	pollsWithoutChanges := 0
+	lastRevisionCount := len(revisions)
+	err = wait.PollImmediate(1*time.Second, 10*time.Minute, func() (bool, error) {
```
deads2k: Why don't we just read the kubeapiserver resource and see if the number is greater than 40.  If so, we're probably hot looping?

damemi: You mean `LatestAvailableRevision`? That might not update quick enough if we're creating a ton of revisions, but I think the configmaps will. I can just check those for a revision count instead of this loop

mfojtik:
/lgtm
/hold

(to answer the question)

deads2k:
if mfojtik likes it, that's good enough

/hold cancel

test/e2e/operator_test.go
```diff
@@ -37,3 +44,79 @@ func TestOperandImageVersion(t *testing.T) {
...
+	operatorConfigClient, err := operatorversionedclient.NewForConfig(kubeConfig)
+	require.NoError(t, err)
+	operatorConfigInformers := operatorv1informers.NewSharedInformerFactory(operatorConfigClient, 10*time.Minute)
+	operatorClient := &operatorclient.OperatorClient{
+		Informers: operatorConfigInformers,
+		Client:    operatorConfigClient.OperatorV1(),
+	}
+
+	// Get current revision limits
+	operatorSpec, _, _, err := operatorClient.GetStaticPodOperatorState()
```
sanchezl: you never check for `err` here

damemi: thanks, updated

PR #444, opened 2019-04-29
[WIP] - Add fix-certs cmd

tnozicka:
Requires:
 - [ ] https://github.com/openshift/cluster-kube-apiserver-operator/pull/448

/cc @mfojtik @deads2k

PR #448, opened 2019-04-30
Add recovery-apiserver command

tnozicka:
Split from https://github.com/openshift/cluster-kube-apiserver-operator/pull/444

adds `cluster-kube-apiserver-operator recovery-apiserver create/destroy` commands

/cc @deads2k

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,3 +11,6 @@ COPY --from=builder /go/src/github.com/openshift/cluster-kube-apiserver-operator
 COPY manifests/*.yaml /manifests
 COPY manifests/image-references /manifests
 LABEL io.openshift.release.operator true
+# FIXME: entrypoint shouldn't be bash but the binary (needs fixing the chain)
+# ENTRYPOINT ["/usr/bin/cluster-kube-apiserver-operator"]
+
--- a/Dockerfile.rhel7
+++ b/Dockerfile.rhel7
@@ -11,3 +11,5 @@ COPY --from=builder /go/src/github.com/openshift/cluster-kube-apiserver-operator
 COPY manifests/*.yaml /manifests
 COPY manifests/image-references /manifests
 LABEL io.openshift.release.operator true
+# FIXME: entrypoint shouldn't be bash but the binary (needs fixing the chain)
+# ENTRYPOINT ["/usr/bin/cluster-kube-apiserver-operator"]
--- a/bindata/v3.11.0/kube-apiserver/recovery-config.yaml
+++ b/bindata/v3.11.0/kube-apiserver/recovery-config.yaml
@@ -1,0 +1,38 @@
+apiVersion: kubecontrolplane.config.openshift.io/v1
+kind: KubeAPIServerConfig
+apiServerArguments:
+  storage-backend:
+  - etcd3
+  storage-media-type:
+  - application/vnd.kubernetes.protobuf
+servingInfo:
+  bindAddress: 127.0.0.1:7443
+  bindNetwork: tcp4
+  certFile: /etc/kubernetes/static-pod-resources/serving-ca.crt
+  keyFile: /etc/kubernetes/static-pod-resources/serving-ca.key
+  clientCA: /etc/kubernetes/static-pod-resources/serving-ca.crt
+storageConfig:
+  keyFile: /etc/kubernetes/static-pod-resources/etcd-client.key
+  certFile: /etc/kubernetes/static-pod-resources/etcd-client.crt
+  ca: /etc/kubernetes/static-pod-resources/etcd-serving-ca-bundle.crt
+  storagePrefix: openshift.io
+  urls:
+  - "https://localhost:2379"
+
+# Make hypershift happy.
+# (Everything bellow this line is just to provide some certs file
+# because hypershift tries to read those even if you don't want to set them up.)
+authConfig:
+  oauthMetadataFile: ""
+  requestHeader:
+    clientCA: /etc/kubernetes/static-pod-resources/serving-ca.crt
+serviceAccountPublicKeyFiles:
+- /etc/kubernetes/static-pod-resources/serving-ca.crt
+kubeletClientInfo:
+  ca: /etc/kubernetes/static-pod-resources/serving-ca.crt
+  certFile: /etc/kubernetes/static-pod-resources/serving-ca.crt
+  keyFile: /etc/kubernetes/static-pod-resources/serving-ca.key
+aggregatorConfig:
+  proxyClientInfo:
+    certFile: /etc/kubernetes/static-pod-resources/serving-ca.crt
+    keyFile: /etc/kubernetes/static-pod-resources/serving-ca.key
--- a/bindata/v3.11.0/kube-apiserver/recovery-pod.yaml
+++ b/bindata/v3.11.0/kube-apiserver/recovery-pod.yaml
@@ -1,0 +1,34 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  namespace: openshift-kube-apiserver
+  name: kube-apiserver-recovery
+  labels:
+    revision: "recovery"
+spec:
+  containers:
+  - name: kube-apiserver-recovery
+    image: "{{ .KubeApiserverImage }}"
+    imagePullPolicy: IfNotPresent
+    terminationMessagePolicy: FallbackToLogsOnError
+    command: ["hypershift", "openshift-kube-apiserver"]
+    args:
+    - --config=/etc/kubernetes/static-pod-resources/config.yaml
+    resources:
+      requests:
+        memory: 1Gi
+        cpu: 150m
+    ports:
+    - containerPort: 7443
+    volumeMounts:
+    - mountPath: /etc/kubernetes/static-pod-resources
+      name: resource-dir
+  terminationGracePeriodSeconds: 0
+  hostNetwork: true
+  priorityClassName: system-node-critical
+  tolerations:
+  - operator: "Exists"
+  volumes:
+  - hostPath:
+      path: "{{ .ResourceDir }}"
+    name: resource-dir
--- a/pkg/cmd/recoveryapiserver/create.go
+++ b/pkg/cmd/recoveryapiserver/create.go
@@ -1,0 +1,117 @@
+package recoveryapiserver
+
+import (
+	"context"
+	"fmt"
+	"path"
+	"time"
+
+	"github.com/spf13/cobra"
+	"k8s.io/apiserver/pkg/server"
+	"k8s.io/klog"
+
+	"k8s.io/client-go/tools/watch"
+
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/recovery"
+)
+
+type CreateOptions struct {
+	Options
+
+	StaticPodResourcesDir string
+	Timeout               time.Duration
+	Wait                  bool
+}
+
+func NewCreateCommand() *cobra.Command {
+	o := &CreateOptions{
+		Options:               NewDefaultOptions(),
+		StaticPodResourcesDir: "/etc/kubernetes/static-pod-resources",
+		Timeout:               5 * time.Minute,
+		Wait:                  true,
+	}
+
+	cmd := &cobra.Command{
+		Use: "create",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			err := o.Complete()
+			if err != nil {
+				return err
+			}
+
+			err = o.Validate()
+			if err != nil {
+				return err
+			}
+
+			err = o.Run()
+			if err != nil {
+				return err
+			}
+
+			return nil
+		},
+		SilenceErrors: true,
+		SilenceUsage:  true,
+	}
+
+	cmd.Flags().DurationVar(&o.Timeout, "timeout", o.Timeout, "startup timeout, 0 means infinite")
+	cmd.Flags().BoolVar(&o.Wait, "wait", o.Wait, "wait for recovery apiserver to become ready")
+
+	return cmd
+}
+
+func (o *CreateOptions) Complete() error {
+	err := o.Options.Complete()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (o *CreateOptions) Validate() error {
+	err := o.Options.Validate()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (o *CreateOptions) Run() error {
+	ctx, cancel := watch.ContextWithOptionalTimeout(context.TODO(), o.Timeout)
+	defer cancel()
+
+	signalHandler := server.SetupSignalHandler()
+	go func() {
+		<-signalHandler
+		cancel()
+	}()
+
+	recoveryApiserver := &recovery.Apiserver{
+		PodManifestDir:        o.PodManifestDir,
+		StaticPodResourcesDir: o.StaticPodResourcesDir,
+	}
+
+	err := recoveryApiserver.Create()
+	if err != nil {
+		return fmt.Errorf("failed to create recovery apiserver: %v", err)
+	}
+
+	kubeconfigPath := path.Join(recoveryApiserver.GetRecoveryResourcesDir(), recovery.AdminKubeconfigFileName)
+	klog.Infof("system:admin kubeconfig written to %q", kubeconfigPath)
+
+	if !o.Wait {
+		return nil
+	}
+
+	klog.Infof("Waiting for recovery apiserver to come up.")
+	err = recoveryApiserver.WaitForHealthz(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to wait for recovery apiserver to be ready: %v", err)
+	}
+	klog.Infof("Recovery apiserver is up.")
+
+	return nil
+}
--- a/pkg/cmd/recoveryapiserver/lib.go
+++ b/pkg/cmd/recoveryapiserver/lib.go
@@ -1,0 +1,20 @@
+package recoveryapiserver
+
+// Options are shared by other subcommands
+type Options struct {
+	PodManifestDir string
+}
+
+func NewDefaultOptions() Options {
+	return Options{
+		PodManifestDir: "/etc/kubernetes/manifests",
+	}
+}
+
+func (o *Options) Complete() error {
+	return nil
+}
+
+func (o *Options) Validate() error {
+	return nil
+}
--- a/pkg/operator/v311_00_assets/bindata.go
+++ b/pkg/operator/v311_00_assets/bindata.go
@@ -6,6 +6,8 @@
 // bindata/v3.11.0/kube-apiserver/ns.yaml
 // bindata/v3.11.0/kube-apiserver/pod-cm.yaml
 // bindata/v3.11.0/kube-apiserver/pod.yaml
+// bindata/v3.11.0/kube-apiserver/recovery-config.yaml
+// bindata/v3.11.0/kube-apiserver/recovery-pod.yaml
 // bindata/v3.11.0/kube-apiserver/svc.yaml
 // DO NOT EDIT!
 
@@ -431,6 +433,112 @@ func v3110KubeApiserverPodYaml() (*asset, error) {
 	return a, nil
 }
 
+var _v3110KubeApiserverRecoveryConfigYaml = []byte(`apiVersion: kubecontrolplane.config.openshift.io/v1
+kind: KubeAPIServerConfig
+apiServerArguments:
+  storage-backend:
+  - etcd3
+  storage-media-type:
+  - application/vnd.kubernetes.protobuf
+servingInfo:
+  bindAddress: 127.0.0.1:7443
+  bindNetwork: tcp4
+  certFile: /etc/kubernetes/static-pod-resources/serving-ca.crt
+  keyFile: /etc/kubernetes/static-pod-resources/serving-ca.key
+  clientCA: /etc/kubernetes/static-pod-resources/serving-ca.crt
+storageConfig:
+  keyFile: /etc/kubernetes/static-pod-resources/etcd-client.key
+  certFile: /etc/kubernetes/static-pod-resources/etcd-client.crt
+  ca: /etc/kubernetes/static-pod-resources/etcd-serving-ca-bundle.crt
+  storagePrefix: openshift.io
+  urls:
+  - "https://localhost:2379"
+
+# Make hypershift happy.
+# (Everything bellow this line is just to provide some certs file
+# because hypershift tries to read those even if you don't want to set them up.)
+authConfig:
+  oauthMetadataFile: ""
+  requestHeader:
+    clientCA: /etc/kubernetes/static-pod-resources/serving-ca.crt
+serviceAccountPublicKeyFiles:
+- /etc/kubernetes/static-pod-resources/serving-ca.crt
+kubeletClientInfo:
+  ca: /etc/kubernetes/static-pod-resources/serving-ca.crt
+  certFile: /etc/kubernetes/static-pod-resources/serving-ca.crt
+  keyFile: /etc/kubernetes/static-pod-resources/serving-ca.key
+aggregatorConfig:
+  proxyClientInfo:
+    certFile: /etc/kubernetes/static-pod-resources/serving-ca.crt
+    keyFile: /etc/kubernetes/static-pod-resources/serving-ca.key
+`)
+
+func v3110KubeApiserverRecoveryConfigYamlBytes() ([]byte, error) {
+	return _v3110KubeApiserverRecoveryConfigYaml, nil
+}
+
+func v3110KubeApiserverRecoveryConfigYaml() (*asset, error) {
+	bytes, err := v3110KubeApiserverRecoveryConfigYamlBytes()
+	if err != nil {
+		return nil, err
+	}
+
+	info := bindataFileInfo{name: "v3.11.0/kube-apiserver/recovery-config.yaml", size: 0, mode: os.FileMode(0), modTime: time.Unix(0, 0)}
+	a := &asset{bytes: bytes, info: info}
+	return a, nil
+}
+
+var _v3110KubeApiserverRecoveryPodYaml = []byte(`apiVersion: v1
+kind: Pod
+metadata:
+  namespace: openshift-kube-apiserver
+  name: kube-apiserver-recovery
+  labels:
+    revision: "recovery"
+spec:
+  containers:
+  - name: kube-apiserver-recovery
+    image: "{{ .KubeApiserverImage }}"
+    imagePullPolicy: IfNotPresent
+    terminationMessagePolicy: FallbackToLogsOnError
+    command: ["hypershift", "openshift-kube-apiserver"]
+    args:
+    - --config=/etc/kubernetes/static-pod-resources/config.yaml
+    resources:
+      requests:
+        memory: 1Gi
+        cpu: 150m
+    ports:
+    - containerPort: 7443
+    volumeMounts:
+    - mountPath: /etc/kubernetes/static-pod-resources
+      name: resource-dir
+  terminationGracePeriodSeconds: 0
+  hostNetwork: true
+  priorityClassName: system-node-critical
+  tolerations:
+  - operator: "Exists"
+  volumes:
+  - hostPath:
+      path: "{{ .ResourceDir }}"
+    name: resource-dir
+`)
+
+func v3110KubeApiserverRecoveryPodYamlBytes() ([]byte, error) {
+	return _v3110KubeApiserverRecoveryPodYaml, nil
+}
+
+func v3110KubeApiserverRecoveryPodYaml() (*asset, error) {
+	bytes, err := v3110KubeApiserverRecoveryPodYamlBytes()
+	if err != nil {
+		return nil, err
+	}
+
+	info := bindataFileInfo{name: "v3.11.0/kube-apiserver/recovery-pod.yaml", size: 0, mode: os.FileMode(0), modTime: time.Unix(0, 0)}
+	a := &asset{bytes: bytes, info: info}
+	return a, nil
+}
+
 var _v3110KubeApiserverSvcYaml = []byte(`apiVersion: v1
 kind: Service
 metadata:
@@ -517,13 +625,15 @@ func AssetNames() []string {
 
 // _bindata is a table, holding each asset generator, mapped to its name.
 var _bindata = map[string]func() (*asset, error){
-	"v3.11.0/kube-apiserver/cm.yaml":            v3110KubeApiserverCmYaml,
-	"v3.11.0/kube-apiserver/defaultconfig.yaml": v3110KubeApiserverDefaultconfigYaml,
-	"v3.11.0/kube-apiserver/kubeconfig-cm.yaml": v3110KubeApiserverKubeconfigCmYaml,
-	"v3.11.0/kube-apiserver/ns.yaml":            v3110KubeApiserverNsYaml,
-	"v3.11.0/kube-apiserver/pod-cm.yaml":        v3110KubeApiserverPodCmYaml,
-	"v3.11.0/kube-apiserver/pod.yaml":           v3110KubeApiserverPodYaml,
-	"v3.11.0/kube-apiserver/svc.yaml":           v3110KubeApiserverSvcYaml,
+	"v3.11.0/kube-apiserver/cm.yaml":              v3110KubeApiserverCmYaml,
+	"v3.11.0/kube-apiserver/defaultconfig.yaml":   v3110KubeApiserverDefaultconfigYaml,
+	"v3.11.0/kube-apiserver/kubeconfig-cm.yaml":   v3110KubeApiserverKubeconfigCmYaml,
+	"v3.11.0/kube-apiserver/ns.yaml":              v3110KubeApiserverNsYaml,
+	"v3.11.0/kube-apiserver/pod-cm.yaml":          v3110KubeApiserverPodCmYaml,
+	"v3.11.0/kube-apiserver/pod.yaml":             v3110KubeApiserverPodYaml,
+	"v3.11.0/kube-apiserver/recovery-config.yaml": v3110KubeApiserverRecoveryConfigYaml,
+	"v3.11.0/kube-apiserver/recovery-pod.yaml":    v3110KubeApiserverRecoveryPodYaml,
+	"v3.11.0/kube-apiserver/svc.yaml":             v3110KubeApiserverSvcYaml,
 }
 
 // AssetDir returns the file names below a certain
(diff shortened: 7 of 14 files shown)


Dockerfile.rhel7
```diff
@@ -11,3 +11,4 @@ COPY --from=builder /go/src/github.com/openshift/cluster-kube-apiserver-operator
 COPY manifests/*.yaml /manifests
 COPY manifests/image-references /manifests
 LABEL io.openshift.release.operator true
+ENTRYPOINT ["/usr/bin/cluster-kube-apiserver-operator"]
```
tnozicka: we may need to contact the ART, not sure what's the policy lately
/hold

tnozicka:
need memory bump for unit test https://github.com/openshift/release/pull/3640

Dockerfile
```diff
@@ -11,3 +11,4 @@ COPY --from=builder /go/src/github.com/openshift/cluster-kube-apiserver-operator
 COPY manifests/*.yaml /manifests
 COPY manifests/image-references /manifests
 LABEL io.openshift.release.operator true
+ENTRYPOINT ["/usr/bin/cluster-kube-apiserver-operator"]
```
deads2k: this shouldn't be required.  Snip it from your pull.

bindata/v3.11.0/kube-apiserver/recovery-config.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: kubecontrolplane.config.openshift.io/v1
+kind: KubeAPIServerConfig
+apiServerArguments:
+  storage-backend:
+  - etcd3
+  storage-media-type:
+  - application/vnd.kubernetes.protobuf
+servingInfo:
+  bindAddress: 0.0.0.0:7443
```
deads2k: no, must be specifically localhost. 127.0.0.1

bindata/v3.11.0/kube-apiserver/recovery-config.yaml
```diff
@@ -0,0 +1,36 @@
...
+  ca: /etc/kubernetes/static-pod-resources/etcd-serving-ca-bundle.crt
+  storagePrefix: openshift.io
+  urls:
+  - "https://localhost:2379"
+
+# Make hypershift happy
+authConfig:
+  oauthMetadataFile: ""
+  requestHeader:
+    clientCA: /etc/kubernetes/static-pod-resources/serving-ca.crt
```
deads2k: find the correct client-ca.  it should be present

bindata/v3.11.0/kube-apiserver/recovery-config.yaml
```diff
@@ -0,0 +1,36 @@
...
+serviceAccountPublicKeyFiles:
+- /etc/kubernetes/static-pod-resources/serving-ca.crt
```
deads2k: required?  icky, but acceptable I suppose

bindata/v3.11.0/kube-apiserver/recovery-config.yaml
```diff
@@ -0,0 +1,36 @@
...
+kubeletClientInfo:
+  ca: /etc/kubernetes/static-pod-resources/serving-ca.crt
+  certFile: /etc/kubernetes/static-pod-resources/serving-ca.crt
+  keyFile: /etc/kubernetes/static-pod-resources/serving-ca.key
+aggregatorConfig:
+  proxyClientInfo:
+    certFile: /etc/kubernetes/static-pod-resources/serving-ca.crt
```
deads2k: this should be loadable

bindata/v3.11.0/kube-apiserver/recovery-config.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+# Make hypershift happy
+authConfig:
+  oauthMetadataFile: ""
+  requestHeader:
+    clientCA: /etc/kubernetes/static-pod-resources/serving-ca.crt
+serviceAccountPublicKeyFiles:
+- /etc/kubernetes/static-pod-resources/serving-ca.crt
+kubeletClientInfo:
+  ca: /etc/kubernetes/static-pod-resources/serving-ca.crt
```
deads2k: document that you know it's wrong and explain why

bindata/v3.11.0/kube-apiserver/recovery-pod.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  namespace: kube-system  # for disaster recovery we need namespace that is always present
```
deads2k: no you don't

bindata/v3.11.0/kube-apiserver/recovery-pod.yaml
```diff
@@ -0,0 +1,34 @@
...
+  name: kube-apiserver-recovery
+  labels:
+    revision: "recovery"
+spec:
+  containers:
+  - name: kube-apiserver-recovery
+    image: "{{ .KubeApiserverImage }}"
+    imagePullPolicy: IfNotPresent
+    terminationMessagePolicy: FallbackToLogsOnError
+    command: ["hypershift", "openshift-kube-apiserver"]
```
deads2k: @sttts I think it may be simpler to launch `hyperkube kube-apiserver` and avoid any entanglments.  preference?

Dockerfile
tnozicka: Why? This is how you build tools packaged in containers meant to be run by users. This is not a bash image (which is the entry point currently) but cluster-kube-apiserver-operator image.

```bash
podman run -it --rm --entrypoint=/usr/bin/cluster-kube-apiserver-operator registry.svc.ci.openshift.org/ocp/4.1-2019-04-29-113133:cluster-kube-apiserver-operator recovery-apiserver create
```
vs.
```bash
podman run -it --rm registry.svc.ci.openshift.org/ocp/4.1-2019-04-29-113133:cluster-kube-apiserver-operator recovery-apiserver create
```

bindata/v3.11.0/kube-apiserver/recovery-config.yaml
tnozicka: I don't want to setup oauth. Everything bellow line `# Make hypershift happy` is just to provide **a** cert file so hypershift doesn't fail while converting the config to flags.

bindata/v3.11.0/kube-apiserver/recovery-config.yaml
tnozicka: same here, will expand on the comment above

bindata/v3.11.0/kube-apiserver/recovery-config.yaml
tnozicka: I though we don't want to set up aggregation for recovery apiserver

bindata/v3.11.0/kube-apiserver/recovery-pod.yaml
tnozicka: yeah, this will come up, I was thinking about the other phase where kubelet syncs pod status into the apiserver but we don't ever setup the credentials, so it will never get as far to spit those errors.

bindata/v3.11.0/kube-apiserver/recovery-pod.yaml
tnozicka: Do we use it anywhere else? I'd not want to be the only one exploring that path and running it differently then the regular kube-apiserver.

bindata/v3.11.0/kube-apiserver/recovery-config.yaml
tnozicka: it is secured with certs auth already, but sure

bindata/v3.11.0/kube-apiserver/recovery-config.yaml
tnozicka: same here

Dockerfile
tnozicka: bootkube is mishandling it :/ which is a bug, but in the interest of time I am going to drop it for now

bindata/v3.11.0/kube-apiserver/recovery-pod.yaml
sttts: have no strong opinion either. If this works, let's leave it.

bindata/v3.11.0/kube-apiserver/recovery-pod.yaml
```diff
@@ -0,0 +1,34 @@
...
+    args:
+    - --config=/etc/kubernetes/static-pod-resources/config.yaml
```
sttts: this is the same config as the static pod? How does this work with 7443 below? Would have expected `recovery-config.yaml`.

pkg/cmd/recoveryapiserver/create.go
```diff
@@ -0,0 +1,122 @@
...
+
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/recovery"
+)
+
+type CreateOptions struct {
+	Options
+
+	KubeApiserverImage string
+	Timeout            time.Duration
+	Wait               bool
```
sttts: wait for what?

pkg/cmd/recoveryapiserver/create.go
```diff
@@ -0,0 +1,122 @@
...
+	"k8s.io/client-go/tools/watch"
+
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/recovery"
+)
+
+type CreateOptions struct {
+	Options
+
+	KubeApiserverImage string
+	Timeout            time.Duration
```
sttts: timeout of what?

pkg/cmd/recoveryapiserver/create.go
```diff
@@ -0,0 +1,122 @@
...
+	Wait               bool
+}
+
+func NewCreateCommand() *cobra.Command {
+	o := &CreateOptions{
+		Options:            NewDefaultOptions(),
+		KubeApiserverImage: "", // TODO: set the public pullspec
```
sttts: what is the plan where to get this from?

pkg/cmd/recoveryapiserver/create.go
```diff
@@ -0,0 +1,122 @@
...
+			return nil
+		},
+	}
+
+	// TODO: add/prettify descriptions
+	AddFlags(cmd, &o.Options)
+
+	cmd.Flags().StringVar(&o.KubeApiserverImage, "kube-apiserver-image", o.KubeApiserverImage, "")
+	cmd.Flags().DurationVar(&o.Timeout, "timeout", o.Timeout, "timeout, 0 means infinite")
+	cmd.Flags().BoolVar(&o.Wait, "wait", o.Wait, "wait for recovery apiserver to become ready")
```
sttts: instead of what? returning immediately, leaving the pod launching?

pkg/cmd/recoveryapiserver/create.go
```diff
@@ -0,0 +1,122 @@
...
+
+			return nil
+		},
+	}
+
+	// TODO: add/prettify descriptions
+	AddFlags(cmd, &o.Options)
+
+	cmd.Flags().StringVar(&o.KubeApiserverImage, "kube-apiserver-image", o.KubeApiserverImage, "")
+	cmd.Flags().DurationVar(&o.Timeout, "timeout", o.Timeout, "timeout, 0 means infinite")
```
sttts: startup-timeout

pkg/cmd/recoveryapiserver/lib.go
```diff
@@ -0,0 +1,35 @@
+package recoveryapiserver
+
+import (
+	"errors"
+
+	"github.com/spf13/cobra"
+)
+
+type Options struct {
```
sttts: this are shared options, right? Would better call it like that. Was confused when reading the code.

pkg/recovery/apiserver.go
```diff
@@ -0,0 +1,438 @@
...
+func init() {
+	metav1.AddToGroupVersion(Scheme, schema.GroupVersion{Version: "v1"})
+	utilruntime.Must(corev1.AddToScheme(Scheme))
+}
+
+type Apiserver struct {
+	KubeApiserverImage string
+	PodManifestDir     string
+	ResourceDirPath    string
+	RecoveryDirPath    string
```
sttts: I know the other 3, but what is this?

pkg/recovery/apiserver.go
```diff
@@ -0,0 +1,438 @@
...
+	restConfig             *rest.Config
+	kubeClientSet          *kubernetes.Clientset
+}
+
+func (s *Apiserver) GetCurrentApiserverManifest() (*corev1.Pod, error) {
+	if s.kubeApiserverStaticPod != nil {
+		return s.kubeApiserverStaticPod, nil
+	}
+
+	kubeApiserverManifestPath := path.Join(s.PodManifestDir, KubeApiserverStaticPodFileName)
```
sttts: filepath.Join

pkg/recovery/apiserver.go
```diff
@@ -0,0 +1,438 @@
...
+	PodManifestDir     string
+	ResourceDirPath    string
+	RecoveryDirPath    string
+
+	kubeApiserverStaticPod *corev1.Pod
+	restConfig             *rest.Config
+	kubeClientSet          *kubernetes.Clientset
+}
+
+func (s *Apiserver) GetCurrentApiserverManifest() (*corev1.Pod, error) {
```
sttts: `s/Get//`

pkg/recovery/apiserver.go
```diff
@@ -0,0 +1,438 @@
...
+	if s.kubeApiserverStaticPod != nil {
+		return s.kubeApiserverStaticPod, nil
+	}
+
+	kubeApiserverManifestPath := path.Join(s.PodManifestDir, KubeApiserverStaticPodFileName)
+	f, err := os.OpenFile(kubeApiserverManifestPath, os.O_RDONLY, 0644)
```
sttts: `ioutil.ReadFile`, much simpler

pkg/recovery/apiserver.go
```diff
@@ -0,0 +1,438 @@
...
+		return nil, fmt.Errorf("failed to decode file %q: %v", kubeApiserverManifestPath, err)
+	}
+
+	// TODO: support conversions if the object is in different but convertible version
+	kubeApiserverStaticPod, ok := obj.(*corev1.Pod)
+	if !ok {
+		return nil, fmt.Errorf("unsupported type: kubeApiserverStaticPodManifest is not type *corev1.Pod but %T", obj)
+	}
+
+	s.kubeApiserverStaticPod = kubeApiserverStaticPod
```
sttts: side effects in getters are bad. Don't do this. Set the value outside at the call side.

pkg/recovery/apiserver.go
```diff
@@ -0,0 +1,438 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to read apiserver pod manifest: %v", err)
+	}
+
+	resourceDir, err := getVolumeHostPathPath("resource-dir", pod.Spec.Volumes)
+	if err != nil {
+		return "", fmt.Errorf("failed to find resource-dir: %v", err)
+	}
+
+	s.ResourceDirPath = *resourceDir
```
sttts: as above

pkg/recovery/apiserver.go
```diff
@@ -0,0 +1,438 @@
...
+	}
+
+	s.kubeApiserverStaticPod = kubeApiserverStaticPod
+
+	return s.kubeApiserverStaticPod, nil
+}
+
+func (s *Apiserver) GetResourceDirPath() (string, error) {
+	if s.ResourceDirPath != "" {
+		return s.ResourceDirPath, nil
```
sttts: is this really time critical that logic like this makes sense?

pkg/recovery/apiserver.go
```diff
@@ -0,0 +1,438 @@
...
+
+	_, err = io.Copy(destFile, srcFile)
+	if err != nil {
+		return fmt.Errorf("failed to copy file %q into %q: %v", src, dest, err)
+	}
+
+	return nil
+}
+
+func getVolumeHostPathPath(volumeName string, volumes []corev1.Volume) (*string, error) {
```
sttts: why a pointer? Unusual signature. Just return `""`.

sttts: It's even dangerous as you can mutate the pod via this pointer. You probably (hopefully) don't want that.

pkg/recovery/apiserver.go
```diff
@@ -0,0 +1,438 @@
...
+		return s.RecoveryDirPath, nil
+	}
+
+	// Create tmp dir for recovery apiserver
+	recoveryDir, err := ioutil.TempDir("", "kube-recovery-apiserver")
+	if err != nil {
+		return "", fmt.Errorf("failed to create tmpDir: %v", err)
+	}
+
+	s.RecoveryDirPath = recoveryDir
```
sttts: as above

pkg/recovery/apiserver.go
```diff
@@ -0,0 +1,438 @@
...
+	return s.ResourceDirPath, nil
+}
+
+func (s *Apiserver) GetRecoveryDirPath() (string, error) {
+	if s.RecoveryDirPath != "" {
+		return s.RecoveryDirPath, nil
+	}
+
+	// Create tmp dir for recovery apiserver
+	recoveryDir, err := ioutil.TempDir("", "kube-recovery-apiserver")
```
sttts: anybody cleaning this up? If not, add a comment that we accept leaking temp dirs.

pkg/recovery/apiserver.go
```diff
@@ -0,0 +1,438 @@
...
+	recoveryPodBuffer := bytes.NewBuffer(nil)
+	err = t.Execute(recoveryPodBuffer, struct {
+		KubeApiserverImage string
+		ResourceDir        string
+	}{
+		KubeApiserverImage: s.KubeApiserverImage,
+		ResourceDir:        recoveryDirPath,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("fail to execute internal recovery pod template %q: %v", RecoveryPodAsset, err)
```
sttts: failed

pkg/recovery/apiserver.go
```diff
@@ -0,0 +1,438 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to read file %q: %v", kubeApiserverManifestPath, err)
+	}
+
+	obj, err := runtime.Decode(Codecs.UniversalDecoder(corev1.SchemeGroupVersion), buf.Bytes())
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode file %q: %v", kubeApiserverManifestPath, err)
+	}
+
+	// TODO: support conversions if the object is in different but convertible version
```
sttts: there is only core/v1 for the forseeable future.

pkg/recovery/apiserver.go
```diff
@@ -0,0 +1,438 @@
...
+		return fmt.Errorf("failed to get recovery dir: %v", err)
+	}
+
+	// Copy certs for accessing etcd
+	for src, dest := range map[string]string{
+		"secrets/etcd-client/tls.key":              "etcd-client.key",
+		"secrets/etcd-client/tls.crt":              "etcd-client.crt",
+		"configmaps/etcd-serving-ca/ca-bundle.crt": "etcd-serving-ca-bundle.crt",
+	} {
+		err = copyFile(path.Join(resourceDirPath, src), path.Join(recoveryDirPath, dest))
```
sttts: filepath

pkg/recovery/apiserver.go
```diff
@@ -0,0 +1,438 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	// We are creating only temporary certificates to start the recovery apiserver.
+	// A week seem reasonably high for a debug session, while it is easy to create a new one.
+	certValidity := 7 * 24 * time.Hour
```
sttts: this would deserve some log output maybe: "Creating temporary certs with 1 week validity"

pkg/recovery/apiserver.go
sttts: This will contain secrets. Better make it 0700.

pkg/recovery/apiserver.go
sttts: and everywhere below

pkg/operator/v311_00_assets/bindata.go
```diff
@@ -431,6 +433,112 @@ func v3110KubeApiserverPodYaml() (*asset, error) {
...
+  urls:
+  - "https://localhost:2379"
+
+# Make hypershift happy.
+# (Everything bellow this line is just to provide some certs file
+# because hypershift tries to read those even if you don't want to set them up.)
+authConfig:
+  oauthMetadataFile: ""
+  requestHeader:
+    clientCA: /etc/kubernetes/static-pod-resources/serving-ca.crt
```
sttts: # we are lazy: we reuse the serving CA for clients as well

pkg/recovery/apiserver.go
```diff
@@ -0,0 +1,438 @@
...
+	}
+
+	// Create client cert
+	ca := crypto.CA{
+		Config:          rootCaConfig,
+		SerialGenerator: &crypto.RandomSerialGenerator{},
+	}
+
+	// Create client certificates for system:admin
+	clientCert, err := ca.MakeClientCertificateForDuration(&user.DefaultInfo{Name: "system:admin"}, certValidity)
```
sttts: // we are lazy and reuse the serving CA as client CA

pkg/recovery/apiserver.go
```diff
@@ -0,0 +1,438 @@
...
+
+func (s *Apiserver) Destroy() error {
+	recoveryPodManifestPath := path.Join(s.PodManifestDir, RecoveryPodFileName)
+	err := os.Remove(recoveryPodManifestPath)
+	if err != nil {
+		return fmt.Errorf("failed to remove recovery pod manifest %q: %v", recoveryPodManifestPath, err)
+	}
+
+	// TODO: consider removing the recoveryDir
+	// (The dir is tmp by default anyways and putting a wrong path here could be fatal as it would have to recursively delete.)
```
sttts: which sensitive information (after recovery) does this dir contain?

sttts: etcd credentials come to mind. Leaving those laying around is a no-go.

pkg/recovery/apiserver.go
```diff
@@ -0,0 +1,438 @@
...
+		return fmt.Errorf("failed to open file %q: %v", src, err)
+	}
+	defer srcFile.Close()
+
+	srcFileInfo, err := srcFile.Stat()
+	if err != nil {
+		return fmt.Errorf("failed to stat file %q: %v", src, err)
+	}
+
+	if srcFileInfo.IsDir() {
```
sttts: check for IsRegular instead. There are other non-files than dirs.

bindata/v3.11.0/kube-apiserver/recovery-pod.yaml
sttts: saw the config. I guess the config.yaml will be the recovery config at the end with the recovery dir as the `static-pod-resources`.

tnozicka: it is a different one
https://github.com/openshift/cluster-kube-apiserver-operator/blob/3a5f313893a373f7b3f766d3c08ef17ce6f19fed/bindata/v3.11.0/kube-apiserver/recovery-config.yaml#L1
https://github.com/openshift/cluster-kube-apiserver-operator/blob/3a5f313893a373f7b3f766d3c08ef17ce6f19fed/pkg/recovery/apiserver.go#L260-L264

pkg/recovery/apiserver.go
sttts: `RecoveryStatusPodResourceDir`  and everything is clear (as long as your monitor is big enough).

pkg/recovery/apiserver.go
```diff
@@ -0,0 +1,438 @@
...
+
+func init() {
+	metav1.AddToGroupVersion(Scheme, schema.GroupVersion{Version: "v1"})
+	utilruntime.Must(corev1.AddToScheme(Scheme))
+}
+
+type Apiserver struct {
+	KubeApiserverImage string
+	PodManifestDir     string
+	ResourceDirPath    string
```
sttts: LastStaticPodResourceDir

pkg/cmd/recoveryapiserver/create.go
tnozicka: for the apiserver healthz https://github.com/openshift/cluster-kube-apiserver-operator/blob/3a5f313893a373f7b3f766d3c08ef17ce6f19fed/pkg/cmd/recoveryapiserver/create.go#L115

pkg/cmd/recoveryapiserver/create.go
tnozicka: the whole create command, at this point the only asynchronous op is waiting for the server to come up

pkg/cmd/recoveryapiserver/create.go
tnozicka: I think we can autodetect it from the current kubeapiserver if present or you can get that image from the release

pkg/cmd/recoveryapiserver/create.go
tnozicka: instead of just writing the file

pkg/recovery/apiserver.go
tnozicka: the one used to store the recovery-apiserver resources

sttts: just choose a better Go name. Doesn't help so much here in the comments ;-)

pkg/recovery/apiserver.go
tnozicka: I was definitely doing just that at some iteration to change the hostpath while I was reusing the manifest, maybe it is not used now we have forked the manifest file

pkg/recovery/apiserver.go
tnozicka: it contains both configmaps and secrets so technicaly the dir is 755 and the secret file need to be 600. it is the same as in /etc where others can read the dir and configmaps, not the secret files. we have to make sure secrets have 600 though

pkg/recovery/apiserver.go
tnozicka: they still have the same perms restriction as in etcd, I'll think about other options

pkg/operator/v311_00_assets/bindata.go
tnozicka: is this not enough https://github.com/openshift/cluster-kube-apiserver-operator/pull/448/files#diff-23575013b46cf40eaadce54aa93eb7f7R23 ? :)

pkg/recovery/apiserver.go
tnozicka: except different call sites need different parts initialized

pkg/recovery/apiserver.go
tnozicka: some caller don't need it some don't, better to fail only reading stuff you need

pkg/cmd/recoveryapiserver/create.go
```diff
@@ -0,0 +1,127 @@
...
+		},
+		SilenceErrors: true,
+		SilenceUsage:  true,
+	}
+
+	// TODO: add/prettify descriptions
+	AddFlags(cmd, &o.Options)
+
+	cmd.Flags().StringVar(&o.KubeApiserverImage, "kube-apiserver-image", o.KubeApiserverImage, "image reference for kubernetes apiserver")
+	cmd.Flags().StringVar(&o.StaticPodResourcesDir, "static-pod-resources", o.StaticPodResourcesDir, "path to store a directory containing static pod resources for recovery apiserver manifest")
```
deads2k: This should not be an arg. I see no reason for a user to set this.  We don't allow directory choice for anything else.

pkg/cmd/recoveryapiserver/create.go
```diff
@@ -0,0 +1,127 @@
...
+			return nil
+		},
+		SilenceErrors: true,
+		SilenceUsage:  true,
+	}
+
+	// TODO: add/prettify descriptions
+	AddFlags(cmd, &o.Options)
+
+	cmd.Flags().StringVar(&o.KubeApiserverImage, "kube-apiserver-image", o.KubeApiserverImage, "image reference for kubernetes apiserver")
```
deads2k: I see no reason to allow this to be set. If a user wants to change it, they can manually edit the static pod manifest that you create.

pkg/recovery/apiserver.go
```diff
@@ -0,0 +1,392 @@
...
+		return nil, fmt.Errorf("failed to decode file %q: %v", kubeApiserverManifestPath, err)
+	}
+
+	s.kubeApiserverStaticPod = pod
+
+	return s.kubeApiserverStaticPod, nil
+}
+
+func (s *Apiserver) GetResourceDirPath() (string, error) {
+	if s.ResourceDirPath != "" {
```
deads2k: project standard is the len check

pkg/recovery/apiserver.go
deads2k: > some caller don't need it some don't, better to fail only reading stuff you need

I agree with sttts.  Caching is harder for me to think about as I read than non-caching.

pkg/recovery/apiserver.go
```diff
@@ -0,0 +1,392 @@
...
+	return s.ResourceDirPath, nil
+}
+
+func (s *Apiserver) GetRecoveryDir() string {
+	return s.recoveryDir
+}
+
+func (s *Apiserver) RestConfig() (*rest.Config, error) {
+	if s.restConfig == nil {
+		return nil, errors.New("no rest config is set yet")
```
deads2k: Do we really just want a `NewAPIServer` func that requires a non-nil restConfig?

pkg/recovery/apiserver.go
```diff
@@ -0,0 +1,392 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("fail to parse internal recovery pod template %q: %v", RecoveryPodAsset, err)
+	}
+
+	recoveryPodBuffer := bytes.NewBuffer(nil)
+	err = t.Execute(recoveryPodBuffer, struct {
+		KubeApiserverImage string
+		ResourceDir        string
+	}{
+		KubeApiserverImage: s.KubeApiserverImage,
```
deads2k: I expected you to get this value from the existing kube-apiserver pod manifest that you're already reading.

pkg/recovery/apiserver.go
deads2k: > // we are lazy and reuse the serving CA as client CA

I really hate that, but I won't fight it here for now.

pkg/cmd/recoveryapiserver/create.go
```diff
@@ -0,0 +1,127 @@
...
+	cmd.Flags().StringVar(&o.StaticPodResourcesDir, "static-pod-resources", o.StaticPodResourcesDir, "path to store a directory containing static pod resources for recovery apiserver manifest")
+	cmd.Flags().DurationVar(&o.Timeout, "timeout", o.Timeout, "startup timeout, 0 means infinite")
+	cmd.Flags().BoolVar(&o.Wait, "wait", o.Wait, "wait for recovery apiserver to become ready")
```
deads2k: This command should always block until we give up on healthz or until the user cancels it.  Imagine it as the old `openshift start` logic that runs a server until the user terminates it.

pkg/recovery/apiserver.go
tnozicka: done

pkg/cmd/recoveryapiserver/create.go
tnozicka: this makes it impossible to test it on your machine, but i'll drop it

pkg/cmd/recoveryapiserver/create.go
tnozicka: ok

pkg/cmd/recoveryapiserver/create.go
tnozicka: I don't really see this being an issue, we may have a script in the future, that already handles checking healthz or want to talk to apiserver before that. `openshift start` is a bit different because the process is actually the server. This just creates a config for another tool to run it and a distinct command to delete it

tnozicka: and the default is to wait

pkg/recovery/apiserver.go
tnozicka: Was not intended as caching initially but as deferred loading; the other PR adds another caller creating some distinct paths. removed most of it because it was not needed anymore (I guess I'll find that out when I do the rebase :)).

pkg/recovery/apiserver.go
tnozicka: we are now using /etc/kubernetes/manifests and destroying after

pkg/recovery/apiserver.go
tnozicka:
```golang
// A FileInfo describes a file and is returned by Stat and Lstat.
type FileInfo interface {
	Name() string       // base name of the file
	Size() int64        // length in bytes for regular files; system-dependent for others
	Mode() FileMode     // file mode bits
	ModTime() time.Time // modification time
	IsDir() bool        // abbreviation for Mode().IsDir()
	Sys() interface{}   // underlying data source (can return nil)
}
```

also symlinks are fine I guess

pkg/recovery/apiserver.go
tnozicka: that would mean going through create which would pose hard to implement destroy method

pkg/recovery/apiserver.go
tnozicka: done

pkg/cmd/recoveryapiserver/create.go
```diff
@@ -0,0 +1,110 @@
...
+				return err
+			}
+
+			return nil
+		},
+		SilenceErrors: true,
+		SilenceUsage:  true,
+	}
+
+	cmd.Flags().DurationVar(&o.Timeout, "timeout", o.Timeout, "startup timeout, 0 means infinite")
```
deads2k: These two flags need to be removed.

pkg/cmd/recoveryapiserver/create.go
deads2k: It's not about defaults.  it's about not having options we don't need or want.  These should not exist at all.

deads2k: > I don't really see this being an issue, we may have a script in the future, that already handles checking healthz or want to talk to apiserver before that. `openshift start` is a bit different because the process is actually the server. This just creates a config for another tool to run it and a distinct command to delete it

It should be a single command that runs a server until you cancel it and then the process terminates and cleans up.

pkg/recovery/apiserver.go
```diff
@@ -0,0 +1,374 @@
...
+	}
+
+	klog.Infof("Deleting resource-dir %q", resourceDirPath)
+	err = os.RemoveAll(resourceDirPath)
+	if err != nil {
+		return fmt.Errorf("failed to remove recovery pod manifest %q: %v", recoveryPodManifestPath, err)
+	}
+
+	klog.Infof("Deleting recovery pod manifest %q", recoveryPodManifestPath)
+	err = os.Remove(recoveryPodManifestPath)
```
deads2k: this delete should logically happen first

deads2k: > this delete should logically happen first

nm. I understand the indempotency.

tnozicka: depends on the logic if it fails in between you leak the dir and second run won't help. Non-working recovery apiserver that you are trying to destroy seems better

deads2k:
merged in https://github.com/openshift/cluster-kube-apiserver-operator/pull/461

PR #461, opened 2019-05-06
 generate certificates synchronously and do not start a kube-apiserver

deads2k:
evolution of #444 

Most notable changes:
 1. fully synchronous, there are no "wait for a little while" aspects anymore and no gofuncs
 2. zero arg.  There's nothing to configure anymore
 3. uses an already existent recovery-kube-apiserver and doesn't affect its running state
 4. updates certs for the actual kube-apiserver static pod, not the recovery apiserver pod

@mfojtik @tnozicka @sttts


pkg/cmd/recoveryapiserver/create.go
```diff
@@ -19,16 +18,12 @@ type CreateOptions struct {
 	Options
 
 	StaticPodResourcesDir string
-	Timeout               time.Duration
-	Wait                  bool
```
tnozicka: @deads2k I think you should have kept at least the struct fileds or unit testing becomes real pain

pkg/cmd/fixcerts/fixcerts.go
```diff
@@ -37,6 +37,8 @@ import (
 )
 
 const (
+	RecoveryPodKubeConfig                  = "/etc/kubernetes/static-pod-resources/recovery-kube-apiserver-pod/admin.kubeconfig"
```
tnozicka: this kills unit tests too

PR #464, opened 2019-05-07
Bug 1707877: Use internal url only for internal load balancer

damemi:

PR #465, opened 2019-05-07
Use internal url everywhere

damemi:
Switch everything to using the new internal url field, on the way to https://github.com/openshift/cluster-kube-apiserver-operator/pull/464


deads2k:
relies on https://github.com/openshift/installer/pull/1718

/lgtm

PR #467, opened 2019-05-07
don't wait for healthz on recovery-apiserver and don't destroy on shutdown

deads2k:
Eliminates all waiting, just creates manifests and exits to avoid hangs, timeouts, failures on usage, weird blocking behavior for a create when usage of the thing is not controlled.


smarterclayton:
As we get closer to release, please ensure code changes have a bug and the bug is associated in the PR title - follow the conventions described in previous emails about how to associate bugs with PRs.  The PR title must be `Bug XXXX: <description>`.

PR #469, opened 2019-05-08
Bug 1707839: regenerate-certificates needs to address derivative resources

deads2k:
We need to make the client-ca correct or kubelets cannot connect to the kube-apiserver.


deads2k:
This is built on current head and holding up disaster recovery.

Issue #470, opened 2019-05-08
regenerate-certificates has too many CA's in csr-signer


rphillips:
Running `cluster-kube-apiserver-operator regenerate-certificates` on a failed master causes the `/etc/kubernetes/static-pod-resources/kube-controller-manager-pod-5/secrets/csr-signer/tls.crt` file to contain two certificates. The controller-manager CSR signer then fails signing certificates with this error:

```
E0509 20:04:43.804526       1 controllermanager.go:541] Error starting "csrsigning"
F0509 20:04:43.804566       1 controllermanager.go:240] error starting controllers: failed to start certificate controller: error parsing CA cert file "/etc/kubernetes/static-pod-resources/secrets/cs
r-signer/tls.crt": {"code":1003,"message":"the PEM file should contain only one object"}
```

Manually removing the offending certificates from tls.crt and rebooting the node fixes the issue allowing the controller-manager to sign certificates.

/cc @tnozicka

wking:
Can we close this with #471 landed?

rphillips:
I was verifying it and ran into an issue which may be fixed here: https://github.com/openshift/cluster-kube-apiserver-operator/pull/472. I'll close this once verified.

rphillips:
Works!

PR #471, opened 2019-05-09
Fix csr-signer location to target namespace

tnozicka:
https://github.com/openshift/cluster-kube-apiserver-operator/pull/469 has fixed the csr-signer certificate format, but it forgot to use the new file

/cc @deads2k @soltysh


deads2k:
/lgtm

If we get it, great.  If we don't, put it in a zstream.

rphillips:
Fixes https://github.com/openshift/cluster-kube-apiserver-operator/issues/470

PR #473, opened 2019-05-14
Bug 1709869: tls: add openshift service names

sanchezl:


smarterclayton:
Be sure to put the correct bug name before this merges

PR #475, opened 2019-05-15
Bug 1710293 - Don't explicitly set storagePrefix to use the default value

soltysh:
/assign @smarterclayton @mfojtik @sttts @deads2k @derekwaynecarr 

Not setting this value will cause us to default it from here: https://github.com/openshift/origin/blob/7599337aeefb8314f9b5bb832b42639d4e28ce26/pkg/cmd/openshift-kube-apiserver/configdefault/kubecontrolplane_default.go#L58


mfojtik:
/lgtm

Same comment https://github.com/openshift/cluster-kube-apiserver-operator/pull/476#issuecomment-492743971

tnozicka:
/override ci/prow/e2e-aws-upgrade
based on https://github.com/openshift/cluster-kube-apiserver-operator/pull/476#issuecomment-492774939

PR #476, opened 2019-05-15
Bug 1710293 - Don't explicitly set storagePrefix to use the default value

soltysh:
/assign @smarterclayton @mfojtik @sttts @deads2k @derekwaynecarr 

Not setting this value will cause us to default it from here: https://github.com/openshift/origin/blob/7599337aeefb8314f9b5bb832b42639d4e28ce26/pkg/cmd/openshift-kube-apiserver/configdefault/kubecontrolplane_default.go#L58

### DIFF
--- a/bindata/v3.11.0/kube-apiserver/defaultconfig.yaml
+++ b/bindata/v3.11.0/kube-apiserver/defaultconfig.yaml
@@ -108,7 +108,6 @@ storageConfig:
   ca: /etc/kubernetes/static-pod-resources/configmaps/etcd-serving-ca/ca-bundle.crt
   certFile: /etc/kubernetes/static-pod-resources/secrets/etcd-client/tls.crt
   keyFile: /etc/kubernetes/static-pod-resources/secrets/etcd-client/tls.key
-  storagePrefix: openshift.io
   urls: null
 userAgentMatchingConfig:
   defaultRejectionMessage: ""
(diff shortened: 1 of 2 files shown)


bindata/v3.11.0/kube-apiserver/defaultconfig.yaml
```diff
@@ -108,7 +108,6 @@ storageConfig:
   ca: /etc/kubernetes/static-pod-resources/configmaps/etcd-serving-ca/ca-bundle.crt
   certFile: /etc/kubernetes/static-pod-resources/secrets/etcd-client/tls.crt
   keyFile: /etc/kubernetes/static-pod-resources/secrets/etcd-client/tls.key
-  storagePrefix: openshift.io
```
derekwaynecarr: will we remove this field entirely from storage config in a follow-on?

it seems we always hard-code this 
https://github.com/openshift/origin/blob/1a61ff214ee8a193e4b7331efe8ce6793b3725ec/pkg/cmd/openshift-kube-apiserver/configdefault/openshiftcontrolplane_default.go#L9

mfojtik: @derekwaynecarr we can remove this in openshift apiserver operator config

mfojtik:
/lgtm

Discussed via Slack, this will break the beta upgrades and existing clusters after receiving an upgrade won't see the kubernetes resources created pre-upgrade.

For that, the e2e-upgrade is failing now and this PR will have to be green button merged (or we need to use /override).

@smarterclayton

smarterclayton:
/override ci/prow/e2e-aws-upgrade

This job deliberately breaks backwards compatibility with no possible migration

PR #478, opened 2019-05-17
Bug 1711185 - [4.1] Remove storage prefix to default to correct one for kube resources in recovery apisever

tnozicka:
Catches up with storage change made here https://github.com/openshift/cluster-kube-apiserver-operator/pull/476

/priority critical-urgent

/cc @openshift/sig-master @derekwaynecarr @smarterclayton


tnozicka:
/hold
( I think for the backport we technically need and architect's +1)

PR #480, opened 2019-05-17
Set serviceSubnet for recovery apiserver

tnozicka:
if not set a lot of these are seen
```
openshift-kube-scheduler-operator   2m2s    Warning   ClusterIPOutOfRange                 service/metrics                                                           Cluster IP 172.30.24.255 is not within the service CIDR 10.0.0.0/24; please recreate service
openshift-kube-scheduler-operator   25s     Warning   ClusterIPNotAllocated               service/metrics                                                           Cluster IP 172.30.24.255 is not allocated; repairing
```

/cc @sttts @deads2k 
/cherrypick release-4.1

### DIFF
--- a/bindata/v4.1.0/kube-apiserver/recovery-config.yaml
+++ b/bindata/v4.1.0/kube-apiserver/recovery-config.yaml
@@ -17,6 +17,7 @@ storageConfig:
   ca: /etc/kubernetes/static-pod-resources/etcd-serving-ca-bundle.crt
   urls:
   - "https://localhost:2379"
+servicesSubnet: 172.30.0.0/16
 
 # Make hypershift happy.
 # (Everything bellow this line is just to provide some certs file
--- a/pkg/operator/v410_00_assets/bindata.go
+++ b/pkg/operator/v410_00_assets/bindata.go
@@ -451,6 +451,7 @@ storageConfig:
   ca: /etc/kubernetes/static-pod-resources/etcd-serving-ca-bundle.crt
   urls:
   - "https://localhost:2379"
+servicesSubnet: 172.30.0.0/16
 
 # Make hypershift happy.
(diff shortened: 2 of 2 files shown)


bindata/v4.1.0/kube-apiserver/recovery-config.yaml
```diff
@@ -17,6 +17,7 @@ storageConfig:
   ca: /etc/kubernetes/static-pod-resources/etcd-serving-ca-bundle.crt
   urls:
   - "https://localhost:2379"
+servicesSubnet: 172.30.0.0/16
```
mfojtik: /cc @squeed

squeed: NACK, this is a user-configurable range.

squeed:
This is a user-configurable range, so you'll have to get it from either the network configuration (`network.config.openshift.io` or the install config).

I realize that you might have a chicken-and-egg problem here, though. If so, let's figure it out. But you definitely can't start an apiserver with the wrong service cidr unless you are **1000%** sure no Services will be created.

tnozicka:
no apiservices will be created. We already have a wrong service range by not specifying it, at least specifying the default once correctly seems helpful in limiting the number of error seen in the recovery apiserver.


> This is a user-configurable range, so you'll have to get it from either the network configuration (network.config.openshift.io or the install config).

Wonder if the operator recreates all the services in the cluster when this is changed...

tnozicka:
I guess we can parse it out from mounts and the config file of the real apiserver on disk in the future, this at least set the default one correctly

sttts:
I agree with @squeed that using the default is not an option here. We can either load the latest value from /etc/static-pod-resources.... or we make the installer pod to write out the value somewhere, like a recovery config. IMO the later would be more elegant and stable.

PR #483, opened 2019-05-28
Bug 1714158 - Limit races between recovery tools and installer pods and cert-syncer

tnozicka:
It makes `regenerate-certificates` command to create a hold file `/etc/kubernetes/manifests/INSTALLER_HOLD`. 
If an installer pod sees the file when it is starting, it exits because it could rollout a new revision before all the certs are fixed in the api. `regenerate-certificates` force triggers necessary rollouts when it is done.
If cert-syncer pod sees the file when starting, it exits. It uses caches so re-queue would be risky. Kubelet will manage backoff and restart.

Requires:
 - https://github.com/openshift/library-go/pull/420
/hold

TODO:
- [x] Extract library-go changes and re-vendor back
- [x] Emit event when pods run into hold file

### DIFF
--- a/bindata/v4.1.0/kube-apiserver/pod.yaml
+++ b/bindata/v4.1.0/kube-apiserver/pod.yaml
@@ -86,6 +86,7 @@ spec:
       - --kubeconfig=/etc/kubernetes/static-pod-resources/configmaps/kube-apiserver-cert-syncer-kubeconfig/kubeconfig
       - --namespace=$(POD_NAMESPACE)
       - --destination-dir=/etc/kubernetes/static-pod-certs
+      - --pod-manifest-dir=/etc/kubernetes/manifests
     resources:
       requests:
         memory: 50Mi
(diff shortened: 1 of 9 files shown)


deads2k:
Can you write in the description what this does?  I think the goal is to write a file to some location (describe which with an example), that will cause installer pods to fail that will cause the operator to go degraded (you need to ensure a good message for how to fix it), to avoid a case where the cert syncer will be stomping files already present.

This is technically plausible, but have we observed an issue in the wild?  Because you updated the resources in etcd, it seems unlikely that we'll stomp back a bad config.

tnozicka:
@deads2k I've updated the description. Also I have seen this in the wild (cert-syncer race), QA hit it as well. 

btw. cert-syncer uses caches and on change overwrites all files, not just the changed one.

tnozicka:
/hold
(for https://github.com/openshift/library-go/pull/420)

PR #491, opened 2019-06-03
Bug 1716426 : Fix typo in admission pluginConfig

mfojtik:
Fix: https://bugzilla.redhat.com/show_bug.cgi?id=1716426

### DIFF
--- a/bindata/v4.1.0/kube-apiserver/defaultconfig.yaml
+++ b/bindata/v4.1.0/kube-apiserver/defaultconfig.yaml
@@ -1,21 +1,13 @@
 apiVersion: kubecontrolplane.config.openshift.io/v1
 kind: KubeAPIServerConfig
 admission:
-  pluginConfigs:
+  pluginConfig:
     network.openshift.io/ExternalIPRanger:
       configuration:
         allowIngressIP: true
         apiVersion: network.openshift.io/v1
         externalIPNetworkCIDRs: null
         kind: ExternalIPRangerAdmissionConfig
-      location: ""
-    network.openshift.io/RestrictedEndpointsAdmission:
-      configuration:
-        apiVersion: network.openshift.io/v1
-        kind: RestrictedEndpointsAdmissionConfig
-        restrictedCIDRs:
-        - 10.3.0.0/16 # ServiceCIDR
-        - 10.2.0.0/16 # ClusterCIDR
       location: ""
 aggregatorConfig:
   proxyClientInfo:
--- a/pkg/operator/configobservation/network/observe_network.go
+++ b/pkg/operator/configobservation/network/observe_network.go
@@ -15,7 +15,7 @@ func ObserveRestrictedCIDRs(genericListers configobserver.Listers, recorder even
 	listers := genericListers.(configobservation.Listers)
 
 	var errs []error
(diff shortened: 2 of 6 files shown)


bindata/v4.1.0/kube-apiserver/defaultconfig.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: kubecontrolplane.config.openshift.io/v1
 kind: KubeAPIServerConfig
 admission:
-  pluginConfigs:
+  pluginConfig:
```
danwinship: looking further down:

    network.openshift.io/RestrictedEndpointsAdmission:
      configuration:
        apiVersion: network.openshift.io/v1
        kind: RestrictedEndpointsAdmissionConfig
        restrictedCIDRs:
        - 10.3.0.0/16 # ServiceCIDR
        - 10.2.0.0/16 # ClusterCIDR

neither of those CIDRs is correct. I think the values here will just get overwritten at runtime with the correct values read from the network config status, but it's weird to have explicitly wrong values in there by default...

mfojtik: @danwinship what are the correct defaults?

mfojtik: we will remove this config and wait for network operator to set it.

deads2k: To be clear, he's talking about this code here: https://github.com/openshift/cluster-kube-apiserver-operator/tree/master/pkg/operator/configobservation/network

danwinship: `172.30.0.0/16` for ServiceCIDR and `10.128.0.0/14` for ClusterCIDR, but they can be overridden from install-config.yaml

deads2k: @mfojtik is that code typo'd too? https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/pkg/operator/configobservation/network/observe_network.go#L18

mfojtik: yeah, that looks bugged as well... fixed.

mfojtik:
@danwinship @deads2k updated

squeed:
Not there yet:
`F0603 19:36:25.026838       1 cmd.go:66] failed to initialize admission: couldn't init admission plugin "network.openshift.io/RestrictedEndpointsAdmission": error reading config: Object 'Kind' is missing in '{"restrictedCIDRs":["10.128.0.0/14","172.30.0.0/16"]}'`

deads2k:
/lgtm
/hold

holding so it doesn't cycle over and over.

@squeed for the future, have a look at the aos-devel thread with Derek and myself called "admission plugins and you" for a list of ownership and how we're handling integration for kube-apiserver and openshift-apiserver configuration in support of various features.  Everyone has a pet setting and we're looking for feature owners to own enabling their own features.

mfojtik:
Replaced by https://github.com/openshift/cluster-kube-apiserver-operator/pull/494

PR #492, opened 2019-06-03
Bug 1716622: start informers after lister registration

sanchezl:
Picking up https://github.com/openshift/library-go/pull/435
To fix https://bugzilla.redhat.com/show_bug.cgi?id=1716622


sanchezl:
/hold

fakebump until https://github.com/openshift/library-go/pull/435 merges

glide.yaml
```diff
@@ -23,7 +23,8 @@ import:
 - package: github.com/openshift/client-go
   version: master
 - package: github.com/openshift/library-go
-  version: master
```
deads2k: hold until this can just be release-4.1

glide.yaml
```diff
@@ -23,7 +23,8 @@ import:
 - package: github.com/openshift/client-go
   version: master
```
deads2k: make this release-4.1

sanchezl: done
also pinned api

PR #493, opened 2019-06-04
Set short rotation period

soltysh:
The bug to get this reverted for 4.2 is https://bugzilla.redhat.com/show_bug.cgi?id=1716978

This should set base cert rotation for 1m, which means that most certs will be rotated every 15 minutes, with validity for 30 mins. 

/assign @deads2k  @mfojtik


smarterclayton:
It looks like this broke CI, we are testing the revert.

We should identify how something this disruptive was able to merge (if this was the culprit)

PR #494, opened 2019-06-04
 Bug 1716426 : Fix admission plugin configuration

squeed:
This takes over #491, which fixes admission plugin configuration transmutation, but uncovered a bug in the network observer.

This PR also fixes the network observer.

### DIFF
--- a/bindata/v4.1.0/kube-apiserver/defaultconfig.yaml
+++ b/bindata/v4.1.0/kube-apiserver/defaultconfig.yaml
@@ -1,21 +1,13 @@
 apiVersion: kubecontrolplane.config.openshift.io/v1
 kind: KubeAPIServerConfig
 admission:
-  pluginConfigs:
+  pluginConfig:
     network.openshift.io/ExternalIPRanger:
       configuration:
         allowIngressIP: true
         apiVersion: network.openshift.io/v1
         externalIPNetworkCIDRs: null
         kind: ExternalIPRangerAdmissionConfig
-      location: ""
-    network.openshift.io/RestrictedEndpointsAdmission:
-      configuration:
-        apiVersion: network.openshift.io/v1
-        kind: RestrictedEndpointsAdmissionConfig
-        restrictedCIDRs:
-        - 10.3.0.0/16 # ServiceCIDR
-        - 10.2.0.0/16 # ClusterCIDR
       location: ""
 aggregatorConfig:
   proxyClientInfo:
--- a/pkg/operator/configobservation/network/observe_network.go
+++ b/pkg/operator/configobservation/network/observe_network.go
@@ -15,7 +15,7 @@ func ObserveRestrictedCIDRs(genericListers configobserver.Listers, recorder even
 	listers := genericListers.(configobservation.Listers)
 
 	var errs []error
-	restrictedCIDRsPath := []string{"admissionPluginConfig", "network.openshift.io/RestrictedEndpointsAdmission", "configuration", "restrictedCIDRs"}
+	restrictedCIDRsPath := []string{"admission", "pluginConfig", "network.openshift.io/RestrictedEndpointsAdmission", "configuration", "restrictedCIDRs"}
 
 	previouslyObservedConfig := map[string]interface{}{}
 	if currentRestrictedCIDRBs, _, err := unstructured.NestedStringSlice(existingConfig, restrictedCIDRsPath...); len(currentRestrictedCIDRBs) > 0 {
--- a/pkg/operator/configobservation/network/observe_network.go
+++ b/pkg/operator/configobservation/network/observe_network.go
@@ -10,57 +10,94 @@ import (
 	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/configobservation"
 )
 
-// ObserveRestrictedCIDRs observes list of restrictedCIDRs.
+// ObserveRestrictedCIDRs generates two configuration
 func ObserveRestrictedCIDRs(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
 	listers := genericListers.(configobservation.Listers)
 
 	var errs []error
-	restrictedCIDRsPath := []string{"admission", "pluginConfig", "network.openshift.io/RestrictedEndpointsAdmission", "configuration", "restrictedCIDRs"}
+	configPath := []string{"admission", "pluginConfig", "network.openshift.io/RestrictedEndpointsAdmission", "configuration"}
 
-	previouslyObservedConfig := map[string]interface{}{}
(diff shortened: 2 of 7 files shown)


pkg/operator/configobservation/network/observe_network.go
```diff
@@ -10,57 +10,94 @@ import (
...
-	//	    restrictedCIDRs:
-	//	    - 10.3.0.0/16 # ServiceCIDR
-	//	    - 10.2.0.0/16 # ClusterCIDR
-	//  servicesSubnet: 10.3.0.0/16
+	//  admission:
+	//    pluginConfig:
+	//      network.openshift.io/RestrictedEndpointsAdmission:
+	//        version: network.openshift.io/v1
+	//        kind: RestrictedEndpointsAdmission
+	//	      configuration:
```
deads2k: nesting looks really weird.

pkg/operator/configobservation/network/observe_network.go
```diff
@@ -10,57 +10,94 @@ import (
...
-			errs = append(errs, err)
-		}
+	admissionControllerConfig := unstructured.Unstructured{}
+	prev, ok, err := unstructured.NestedMap(existingConfig, configPath...)
+	if err != nil {
+		errs = append(errs, err)
+	}
+	if ok && err != nil {
+		// If we ever bump the API version, we'll have to do conversion here.
+		admissionControllerConfig.SetUnstructuredContent(prev)
```
deads2k: this reads oddly to me.  Why are we taking the old config?

deads2k:
looks plausible.  A few questions

pkg/operator/configobservation/network/observe_network.go
squeed: I was confused too. This is a straight port of the old code. It seemed to preserve previous so that it could return it in case it wasn't able to retrieve the network configuration. I'm not sure how useful that actually is.

pkg/operator/configobservation/network/observe_network.go
squeed: huh, code had some tabs mixed in there.

pkg/operator/configobservation/network/observe_network.go
```diff
@@ -10,57 +10,93 @@ import (
...
+	//    pluginConfig:
+	//      network.openshift.io/RestrictedEndpointsAdmission:
+	//        version: network.openshift.io/v1
+	//        kind: RestrictedEndpointsAdmission
+	//        configuration:
+	//          restrictedCIDRs:
+	//            - 10.3.0.0/16 # ServiceCIDR
+	//            - 10.2.0.0/16 # ClusterCIDR
 	restrictedCIDRs := clusterCIDRs
 	if len(serviceCIDR) > 0 {
```
deads2k: if you don't have restrictedcidrs or you don't have servicecidr, you should return what you should have.

pkg/operator/configobservation/network/observe_network.go
```diff
@@ -10,57 +10,93 @@ import (
...
-	//    network.openshift.io/RestrictedEndpointsAdmission:
-	//	  configuration:
-	//	    restrictedCIDRs:
-	//	    - 10.3.0.0/16 # ServiceCIDR
-	//	    - 10.2.0.0/16 # ClusterCIDR
-	//  servicesSubnet: 10.3.0.0/16
+	//  admission:
+	//    pluginConfig:
+	//      network.openshift.io/RestrictedEndpointsAdmission:
+	//        version: network.openshift.io/v1
```
deads2k: this nesting doesn't match previous.

squeed: because it was wrong :-)

bindata/v4.1.0/kube-apiserver/defaultconfig.yaml
```diff
@@ -1,22 +1,14 @@
 apiVersion: kubecontrolplane.config.openshift.io/v1
 kind: KubeAPIServerConfig
 admission:
-  pluginConfigs:
+  pluginConfig:
     network.openshift.io/ExternalIPRanger:
```
danwinship: do we know how to test if this is working now as well?

pkg/operator/configobservation/network/observe_network.go
```diff
@@ -10,57 +10,93 @@ import (
...
+
+	observedConfig := map[string]interface{}{}
+	unstructured.SetNestedMap(observedConfig, admissionControllerConfig.Object, configPath...)
+
+	return observedConfig, errs
+}
+
+// ObserveServicesSubnet watches the network configuration and generates the
+// servicesSubnet
+func ObserveServicesSubnet(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
```
danwinship: This appears to only be used from TestObserveServicesSubnet()?

pkg/operator/targetconfigcontroller/targetconfigcontroller_test.go
```diff
@@ -35,6 +35,7 @@ func TestIsRequiredConfigPresent(t *testing.T) {
 		     }
 		   ]
 		 },
+         "admission": {"pluginConfig": { "network.openshift.io/RestrictedEndpointsAdmission": {}}},
 		 "storageConfig": {
```
danwinship: indentation issues

pkg/operator/configobservation/network/observe_network.go
squeed: fixed

pkg/operator/configobservation/network/observe_network.go
squeed: fixed

pkg/operator/configobservation/network/observe_network.go
squeed: good catch, fixed.

squeed:
@danwinship @deads2k updated, thanks.

squeed:
Looks like we've got it now:

```
sh-4.2# cat /tmp/kubeapiserver-admission-config.yaml587128050 
apiVersion: apiserver.k8s.io/v1alpha1
kind: AdmissionConfiguration
plugins:
- configuration:
    allowIngressIP: true
    apiVersion: network.openshift.io/v1
    externalIPNetworkCIDRs: null
    kind: ExternalIPRangerAdmissionConfig
  name: network.openshift.io/ExternalIPRanger
  path: ""
- configuration:
    apiVersion: network.openshift.io/v1
    kind: RestrictedEndpointsAdmissionConfig
    restrictedCIDRs:
    - 10.128.0.0/14
    - 172.30.0.0/16
  name: network.openshift.io/RestrictedEndpointsAdmission
  path: ""
```

deads2k:
@sanchezl did someone mess with our job in openshfit/release?

squeed:
Before we cherry-pick this, I want to make sure we didn't break external ips.

squeed:
Yes, this breaks external IPs... but we never noticed because the admission controller was broken.

We can fix this with an API change in 4.2, and the backport will just disable it.

PR #496, opened 2019-06-05
Bug 1716426: Backport admission controller fixes

squeed:
This backports the the admission controller fixes in #494, along with an extra commit that leaves the ExternalIPRanger disabled.


squeed:
cc @mfojtik @danwinship

PR #499, opened 2019-06-06
Bug 1718052: Do not log OAuth tokens in audit logs

enj:
Even the metadata level is not safe as metadata.name is the secret.

@deads2k


wking:
/hold

Tgis keeps dying on an existing A record.  `oc delete project $YOUR_PROJECT` to get a freshlt-randomized project name, and tgen pull tge hold.

PR #501, opened 2019-06-11
manifests: add apiserver service monitor

s-urbaniak:
tldr: we want to move the service monitor creation for the apiserver out of cluster monitoring operator into the operator repos themselves.

/cc @sttts @brancz 

Fixes MON-670

### DIFF
--- a/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
+++ b/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
@@ -1,0 +1,37 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    k8s-app: apiserver
+  name: kube-apiserver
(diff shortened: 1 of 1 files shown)


manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    k8s-app: apiserver
+  name: kube-apiserver
+  namespace: openshift-monitoring
```
s-urbaniak: side note: once https://github.com/openshift/cluster-kube-apiserver-operator/pull/502 is merged and https://github.com/openshift/cluster-monitoring-operator/blob/master/assets/prometheus-k8s/service-monitor-apiserver.yaml removed, we can switch here to the `openshift-kube-apiserver` namespace.

PR #506, opened 2019-06-12
sets feature-gates in bootstrap-config-overrides for kube-apiserver-o…

p0lyn0mial:
explicitly sets the following features gates:
```
"ExperimentalCriticalPodAnnotation=true"
"RotateKubeletServerCertificate=true"
"SupportPodPidsLimit=true"
"LocalStorageCapacityIsolation=false"
```

### DIFF
--- a/bindata/bootkube/config/bootstrap-config-overrides.yaml
+++ b/bindata/bootkube/config/bootstrap-config-overrides.yaml
@@ -54,3 +54,9 @@ admission:
         - {{.}}{{end}}{{range .ClusterCIDR}}
         - {{.}}{{end}}
     {{end}}
+apiServerArguments:
+  feature-gates:
+    - "ExperimentalCriticalPodAnnotation=true"
+    - "RotateKubeletServerCertificate=true"
+    - "SupportPodPidsLimit=true"
+    - "LocalStorageCapacityIsolation=false"
--- a/pkg/cmd/render/render_test.go
+++ b/pkg/cmd/render/render_test.go
@@ -1,8 +1,19 @@
 package render
 
 import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
 	"reflect"
+	"strings"
 	"testing"
+
+	"github.com/ghodss/yaml"
+
+	configv1 "github.com/openshift/api/config/v1"
+	kubecontrolplanev1 "github.com/openshift/api/kubecontrolplane/v1"
 )
 
 var (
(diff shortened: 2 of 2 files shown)


bindata/bootkube/config/bootstrap-config-overrides.yaml
```diff
@@ -54,3 +54,9 @@ admission:
         - {{.}}{{end}}{{range .ClusterCIDR}}
         - {{.}}{{end}}
     {{end}}
+apiServerArguments:
+  feature-gates:
+    - "ExperimentalCriticalPodAnnotation=true"
+    - "RotateKubeletServerCertificate=true"
+    - "SupportPodPidsLimit=true"
+    - "LocalStorageCapacityIsolation=false"
```
sttts: @sjenning can you take a look? Every time we touch feature gates in the observers, we have to do the same here. Otherwise during bootstrapping the gates are wrong.

sttts:
/approve
lgtm

Leaving final lgtm to @sjenning.

bindata/bootkube/config/bootstrap-config-overrides.yaml
sjenning: does the apiserver care anything about these feature gates?

sjenning: as @sttts, we would need to keep these in sync with https://github.com/openshift/api/blob/master/config/v1/types_feature.go#L57-L90

sttts: @sjenning this is necessary for all operators where the operand uses those gates. And yes @ sync. We (as the non-owners of the features) don‘t really know where they influence which operand. The goal of the unit test is to make those own the bootstrap gates who already own the gates during normal operation.

sttts: @sjenning there is a bz where we suspect that during the bootstrap minutes objects were created which failed to work later.

pkg/cmd/render/render_test.go
```diff
@@ -109,3 +119,131 @@ func TestDiscoverCIDRs(t *testing.T) {
...
+				"--templates-input-dir=" + templateDir,
+				"--asset-output-dir=",
+				"--config-output-file=",
+			},
+			testFunction: func(cfg *kubecontrolplanev1.KubeAPIServerConfig) error {
+				actualGates, ok := cfg.APIServerArguments["feature-gates"]
+				if !ok {
+					return fmt.Errorf("missing \"feature-gates\" entry in APIServerArguments")
+				}
+				expectedGates := []string{
```
deads2k: get this from https://github.com/openshift/api/blob/master/config/v1/types_feature.go#L68-L78 please

p0lyn0mial: sure, np.

sjenning:
now that the test compares to the default featuregate set in openshift/api
/lgtm

PR #509, opened 2019-06-17
ConfigObservation: add ExternalIPRanger admission controller 

squeed:
API PR: https://github.com/openshift/api/pull/347

This PR adds support for configuring the ExternalIPRanger admission controller from the new ExternalIPs field on the Network configuration object.

### DIFF
--- a/pkg/operator/configobservation/network/observe_network.go
+++ b/pkg/operator/configobservation/network/observe_network.go
@@ -10,7 +10,8 @@ import (
 	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/configobservation"
 )
 
-// ObserveRestrictedCIDRs generates two configuration
+// ObserveRestrictedCIDRs watches the network configuration and updates the
+// RestrictedEndpointsAdmission controller config.
 func ObserveRestrictedCIDRs(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
 	listers := genericListers.(configobservation.Listers)
 
@@ -52,7 +53,7 @@ func ObserveRestrictedCIDRs(genericListers configobserver.Listers, recorder even
 	//      network.openshift.io/RestrictedEndpointsAdmission:
 	//        configuration:
 	//          version: network.openshift.io/v1
-	//          kind: RestrictedEndpointsAdmission
+	//          kind: RestrictedEndpointsAdmissionConfig
 	//          restrictedCIDRs:
 	//            - 10.3.0.0/16 # ServiceCIDR
 	//            - 10.2.0.0/16 # ClusterCIDR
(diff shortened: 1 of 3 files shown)


glide.yaml
```diff
@@ -33,6 +33,8 @@ import:
   version: v1.0.0
 - package: github.com/stretchr/testify
   version: c679ae2cc0cb27ec3293fea7e254e47386f05d69
+- package: github.com/onsi/gomega
```
danwinship: this is pulling in a lot of stuff just to make the test case a little shorter...

pkg/operator/configobservation/network/observe_network.go
```diff
@@ -101,3 +101,77 @@ func ObserveServicesSubnet(genericListers configobserver.Listers, recorder event
...
+
+	conf := []string{}
+	for _, cidr := range externalIPPolicy.RejectedCIDRs {
+		conf = append(conf, "!"+cidr)
+	}
+	for _, cidr := range externalIPPolicy.AllowedCIDRs {
+		conf = append(conf, cidr)
+	}
+
+	allowIngressIP := len(ingressIPs) > 0
```
danwinship: Is this how we set the option in 3.x?

vendor/github.com/openshift/library-go/pkg/operator/configobserver/network/observe_network.go
```diff
@@ -3,7 +3,7 @@ package network
 import (
```
danwinship: This change needs to be made in library-go but you don't seem to have an outstanding PR there?

squeed: was waiting for the API change to merge, but yes.

glide.yaml
squeed: more like, glide is annoying and can't compute minimal updates :-/

pkg/operator/configobservation/network/observe_network.go
squeed: more-or-less: https://github.com/openshift/origin/blob/release-3.11/pkg/cmd/openshift-kube-apiserver/conversion.go#L30

squeed:
OK, updated after adding support to the openshift-controller-manager-operator too.

Blocked until
https://github.com/openshift/api/pull/347
https://github.com/openshift/library-go/pull/465
are merged.

glide.yaml
danwinship: Even ignoring stuff other than gomega being pulled in, I meant that pulling in gomega just to use in a few test cases seems excessive. You could just write the test cases not using gomega, like all the existing test cases.

pkg/operator/configobservation/network/observe_network.go
```diff
@@ -101,3 +102,84 @@ func ObserveServicesSubnet(genericListers configobserver.Listers, recorder event
...
+	//        configuration:
+	//          version: network.openshift.io/v1
+	//          kind: ExternalIPRangerAdmissionConfig
+	//          externalIPNetworkCIDRs: [...]
+	configPath := []string{"admission", "pluginConfig", "network.openshift.io/ExternalIPRanger", "configuration"}
+
+	// Need to handle 3 cases:
+	// 1. retrieval error: pass through existing, if it exists
+	// 2. Null egress policy: disable the admission controller by returning no config
+	// 3. Egress policy: enable the egress controller, pass through configuration.
```
danwinship: "egress" in the comments is wrong. ("**e**xternalip" + "in**gress**" = "egress"?)

squeed: hah. fixed.

squeed:
OK, API, library changes merged and vendored. Ready for final review + merging.

danwinship:
> vendor: add gomega (useful assetion library), update openshift/api

typo ("assetion")

Have you tested that this works? Is there a jira card or something to ensure we get QE testing of this feature so we don't break it again?

glide.yaml
sttts: Please put glide.yaml changes in a separate commit, then bump in its own commit. Make it easier to review.

sttts: How about this assert library we use here and there?

pkg/operator/configobservation/network/observe_network_test.go
```diff
@@ -3,18 +3,21 @@ package network
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-
 	"k8s.io/client-go/tools/cache"
 
 	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
 	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/configobservation"
 	"github.com/openshift/library-go/pkg/operator/events"
+
+	. "github.com/onsi/gomega"
```
sttts: use github.com/stretchr/testify/assert instead of this heavy weight

glide.yaml
squeed: Do you mean to do `glide update` in commit 1, then edit glide.yaml and `glide update` in commit 2, or edit glide.yaml in commit 1, and `glide update` in commit 2?  Either way, sure I'll use testify/assert.

pkg/operator/configobservation/network/observe_network_test.go
squeed: done

squeed:
OK, rebased and tested this pretty thoroughly. Seems good. Ready for final review, @stts @danwinship.

squeed:
All green.

pkg/operator/configobservation/network/observe_network_test.go
```diff
@@ -61,17 +52,18 @@ func TestObserveRestrictedCIDRs(t *testing.T) {
...
-		t.Fatal(err)
-	}
-	if restrictedCIDRs[0] != "podCIDR" {
-		t.Error(restrictedCIDRs[0])
-	}
-	if restrictedCIDRs[1] != "serviceCIDR" {
-		t.Error(restrictedCIDRs[1])
-	}
+	assert.Empty(t, errors)
+	shouldMatchYaml(t, result, `
```
sttts: can you strictly unmarshal this into a real Golang config type everywhere? So we catch API type incompatibility and typos early.

squeed: That would require either copy-pasting the type or vendoring in kubernetes, Which would you prefer?

sttts: oh, it's just deep in k/k ? that wouldn't make it easy. And no, we don't want to vendor k/k :)

squeed:
/retest
(ci issues, something with images)

PR #510, opened 2019-06-20
allow --new-image for step 1 of 7

deads2k:
We want to use hyperkube instead of hypershift.  To do so in an orderly fashion will take a few steps with pulls to different repos.

`--newImage` allows us to change our image to one with a different binary without breaking functionality at intermediate steps
1. allow `--newImage`
2. installer passes `--image=old` and `--new-image=new`
3. operator render uses `--new-image` only
4. installer passes `--image=new` and `--new-image=new`
5. operator switches back to `--image`
6. installer passes `--image=new`
7. operator removes `--new-image`

### DIFF
--- a/pkg/cmd/render/render.go
+++ b/pkg/cmd/render/render.go
@@ -32,6 +32,16 @@ type renderOpts struct {
 	etcdServerURLs    []string
 	etcdServingCA     string
 	clusterConfigFile string
+
+	// newImage allows us to change our image to one with a different binary without breaking functionality at intermediate steps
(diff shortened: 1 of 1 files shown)


smarterclayton:
Won’t this break upgrades after it merges?  Or everything in the payload will be internally consistent?

deads2k:
> Won’t this break upgrades after it merges? Or everything in the payload will be internally consistent?

No. Everything in the payload will match.  This changes initial rendering only and each installer/operator tuple matches.

/retest

PR #514, opened 2019-07-01
Use new crd-schema-gen in Makefile

damemi:
This is an example of how we could use the crd-schema-gen hosted image in our makefiles, and ultimately remove the dependency on the library-go scripts which were based on hacked and patched forks of dependencies. Ideally I think we could just add these commands to library-go's alpha-build-machinery helpers and import from there.


damemi:
Note that this won't work right now because of https://github.com/openshift/api/issues/362

Makefile
```diff
@@ -40,11 +40,14 @@ test-e2e: GO_TEST_PACKAGES :=./test/e2e/...
 test-e2e: GO_TEST_FLAGS += -v
 test-e2e: test-unit
 
+PKG := ${PWD:${GOPATH}/%=%}
 update-codegen-crds:
-	go run ./vendor/github.com/openshift/library-go/cmd/crd-schema-gen/main.go --domain openshift.io --apis-dir vendor/github.com/openshift/api
+	docker pull registry.svc.ci.openshift.org/ocp/4.2:crd-schema-gen
```
sttts: is the pull necessary? Isn't there some option for docker run?

damemi: You are probably right, I'll look

damemi:
`make: podman: Command not found`, also tried with `docker`, what command should I use to run containers in makefiles?

tnozicka:
Given our tests run as containers, it sounds like you are trying to do nested docker which you can't (easily), it may be possible with podman

we'd need to extend the golang image with podman in a step or ask it to be included there (which I am not sure of, but there are already other tools there)
https://github.com/openshift/release/blob/master/ci-operator/config/openshift/cluster-kube-apiserver-operator/openshift-cluster-kube-apiserver-operator-master.yaml#L7

I'd like to see the nested podman tested manually first

damemi:
bumped this to pull in API fixes and ran the new generator, looks like the missing top-level descriptions are there now but this also seemed to generate validations for ObjectMeta, which was apparently missing before. Does that look right?

Makefile
```diff
@@ -40,11 +40,13 @@ test-e2e: GO_TEST_PACKAGES :=./test/e2e/...
...
 test-e2e: test-unit
 
+PKG := ${PWD:${GOPATH}/%=%}
 update-codegen-crds:
-	go run ./vendor/github.com/openshift/library-go/cmd/crd-schema-gen/main.go --domain openshift.io --apis-dir vendor/github.com/openshift/api
+	docker run -v ${PWD}:/go/$(PKG):Z -w /go/$(PKG) registry.svc.ci.openshift.org/ocp/4.2:crd-schema-gen --apis-dir vendor/github.com/openshift/api/operator/v1
 update-codegen: update-codegen-crds
 verify-codegen-crds:
-	go run ./vendor/github.com/openshift/library-go/cmd/crd-schema-gen/main.go --domain openshift.io --apis-dir vendor/github.com/openshift/api --verify-only
+	go get github.com/openshift/crd-schema-gen/... && go install github.com/openshift/crd-schema-gen/...
```
sttts: this must be versioned.

Makefile
```diff
@@ -44,8 +44,10 @@ PKG := ${PWD:${GOPATH}/%=%}
 update-codegen-crds:
 	docker run -v ${PWD}:/go/$(PKG):Z -w /go/$(PKG) registry.svc.ci.openshift.org/ocp/4.2:crd-schema-gen --apis-dir vendor/github.com/openshift/api/operator/v1
 update-codegen: update-codegen-crds
+
+TMP_GOPATH :=$(shell mktemp -d)
```
sttts: who deletes it?

Makefile
```diff
@@ -44,8 +44,10 @@ PKG := ${PWD:${GOPATH}/%=%}
...
 verify-codegen-crds:
-	go get github.com/openshift/crd-schema-gen/... && go install github.com/openshift/crd-schema-gen/...
+	export GOPATH=$(TMP_GOPATH) && export GOBIN=$(TMP_GOPATH)/bin && git clone -b release-4.2 --single-branch --depth 1 https://github.com/openshift/crd-schema-gen.git $(TMP_GOPATH)/src/github.com/openshift/crd-schema-gen && go install $(TMP_GOPATH)/src/github.com/openshift/crd-schema-gen/cmd/crd-schema-gen
```
sttts: `GOPATH=$(TMP_GOPATH) GOBIN=... go install ...` is enough. No need to export. git does not use them.

sttts: sure about GOBIN? We have to call it from there as well.

Makefile
```diff
@@ -40,11 +40,18 @@ test-e2e: GO_TEST_PACKAGES :=./test/e2e/...
...
-	go run ./vendor/github.com/openshift/library-go/cmd/crd-schema-gen/main.go --domain openshift.io --apis-dir vendor/github.com/openshift/api
+	git clone -b release-4.2 --single-branch --depth 1 https://github.com/openshift/crd-schema-gen.git $(TMP_GOPATH)/src/github.com/openshift/crd-schema-gen
+	GOPATH=$(TMP_GOPATH) go install $(TMP_GOPATH)/src/github.com/openshift/crd-schema-gen/cmd/crd-schema-gen
+	crd-schema-gen --apis-dir vendor/github.com/openshift/api/operator/v1
 update-codegen: update-codegen-crds
+
+TMP_GOPATH :=$(shell mktemp -d)
 verify-codegen-crds:
-	go run ./vendor/github.com/openshift/library-go/cmd/crd-schema-gen/main.go --domain openshift.io --apis-dir vendor/github.com/openshift/api --verify-only
+	git clone -b release-4.2 --single-branch --depth 1 https://github.com/openshift/crd-schema-gen.git $(TMP_GOPATH)/src/github.com/openshift/crd-schema-gen
```
sttts: thanks for the line breaks :)

Makefile
```diff
@@ -40,11 +40,18 @@ test-e2e: GO_TEST_PACKAGES :=./test/e2e/...
 test-e2e: GO_TEST_FLAGS += -v
 test-e2e: test-unit
 
+PKG := ${PWD:${GOPATH}/%=%}
 update-codegen-crds:
-	go run ./vendor/github.com/openshift/library-go/cmd/crd-schema-gen/main.go --domain openshift.io --apis-dir vendor/github.com/openshift/api
+	git clone -b release-4.2 --single-branch --depth 1 https://github.com/openshift/crd-schema-gen.git $(TMP_GOPATH)/src/github.com/openshift/crd-schema-gen
```
sttts: move both lines into a make target `crd-schema-gen`

Makefile
tnozicka: some people like @stlaz have set GOBIN explicitly and without setting it to temp we install it to the global GOBIN instead the one in temporary GOPATH

Makefile
```diff
@@ -40,11 +40,18 @@ test-e2e: GO_TEST_PACKAGES :=./test/e2e/...
 test-e2e: GO_TEST_FLAGS += -v
 test-e2e: test-unit
 
+PKG := ${PWD:${GOPATH}/%=%}
```
tnozicka: ihttps://github.com/openshift/library-go/blob/master/alpha-build-machinery/make/lib/golang.mk#L3 ?

Makefile
tnozicka: systemd

Makefile
```diff
@@ -40,11 +40,18 @@ test-e2e: GO_TEST_PACKAGES :=./test/e2e/...
...
-	go run ./vendor/github.com/openshift/library-go/cmd/crd-schema-gen/main.go --domain openshift.io --apis-dir vendor/github.com/openshift/api
-update-codegen: update-codegen-crds
-verify-codegen-crds:
-	go run ./vendor/github.com/openshift/library-go/cmd/crd-schema-gen/main.go --domain openshift.io --apis-dir vendor/github.com/openshift/api --verify-only
-verify-codegen: verify-codegen-crds
+PKG := ${PWD:${GOPATH}/%=%}
+crd-schema-gen:
+	git clone -b release-4.2 --single-branch --depth 1 https://github.com/openshift/crd-schema-gen.git $(TMP_GOPATH)/src/github.com/openshift/crd-schema-gen
+	GOPATH=$(TMP_GOPATH) go install $(TMP_GOPATH)/src/github.com/openshift/crd-schema-gen/cmd/crd-schema-gen
+	crd-schema-gen --apis-dir vendor/github.com/openshift/api/operator/v1
```
tnozicka: by convention we end phony target with `.PHONY: target-name

tnozicka: I see that was preexisting, so it's up to you to fix it

Makefile
```diff
@@ -40,11 +40,18 @@ test-e2e: GO_TEST_PACKAGES :=./test/e2e/...
...
+update-codegen: crd-schema-gen
+
+TMP_GOPATH :=$(shell mktemp -d)
+crd-schema-verify:
+	git clone -b release-4.2 --single-branch --depth 1 https://github.com/openshift/crd-schema-gen.git $(TMP_GOPATH)/src/github.com/openshift/crd-schema-gen
+	GOPATH=$(TMP_GOPATH) go install $(TMP_GOPATH)/src/github.com/openshift/crd-schema-gen/cmd/crd-schema-gen
```
tnozicka: set GOBIN

Makefile
```diff
@@ -40,11 +40,18 @@ test-e2e: GO_TEST_PACKAGES :=./test/e2e/...
...
+PKG := ${PWD:${GOPATH}/%=%}
+crd-schema-gen:
+	git clone -b release-4.2 --single-branch --depth 1 https://github.com/openshift/crd-schema-gen.git $(TMP_GOPATH)/src/github.com/openshift/crd-schema-gen
+	GOPATH=$(TMP_GOPATH) go install $(TMP_GOPATH)/src/github.com/openshift/crd-schema-gen/cmd/crd-schema-gen
+	crd-schema-gen --apis-dir vendor/github.com/openshift/api/operator/v1
+update-codegen: crd-schema-gen
+
+TMP_GOPATH :=$(shell mktemp -d)
+crd-schema-verify:
+	git clone -b release-4.2 --single-branch --depth 1 https://github.com/openshift/crd-schema-gen.git $(TMP_GOPATH)/src/github.com/openshift/crd-schema-gen
```
tnozicka: set dependency on `crd-schema-gen` instead installing it again

Makefile
damemi: this was leftover from when I was running the container instead of cloning and building, removed

Makefile
damemi: Is adding it to the .PHONY list at the end good?

Makefile
```diff
@@ -40,11 +40,17 @@ test-e2e: GO_TEST_PACKAGES :=./test/e2e/...
 test-e2e: GO_TEST_FLAGS += -v
 test-e2e: test-unit
 
-update-codegen-crds:
-	go run ./vendor/github.com/openshift/library-go/cmd/crd-schema-gen/main.go --domain openshift.io --apis-dir vendor/github.com/openshift/api
+CRD_GOPATH :=$(shell mktemp -d)
```
tnozicka: this will create a tmpdir on every invocation of the makefile even if nor related to crd, make it target specific

Makefile
tnozicka: it works the same way, our convention is to use it as `{}` to scope the target

Makefile
```diff
@@ -40,11 +40,18 @@ test-e2e: GO_TEST_PACKAGES :=./test/e2e/...
 test-e2e: GO_TEST_FLAGS += -v
 test-e2e: test-unit
 
-update-codegen-crds:
-	go run ./vendor/github.com/openshift/library-go/cmd/crd-schema-gen/main.go --domain openshift.io --apis-dir vendor/github.com/openshift/api
+crd-schema-gen:
```
tnozicka: install-crd-schema-gen?

Makefile
```diff
@@ -40,11 +40,19 @@ test-e2e: GO_TEST_PACKAGES :=./test/e2e/...
 test-e2e: GO_TEST_FLAGS += -v
 test-e2e: test-unit
 
-update-codegen-crds:
-	go run ./vendor/github.com/openshift/library-go/cmd/crd-schema-gen/main.go --domain openshift.io --apis-dir vendor/github.com/openshift/api
+CRD_GEN_VERSION := v1.0.0
```
sttts: CRD_SCHEMA_GEN_VERSION

Makefile
```diff
@@ -40,11 +40,19 @@ test-e2e: GO_TEST_PACKAGES :=./test/e2e/...
...
+crd-schema-gen:
+	git clone -b $(CRD_GEN_VERSION) --single-branch --depth 1 https://github.com/openshift/crd-schema-gen.git $(CRD_GOPATH)/src/github.com/openshift/crd-schema-gen
+	GOPATH=$(CRD_GOPATH) GOBIN=$(CRD_GOPATH)/bin go install $(CRD_GOPATH)/src/github.com/openshift/crd-schema-gen/cmd/crd-schema-gen
+
+update-codegen-crds: CRD_GOPATH :=$(shell mktemp -d)
```
sttts: CRD_SCHEMA_GEN_GOPATH

PR #515, opened 2019-07-02
APIServer config: Add observer for the extraCORSAllowedOrigins

stlaz:
This adds observation for the `extraCORSAllowedOrigins` field of the `APIServer` config resource.

Depends on https://github.com/openshift/api/pull/361

TODO:
- [x] e2e-operator test


stlaz:
I might be missing something as the `observedConfig` did not seem to pick up the change done to `APIServer.spec.extraCORSAllowedOrigins` when I was trying this on my cluster. The unit tests seem to work, though, and the input expectations seem to be in line with other APIServer observed fields. Am I missing something?

pkg/operator/configobservation/apiserver/observe_cors.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// ObserveDefaultUserServingCertificate returns an ObserveConfigFunc that observes user configuration
+// of CORSAllowedOrigins
+var ObserveExtraCORSAllowedOrigins configobserver.ObserveConfigFunc = (&apiServerObserver{
```
deads2k: please don't re-use apiServerObserver .  I think that pattern is overkill for your needs.  try to model it on an easier one like `ObserveInternalRegistryHostname`

pkg/operator/configobservation/apiserver/observe_cors.go
```diff
@@ -0,0 +1,55 @@
...
+	apiServer, err := listers.APIServerLister.Get("cluster")
+	if errors.IsNotFound(err) {
+		klog.Warningf("apiserver.config.openshift.io/cluster: not found")
+		return observedConfig, errs
+	}
+	if err != nil {
+		return prevObservedConfig, errs
+	}
+
+	newCORSAllowedOrigins := append(clusterDefaultCORSALlowedOrigins, apiServer.Spec.ExtraCORSAllowedOrigins...)
```
sttts: clusterDefaultCORSALlowedOrigins must go to line 24.

stlaz: This should now be fixed

test/e2e/user_cors_test.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	for i, tc := range testCases {
+		t.Logf("running test %d", i)
+		updateAPIServerClusterConfigSpec(configClient, func(apiserver *configv1.APIServer) {
+			apiserver.Spec.AdditionalCORSAllowedOrigins = tc.additionalCORS
+		})
+
+		var currentCORS []string
+		err = wait.PollImmediate(time.Second, 30*time.Minute, func() (bool, error) {
```
stlaz: This was an unreasonable timeout caused by typo, although the test still receives `[]string(nil)` from the observed config, I have to find out why that is.

test/e2e/user_cors_test.go
```diff
@@ -0,0 +1,105 @@
...
+		{
+			expectedConfigCORS: []string{
+				`//127\.0\.0\.1(:|$)`,
+				`//localhost(:|$)`,
+			},
+		},
+	}
+
+	for i, tc := range testCases {
+		t.Logf("running test %d", i)
```
sttts: `t.Run(fmt.Sprintf(%v, tc.additionalCORS), func(t *testing.T) { ...`

test/e2e/user_cors_test.go
sttts: for this kind of test timeouts there is `wait.ForeverTestTimeout`.

stlaz:
Thanks, I addressed the comments and fixed an error which would cause wrong current CORS in case of an error, the test is passing locally for me.

stlaz:
Rebased on top of current API

pkg/operator/configobservation/apiserver/observe_cors.go
```diff
@@ -0,0 +1,65 @@
...
+	`//127\.0\.0\.1(:|$)`,
+	`//localhost(:|$)`,
+}
+
+func ObserveAdditionalCORSAllowedOrigins(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
+	const corsAllowedOriginsPath = "corsAllowedOrigins"
+
+	listers := genericListers.(configobservation.Listers)
+	errs := []error{}
+	prevObservedConfig := map[string]interface{}{}
```
sttts: shouldn't we put clusterDefaultCORSALlowedOrigins in here?

pkg/operator/configobservation/apiserver/observe_cors.go
```diff
@@ -0,0 +1,65 @@
...
+
+	currentCORSAllowedOrigins, _, err := unstructured.NestedStringSlice(existingConfig, corsAllowedOriginsPath)
+	if err != nil {
+		return prevObservedConfig, append(errs, err)
+	}
```
sttts: if this worked fine, shouldn't we use `currentCORSAllowedOrigins` instead of `prevObservedConfig` instead in the error cases below?

pkg/operator/configobservation/apiserver/observe_cors.go
stlaz: I'll add that there, I guess it's safer to go with the cluster defaults so that at least have some CORS working

pkg/operator/configobservation/apiserver/observe_cors.go
stlaz: You're right, returning `existingConfig` seems safer given the error that could appear while trying to enrich the `currentCORSAllowedOrigins` with cluster defaults, I'll modify it as such.

pkg/operator/configobservation/apiserver/observe_cors.go
```diff
@@ -0,0 +1,60 @@
...
+	currentCORSAllowedOrigins, _, err := unstructured.NestedStringSlice(existingConfig, corsAllowedOriginsPath)
+	if err != nil {
+		return defaultConfig, append(errs, err)
+	}
+	currentCORSSet := sets.NewString(currentCORSAllowedOrigins...)
+	currentCORSSet.Insert(clusterDefaultCORSALlowedOrigins...)
+
+	observedConfig := map[string]interface{}{}
+	apiServer, err := listers.APIServerLister.Get("cluster")
+	newCORSSet := sets.NewString(clusterDefaultCORSALlowedOrigins...)
```
sttts: nit: move that down

stlaz:
@sttts addressed the last comment, lets not allocate vars if not necessary

PR #517, opened 2019-07-04
Introduce gosec for security checks

JAORMX:
gosec does static code analysis and checks for common security issues in golang codebases.

This PR introduces the tool, and sets it up as a test target in the Makefile.

Warnings were also addressed as part of this PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,11 @@ test-e2e: GO_TEST_PACKAGES :=./test/e2e/...
 test-e2e: GO_TEST_FLAGS += -v
 test-e2e: test-unit
 
+.PHONY: test-sec
+test-sec:
+	@which gosec 2> /dev/null >&1 || { echo "gosec must be installed to lint code";  exit 1; }
+	gosec -severity medium -confidence medium -exclude G304 -quiet ./...
+
 CRD_SCHEMA_GEN_VERSION := v1.0.0
 crd-schema-gen:
 	git clone -b $(CRD_SCHEMA_GEN_VERSION) --single-branch --depth 1 https://github.com/openshift/crd-schema-gen.git $(CRD_SCHEMA_GEN_GOPATH)/src/github.com/openshift/crd-schema-gen
--- a/pkg/cmd/regeneratecerts/regenerate_certificates.go
+++ b/pkg/cmd/regeneratecerts/regenerate_certificates.go
@@ -340,6 +340,9 @@ func (o *Options) Run() error {
 		}
 
 		dir := filepath.Join(def.toplevelDir, def.objectType, def.name)
+		// #nosec
+		// gosec(G301): This contains public certiicates, so it should be
+		// public.
 		err = os.MkdirAll(dir, 0755)
 		if err != nil {
 			return fmt.Errorf("failed to create dir %q: %v", dir, err)
--- a/pkg/recovery/apiserver.go
+++ b/pkg/recovery/apiserver.go
@@ -189,7 +189,7 @@ func (s *Apiserver) Create() error {
 	}
 
 	s.recoveryResourcesDir = filepath.Join(s.StaticPodResourcesDir, "recovery-kube-apiserver-pod")
-	err = os.Mkdir(s.recoveryResourcesDir, 755)
+	err = os.Mkdir(s.recoveryResourcesDir, 0750)
 	if err != nil {
 		if os.IsExist(err) {
 			klog.Errorf("Recovery dir %q already exist. Please use `recovery-apiserver destroy` command or remove the dir manually.", s.recoveryResourcesDir)
--- a/pkg/cmd/insecurereadyz/readyz.go
+++ b/pkg/cmd/insecurereadyz/readyz.go
@@ -71,6 +71,9 @@ func (r *readyzOpts) Complete() error {
 
 // Run contains the logic of the insecure-readyz command.
 func (r *readyzOpts) Run() error {
+	// #nosec
+	// gosec(G402): This is already assumed to skip the verification and that's
+	// the intent. So we can ignore this warning.
 	client := &http.Client{Transport: &http.Transport{
 		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
 	}}


Makefile
```diff
@@ -40,6 +40,11 @@ test-e2e: GO_TEST_PACKAGES :=./test/e2e/...
 test-e2e: GO_TEST_FLAGS += -v
 test-e2e: test-unit
 
+.PHONY: test-sec
+test-sec:
+	@which gosec 2> /dev/null >&1 || { echo "gosec must be installed to lint code";  exit 1; }
+	gosec -severity medium --confidence medium -quiet ./...
```
sttts: this belongs into github.com/openshift/library-go/alpha-build-machinery

pkg/recovery/helpers.go
```diff
@@ -92,7 +93,7 @@ func EnsureFileContent(filePath string, data []byte) error {
 			return fmt.Errorf("file %q is a directory", filePath)
 		}
 
-		fileBytes, err := ioutil.ReadFile(filePath)
+		fileBytes, err := ioutil.ReadFile(filepath.Clean(filePath))
```
sttts: why does this contribute to security?

pkg/cmd/regeneratecerts/regenerate_certificates.go
```diff
@@ -340,7 +340,7 @@ func (o *Options) Run() error {
 		}
 
 		dir := filepath.Join(def.toplevelDir, def.objectType, def.name)
-		err = os.MkdirAll(dir, 0755)
+		err = os.MkdirAll(dir, 0750)
```
sttts: how can static analysis confirm that this makes sense? Maybe it even does, am just curious.

Makefile
sttts: @tnozicka ^^

pkg/recovery/apiserver.go
```diff
@@ -189,7 +189,7 @@ func (s *Apiserver) Create() error {
 	}
 
 	s.recoveryResourcesDir = filepath.Join(s.StaticPodResourcesDir, "recovery-kube-apiserver-pod")
-	err = os.Mkdir(s.recoveryResourcesDir, 755)
+	err = os.Mkdir(s.recoveryResourcesDir, 0750)
```
sttts: wow, this was an actual bug :+1:

pkg/cmd/regeneratecerts/regenerate_certificates.go
JAORMX: static code analysis will issue false positives. This might or might not make sense; however, in many cases the intent is not to make these files readable by everyone. So, this is a good default to have unless you really want to share this directory.

pkg/recovery/helpers.go
JAORMX: The intent is to sanitize file paths that might come from the user. If this is useless, we can tag it as such with the comment:

`// #nosec`

sttts: in this case we sanitize paths which are passed to a Linux system call. Don't think this adds anything. Linux will do the same logic.

pkg/cmd/regeneratecerts/regenerate_certificates.go
sttts: @tnozicka can judge here whether it is good.

Makefile
JAORMX: @tnozicka any feedback?

pkg/cmd/regeneratecerts/regenerate_certificates.go
JAORMX: any more feedback?

Makefile
sttts: @tnozicka ^^^ ping

pkg/cmd/regeneratecerts/regenerate_certificates.go
tnozicka: there are configmaps (public CAs in this dir) so it makes sense for the dir itself to be public

pkg/recovery/apiserver.go
tnozicka: @sttts why? the folder contains both configmaps and secrets so it should be public. The actual secret files should be private an I hope they are.

pkg/recovery/apiserver.go
```diff
@@ -334,6 +334,7 @@ func (s *Apiserver) Destroy() error {
 }
 
 func copyFile(src, dest string) error {
+	// #nosec
```
tnozicka: I am skeptical to polluting code like this

Makefile
tnozicka: If we decide to use the tool, library-go should be the location.
I admit I don't have any experience with the tool but sounds like a linter. If we hook it up into makefile and nobody uses it it will just rot. We can't hook it into CI because of false positives. For the same reason we don't run golint.

pkg/recovery/apiserver.go
tnozicka: ok, so the octal vs. decimal is a bug, not the 5 at the end

pkg/recovery/apiserver.go
JAORMX: If we don't think this is a relevant check, we can disable it so we don't need to add these flags in this case.

Makefile
JAORMX: static analysis tools tend to give false positives. Hence why I'm not proposing this tool for CI, but for local testing. perhaps at some point it could be part of a common lint target or something of the sort.

pkg/cmd/regeneratecerts/regenerate_certificates.go
JAORMX: sure, I can change these. Also, if we think the rule doesn't make sense we can exclude it entirely.

pkg/recovery/apiserver.go
JAORMX: @sttts @tnozicka while they "should" be private, why is it harmful to still reduce the permissions given to this directory?

PR #522, opened 2019-07-12
Add strict config unmarshal step before creating a ConfigMap

sttts:
Catch observer bugs and user typos early.

### DIFF
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"k8s.io/klog"
+	"sigs.k8s.io/yaml"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -233,6 +236,17 @@ func manageKubeAPIServerConfig(client coreclientv1.ConfigMapsGetter, recorder ev
 	if err != nil {
 		return nil, false, err
 	}
+
+	// sanity check that we create a valid config
```
sttts: @deads2k we might want to exclude the unsupported config overrides here. Wdyt?

sttts: For those I am fine with non-strict unmarshal. But the observers concern me most because of tons of unstructured logic there.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -222,17 +224,42 @@ func createTargetConfig(c TargetConfigController, recorder events.Recorder, oper
 	return false, nil
 }
 
+// extraConfig are additional config fields in the observed config which are only used inside the
+// operator. They are pruned before sending the config to the operand.
+type extraConfig struct {
+}
```
sttts: @deads2k add anything additional here. It will be pruned from ConfigMap.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -222,17 +224,42 @@ func createTargetConfig(c TargetConfigController, recorder events.Recorder, oper
 	return false, nil
 }
 
+// extraConfig are additional config fields in the observed config which are only used inside the
+// operator. They are pruned before sending the config to the operand.
+type extraConfig struct {
```
enj: Can you provide an example of what may go here in the future?

sttts:
Replaced with https://github.com/openshift/cluster-kube-apiserver-operator/pull/686.

PR #524, opened 2019-07-12
Bug 1729242: OCP 4 AWS privilege escalation vulnerability by running pods on masters

ravisantoshgudimetla:
Make worker label as defaultNodeSelector to avoid pods getting scheduled to master nodes.

/cc @deads2k @enj @sttts

### DIFF
--- a/pkg/operator/configobservation/scheduler/observe_scheduler.go
+++ b/pkg/operator/configobservation/scheduler/observe_scheduler.go
@@ -1,6 +1,7 @@
 package scheduler
 
 import (
+	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/configobservation"
 	"github.com/openshift/library-go/pkg/operator/configobserver"
 	"github.com/openshift/library-go/pkg/operator/events"
@@ -9,6 +10,8 @@ import (
 
 	"k8s.io/klog"
 )
+
+const workerNodeSelector = "node-role.kubernetes.io/worker=\"\""
 
 // ObserveDefaultNodeSelector reads the defaultNodeSelector from the scheduler configuration instance cluster
 func ObserveDefaultNodeSelector(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
@@ -36,7 +39,12 @@ func ObserveDefaultNodeSelector(genericListers configobserver.Listers, recorder 
 	if err != nil {
 		return prevObservedConfig, errs
 	}
-
+	if schedulerConfig.Spec == (configv1.SchedulerSpec{}) {
+		if err := unstructured.SetNestedField(observedConfig, workerNodeSelector, defaultNodeSelectorPath...); err != nil {
+			errs = append(errs, err)
+		}
+		klog.V(4).Infof("Found an empty scheduler spec, so setting the defaultNodeSelector to worker nodes")
+	}
 	defaultNodeSelector := schedulerConfig.Spec.DefaultNodeSelector
 	if len(defaultNodeSelector) > 0 {
 		if err := unstructured.SetNestedField(observedConfig, defaultNodeSelector, defaultNodeSelectorPath...); err != nil {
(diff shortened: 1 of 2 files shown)


bindata/v4.1.0/kube-apiserver/defaultconfig.yaml
```diff
@@ -84,7 +84,7 @@ kubeletClientInfo:
   keyFile: /etc/kubernetes/static-pod-resources/secrets/kubelet-client/tls.key
   port: 10250
 projectConfig:
-  defaultNodeSelector: ""
+  defaultNodeSelector: "node-role.kubernetes.io/worker" # This value can be overriden in defaultNodeSelector field in scheduler CR
```
deads2k: this should be a config observer.

pkg/operator/configobservation/scheduler/observe_scheduler.go
```diff
@@ -36,7 +39,12 @@ func ObserveDefaultNodeSelector(genericListers configobserver.Listers, recorder
 	if err != nil {
 		return prevObservedConfig, errs
 	}
-
+	if schedulerConfig.Spec == (configv1.SchedulerSpec{}) {
```
enj: I prefer the defaulting pattern used in:

https://github.com/openshift/cluster-kube-apiserver-operator/blob/3568cbf5b403833349f73eb5de226ad60e10b5d5/pkg/operator/configobservation/auth/auth_metadata.go#L52

enj:
/hold

Per David's request to continue review on Monday.

pkg/operator/configobservation/scheduler/observe_scheduler.go
```diff
@@ -10,6 +11,8 @@ import (
 	"k8s.io/klog"
 )
 
+const workerNodeSelector = "node-role.kubernetes.io/worker=\"\""
```
smarterclayton: This is not guaranteed to exist in all clusters.

Node roles are not an api.

ravisantoshgudimetla: IIUC, there are 2 ways to distinguish master nodes from worker nodes:

- role label
- Master taints

So, are you suggesting, we cannot rely on role label which the `defaultNodeSelector` is based of? IIRC, we used to have something like `type=Compute` as defaultNodeSelector in 3.11. 

This is not to say that we cannot have a validation against pods have master tolerations

deads2k:
I'm a -1 on resolving the issue this way.  The mechanism for steering workloads away from masters is taints and restricting tolerations.  I thought this is what PodTolerationRestrictions did, but that plugin is very confused.  Mapping tolerations to rbac permissions for serviceaccounts seems more practical.

sfowl:
CVE-2019-10200 was assigned to this issue.

pkg/operator/configobservation/scheduler/observe_scheduler.go
jkupferer: @smarterclayton Can you explain what you mean by "Node roles are not an api and no business logic should based on node roles."? It has been common practice with OpenShift in the field to add role labels to nodes and separate workload based on these labels by using node selectors and the `openshift.io/node-selector` annotation. This has been a very effective design pattern, what is inappropriate about this? Should we encourage use of a different label?

jkupferer:
Can we get an update on where we are at on this issue?

We need this resolved both for the potential security impact and also to clarify how users should isolate workload when there is a security concern.

In my experience as an openshift consultant and instructor, taints and tolerations are not well suited for security isolation because most users find these difficult to understand and implement. While it may be possible to resolve this with podTolerationRestrictions, this is likely going to be hopelessly complicated for many OpenShift administrators. Node selectors are much easier for our users and administrators to understand and node selectors on namespaces are easy to audit with automation.

imho, we should label all masters as `node-role.kubernetes.io/master` everything else as workers, `node-role.kubernetes.io/worker`, and set the default for the defaultNodeSeloctor to run on workers as this PR was aiming to do. In the case where we want workload to run on the masters, we simply add the worker label to the masters, so it really shouldn't limit node topologies.

soltysh:
I think https://bugzilla.redhat.com/show_bug.cgi?id=1730165#c8 has the answer. In short version 2 of https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html should be able to solve the issue.
/close

PR #526, opened 2019-07-18
Bug 1731232: Report infrastructure provider metric from the operator

smarterclayton:
The infrastructure provider metric summarizes the desired type and
region (if any) to prometheus, based on the desired config. If type
is not recognized it is passed along, if it is empty it is set as
empty.

### DIFF
--- a/pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
+++ b/pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
@@ -1,13 +1,14 @@
 package configobservercontroller
 
 import (
-	"github.com/openshift/library-go/pkg/operator/configobserver/cloudprovider"
-	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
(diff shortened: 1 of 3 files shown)


pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -115,5 +119,47 @@ func NewConfigObserver(
...
 
+	registerMetrics(configInformer)
+
 	return c
 }
+
+// registerMetrics exposes core platform metrics that relate to the configuration
+// of Kubernetes.
+// TODO: in the future this may move to cluster-config-operator.
+func registerMetrics(configInformer configinformers.SharedInformerFactory) {
```
deads2k: since you have two, start us a little config metrics package called from starter.go?  We'll lose track of it here.

smarterclayton: sure

PR #527, opened 2019-07-18
Bug 1731228: Expose feature set field as a metric for alerting

smarterclayton:
A cluster that has specific featureSets configured is blocked from 
upgrading, which should be an alert. Expose a metric from the KAO which
allows the cluster to alert on this state.

### DIFF
--- a/pkg/operator/configmetrics/configmetrics.go
+++ b/pkg/operator/configmetrics/configmetrics.go
@@ -1,0 +1,48 @@
+package configmetrics
+
+import (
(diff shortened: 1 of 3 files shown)


deads2k:
fine with idea, not as fine with the spot. Because the informer is shared, we can actually register the metrics in a less ugly spot by passing the same informer right here-ish https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/pkg/operator/starter.go#L153

PR #528, opened 2019-07-18
bindata/bootkube: fix image for http-check-container

abhinavdahiya:
/cc @sttts @deads2k


abhinavdahiya:
/retest

```
one of the worker node went NotReady due to PLEG is not healthy: pleg was last seen active ~30m ago; threshold is 3m0s.
```

PR #530, opened 2019-07-24
Start observing the Proxy object

stlaz:
Adds an observer for the Proxy object, pulls its fields and passes
them to the environment of the deployed kube-apiserver container.

### DIFF
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -7,6 +7,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/ghodss/yaml"
+
 	"k8s.io/klog"
 
 	corev1 "k8s.io/api/core/v1"
@@ -20,6 +22,7 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
 
+	kubecontrolplanev1 "github.com/openshift/api/kubecontrolplane/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
@@ -222,7 +225,15 @@ func manageKubeAPIServerConfig(client coreclientv1.ConfigMapsGetter, recorder ev
 		".oauthConfig": RemoveConfig,
 	}
 
-	requiredConfigMap, _, err := resourcemerge.MergeConfigMap(configMap, "config.yaml", specialMergeRules, defaultConfig, operatorSpec.ObservedConfig.Raw, operatorSpec.UnsupportedConfigOverrides.Raw)
+	requiredConfigMap, _, err := resourcemerge.MergePrunedConfigMap(
+		&kubecontrolplanev1.KubeAPIServerConfig{},
+		configMap,
+		"config.yaml",
+		specialMergeRules,
+		defaultConfig,
+		operatorSpec.ObservedConfig.Raw,
+		operatorSpec.UnsupportedConfigOverrides.Raw,
+	)
 	if err != nil {
 		return nil, false, err
 	}
@@ -269,6 +280,20 @@ func managePod(client coreclientv1.ConfigMapsGetter, recorder events.Recorder, o
 		v = 2
 	}
 	required.Spec.Containers[0].Args = append(required.Spec.Containers[0].Args, fmt.Sprintf("-v=%d", v))
+
+	var observedConfig map[string]interface{}
+	if err := yaml.Unmarshal(operatorSpec.ObservedConfig.Raw, &observedConfig); err != nil {
+		return nil, false, fmt.Errorf("failed to unmarshal the observedConfig: %v", err)
+	}
+	proxyConfig, _, err := unstructured.NestedStringMap(observedConfig, "targetconfigcontroller", "proxy")
+	if err != nil {
+		return nil, false, fmt.Errorf("couldn't get the proxy config from observedConfig: %v", err)
+	}
+
+	proxyEnvVars := proxyMapToEnvVars(proxyConfig)
+	for i, container := range required.Spec.Containers {
+		required.Spec.Containers[i].Env = append(container.Env, proxyEnvVars...)
+	}
 
 	configMap := resourceread.ReadConfigMapV1OrDie(v410_00_assets.MustAsset("v4.1.0/kube-apiserver/pod-cm.yaml"))
 	configMap.Data["pod.yaml"] = resourceread.WritePodV1OrDie(required)
@@ -404,4 +429,17 @@ func (c *TargetConfigController) namespaceEventHandler() cache.ResourceEventHand
 			}
 		},
 	}
+}
+
+func proxyMapToEnvVars(proxyConfig map[string]string) []corev1.EnvVar {
+	if proxyConfig == nil {
+		return nil
+	}
+
+	envVars := []corev1.EnvVar{}
+	for k, v := range proxyConfig {
+		envVars = append(envVars, corev1.EnvVar{Name: k, Value: v})
+	}
+
+	return envVars
 }
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
 hash: 3242396de82077cf03c242cfe43d626bf4a5b69be0be6ad4de530c7004a6edb2
-updated: 2019-07-31T14:54:14.270029363-04:00
+updated: 2019-08-02T14:02:34.100638375+02:00
 imports:
 - name: github.com/apparentlymart/go-cidr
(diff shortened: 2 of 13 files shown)


mfojtik:
I think this code is generic enough to live in library-go and being re-used by other operators.

pkg/operator/configobservation/proxy/observe_proxy.go
```diff
@@ -0,0 +1,73 @@
...
+	"k8s.io/klog"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/configobservation"
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+)
+
+func ObserveProxyConfig(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
+	const proxyPath = "proxy"
```
mfojtik: move this outside, make a comment describing what this is.

pkg/operator/configobservation/proxy/observe_proxy.go
```diff
@@ -0,0 +1,73 @@
...
+	errs := []error{}
+	prevObservedProxyConfig := map[string]interface{}{}
+
+	// grab the current Proxy config to later check whether it was updated
+	currentProxyMap, _, err := unstructured.NestedStringMap(existingConfig, proxyPath)
+	if err != nil {
+		return prevObservedProxyConfig, append(errs, err)
+	}
+
+	if len(currentProxyMap) > 0 {
```
mfojtik: what happen if somebody delete the proxy configuration because he want to unset the proxy? is that case covered?

pkg/operator/configobservation/proxy/observe_proxy.go
```diff
@@ -0,0 +1,73 @@
...
+		unstructured.SetNestedStringMap(prevObservedProxyConfig, currentProxyMap, proxyPath)
+	}
+
+	observedConfig := map[string]interface{}{}
+	proxyConfig, err := listers.ProxyLister.Get("cluster")
+	if errors.IsNotFound(err) {
+		klog.Warningf("proxy.config.openshift.io/cluster: not found")
```
mfojtik: this might be noisy, I would move it to V(4-5) ?

mfojtik: check the service account one, use the same level

pkg/operator/configobservation/proxy/observe_proxy.go
```diff
@@ -0,0 +1,73 @@
...
+		recorder.Eventf("ObserveProxyConfig", "proxy changed to %q", newProxyMap)
+	}
+
+	return observedConfig, errs
+}
+
+func proxyToMap(proxy *configv1.Proxy) map[string]string {
+	proxyMap := map[string]string{}
+
+	if noProxy := proxy.Spec.NoProxy; len(noProxy) > 0 {
```
mfojtik: do we need to validate these fields? "i made a typo in proxy and now my cluster is hosed"

mfojtik: @danehans probably higher level question ^

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -277,6 +278,16 @@ func managePod(client coreclientv1.ConfigMapsGetter, recorder events.Recorder, o
...
 
+	var observedConfig map[string]interface{}
+	if err := yaml.Unmarshal(operatorConfig.Spec.ObservedConfig.Raw, &observedConfig); err != nil {
+		return nil, false, fmt.Errorf("failed to unmarshal the observedConfig: %v", err)
+	}
+	proxyConfig, _, err := unstructured.NestedStringMap(observedConfig, "proxy")
+	if err != nil {
+		return nil, false, fmt.Errorf("couldn't get the proxy config from observedConfig: %v", err)
+	}
+	required.Spec.Containers[0].Env = append(required.Spec.Containers[0].Env, proxyMapToEnvVars(proxyConfig)...)
```
mfojtik: do we want to check if the proxy var are already set?

pkg/operator/configobservation/proxy/observe_proxy.go
sttts: They are validated by the proxy controller inside of the network operator.

pkg/operator/configobservation/proxy/observe_proxy.go
sttts: this is our config, not the proxy object. It's fine.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -277,6 +278,16 @@ func managePod(client coreclientv1.ConfigMapsGetter, recorder events.Recorder, o
 	}
 	required.Spec.Containers[0].Args = append(required.Spec.Containers[0].Args, fmt.Sprintf("-v=%d", v))
 
+	var observedConfig map[string]interface{}
+	if err := yaml.Unmarshal(operatorConfig.Spec.ObservedConfig.Raw, &observedConfig); err != nil {
+		return nil, false, fmt.Errorf("failed to unmarshal the observedConfig: %v", err)
+	}
+	proxyConfig, _, err := unstructured.NestedStringMap(observedConfig, "proxy")
```
sttts: would choose a more unique field. Not that we add a proxy field to the official type and this break down.

targetconfigcontroller.proxy ?

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
sttts: will the last one win?

The blueprint for this comes from us. The user has no influence on it. So I don't think we need further validation.

mfojtik: @sttts if the required.Spec.Containers[0].Env already have the proxy vars from us, is this going to keep adding them over and over? ;-) I think that was my question. Agree, user won't be able to change this (we will stomp).

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -21,6 +21,7 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
 
+	"github.com/ghodss/yaml"
```
sttts: nit: order

pkg/operator/configobservation/proxy/observe_proxy.go
sttts: Can we instead share this path somehow between config observation and target config controller? Would make it clear that these two are interconnected.

pkg/operator/configobservation/proxy/observe_proxy.go
```diff
@@ -0,0 +1,73 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/klog"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/configobservation"
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+)
+
+func ObserveProxyConfig(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
```
sttts: godoc

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
stlaz: It's a good point, although I'd say we may want to decide that once we have the need to deviate from whatever there is in the `Proxy` object, as at least for the `noProxy` field, that's going to be fed with some cluster defaults from the very start if I understand the design of the feature properly.

stlaz:
@mfojtik was right to request the observer in library-go, here be the PR: https://github.com/openshift/library-go/pull/485

stlaz:
This now uses the not-yet merged code from lib-go, appears to work

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
stlaz: Went with `targetconfigcontroller.proxy` then

pkg/operator/configobservation/proxy/observe_proxy.go
stlaz: The function used here still kept the `klog` but I removed this logging completely in the library-go PR

stlaz:
Pulled in the latest library-go to get the proxy observer from there

stlaz:
Grabbed the `MergePrunedConfigMap()` method from library-go to auto-prune the fields of observed config that don't belong to the config object.

PR #531, opened 2019-07-24
Bug 1732945: Report infra provider and feature set metrics

smarterclayton:
The infrastructure provider metric summarizes the desired type and
region (if any) to prometheus, based on the desired config. If type
is not recognized it is passed along, if it is empty it is set as
empty.

The featureset metric reports what featureset is in use and 1 if it
is "supported" and 0 if it is "not supported" (in the tech preview
sense).


smarterclayton:
@deads2k @sttts is verify-reps supposed to be broken?

jwforres:
#535 is open to fix the verify-deps issue you are seeing

That aside, I see two different PRs both attempting to backport this fix. This one, and #537 cc @p0lyn0mial 

I can't approve both of them. Let me know which one of these you actually want approved.

p0lyn0mial:
> I can't approve both of them. Let me know which one of these you actually want approved.

let's take this one.

tnozicka:
> @deads2k @sttts is verify-reps supposed to be broken?

can you pick commits from https://github.com/openshift/cluster-kube-apiserver-operator/pull/535 (@eparis didn't want it to go separately so we don't spin an image with no changes)

smarterclayton:
Picked #535 into this PR

smarterclayton:
Can I get approval and lgtm on this now that we're back to green?

PR #536, opened 2019-07-29
switch to generic client

deads2k:
Switches us onto the standard dynamic client for operator resources.


deads2k:
same as the others

PR #538, opened 2019-07-29
block upgrades on bad featuregates

deads2k:
I'm not excited about this.  I still think it belongs in the CVO, but every release it isn't present is one more release harder it is to enforce later on.

This adds a simple way to force the ugpradeable condition false on featuregate values.

### DIFF
--- a/pkg/operator/featureupgradablecontroller/feature_upgradeable_controller.go
+++ b/pkg/operator/featureupgradablecontroller/feature_upgradeable_controller.go
@@ -1,0 +1,139 @@
+package featureupgradablecontroller
+
+import (
+	"fmt"
+	"time"
+
+	"k8s.io/klog"
+
+	configv1 "github.com/openshift/api/config/v1"
+	operatorv1 "github.com/openshift/api/operator/v1"
+	configinformers "github.com/openshift/client-go/config/informers/externalversions"
+	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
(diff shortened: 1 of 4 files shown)


enj:
LGTM, certainly better than not enforcing feature gates.

derekwaynecarr:
there was an agreement that the cvo would handle this and it was to be treated as a release blocking bug.  see: https://bugzilla.redhat.com/show_bug.cgi?id=1730401

pkg/operator/featureupgradablecontroller/feature_upgradeable_controller.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	return nil
+}
+
+func newUpgradeableCondition(featureGates *configv1.FeatureGate) operatorv1.OperatorCondition {
+	if featureGatesAllowingUpgrade.Has(string(featureGates.Spec.FeatureSet)) {
+		return operatorv1.OperatorCondition{
+			Type:   "FeatureGatesUpgradeable",
+			Reason: "AllowedFeatureGates_" + string(featureGates.Spec.FeatureSet),
```
sttts: how will this look like as a string?

sttts:
I expected something more ugly. Sgtm.

pkg/operator/featureupgradablecontroller/feature_upgradeable_controller.go
deads2k: > how will this look like as a string?

A little weird, but not distractingly bad.

deads2k:
> I expected something more ugly. Sgtm.

Do this just in case the CVO has trouble getting it done and backported, but then we revert after?

sttts:
> Do this just in case the CVO has trouble getting it done and backported, but then we revert after?

Seems good. Create a blocker bugzilla to remind us.

PR #539, opened 2019-07-30
add cert rotate upgrade prevention

deads2k:
We named this "unsupported-cert-rotation-config".  This adds logic to give that some teeth.

/assign @enj


deads2k:
@enj now with working e2e.  just gofmt remaining.

enj:
LGTM minus formatting.  Feel free to tag when ready.

PR #540, opened 2019-07-30
block upgrades on techpreviewnoupgrade featuregates

deads2k:
cherrypick of https://github.com/openshift/cluster-kube-apiserver-operator/pull/538 to prevent upgrades on techpreviewnoupgrade featuregates.  The CVO will natively enforce this soon, but we want to prevent further upgrades that we can't support.


tnozicka:
rebased in https://github.com/openshift/cluster-kube-apiserver-operator/pull/548

eparis:
closing since we have 548

PR #541, opened 2019-07-31
Bug 1734622: bump(*)

enj:


tnozicka:
the BZ here is for 4.1 branch not master (4.2) - a) I'd be afraid this closes the BZ before the actual backport, b) QA is missing a BZ to verify it in 4.2

PR #545, opened 2019-08-05
Bug 1732945: Fallback to platform type on 4.1 infra metrics

smarterclayton:
4.1 clusters do not set infrastructure.status.platformStatus, causing the metric for cluster infrastructure to be empty. If platformstatus is nil, fallback to infrastructure type. A bug has been opened to ensure 4.2 clusters have platformStatus set during upgrade.

The previous PR #531 did not cover this difference and so reported an empty metric.

### DIFF
--- a/pkg/operator/configmetrics/configmetrics.go
+++ b/pkg/operator/configmetrics/configmetrics.go
@@ -59,6 +59,22 @@ func (m *configMetrics) Collect(ch chan<- prometheus.Metric) {
 			}
 			g.Set(value)
 			ch <- g
+		} else {
+			// if the parent has not reported platform status yet
+			infraType := infra.Status.Platform
+			var g prometheus.Gauge
+			var value float64 = 1
+			switch {
+			// it is illegal to set type to empty string, so let the default case handle
+			// empty string (so we can detect it) while preserving the constant None here
+			case infraType == configv1.NonePlatformType:
+				g = m.cloudProvider.WithLabelValues(string(infraType), "")
+				value = 0
+			default:
+				g = m.cloudProvider.WithLabelValues(string(infraType), "")
+			}
+			g.Set(value)
+			ch <- g
 		}
 	}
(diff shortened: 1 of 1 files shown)


smarterclayton:
Verified that https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-kube-apiserver-operator/545/pull-ci-openshift-cluster-kube-apiserver-operator-release-4.1-e2e-aws/27/artifacts/e2e-aws/metrics/prometheus.tar contains cluster_infrastructure_provider{type="AWS"}

This is ready for review / approval / cherrypick-approval

pkg/operator/configmetrics/configmetrics.go
```diff
@@ -59,6 +59,22 @@ func (m *configMetrics) Collect(ch chan<- prometheus.Metric) {
...
 			g.Set(value)
 			ch <- g
+		} else {
+			// if the parent has not reported platform status yet
+			infraType := infra.Status.Platform
+			var g prometheus.Gauge
+			var value float64 = 1
+			switch {
+			// it is illegal to set type to empty string, so let the default case handle
+			// empty string (so we can detect it) while preserving the constant None here
```
crawford: I'm confused by this. So if we find `None`, we set the value to `0`. Otherwise, it's set to `1` (including the empty string). Why bother setting the value to `0` at all? We can already distinguish between <empty> and `None`.

smarterclayton: The query we'd want to do most often is "how many non-cloud provider clusters do we have" and "how many cloud-provider clusters do we have" so 1 vs 0 is that check.  The error state is not the same as having no cloud provider, so I'd rather it be treated as a bug when doing the naive queries (the query is `count by (_id) (cloud_provider == 1)` or `count by (_id) (cloud_provider == 0)`.  I didn't want someone reading this to assume "" == "None" because it is not.

PR #546, opened 2019-08-06
fix `make images`

stlaz:
The internals of the `make images` now take four arguments:
- target name
- tag
- file
- context

/cc @sttts @tnozicka


stlaz:
same as https://github.com/openshift/cluster-kube-apiserver-operator/pull/551, closing

PR #551, opened 2019-08-14
updateMakefile to fix 'make images'

sallyom:
@deads2k 
this PR fixes 'make images' (this is only run locally AFAIK, but was broken nonetheless)
`make images` requires `image name` parameter

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,13 +9,15 @@ include $(addprefix ./vendor/github.com/openshift/library-go/alpha-build-machine
 	targets/openshift/images.mk \
 )
 
+IMAGE_REGISTRY :=registry.svc.ci.openshift.org
+
 # This will call a macro called "build-image" which will generate image specific targets based on the parameters:
 # $0 - macro name
-# $1 - target suffix
-# $2 - Dockerfile path
-# $3 - context directory for image build
-# It will generate target "image-$(1)" for builing the image an binding it as a prerequisite to target "images".
-$(call build-image,origin-cluster-kube-apiserver-operator,./Dockerfile,.)
+# $1 - target name
+# $2 - image ref
+# $3 - Dockerfile path
+# $4 - context directory for image build
+$(call build-image,ocp-cluster-kube-apiserver-operator,$(IMAGE_REGISTRY)/ocp/4.2:cluster-kube-apiserver-operator, ./Dockerfile.rhel7,.)
 
 # This will call a macro called "add-bindata" which will generate bindata specific targets based on the parameters:
 # $0 - macro name


manifests/image-references
```diff
@@ -5,7 +5,7 @@ spec:
   - name: cluster-kube-apiserver-operator
     from:
       kind: DockerImage
-      name: docker.io/openshift/origin-cluster-kube-apiserver-operator:v4.0
+      name: quay.io/openshift/origin-cluster-kube-apiserver-operator:v4.0
```
sallyom: @deads2k this change shouldn't break anything? service-ca-operator, cluster-authentication-operator reference quay.io

Makefile
```diff
@@ -12,10 +12,11 @@ include $(addprefix ./vendor/github.com/openshift/library-go/alpha-build-machine
...
 # $0 - macro name
 # $1 - target suffix
-# $2 - Dockerfile path
-# $3 - context directory for image build
+# $2 - image name
+# $3 - Dockerfile path
+# $4 - context directory for image build
 # It will generate target "image-$(1)" for builing the image an binding it as a prerequisite to target "images".
-$(call build-image,origin-cluster-kube-apiserver-operator,./Dockerfile,.)
+$(call build-image,origin-$(go_build_binaries),docker.io/openshift/origin-$(go_build_binaries),./Dockerfile,.)
```
sallyom: matching what's in manifests/image-references `docker.io` but shouldn't it be `quay.io`?

mfojtik: i would go with quay.io if that works

tnozicka: I'd go with `registry.svc.ci.openshift.org` here is the new updated structure https://github.com/openshift/oc/blob/55a3aad4d40cba59d5367f15b850c36bafba4ed1/Makefile#L45

sallyom: I've updated to match the oc pattern.  AFAICT nobody (ART) uses the make cmd to build images, but if we switch to Dockerfile here rather than Dockerfile.rhel7, devs _could_ use the make cmd to build a local image.   As/is, the Dockerfile.rhel7 I get this:
```console
$ make image-ocp-cluster-kube-apiserver-operator 
imagebuilder --allow-pull -t registry.svc.ci.openshift.org/ocp/4.2:cluster-kube-apiserver-operator -f ./Dockerfile.rhel7 .
--> Image registry.svc.ci.openshift.org/ocp/builder:golang-1.12 was not found, pulling ...
unable to pull image (from: registry.svc.ci.openshift.org/ocp/builder, tag: golang-1.12): unauthorized: authentication required
make: *** [Makefile:20: image-ocp-cluster-kube-apiserver-operator] Error 1
```
I think most devs don't have proper access for:
 `registry.svc.ci.openshift.org/ocp/builder:golang-1.12 AS builder` and `registry.svc.ci.openshift.org/ocp/4.0:base` 
while we _can_ access:
`registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder` and `registry.svc.ci.openshift.org/openshift/origin-v4.0:base`

PR #552, opened 2019-08-19
Inject kube-apiserver pods trust stores with trusted ca bundle

stlaz:
Adds the trusted store injection and a watchdog that will restart
the kube-apiserver process on the trusted CA bundle change.

### DIFF
--- a/bindata/v4.1.0/kube-apiserver/pod.yaml
+++ b/bindata/v4.1.0/kube-apiserver/pod.yaml
@@ -31,9 +31,14 @@ spec:
     image: ${IMAGE}
     imagePullPolicy: IfNotPresent
     terminationMessagePolicy: FallbackToLogsOnError
-    command: ["hyperkube", "kube-apiserver"]
+    command: ["/bin/bash", "-ec"]
     args:
-    - --openshift-config=/etc/kubernetes/static-pod-resources/configmaps/config/config.yaml
+        - |
+          if [ -f /etc/kubernetes/static-pod-resources/kube-apiserver-certs/configmaps/trusted-ca-bundle/tls-ca-bundle.pem ]; then
+            echo "Copying system trust bundle"
+            cp -f /etc/kubernetes/static-pod-resources/kube-apiserver-certs/configmaps/trusted-ca-bundle/tls-ca-bundle.pem /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
+          fi
+          exec hyperkube kube-apiserver --openshift-config=/etc/kubernetes/static-pod-resources/configmaps/config/config.yaml
     resources:
       requests:
         memory: 1Gi
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -204,6 +204,9 @@ var RevisionSecrets = []revision.RevisionResource{
 var CertConfigMaps = []revision.RevisionResource{
 	{Name: "aggregator-client-ca"},
 	{Name: "client-ca"},
+
+	// this is a copy of trusted-ca-bundle CM but with key modified to "tls-ca-bundle.pem" so that we can mount it the way we need
+	{Name: "trusted-ca-bundle", Optional: true},
 }
 
 var CertSecrets = []revision.RevisionResource{
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -12,6 +12,8 @@ import (
 	"k8s.io/klog"
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -194,6 +196,11 @@ func createTargetConfig(c TargetConfigController, recorder events.Recorder, oper
 		errors = append(errors, fmt.Errorf("%q: %v", "configmap/kube-apiserver-server-ca", err))
 	}
 
+	err = ensureKubeAPIServerTrustedCA(c.kubeClient.CoreV1(), recorder)
+	if err != nil {
+		errors = append(errors, fmt.Errorf("%q: %v", "configmap/trusted-ca-bundle", err))
+	}
+
 	if len(errors) > 0 {
 		condition := operatorv1.OperatorCondition{
 			Type:    "TargetConfigControllerDegraded",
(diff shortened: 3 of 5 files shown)


stlaz:
In order for this to work correctly, we'll have to add the injected cm somewhere into https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/pkg/operator/starter.go#L180.

We'll need to get a CM with the proper key there somewhere, possibly adding a CM apply from the targetconfigcontroller that'll read it from the CM that's being injected :thinking:

bindata/v4.1.0/kube-apiserver/pod.yaml
```diff
@@ -113,6 +115,39 @@ spec:
       requests:
         memory: 50Mi
         cpu: 10m
+  - name: kube-apiserver-watchdog-REVISION
+    command: ["cluster-kube-apiserver-operator", "file-watcher-watchdog"]
```
mfojtik: you need `shareProcessNamespace: true` in spec (https://github.com/openshift/cluster-kube-controller-manager-operator/pull/254/files#diff-ad2ab26372ca50cd65844a937511d0a4R11)

mfojtik:
This need e2e test.

bindata/v4.1.0/kube-apiserver/pod.yaml
stlaz: Yeah, I just noticed, thanks though :+1:

mfojtik:
You can't wire file watchdog to kube api server pod, because when the configmap change, the watchdog will deliberately kill all 3 api servers immediately (to restart them).

I think we need to roll new revision when the config map change. To do that you need to add the configmap with CA into `RevisionConfigMaps` and then create optional volume mount in kube apiserver pod.

@sttts @deads2k correct?

stlaz:
@mfojtik: seeing the issue I was mentioning earlier that prevents kube-apiserver pods from running:
```
'MountVolume.SetUp failed for volume "trusted-ca-bundle" : configmaps "trusted-ca-bundle-2"
    is forbidden: User "system:node:ip-10-0-149-86.ec2.internal" cannot get resource
    "configmaps" in API group "" in the namespace "openshift-kube-apiserver": no relationship
    found between node "ip-10-0-149-86.ec2.internal" and this object'
```

bindata/v4.1.0/kube-apiserver/pod.yaml
```diff
@@ -128,3 +130,6 @@ spec:
   - hostPath:
       path: /var/log/kube-apiserver
     name: audit-dir
+  - hostPath:
+      path: /etc/kubernetes/static-pod-resources/configmaps/trusted-ca-bundle-REVISION
```
mfojtik: why this need to be revisioned?

pkg/operator/starter.go
```diff
@@ -190,6 +190,9 @@ var RevisionConfigMaps = []revision.RevisionResource{
 	{Name: "kube-apiserver-server-ca", Optional: true},
 	{Name: "kubelet-serving-ca"},
 	{Name: "sa-token-signing-certs"},
+
+	// this is a copy of trusted-ca-bundle CM but with key modified to "tls-ca-bundle.pem" so that we can mount it the way we need
+	{Name: "trusted-ca-bundle"},
```
mfojtik: I think you need this in `CertConfigMaps` /cc @deads2k

bindata/v4.1.0/kube-apiserver/pod.yaml
```diff
@@ -47,6 +47,8 @@ spec:
       name: cert-dir
     - mountPath: /var/log/kube-apiserver
       name: audit-dir
+    - mountPath: /etc/pki/ca-trust/extracted/pem/
```
mfojtik: is this overriding existing directory? in case there is no trusted-ca-bundle config map, we erase that dir?

mfojtik: (i wonder if this should be handled in the installer pod controller)

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -343,6 +348,28 @@ func manageKubeAPIServerCABundle(lister corev1listers.ConfigMapLister, client co
 	return resourceapply.ApplyConfigMap(client, recorder, requiredConfigMap)
 }
 
+func syncTrustedCABundle(lister corev1listers.ConfigMapLister, client coreclientv1.ConfigMapsGetter, recorder events.Recorder) (bool, error) {
```
mfojtik: where do you use the `bool` return value?

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -343,6 +348,28 @@ func manageKubeAPIServerCABundle(lister corev1listers.ConfigMapLister, client co
...
+	trustedCACM, err := lister.ConfigMaps(operatorclient.OperatorNamespace).Get("trusted-ca-bundle")
+	if err != nil {
+		return false, err
+	}
+
+	requiredConfigMap := resourceread.ReadConfigMapV1OrDie(v410_00_assets.MustAsset("v4.1.0/kube-apiserver/trustedcacm.yaml"))
+
+	caData, ok := trustedCACM.Data["ca-bundle.crt"]
```
mfojtik: do we care about `ok && len() >0` ?

bindata/v4.1.0/kube-apiserver/pod.yaml
stlaz: That's a good point, I think the hostPath directory will only appear once the appropriate CM exists. I guess setting `type: Directory` for the matching `volume` might solve the problem of the `hostPath` directory not existing.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
stlaz: I think we should, I'll add that.

bindata/v4.1.0/kube-apiserver/pod.yaml
stlaz: I don't think it has to be, it only felt consistent with the `kube-apiserver-pod` resource few lines above

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
stlaz: I was originally merely following the pattern of the `manage*` function, I'll remove that as it does not feel necessary

stlaz:
/test e2e-aws
Looks like installer bug:
```
level=error msg="Error: Provider produced inconsistent result after apply"
level=error
level=error msg="When applying changes to module.bootstrap.aws_s3_bucket.ignition, provider"
level=error msg="\"aws\" produced an unexpected new value for was present, but now absent."
level=error
level=error msg="This is a bug in the provider, which should be reported in the provider's own"
level=error msg="issue tracker."
```
@wking

wking:
Shared [upstream][1].

[1]: https://github.com/terraform-providers/terraform-provider-aws/issues/9725#issuecomment-523904054

stlaz:
/hold
I was suggested to use bash wrapper to wait for the file mount location to exist and then just copy to proper path, probably a better solution

stlaz:
/hold cancel
actually, now that I think of it, that bash hack probably won't work here because there would be noone to inject the CM to the hostpath

stlaz:
The dir check in hostPath seemed to have failed, I wonder if removing revision from the path will help

bindata/v4.1.0/kube-apiserver/pod.yaml
danehans: @stlaz the mount should be optional.

stlaz:
Pods are still getting stuck in unitialized on the mount, it would appear the directory just does not appear...

pkg/operator/starter.go
stlaz: I guess you're right, in the logs, I can see that the dirs created are:
```
"/etc/kubernetes/static-pod-resources/kube-apiserver-certs/configmaps/client-ca"
```
but on the other hand also
```
"/etc/kubernetes/static-pod-resources/kube-apiserver-pod-2/configmaps/trusted-ca-bundle"
```
which would explain why the `Directory` hostMount assumption fails, I'll move this to the `CertConfigMaps` then.

bindata/v4.1.0/kube-apiserver/pod.yaml
stlaz: @danehans I don't think there is a way how that could work here

stlaz: Nvm, we found a way around it to make it optional

stlaz:
/retest
The failures don’t seem to be related to this PR

bindata/v4.1.0/kube-apiserver/pod.yaml
sttts: These files are in a revisioned directory. But due to the uniform mount point, that's not visible here.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -343,6 +350,28 @@ func manageKubeAPIServerCABundle(lister corev1listers.ConfigMapLister, client co
 	return resourceapply.ApplyConfigMap(client, recorder, requiredConfigMap)
 }
 
+func ensureKubeAPIServerTrustedCA(client coreclientv1.CoreV1Interface, recorder events.Recorder) error {
```
mfojtik: I do believe these should have comments describing their purpose.

mfojtik:
/lgtm

I don't like the bash in the pod spec, but for can't think of simpler solution for now (we can improve this in future).

PR #557, opened 2019-08-26
Bug 1714158: Prevent cert-syncer to act on stale data

tnozicka:
Requires:
 - [x] https://github.com/openshift/library-go/pull/512

picking up a backport of https://github.com/openshift/cluster-kube-apiserver-operator/pull/520/commits/d0f4b034e2998eb76530e298febbf5375c7df757 from https://github.com/openshift/cluster-kube-apiserver-operator/pull/520 to fix breaking api change in https://github.com/openshift/api/pull/375 we got with the bump

/cc @sttts


vendor/github.com/openshift/library-go/pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -79,18 +85,69 @@ func (c *CertSyncController) sync() error {
 		case apierrors.IsNotFound(err) && !cm.Optional:
 			errors = append(errors, err)
 			continue
+
 		case apierrors.IsNotFound(err) && cm.Optional:
+			// Check with the live call it is really missing
```
tnozicka: here it is

pkg/operator/configobservation/apiserver/observe_apiserver.go
```diff
@@ -62,17 +62,6 @@ var ObserveUserClientCABundle configobserver.ObserveConfigFunc = (&apiServerObse
 	resourceType:  corev1.ConfigMap{},
 }).observe
 
-// ObserveDefaultUserServingCertificate returns an ObserveConfigFunc that observes user managed TLS cert info for
-// serving secure traffic.
-// We are disabling this because it doesn't work today and customers aren't going to be able to get the kube service network options right.
-// Customers may only use SNI.  I'm leaving this code in case we ever come up with a way to make an SNI-like thing based on IPs.
-var ObserveDefaultUserServingCertificate configobserver.ObserveConfigFunc = (&apiServerObserver{
```
sttts: why is this commit good? removal without explanation in the commit msg is not good.

tnozicka: this is bump of openshift/api not something this PR introduced

sttts: but it changes pkg/operator/configobservation/apiserver/observe_apiserver.go, not vendor/openshift/api.

tnozicka: this was a backport of breaking api change which just disabled the code, but left it there. this PR won the first bump and dealing with incompatibilities

mfojtik:
The 4.2 is in the right state with right priority and the fix in master soaked, approving.

PR #559, opened 2019-08-29
only use the internalIP to speak to kubelets

deads2k:
This allows the NO_PROXY setting to only list internal node CIDRs which reduces churn as nodes are added and removed.

I'm not completely sure of all the downstream impacts from this choice.


sjenning:
i don't see a complication as long as the kubelet serving cert is valid for that IP (e2e would have failed if it was not)

danehans:
@mfojtik @sttts PTAL when you have a moment. This PR can greatly simplify how we manage no_proxy across the the supported install types.

PR #564, opened 2019-09-12
Bug 1751841: null externalip policy should be default-deny

squeed:
Previously, a null externalip policy would allow all external IPs. We shouldn't actually do this. Null policy should mean blocked by default.


squeed:
hey @danwinship could you take a look?

sttts:
/hold

for @danwinship

danwinship:
lgtm
/hold cancel

PR #565, opened 2019-09-12
Adding UnsupportedOverrides test for BZ 1748501

bradmwilliams:


deads2k:
> As much as I like more test coverage, the contents of `unsupportedOverrides` is called unsupported for a reason: it can change, break, go away any time.

I agree with sentiment.  In this particular case, it's more about knowing that we broke it and informed @bradmwilliams, not about him getting to pin the behavior.  I've already indicated that pieces of this config may simply disappear, but we'll know that we're doing it to him.

sttts:
> In this particular case, it's more about knowing that we broke it and informed @bradmwilliams, not about him getting to pin the behavior.

Ok, if the expectation is clear, I am fine with this. Maybe add a bold comment on-top of the test that is to notice breakage. In general dedicated will have to watch the internal config API type changes of releases and prepare early for incompatibilities.

bradmwilliams:
I have added the following comment to the top of the test:
```java
/*
                                BREAKAGE NOTIFICATION TEST
  The purpose of this test is provde a general notification if and when UnsupportedConfigOverrides
  either become broken and/or are removed all together.  The OpenShift Starter environments utilize
  these configurations and we'd like to be notified when this tests fails.  The following email
  address can be used to alert us of any issues if and when they arrive:
                        openshift-cr@redhat.com

  THIS TEST CAN SAFELY BE SKIPPED ONCE A NOTIFICATION HAS BEEN SENT!
*/
```

test/e2e/unsupported_overrides_test.go
```diff
@@ -0,0 +1,120 @@
+/*
+				BREAKAGE NOTIFICATION TEST
+  The purpose of this test is provde a general notification if and when UnsupportedConfigOverrides
+  either become broken and/or are removed all together.  The OpenShift Starter environments utilize
+  these configurations and we'd like to be notified when this tests fails.  The following email
+  address can be used to alert us of any issues if and when they arrive:
+			openshift-cr@redhat.com
+
+  THIS TEST CAN SAFELY BE SKIPPED ONCE A NOTIFICATION HAS BEEN SENT!
+*/
```
sttts: would rather move that above the test. Here at the top of the file everybody misses it.

bradmwilliams: Done!

PR #571, opened 2019-09-23
Add a controller to check pod creation errors

stlaz:
This controller is supposed to help us easily track errors that
appear during kube-apiserver pods' bootstrapping simply by
looking at the operator's status.

cc @mfojtik @sttts @deads2k


stlaz:
superseeded by https://github.com/openshift/library-go/pull/550

PR #572, opened 2019-09-23
DO NOT MERGE - Test with new feature gates on by default

smarterclayton:
These three gates remove the linkage between node-role master and
cluster behavior and allow us to run 3 node clusters on cloud providers
out of the box.

Testing here before opening openshift/api PR


smarterclayton:
The impact of these is minimal for our configurations and we plan to enable in 1.17 upstream by default.  However, in terms of our ability to test metal and edge workloads on cloud (colocated) this is pretty important).

PR #580, opened 2019-10-01
encryption: split shared code

sttts:
Pure file split, no functional or symbol changes.

Follow-up will be to move all of these in packages. But we wait for the controllers to merge to avoid package move pain.


pkg/operator/encryption/controllers.go
```diff
@@ -0,0 +1,71 @@
+package encryption
+
+import (
+	"k8s.io/client-go/tools/cache"
+
+	"github.com/openshift/library-go/pkg/operator/management"
+	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
```
mfojtik: nit: move these two down

pkg/operator/encryption/crypt.go
```diff
@@ -0,0 +1,26 @@
...
+		panic(err) // rand should never fail
+	}
+	return b
+}
+
+func newIdentityKey() []byte {
+	return make([]byte, 16) // the key is not used to perform encryption but must be a valid AES key
+}
+
+var emptyStaticIdentityKey = base64.StdEncoding.EncodeToString(newIdentityKey())
```
mfojtik: nit: move this up?

pkg/operator/encryption/secrets.go
```diff
@@ -0,0 +1,211 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	apiserverconfigv1 "k8s.io/apiserver/pkg/apis/config/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/klog"
+
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/operatorclient"
+)
+
+func findSecretForKey(key keyAndMode, secrets []*corev1.Secret, targetNamespace string) *corev1.Secret {
+	if key == (keyAndMode{}) {
```
mfojtik: hrm, why the () ?

sttts: we discussed this before. Some Golang parser issue. It does not understand that the `{}` belong to the type.

PR #582, opened 2019-10-03
Address some TODOs for getDesiredEncryptionState

enj:
1. Error when referenced encryption secrets get deleted
2. Preserve write keys even if the revisioned config is deleted

@mfojtik @sttts @p0lyn0mial

### DIFF
--- a/pkg/operator/encryption/state.go
+++ b/pkg/operator/encryption/state.go
@@ -52,7 +52,7 @@ func getCurrentEncryptionConfig(secrets corev1client.SecretInterface, revision s
 //   3. Once every resource honors the same write key AND that write key has migrated every requested resource,
 //      all non-write keys should be removed.
 // TODO unit tests
-func getDesiredEncryptionState(encryptionConfig *apiserverconfigv1.EncryptionConfiguration, targetNamespace string, encryptionSecretList *corev1.SecretList, encryptedGRs map[schema.GroupResource]bool) groupResourcesState {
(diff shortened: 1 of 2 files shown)


pkg/operator/encryption/state.go
```diff
@@ -74,16 +74,22 @@ func getDesiredEncryptionState(encryptionConfig *apiserverconfigv1.EncryptionCon
 	// this is our output from the for loop below
 	encryptionState := groupResourcesState{}
 
-	// TODO if we cannot find the secrets associated with the keys, they may have been deleted
-	// we could try to use the key data in the revision to recover or just go degraded permanently
 	for gr, keys := range getGRsActualKeys(encryptionConfig) {
-		writeSecret := findSecretForKey(keys.writeKey, encryptionSecrets, targetNamespace) // TODO handle nil as error when hasWriteKey == true?
+		writeSecret := findSecretForKey(keys.writeKey, encryptionSecrets, targetNamespace)
```
p0lyn0mial: wouldn't returning an error when a key hasn't been found make more sense?

enj: The caller has to determine if not finding a key is an error though.  At most I would expect `findSecretForKey` to return a `bool` to indicate if it found it, but the `nil` check seems enough.

enj:
@mfojtik @sttts @p0lyn0mial you probably care about this sooner rather than later.

/close

PR #585, opened 2019-10-07
encryption: add migration controller

sttts:


pkg/operator/encryption/migration_controller.go
```diff
@@ -0,0 +1,364 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"reflect"
+	"strings"
+	"time"
+
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
```
p0lyn0mial: nit: group imports

PR #588, opened 2019-10-08
encryption: add prune controller

sttts:
This include a getDesiredEncryptionState cleanup with:

- move encryption config into the focus, secrets only give additional info about migration
- by that handle missing secrets (don't panic as before)
- fix bug that all existing keys are added as read keys
- in STEP 2, only add missing read-keys, never drop. We only drop in the last step after migration has finished. This makes us resilient to secret deletion. Conversely, when the config is deleted and we still have the secrets (and the resource list did not change, during upgrade it is a bad time to delete the config), we can reconstruct the config.
- added getDesiredEncryptionState unit tests
- make getDesiredEncryptionState handle nil configs correctly
- add state controller

### DIFF
--- a/pkg/operator/encryption/prune_controller.go
+++ b/pkg/operator/encryption/prune_controller.go
@@ -106,13 +106,12 @@ func (c *pruneController) sync() error {
 }
 
 func (c *pruneController) deleteOldMigratedSecrets() error {
-	_, desiredEncryptionConfig, _, err := getEncryptionConfigAndState(c.podClient, c.secretClient, c.targetNamespace, c.encryptionSecretSelector, c.encryptedGRs)
+	_, desiredEncryptionConfig, isProgressingReason, err := getEncryptionConfigAndState(c.podClient, c.secretClient, c.targetNamespace, c.encryptionSecretSelector, c.encryptedGRs)
 	if err != nil {
 		return err
 	}
-	// do not try to delete anything during a transition state
-	if len(desiredEncryptionConfig) == 0 {
-		// we are woken up when the pods or secrets change
+	if len(isProgressingReason) > 0 {
+		c.queue.AddAfter(migrationWorkKey, 2*time.Minute)
 		return nil
 	}
 
--- a/pkg/operator/encryption/prune_controller.go
+++ b/pkg/operator/encryption/prune_controller.go
@@ -1,0 +1,227 @@
+package encryption
+
+import (
+	"fmt"
+	"sort"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/operatorclient"
+	"github.com/openshift/library-go/pkg/operator/events"
+	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+const (
+	pruneWorkKey        = "key"
+	keepNumberOfSecrets = 10
+)
+
+// pruneController prevents an unbounded growth of old encryption keys.
+// For a given resource, if there are more than ten keys which have been migrated,
+// this controller will delete the oldest migrated keys until there are ten migrated
+// keys total.  These keys are safe to delete since no data in etcd is encrypted using
+// them.  Keeping a small number of old keys around is meant to help facilitate
+// decryption of old backups (and general precaution).
+type pruneController struct {
+	operatorClient operatorv1helpers.StaticPodOperatorClient
+
+	queue         workqueue.RateLimitingInterface
+	eventRecorder events.Recorder
+
+	preRunCachesSynced []cache.InformerSynced
+
+	encryptedGRs map[schema.GroupResource]bool
+
+	targetNamespace          string
+	encryptionSecretSelector metav1.ListOptions
+
+	podClient    corev1client.PodsGetter
+	secretClient corev1client.SecretsGetter
+}
+
+func newPruneController(
+	targetNamespace string,
+	operatorClient operatorv1helpers.StaticPodOperatorClient,
+	kubeInformersForNamespaces operatorv1helpers.KubeInformersForNamespaces,
+	podClient corev1client.PodsGetter,
+	secretClient corev1client.SecretsGetter,
+	encryptionSecretSelector metav1.ListOptions,
+	eventRecorder events.Recorder,
+	encryptedGRs map[schema.GroupResource]bool,
+) *pruneController {
+	c := &pruneController{
+		operatorClient: operatorClient,
+
+		queue:         workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "EncryptionPruneController"),
+		eventRecorder: eventRecorder.WithComponentSuffix("encryption-prune-controller"), // TODO unused
+
+		encryptedGRs:    encryptedGRs,
+		targetNamespace: targetNamespace,
+
+		encryptionSecretSelector: encryptionSecretSelector,
+		podClient:                podClient,
+		secretClient:             secretClient,
+	}
+
+	c.preRunCachesSynced = setUpAllEncryptionInformers(operatorClient, targetNamespace, kubeInformersForNamespaces, c.eventHandler())
+
+	return c
+}
+
+func (c *pruneController) sync() error {
+	if ready, err := shouldRunEncryptionController(c.operatorClient); err != nil || !ready {
+		return err // we will get re-kicked when the operator status updates
+	}
+
+	configError := c.deleteOldMigratedSecrets()
+
+	// update failing condition
+	cond := operatorv1.OperatorCondition{
+		Type:   "EncryptionPruneControllerDegraded",
+		Status: operatorv1.ConditionFalse,
+	}
+	if configError != nil {
+		cond.Status = operatorv1.ConditionTrue
+		cond.Reason = "Error"
+		cond.Message = configError.Error()
+	}
+	if _, _, updateError := operatorv1helpers.UpdateStaticPodStatus(c.operatorClient, operatorv1helpers.UpdateStaticPodConditionFn(cond)); updateError != nil {
+		return updateError
+	}
+
+	return configError
+}
+
+func (c *pruneController) deleteOldMigratedSecrets() error {
+	_, desiredEncryptionConfig, _, err := getEncryptionConfigAndState(c.podClient, c.secretClient, c.targetNamespace, c.encryptionSecretSelector, c.encryptedGRs)
+	if err != nil {
+		return err
+	}
+	// do not try to delete anything during a transition state
+	if len(desiredEncryptionConfig) == 0 {
+		// we are woken up when the pods or secrets change
+		return nil
+	}
+
+	usedSecrets := make([]*corev1.Secret, 0, len(desiredEncryptionConfig))
+	for _, grKeys := range desiredEncryptionConfig {
+		usedSecrets = append(usedSecrets, grKeys.readSecrets...)
+	}
+
+	allkeys, err := c.secretClient.Secrets(operatorclient.GlobalMachineSpecifiedConfigNamespace).List(c.encryptionSecretSelector)
+	if err != nil {
+		return err
+	}
+
+	// sort by keyID
+	encryptionSecrets := make([]*corev1.Secret, 0, len(allkeys.Items))
+	for _, s := range allkeys.Items {
+		encryptionSecrets = append(encryptionSecrets, s.DeepCopy()) // don't use &s because it constant through-out the loop
+	}
+	sort.Slice(encryptionSecrets, func(i, j int) bool {
+		// it is fine to ignore the validKeyID bool here because we filtered out invalid secrets in the loop above
+		iKeyID, _ := secretToKeyID(encryptionSecrets[i])
+		jKeyID, _ := secretToKeyID(encryptionSecrets[j])
+		return iKeyID > jKeyID
+	})
+
+	var deleteErrs []error
+	skippedKeys := 0
+	for _, s := range encryptionSecrets {
+		if hasSecret(usedSecrets, *s) {
+			continue
+		}
+
+		// skip the most recent unused secrets around
+		if skippedKeys < keepNumberOfSecrets {
+			skippedKeys++
+			continue
+		}
+
+		// any secret that isn't a read key isn't used.  just delete them.
+		// two phase delete: finalizer, then delete
+
+		// remove our finalizer if it is present
+		secret := s.DeepCopy()
+		if finalizers := sets.NewString(secret.Finalizers...); finalizers.Has(encryptionSecretFinalizer) {
+			delete(finalizers, encryptionSecretFinalizer)
+			secret.Finalizers = finalizers.List()
+			var updateErr error
+			secret, updateErr = c.secretClient.Secrets(operatorclient.GlobalMachineSpecifiedConfigNamespace).Update(secret)
+			deleteErrs = append(deleteErrs, updateErr)
+			if updateErr != nil {
+				continue
+			}
+		}
+
+		// remove the actual secret
+		if err := c.secretClient.Secrets(operatorclient.GlobalMachineSpecifiedConfigNamespace).Delete(secret.Name, nil); err != nil {
+			deleteErrs = append(deleteErrs, err)
+		} else {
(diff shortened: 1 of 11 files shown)


pkg/operator/encryption/state.go
```diff
@@ -43,175 +45,241 @@ func getCurrentEncryptionConfig(secrets corev1client.SecretInterface, revision s
...
 
+	//
+	// STEP 2: verify to have all necessary read-keys. If not, deploy and wait for stability.
+	//
+	// Note: only keysWithPotentiallyPersistedData are considered. There might be more which are not pruned yet by the pruning controller.
+	//
+	// TODO: allow removing resources (e.g. on downgrades) and transition back to identity.
+	// TODO: if the operand namespace gets deleted, this code causes us to go through a cycle of identity as the write key
 	allReadSecretsAsExpected := true
+	currentlyEncryptedGRs := make([]schema.GroupResource, 0, len(desiredEncryptionState))
```
p0lyn0mial: `currentlyEncryptedGRs` could be calculated from `desiredEncryptionState` before adding `toBeEncryptedGRs`

pkg/operator/encryption/state.go
```diff
@@ -43,175 +45,241 @@ func getCurrentEncryptionConfig(secrets corev1client.SecretInterface, revision s
...
+	for gr := range getGRsActualKeys(oldEncryptionConfig) {
+		currentlyEncryptedGRs = append(currentlyEncryptedGRs, gr)
+	}
+	expectedReadSecrets := keysWithPotentiallyPersistedData(currentlyEncryptedGRs, encryptionSecrets)
```
p0lyn0mial: okay, so this is an improvement, we take all the keys up that haven't been used for migration and additionally the one that has been used.

pkg/operator/encryption/state.go
```diff
@@ -43,175 +45,241 @@ func getCurrentEncryptionConfig(secrets corev1client.SecretInterface, revision s
...
-	// TODO if the operand namespace gets deleted, this code causes us to go through a cycle of identity as the write key
 	if !allReadSecretsAsExpected {
 		klog.V(4).Infof("not all read secrets in sync")
-		for gr := range encryptionState {
-			grState := encryptionState[gr]
-			grState.readSecrets = encryptionSecrets
-			encryptionState[gr] = grState
+		for gr := range desiredEncryptionState {
+			grState := desiredEncryptionState[gr]
+			// TODO: here we might lose read key if the "wrong" secrets are deleted
```
p0lyn0mial: what do you mean by that? The secrets are protected by a finalizer.

pkg/operator/encryption/state.go
```diff
@@ -43,175 +45,241 @@ func getCurrentEncryptionConfig(secrets corev1client.SecretInterface, revision s
...
 	}
 
-	// at this point all of our read and write secrets are the same.  We need to inspect the write secret to
-	// ensure that it has all the expect resources listed as having been migrated.  If this is true, then we
-	// can prune the read secrets down to a single entry.
+	//
+	// STEP 4: with consistent read-keys and write-keys, remove every read-key other than the write-key.
+	//
+	// Note: because read-keys are consistent, currentlyEncryptedGRs equals toBeEncryptedGRs
+	allMigrated, _, reason := migratedFor(currentlyEncryptedGRs, writeSecret)
```
p0lyn0mial: is this needed? I think that `expectedReadSecrets` will hold only `writeSecret` when read-keys are consistent.

pkg/operator/encryption/state.go
```diff
@@ -43,175 +45,241 @@ func getCurrentEncryptionConfig(secrets corev1client.SecretInterface, revision s
...
-			foundCount++
+	if len(missing) > 0 {
+		return false, missing, fmt.Sprintf("secret(s) %s misses resource %s among migrated resources", s.Name, strings.Join(missingStrings, ","))
+	}
+
+	return true, nil, ""
+}
+
+// keysWithPotentiallyPersistedData returns the minimal, recent secrets which have migrated all given GRs.
+func keysWithPotentiallyPersistedData(grs []schema.GroupResource, recentFirstSortedSecrets []*corev1.Secret) []*corev1.Secret {
```
p0lyn0mial: so this function actually`prunes/trims/removes/filters` keys that are no longer needed.

pkg/operator/encryption/state.go
```diff
@@ -43,175 +45,241 @@ func getCurrentEncryptionConfig(secrets corev1client.SecretInterface, revision s
...
-		klog.V(4).Infof("write secret %s is missing migrated resources expected=%s, actual=%s", writeSecret.Name, allResources, migratedResources.Resources)
-		return encryptionState
+	return recentFirstSortedSecrets
+}
+
+func encryptionConfigToEncryptionState(targetNamespace string, c *apiserverconfigv1.EncryptionConfiguration) groupResourcesState {
+	// convert old config to the base of desired state
+	// TODO: fix case with missing secrets (now panics below in hasSecrets)
+	s := groupResourcesState{}
+	currentlyEncryptedGRs := make([]schema.GroupResource, 0, len(s))
```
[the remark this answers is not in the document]
sttts: yes. Fixed in follow-up

pkg/operator/encryption/state.go
```diff
@@ -43,175 +45,241 @@ func getCurrentEncryptionConfig(secrets corev1client.SecretInterface, revision s
...
-	// if we did not find migration indications for all resources, then just wait until we do
-	if foundCount != len(allResources) {
-		klog.V(4).Infof("write secret %s is missing migrated resources expected=%s, actual=%s", writeSecret.Name, allResources, migratedResources.Resources)
-		return encryptionState
+	return recentFirstSortedSecrets
+}
+
+func encryptionConfigToEncryptionState(targetNamespace string, c *apiserverconfigv1.EncryptionConfiguration) groupResourcesState {
+	// convert old config to the base of desired state
+	// TODO: fix case with missing secrets (now panics below in hasSecrets)
```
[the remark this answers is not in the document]
sttts: no, removed in follow-up commit

pkg/operator/encryption/state.go
```diff
@@ -43,175 +45,241 @@ func getCurrentEncryptionConfig(secrets corev1client.SecretInterface, revision s
...
-//   2. Once every resource honors every key, the write key should be the latest key available
-//   3. Once every resource honors the same write key AND that write key has migrated every requested resource,
-//      all non-write keys should be removed.
+//
+// 1. every GR must honor read-key with potentially written data.
+// 2. if (1) is the case, the write-key must be the most recent key.
+// 3. if (1) and (2) are the case, all non-write keys should be removed.
 // TODO unit tests
-func getDesiredEncryptionState(encryptionConfig *apiserverconfigv1.EncryptionConfiguration, targetNamespace string, encryptionSecretList *corev1.SecretList, encryptedGRs map[schema.GroupResource]bool) groupResourcesState {
+func getDesiredEncryptionState(oldEncryptionConfig *apiserverconfigv1.EncryptionConfiguration, targetNamespace string, encryptionSecretList *corev1.SecretList, toBeEncryptedGRs map[schema.GroupResource]bool) groupResourcesState {
```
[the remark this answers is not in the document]
sttts: done already in follow-up :)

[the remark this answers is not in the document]
sttts: I don't like "actual"

pkg/operator/encryption/state.go
```diff
@@ -43,175 +45,241 @@ func getCurrentEncryptionConfig(secrets corev1client.SecretInterface, revision s
...
 	}
+	encryptionSecrets = sortRecentFirst(encryptionSecrets)
 
-	// see if any resource is missing and properly reflect that state so that
-	// we will add the read keys and take it through the transition correctly.
-	for gr := range encryptedGRs {
-		if _, ok := encryptionState[gr]; !ok {
-			encryptionState[gr] = keysState{
+	// add new resources without keys. These resources will trigger STEP 2.
+	for gr := range toBeEncryptedGRs {
```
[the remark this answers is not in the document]
sttts: see follow-up commit

pkg/operator/encryption/state.go
sttts: could, but why?

pkg/operator/encryption/state.go
sttts: fixed in last commit

pkg/operator/encryption/state.go
sttts: but the write key could be set without having finished the migration.

pkg/operator/encryption/state.go
sttts: yes, it takes the prefix of keys (starting with most recent) until it finds one which is migration key

pkg/operator/encryption/prune_controller.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+	if _, _, updateError := operatorv1helpers.UpdateStaticPodStatus(c.operatorClient, operatorv1helpers.UpdateStaticPodConditionFn(cond)); updateError != nil {
+		return updateError
+	}
+
+	return configError
+}
+
+func (c *pruneController) deleteOldMigratedSecrets() error {
+	_, encryptionState, _, err := getEncryptionConfigAndState(c.podClient, c.secretClient, c.targetNamespace, c.encryptionSecretSelector, c.encryptedGRs)
```
p0lyn0mial: `encryptionState` represents the desired encryption config, not the actual one, wouldn't it be better to examine the actual encryption config? (the first returned arg)

pkg/operator/encryption/prune_controller.go
```diff
@@ -0,0 +1,226 @@
...
+	if err != nil {
+		return err
+	}
+	// do not try to delete anything during a transition state
+	if len(encryptionState) == 0 {
+		c.queue.AddAfter(pruneWorkKey, 2*time.Minute)
```
p0lyn0mial: does it make sense? If `encryptionState == 0` there are no secrets and this controller will be notified anyway as soon as a secret will be added.

pkg/operator/encryption/prune_controller.go
```diff
@@ -0,0 +1,226 @@
...
+	sort.Slice(encryptionSecrets, func(i, j int) bool {
+		// it is fine to ignore the validKeyID bool here because we filtered out invalid secrets in the loop above
+		iKeyID, _ := secretToKeyID(encryptionSecrets[i])
+		jKeyID, _ := secretToKeyID(encryptionSecrets[j])
+		return iKeyID > jKeyID
+	})
+
+	var deleteErrs []error
+	skippedKeys := 0
+	for _, s := range encryptionSecrets {
```
p0lyn0mial: do we want to collect and dump (to the log file) the number of keys that were deleted?

pkg/operator/encryption/prune_controller_test.go
```diff
@@ -0,0 +1,235 @@
...
+	apiserverconfigv1 "k8s.io/apiserver/pkg/apis/config/v1"
+	"k8s.io/client-go/kubernetes/fake"
+	clientgotesting "k8s.io/client-go/testing"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+func TestEncryptionPruneController(t *testing.T) {
```
p0lyn0mial: drop `encryption` from the name of the test :) ?

pkg/operator/encryption/prune_controller_test.go
```diff
@@ -0,0 +1,235 @@
...
+				all := []*corev1.Secret{}
+				all = append(all, createMigratedEncryptionKeySecretsWithRndKey(t, 10, ns, "secrets")...)
+				all = append(all, createEncryptionKeySecretWithRawKey(ns, nil, 11, []byte("cfbbae883984944e48d25590abdfd300")))
+				return all
+			}(),
+			expectedActions: []string{"list:pods:kms", "get:secrets:kms", "list:secrets:openshift-config-managed", "list:secrets:openshift-config-managed"},
+		},
+
+		{
+			name:            "14 keys were migrated, 1 of them is used, 10 are kept, 3 most recent are pruned",
```
p0lyn0mial: `the 3 oldest are removed` not `the most recent`

pkg/operator/encryption/prune_controller_test.go
```diff
@@ -0,0 +1,235 @@
...
+				schema.GroupResource{Group: "", Resource: "secrets"}: true,
+			},
+			initialSecrets: createMigratedEncryptionKeySecretsWithRndKey(t, 14, "kms", "secrets"),
+			expectedActions: []string{
+				"list:pods:kms",
+				"get:secrets:kms",
+				"list:secrets:openshift-config-managed",
+				"list:secrets:openshift-config-managed",
+				"update:secrets:openshift-config-managed",
+				"delete:secrets:openshift-config-managed",
```
p0lyn0mial: it would be nice if we could include the name of the deleted resource.

pkg/operator/encryption/prune_controller.go
```diff
@@ -0,0 +1,226 @@
...
+	for _, grKeys := range encryptionState {
+		usedSecrets = append(usedSecrets, grKeys.readSecrets...)
+	}
+
+	allkeys, err := c.secretClient.Secrets(operatorclient.GlobalMachineSpecifiedConfigNamespace).List(c.encryptionSecretSelector)
+	if err != nil {
+		return err
+	}
+
+	// sort by keyID
```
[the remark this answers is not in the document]
sttts: will leave it now. I would like to make a package split when we have everything merged. Then we will improve structure.

pkg/operator/encryption/state_test.go
```diff
@@ -0,0 +1,876 @@
...
+			args{
+				createEncryptionCfgNoWriteKey("1", "NzFlYTdjOTE0MTlhNjhmZDEyMjRmODhkNTAzMTZiNGU=", "configmaps", "secrets"),
+				"kms",
+				nil,
+				[]schema.GroupResource{{Group: "", Resource: "configmaps"}, {Group: "", Resource: "secrets"}},
+			},
+			outputMatchingInputConfig,
+		},
+		{
+			"config exists with only one resource => 2nd resource is added",
```
p0lyn0mial: good test

pkg/operator/encryption/state_test.go
```diff
@@ -0,0 +1,876 @@
...
+									Name:   "1",
+									Secret: base64.StdEncoding.EncodeToString([]byte("71ea7c91419a68fd1224f88d50316b4e")),
+								}},
+							},
+						}},
+					},
+				}}),
+		},
+		{
+			"no config, multiple secrets exists, some migrated => config is recreated, with identity as write key",
```
p0lyn0mial: this essentially turns off the encryption - this is something we would like to fix.

pkg/operator/encryption/prune_controller.go
sttts: If somebody deletes something from the config, or applies with one from the backup, we would prune key that we still need for recovery.

Inverse question: is it bad to use the desired config?

A key not used in the desired config was not returned by `keysWithPotentiallyPersistedData`. Hence, every key we delete here won't have influence on the next call of `getDesiredEncryptionState` and hence we don't influence the state machine.

pkg/operator/encryption/state_test.go
```diff
@@ -0,0 +1,876 @@
...
+									Name:   "3",
+									Secret: base64.StdEncoding.EncodeToString([]byte("3cbfbe7d76876e076b076c659cd895ff")),
+								}},
+							},
+						}},
+					},
+				}}),
+		},
+		{
+			"config exists, only some secret is missing => missing secret is not used as write key, but next most-recent key is",
```
p0lyn0mial: this is a very good test

pkg/operator/encryption/state_test.go
```diff
@@ -0,0 +1,876 @@
...
+						},
+					}},
+				"kms",
+				[]*corev1.Secret{
+					createEncryptionKeySecretWithRawKey("kms", nil, 5, []byte("55b5bcbc85cb857c7c07c56c54983cbcd")),
+				},
+				[]schema.GroupResource{{Group: "", Resource: "configmaps"}, {Group: "", Resource: "secrets"}},
+			},
+			equalsConfig(&apiserverconfigv1.EncryptionConfiguration{
+				// 4 is becoming new write key, not 5!
```
p0lyn0mial: rm the comment

pkg/operator/encryption/prune_controller.go
sttts: also the case of `APIServerRevisionNotConverged`. But we watch the pods and should be notified through them. I think we can safely remove this.

pkg/operator/encryption/state_test.go
```diff
@@ -0,0 +1,876 @@
...
+							},
+						}, {
+							Identity: &apiserverconfigv1.IdentityConfiguration{},
+						}},
+					},
+				},
+			}),
+		},
+		{
+			"config exists, read+write keys are consistent, migrated => old read-keys are pruned",
```
p0lyn0mial: this test answers my previous question.

pkg/operator/encryption/prune_controller.go
sttts: done

pkg/operator/encryption/prune_controller_test.go
sttts: 🎉

pkg/operator/encryption/prune_controller_test.go
sttts: done

pkg/operator/encryption/prune_controller_test.go
sttts: follow-up. Wasn't that your test? :-)

pkg/operator/encryption/state_test.go
sttts: just temporarily, which is ok'ish I think.

pkg/operator/encryption/state_test.go
sttts: done

PR #589, opened 2019-10-08
adds EncryptionConfig To GroupResourceKeys roundtrip tests

p0lyn0mial:


pkg/operator/encryption/state_test.go
```diff
@@ -0,0 +1,339 @@
...
+					},
+				},
+
+				{Group: "", Resource: "configmaps"}: {
+					writeKey: keyAndMode{
+						key: apiserverconfigv1.Key{Name: "", Secret: ""}, mode: "identity",
+					},
+					readKeys: []keyAndMode{
+						{key: apiserverconfigv1.Key{Name: "34", Secret: "MTcxNTgyYTBmY2Q2YzVmZGI2NWNiZjVhM2U5MjQ5ZDc="}, mode: "aescbc"},
+						{key: apiserverconfigv1.Key{Name: "35", Secret: newFakeIdentityEncodedKeyForTest()}, mode: "aescbc"},
```
p0lyn0mial: this should be `aesgcm` - I need to fix that.

sttts: Is this a `/hold`?

p0lyn0mial: ok, let's put it into `hold`, but please feel free to review it anyway.

pkg/operator/encryption/state_test.go
```diff
@@ -0,0 +1,517 @@
...
+			input: func() *apiserverconfigv1.EncryptionConfiguration {
+				keysRes := encryptionKeysResourceTuple{
+					resource: "secrets",
+					keys: []apiserverconfigv1.Key{
+						{
+							Name:   "34",
+							Secret: "MTcxNTgyYTBmY2Q2YzVmZGI2NWNiZjVhM2U5MjQ5ZDc=",
+						},
+
+						// secretsToProviders puts "fakeIdentityProvider" as last
```
p0lyn0mial: will this special key be ever removed?
what happens when a user frequently turns on and off the encryption?

p0lyn0mial: would be nice to add unit tests for ^

pkg/operator/encryption/helpers_test.go
```diff
@@ -238,9 +281,13 @@ func createEncryptionCfgSecretWithWriteKeys(t *testing.T, targetNs string, revis
 	}
 }
 
+// TODO: change the name to encryptionKeysResourceMode
 type encryptionKeysResourceTuple struct {
```
sttts: we have too many key/mode structs

pkg/operator/encryption/state_test.go
sttts: I don't understand that well yet. Yes, at the very least we need heavy unit tests.

PR #591, opened 2019-10-10
encryption: fix encryption bootstrap logic

sttts:
- key controller starts with the first key when the mode changes
- state controller follows key controller only
- when config or keys exists, resetting mode to "" has no effect,
  other than being understood as identity, but leaves encryption
  mechanism active.

### DIFF
--- a/pkg/operator/encryption/state.go
+++ b/pkg/operator/encryption/state.go
@@ -379,7 +379,7 @@ func getEncryptionConfigAndState(
 	secretClient corev1client.SecretsGetter,
 	targetNamespace string,
 	encryptionSecretSelector metav1.ListOptions,
-	encryptedGRs map[schema.GroupResource]bool,
+	encryptedGRs []schema.GroupResource,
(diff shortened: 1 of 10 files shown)


pkg/operator/encryption/state.go
```diff
@@ -26,13 +26,18 @@ func getCurrentEncryptionConfig(secrets corev1client.SecretInterface, revision s
...
 		if errors.IsNotFound(err) {
-			return &apiserverconfigv1.EncryptionConfiguration{}, nil
+			return nil, nil
 		}
 		return nil, err
 	}
 
+	data, ok := encryptionConfigSecret.Data[encryptionConfSecret]
+	if !ok {
+		return nil, nil
```
p0lyn0mial: does it mean that the current encryption config is broken?
shouldn't we report an error in this case?

sttts: We just behave like for a missing config. We can recover.

PR #592, opened 2019-10-10
encryption: wire controllers into operator startup

sttts:


sttts:
TestNamedCertificates flake

/retest

PR #597, opened 2019-10-11
encryption: rewrite migration loop

sttts:
Requires https://github.com/openshift/cluster-kube-apiserver-operator/pull/591.

- don't set progressing condition when waiting
- simplify migration loop without tracker
- iteratively set resource annotation on secrets in order to recover from issues in one resource without repeating already migrated resources


pkg/operator/encryption/migration_controller.go
```diff
@@ -198,46 +196,73 @@ func (c *migrationController) migrateKeysIfNeededAndRevisionStable() (resetProgr
...
+
+	// if we reach this, all migration went fine and we can reset progressing condition
+	return true, nil
+}
+
+func setResourceMigrated(gr schema.GroupResource, s *corev1.Secret) (bool, error) {
+	migratedGRs := migratedGroupResources{}
+	if existing, found := s.Annotations[encryptionSecretMigratedResources]; found {
+		if err := json.Unmarshal([]byte(existing), &migratedGRs); err != nil {
+			// ignore error and just start fresh
```
p0lyn0mial: the worst-case scenario is that we will run migration one more time - why not :)

pkg/operator/encryption/migration_controller.go
```diff
@@ -198,46 +196,73 @@ func (c *migrationController) migrateKeysIfNeededAndRevisionStable() (resetProgr
...
 		}
-		migratedSecret.Annotations[encryptionSecretMigratedResources] = string(migratedResourceBytes)
-		_, _, updateErr := resourceapply.ApplySecret(c.secretClient, c.eventRecorder, migratedSecret)
-		if updateErr != nil {
-			errs = append(errs, updateErr)
+	}
+	alreadyMigrated := false
+	for _, existingGR := range migratedGRs.Resources {
+		if existingGR == gr {
+			alreadyMigrated = true
```
p0lyn0mial: if the matching `gr` was found simply return

pkg/operator/encryption/migration_controller.go
```diff
@@ -198,46 +196,73 @@ func (c *migrationController) migrateKeysIfNeededAndRevisionStable() (resetProgr
...
+			break
 		}
 	}
+	if alreadyMigrated {
```
p0lyn0mial: then this is not needed.

pkg/operator/encryption/migration_controller.go
```diff
@@ -198,46 +196,73 @@ func (c *migrationController) migrateKeysIfNeededAndRevisionStable() (resetProgr
...
+	if s.Annotations == nil {
+		s.Annotations = map[string]string{}
+	}
+
+	s.Annotations[encryptionSecretMigratedTimestamp] = time.Now().Format(time.RFC3339)
+
+	migratedGRs.Resources = append(migratedGRs.Resources, gr)
+	if bs, err := json.Marshal(migratedGRs); err != nil {
+		return false, fmt.Errorf("failed to marshal %s annotation value %#v for key secret %s/%s", encryptionSecretMigratedResources, migratedGRs, s.Namespace, s.Name)
+	} else {
```
p0lyn0mial: drop `else` statement since we always return when `if` condition is satisfied.

pkg/operator/encryption/migration_controller.go
```diff
@@ -198,46 +196,73 @@ func (c *migrationController) migrateKeysIfNeededAndRevisionStable() (resetProgr
...
 	}
+	if alreadyMigrated {
+		return false, nil
+	}
+
+	if s.Annotations == nil {
+		s.Annotations = map[string]string{}
+	}
+
+	s.Annotations[encryptionSecretMigratedTimestamp] = time.Now().Format(time.RFC3339)
```
p0lyn0mial: nit: you could move it close to line 262 where we annotate `encryptionSecretMigratedResources` for readability.

pkg/operator/encryption/migration_controller.go
```diff
@@ -198,46 +196,73 @@ func (c *migrationController) migrateKeysIfNeededAndRevisionStable() (resetProgr
...
-		}
-
 		if needsMigration(writeSecret, gr) {
-			migrationErr := c.runStorageMigration(gr)
-			errs = append(errs, migrationErr)
-			if migrationErr != nil {
-				continue
+			// storage migration takes a long time so we expose that via a distinct status change
+			if err := c.setProgressing(strings.Title(groupToHumanReadable(gr))+strings.Title(gr.Resource), "migrating resource %s.%s to new write key", groupToHumanReadable(gr), gr.Resource); err != nil {
+				return false, err
```
p0lyn0mial: we count continue, migration is more important than setting a status, wdyt?

sttts: we will continue anyway, 2min later. Both ways would be ok I guess, either wait just continue here. Do you have a strong preference?

pkg/operator/encryption/migration_controller.go
```diff
@@ -198,46 +196,73 @@ func (c *migrationController) migrateKeysIfNeededAndRevisionStable() (resetProgr
...
-			return false, err
-		}
-
 		if needsMigration(writeSecret, gr) {
-			migrationErr := c.runStorageMigration(gr)
-			errs = append(errs, migrationErr)
-			if migrationErr != nil {
-				continue
+			// storage migration takes a long time so we expose that via a distinct status change
+			if err := c.setProgressing(strings.Title(groupToHumanReadable(gr))+strings.Title(gr.Resource), "migrating resource %s.%s to new write key", groupToHumanReadable(gr), gr.Resource); err != nil {
```
p0lyn0mial: shouldn't this be the other way around? `migrating resource %s.%s to new write key` as a first arg since it contains formating info.

pkg/operator/encryption/migration_controller.go
sttts: There are more resources waiting to be migrated.

pkg/operator/encryption/migration_controller_test.go
```diff
@@ -168,8 +168,13 @@ func TestMigrationController(t *testing.T) {
 				"get:secrets:kms",
 				"list:secrets:openshift-config-managed",
 				"list:secrets:openshift-config-managed",
+				"get:secrets:openshift-config-managed",
```
p0lyn0mial: oh, wow, fewer events!

pkg/operator/encryption/migration_controller.go
sttts: done

pkg/operator/encryption/migration_controller.go
p0lyn0mial: would prefer to continue stright away but please add a comment if you don't mind.

pkg/operator/encryption/migration_controller.go
sttts: the first is the reason, second the message (with printf semantics)

pkg/operator/encryption/migration_controller_test.go
sttts: one after each resource. So you know where it stands. Poor-men's progress view (will all be replaced with @sanchezl 's storage migration and pretty progress logging in the migration object status).

pkg/operator/encryption/migration_controller.go
sttts: Don't want to risk the green tests now. But put it on the list for TODOs. Have no strong opinion.

PR #599, opened 2019-10-11
add max-in-flight

deads2k:
Setting these higher. The kube-apiserver appears to be fine with more load.


RobertKrawitz:
I certainly have no objection, but this should be tested downscale too (smaller cluster with heavy object load).

PR #601, opened 2019-10-14
encryption: decouple secret names from config

sttts:
Secret names played a role for comparison which secrets are used in the config. This PR changes this by only taking the `<prefix>-<n>` number for sorting, but otherwise just compares mode and keys.

### DIFF
--- a/pkg/operator/encryption/state.go
+++ b/pkg/operator/encryption/state.go
@@ -53,6 +53,7 @@ func getCurrentEncryptionConfig(secrets corev1client.SecretInterface, revision s
 // To do this it compares the current state against the available secrets and to-be-encrypted resources.
 // oldEncryptionConfig can be nil if there is no config yet.
 // If there are no secrets, the identity is set for all resources as write key.
+// It is assumed that encryptionSecrets are all valid.
 //
 // The basic rules are:
 //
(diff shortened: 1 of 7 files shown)


pkg/operator/encryption/state.go
```diff
@@ -253,26 +250,21 @@ func encryptionConfigToEncryptionState(targetNamespace string, c *apiserverconfi
...
 	bs, err := base64.StdEncoding.DecodeString(k.key.Secret)
 	if err != nil {
 		return nil, fmt.Errorf("failed decoding base64 data of key %s", k.key.Name)
 	}
 	return &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      fmt.Sprintf("%s-encryption-%s", targetNamespace, k.key.Name),
-			Namespace: operatorclient.GlobalMachineSpecifiedConfigNamespace,
-			Labels: map[string]string{
-				encryptionSecretComponent: targetNamespace,
```
p0lyn0mial: We are going to list and check all the secrets in a namespace without this label. It means we might report a lot of errors for example `invalid key`. Thoughts?

p0lyn0mial: nvm, those secrets are not stored in the database.

PR #603, opened 2019-10-14
encryption: create new key when key not backed by secret

sttts:
- fake secrets have no namespace. We use this (temporarily until we have a good internal state data structure) to tell the key controller that it should create a new key.
- clean up key controller tests
- fix along the way that not-most-recent keys are used as write keys if there are newer ones in the config already.

### DIFF
--- a/pkg/operator/encryption/key_controller.go
+++ b/pkg/operator/encryption/key_controller.go
@@ -157,7 +157,7 @@ func (c *keyController) checkAndCreateKeys() error {
 	// are missing.
 
 	for _, grKeys := range desiredEncryptionState {
-		keyID, internalReason, ok := needsNewKey(grKeys, currentMode, externalReason)
+		keyID, internalReason, ok := needsNewKey(grKeys, currentMode, externalReason, c.encryptedGRs)
 		if !ok {
 			continue
 		}
@@ -266,13 +266,7 @@ func (c *keyController) getCurrentModeAndExternalReason() (mode, string, error) 
 }
 
 // TODO unit tests
-func needsNewKey(grKeys keysState, currentMode mode, externalReason string) (uint64, string, bool) {
-	// if the length of read secrets is more than one (i.e. we have more than just the write key),
-	// then we haven't successfully migrated and removed old keys so you should wait before generating more keys.
-	if len(grKeys.readSecrets) > 1 {
-		return 0, "", false
-	}
-
+func needsNewKey(grKeys keysState, currentMode mode, externalReason string, encryptedGRs []schema.GroupResource) (uint64, string, bool) {
 	// we always need to have some encryption keys unless we are turned off
 	if len(grKeys.readSecrets) == 0 {
 		return 0, "no-secrets", currentMode != identity
@@ -280,8 +274,19 @@ func needsNewKey(grKeys keysState, currentMode mode, externalReason string) (uin
 
 	latestKey, latestKeyID := grKeys.latestKey()
 
+	// if latest secret has been deleted, we will never be able to migrate to that key.
+	if !secretIsBacked(latestKey) {
+		return latestKeyID, "missing-secret", true
+	}
+
+	// if the length of read secrets is more than one (i.e. we have more than just the write key),
+	// then we haven't successfully migrated and removed old keys so you should wait before generating more keys.
+	if len(grKeys.readSecrets) > 1 {
+		return 0, "", false
+	}
+
 	// we have not migrated the latest key, do nothing until that is complete
-	if len(latestKey.Annotations[encryptionSecretMigratedTimestamp]) == 0 {
+	if allMigrated, _, _ := migratedFor(encryptedGRs, latestKey); !allMigrated {
 		return 0, "", false
 	}
 
--- a/pkg/operator/encryption/key_controller_test.go
+++ b/pkg/operator/encryption/key_controller_test.go
@@ -128,9 +128,36 @@ func TestKeyController(t *testing.T) {
 			},
 		},
 
-		// Verifies if a new key is not created when there is a valid write key in the system.
 		{
-			name: "no-op when a valid write key exists",
+			name: "no-op when a valid write key exists, but is not migrated",
+			targetGRs: []schema.GroupResource{
+				{Group: "", Resource: "secrets"},
+			},
+			initialObjects: []runtime.Object{
+				createDummyKubeAPIPod("kube-apiserver", "kms"),
+				createEncryptionKeySecretWithRawKey("kms", nil, 7, []byte("61def964fb967f5d7c44a2af8dab6865")),
+			},
+			apiServerObjects: apiServerAesCBC,
+			targetNamespace:  "kms",
+			expectedActions:  []string{"list:pods:kms", "get:secrets:kms", "list:secrets:openshift-config-managed"},
+		},
+
+		{
+			name: "no-op when a valid write key exists, is migrated, but not expired",
+			targetGRs: []schema.GroupResource{
+				{Group: "", Resource: "secrets"},
+			},
+			initialObjects: []runtime.Object{
+				createDummyKubeAPIPod("kube-apiserver", "kms"),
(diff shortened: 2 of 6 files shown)


sttts:
/hold

adding some state_test.go tests.

pkg/operator/encryption/key_controller.go
```diff
@@ -266,22 +266,27 @@ func (c *keyController) getCurrentModeAndExternalReason() (mode, string, error)
...
 	// we always need to have some encryption keys unless we are turned off
 	if len(grKeys.readSecrets) == 0 {
 		return 0, "no-secrets", currentMode != identity
 	}
 
 	latestKey, latestKeyID := grKeys.latestKey()
 
+	// if latest secret has been deleted, we will never be able to migrate to that key.
+	if !secretIsBacked(latestKey) {
+		return latestKeyID, "missing-secret", true
```
p0lyn0mial: if the key is missing what will be `latestKeyID`?

pkg/operator/encryption/key_controller.go
```diff
@@ -266,22 +266,27 @@ func (c *keyController) getCurrentModeAndExternalReason() (mode, string, error)
...
+
+	// if the length of read secrets is more than one (i.e. we have more than just the write key),
+	// then we haven't successfully migrated and removed old keys so you should wait before generating more keys.
+	if len(grKeys.readSecrets) > 1 {
+		return 0, "", false
+	}
+
 	// we have not migrated the latest key, do nothing until that is complete
-	if len(latestKey.Annotations[encryptionSecretMigratedTimestamp]) == 0 {
+	if allMigrated, _, _ := migratedFor(encryptedGRs, latestKey); !allMigrated {
```
p0lyn0mial: ok this will also work when a gr is removed from `encryptedGRs`

pkg/operator/encryption/key_controller.go
```diff
@@ -266,22 +266,27 @@ func (c *keyController) getCurrentModeAndExternalReason() (mode, string, error)
...
-		return 0, "", false
-	}
-
+func needsNewKey(grKeys keysState, currentMode mode, externalReason string, encryptedGRs []schema.GroupResource) (uint64, string, bool) {
 	// we always need to have some encryption keys unless we are turned off
 	if len(grKeys.readSecrets) == 0 {
 		return 0, "no-secrets", currentMode != identity
 	}
 
 	latestKey, latestKeyID := grKeys.latestKey()
```
p0lyn0mial: will this function behave correctly when a secret was deleted?

pkg/operator/encryption/key_controller_test.go
```diff
@@ -128,37 +128,48 @@ func TestKeyController(t *testing.T) {
...
+				createMigratedEncryptionKeySecretWithRawKey("kms", []schema.GroupResource{{Group: "", Resource: "secrets"}}, 7, []byte("61def964fb967f5d7c44a2af8dab6865"), time.Now()),
 			},
 			apiServerObjects: apiServerAesCBC,
 			targetNamespace:  "kms",
 			expectedActions:  []string{"list:pods:kms", "get:secrets:kms", "list:secrets:openshift-config-managed"},
 		},
 
-		// Checks if a new write key is created because the previous one was migrated.
+		{
+			name: "creates a new write key because previous one is migrated, but has no migration timestamp",
```
p0lyn0mial: okay but if a secret doesn't have a migration timestamp then we dead lock.

pkg/operator/encryption/prune_controller.go
```diff
@@ -139,9 +139,12 @@ func (c *pruneController) deleteOldMigratedSecrets() error {
 
 	var deleteErrs []error
 	skippedKeys := 0
+NextEncryptionSecret:
 	for _, s := range encryptionSecrets {
-		if hasKeyInSecret(usedSecrets, s) {
-			continue
+		for _, us := range usedSecrets {
+			if equalKeyInSecret(us, s) {
```
p0lyn0mial: why we cannot use `equalKeyAndEqualID`?

pkg/operator/encryption/key_controller.go
sttts: the right one. We have the name, just not the namespace.

pkg/operator/encryption/key_controller.go
sttts: yes. The desired state includes deleted keys with the right IDs. It's what you will see here, and the check below will notice that we have to create a new key in a secret.

pkg/operator/encryption/key_controller_test.go
sttts: no, there is an extra test case. If there is no timestamp, we will behave as it is expired.

pkg/operator/encryption/prune_controller.go
sttts: we possibly could

pkg/operator/encryption/state.go
```diff
@@ -122,16 +131,35 @@ func getDesiredEncryptionState(oldEncryptionConfig *apiserverconfigv1.Encryption
...
 		for _, expected := range expectedReadSecrets {
-			if !hasKeyInSecret(grState.readSecrets, expected) {
+			found := false
+			for _, rs := range grState.readSecrets {
+				if equalKeyInSecret(rs, expected) {
+					found = true
+					break
+				}
+			}
+			if !found {
```
p0lyn0mial: when this can happen? both `rs` and `expected` were taken from `encryptionSecrets`

p0lyn0mial: ah, it means that a key is missing from `desiredEncryptionState`

pkg/operator/encryption/state.go
```diff
@@ -122,16 +131,35 @@ func getDesiredEncryptionState(oldEncryptionConfig *apiserverconfigv1.Encryption
...
 		if changed {
+			grState.readSecrets = sortRecentFirst(grState.readSecrets)
 			desiredEncryptionState[gr] = grState
 		}
+
+		// verify that all read secrets are backed by a real secret, and therefore
+		// in encryptionSecrets. Read secrets are sorted. So encryptionSecrets[0]
+		// is a good next write key because it is most recent.
+		for _, rs := range grState.readSecrets {
+			if !secretIsBacked(rs) {
```
p0lyn0mial: when this can happen? both `rs` and `expected` were taken from `encryptionSecrets`

p0lyn0mial: ic, https://github.com/openshift/cluster-kube-apiserver-operator/pull/603/files#diff-010b01c35a36be66869aae48f2ec96d5R91

pkg/operator/encryption/state_test.go
```diff
@@ -488,15 +488,15 @@ func TestGetDesiredEncryptionState(t *testing.T) {
...
+									Secret: base64.StdEncoding.EncodeToString([]byte("55b5bcbc85cb857c7c07c56c54983cbcd")),
 								}},
 							},
 						}, {
 							AESCBC: &apiserverconfigv1.AESConfiguration{
 								Keys: []apiserverconfigv1.Key{{
-									Name:   "5",
-									Secret: base64.StdEncoding.EncodeToString([]byte("55b5bcbc85cb857c7c07c56c54983cbcd")),
+									Name:   "4",
+									Secret: base64.StdEncoding.EncodeToString([]byte("447907494bßc4897b876c8476bf807bc")),
```
p0lyn0mial: `ß` is not a valid hex-encoded char :)

pkg/operator/encryption/state.go
```diff
@@ -122,16 +131,35 @@ func getDesiredEncryptionState(oldEncryptionConfig *apiserverconfigv1.Encryption
...
 				allReadSecretsAsExpected = false
 				klog.V(4).Infof("%s missing read secret %s", gr, expected.Name)
 			}
 		}
 		if changed {
+			grState.readSecrets = sortRecentFirst(grState.readSecrets)
 			desiredEncryptionState[gr] = grState
 		}
+
+		// verify that all read secrets are backed by a real secret, and therefore
```
p0lyn0mial: is this a hot loop? will we always get into this state if a secret was removed from the database?

sttts:
Part of https://github.com/openshift/cluster-kube-apiserver-operator/pull/606

PR #604, opened 2019-10-14
adds an e2e happy path test case for etcd encryption

p0lyn0mial:


p0lyn0mial:
/hold

need to make sure the test are stable

p0lyn0mial:
/hold cancel

there are some things I'd like to improve but having the tests seems more important

PR #615, opened 2019-10-17
adds an e2e happy path test case for etcd encryption

p0lyn0mial:


test/library/encryption/e_log.go
```diff
@@ -0,0 +1,58 @@
+package encryption
+
+import (
+	"fmt"
+	"os"
+	"testing"
+)
+
+// E is like testing.T except it overloads some methods to print to stdout
```
p0lyn0mial: don't tell me that `go` is not an `OOP` language

test/library/cluster_operator.go
```diff
@@ -3,38 +3,50 @@ package library
...
 )
 
 // WaitForKubeAPIServerClusterOperatorAvailableNotProgressingNotDegraded waits for ClusterOperator/kube-apiserver to report
 // status as active, not progressing, and not failing.
-func WaitForKubeAPIServerClusterOperatorAvailableNotProgressingNotDegraded(t *testing.T, client configclient.ConfigV1Interface) {
-	err := wait.Poll(WaitPollInterval, WaitPollTimeout, func() (bool, error) {
+func WaitForKubeAPIServerClusterOperatorAvailableNotProgressingNotDegraded(t testing.TB, client configv1client.ConfigV1Interface) {
+	t.Helper()
+	t.Log("Waiting for ClusterOperator/kube-apiserver to be Available: true, Progressing: false, Degraded: false, Done: false")
+	time.Sleep(time.Minute) // make sure we are not racing against an initial observation of change
```
sttts: suspiciuos. Why this?

test/library/cluster_operator.go
```diff
@@ -3,38 +3,50 @@ package library
...
 		conditions := clusterOperator.Status.Conditions
-		available := clusteroperatorhelpers.IsStatusConditionPresentAndEqual(conditions, configv1.OperatorAvailable, configv1.ConditionTrue)
-		notProgressing := clusteroperatorhelpers.IsStatusConditionPresentAndEqual(conditions, configv1.OperatorProgressing, configv1.ConditionFalse)
-		notDegraded := clusteroperatorhelpers.IsStatusConditionPresentAndEqual(conditions, configv1.OperatorDegraded, configv1.ConditionFalse)
-		done := available && notProgressing && notDegraded
-		fmt.Printf("ClusterOperator/kube-apiserver: Available: %v  Progressing: %v  Degraded: %v\n", available, !notProgressing, !notDegraded)
+		available := clusteroperatorhelpers.IsStatusConditionTrue(conditions, configv1.OperatorAvailable)
+		notProgressing := clusteroperatorhelpers.IsStatusConditionFalse(conditions, configv1.OperatorProgressing)
+		notDegraded := clusteroperatorhelpers.IsStatusConditionFalse(conditions, configv1.OperatorDegraded)
+		// make sure that we have not been progressing for a while so that multi-stage rollouts are correctly accounted for
```
sttts: what does this mean?

test/library/encryption/assertion.go
```diff
@@ -0,0 +1,116 @@
...
+		return fmt.Errorf("unexpected encrypted state, expected data to be %q but was %q, mode %q", conditionToStr(isEncrypted), conditionToStr(expectedEncrypted), actualMode)
+	}
+	if actualMode != expectedMode {
+		return fmt.Errorf("unexpected encryption mode %q, expected %q, was data encrypted/decrypted with a wrong key", actualMode, expectedMode)
+	}
+
+	return nil
+}
+
+func determineEncryptionMode(data []byte) (string, bool) {
```
sttts: encryptionModeFromEtcdValue

test/library/encryption/e_log.go
sttts: Go is not an OOP language.

test/library/cluster_operator.go
p0lyn0mial: I can remove it, Mo added it.

test/library/cluster_operator.go
p0lyn0mial: the last transition time for `progressing` condition must be older than `1 min`, as some tests want to encrypt more than one `gr` and want to roll out more than one key that condition could change.

test/library/cluster_operator.go
p0lyn0mial: he might have added it because we don't know when encryption has started and it might happen that the condition will be satisfied immediately when encryption hasn't started.

p0lyn0mial: we could try to observe conditions from individual encryption controllers instead of looking at the aggregate condition but it is not that bad.

test/library/cluster_operator.go
p0lyn0mial: it might be a wrong assumption when we run multiple tests ....

test/library/cluster_operator.go
p0lyn0mial: we could also add a condition that would tell us that migration is done (all `GRs` migrated) and watch it.

p0lyn0mial:
/test e2e-aws-operator

make sure it is stable

sttts:
We will replace `WaitForOperatorAndMigrationControllerAvailableNotProgressingNotDegraded` with a proper `Encrypted` condition in a follow-up.

/lgtm
/approve

PR #616, opened 2019-10-18
more e2e tests for encryption

p0lyn0mial:

PR #618, opened 2019-10-19
Encryption e2e tests rotation

p0lyn0mial:


p0lyn0mial:
builds on https://github.com/openshift/cluster-kube-apiserver-operator/pull/616

test/library/encryption/assertion.go
```diff
@@ -10,35 +10,68 @@ import (
...
+
+	configv1 "github.com/openshift/api/config/v1"
 )
 
 var protoEncodingPrefix = []byte{0x6b, 0x38, 0x73, 0x00}
 
+var defaultTargetGRs = []schema.GroupResource{
+	{Group: "", Resource: "secrets"},
+	{Group: "", Resource: "configmaps"},
+}
```
sttts: we need a constant in the operator code and reference that here.

test/library/encryption/e_log.go
```diff
@@ -19,7 +19,7 @@ func NewE(t *testing.T) *E {
 	// that means we don't have any visibility when running the tests from a local machine
 	//
 	// thus std logger will be used when the test are run from a local machine to give instant feedback
-	if len(os.Getenv("PROW_JOB_ID")) == 0 {
+	if len(os.Getenv("PROW_JOB_ID")) != 0 {
 		e.local = true
 	}
```
sttts: is this change correct?

test/library/encryption/helpers.go
```diff
@@ -36,55 +86,194 @@ func GetClients(t testing.TB) (EtcdClient, configv1client.APIServerInterface, op
...
-
-	observedOperatorStateAsString := ""
-	if err := wait.Poll(waitPollInterval, waitPollTimeout, func() (bool, error) {
-		clusterOperator, err := operatorClient.Get("cluster", metav1.GetOptions{})
-		if errors.IsNotFound(err) {
-			t.Log("KubeAPIServer/cluster operator does not yet exist.")
-			return false, nil
+	waitNoKeyPollInterval := 15 * time.Second
+	waitNoKeyPollTimeout := 6 * time.Minute
+	waitDuration := 5 * time.Minute
```
sttts: What are these timeouts? Where do they come from?

test/library/encryption/e_log.go
p0lyn0mial: nope, need to revert it, thx.

test/library/encryption/helpers.go
p0lyn0mial: in some cases, we know that a new key will not be created, for example, running `aescbc` scenario one after another will create only one key. Here the tests wait up to 5 min before checking if resources were migrated.

test/library/encryption/assertion.go
p0lyn0mial: should e2e tests trust production code :)?

I was thinking about listing all resources (directly from etcd) and checking the prefix. We would fail if some resources have the prefix but are not on that list.

sttts: If it doesn't trust that code, just add a test here that your list matches the prod list.

test/library/encryption/helpers.go
sttts: Isn't there a better way to know that? Waiting minutes for nothing to happen is not a good idea IMO.

test/library/encryption/assertion.go
p0lyn0mial: will add the test in a separate PR.

PR #622, opened 2019-10-21
move KubeAPIDown alert to this operator

sjenning:
https://github.com/openshift/cluster-monitoring-operator/pull/516


sjenning:
fyi @csrwng @derekwaynecarr

PR #625, opened 2019-10-23
Add alert for deprecated apis

tnozicka:
/cc @soltysh 

/hold
(just in case, I think this should fail on current offenders)

### DIFF
--- a/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
+++ b/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
@@ -35,3 +35,34 @@ spec:
     matchLabels:
       component: apiserver
       provider: kubernetes
+---
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: kube-apiserver
+  namespace: openshift-kube-apiserver
+spec:
+  groups:
+  - name: using-deprecated-apis
+    rules:
+    - alert: UsingDeprecatedAPIAppsV1Beta1
+      annotations:
+        message: A client in the cluster is using deprecated apps/v1beta1 API that will be removed soon.
+      expr: |
+        apiserver_request_count{group="apps",version="v1beta1"}
+      labels:
+        severity: warning
+    - alert: UsingDeprecatedAPIAppsV1Beta2
+      annotations:
+        message: A client in the cluster is using deprecated apps/v1beta2 API that will be removed soon.
+      expr: |
+        apiserver_request_count{group="apps",version="v1beta2"}
+      labels:
+        severity: warning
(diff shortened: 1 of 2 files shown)


tnozicka:
I have manually checked that the alerts are present in the prometheus dump from the CI run - that should have failed aws-e2e - monitoring team is working on a fix. We should merge when the check is fixed, we get really green e2e and no alerts are firing. (Also means making the offenders to fix their code.)

tnozicka:
e2e fix in https://github.com/openshift/origin/pull/24005

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -35,3 +35,34 @@ spec:
...
+        message: A client in the cluster is using deprecated apps/v1beta2 API that will be removed soon.
+      expr: |
+        apiserver_request_count{group="apps",version="v1beta2"}
+      labels:
+        severity: warning
+    - alert: UsingDeprecatedAPIExtensionsV1Beta1
+      annotations:
+        message: A client in the cluster is using deprecated extensions/v1beta1 API that will be removed soon.
+      expr: |
+        apiserver_request_count{group="extensions",version=~"v1beta1",resource!~"(?:ingresses|)",client!~".*/kube-controller-manager"}
```
soltysh: Any particular reason you do `=~` when matching version here? I doubt it's required.

tnozicka: not really, just wanted to set pattern for easily adding another one if need be

tnozicka: oh, you mean version - that's a leftover from where it was matching v1beta2 as well before I made it separate alert, will fix it tomorrow

sttts:
:+1:

LGTM

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
tnozicka: fixed

tnozicka:
alerts are firing in https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-kube-apiserver-operator/625/pull-ci-openshift-cluster-kube-apiserver-operator-master-e2e-aws/2869/artifacts/e2e-aws/metrics/prometheus.tar , the e2e that should stop merging it is green :(

tnozicka:
ok, this needs to land for feature

/hold cancel

I'll follow up with a high prio BZ to fix the e2e check; the ones for offenders are already filled

tnozicka:
ref https://bugzilla.redhat.com/show_bug.cgi?id=1766638

PR #626, opened 2019-10-25
add privilege to operator init container

haircommander:
CRI-O 1.16 made a change to how selinux labels were handled that increased security. In doing so, it tightened security and needs to be explicitly told to not label a container. In the case of this initContainer, it is trying to edit the permissions of a file on the host system, which is (to SELinux) a security flaw, and thus selinux is denying it.

explicitly add privilege to the initContainer so it can touch the hostPath

Signed-off-by: Peter Hunt <pehunt@redhat.com>

### DIFF
--- a/bindata/v4.1.0/kube-apiserver/pod.yaml
+++ b/bindata/v4.1.0/kube-apiserver/pod.yaml
@@ -17,6 +17,8 @@ spec:
         - mountPath: /var/log/kube-apiserver
           name: audit-dir
       command: ['/usr/bin/timeout', '105', '/bin/bash', '-ec'] # a bit more than 60s for graceful termination + 35s for minimum-termination-duration, 5s extra cri-o's graceful termination period
+      securityContext:
+        privileged: true
       args:
       - |
         echo -n "Fixing audit permissions."
--- a/pkg/operator/v410_00_assets/bindata.go
+++ b/pkg/operator/v410_00_assets/bindata.go
@@ -326,6 +326,8 @@ spec:
         - mountPath: /var/log/kube-apiserver
           name: audit-dir
       command: ['/usr/bin/timeout', '105', '/bin/bash', '-ec'] # a bit more than 60s for graceful termination + 35s for minimum-termination-duration, 5s extra cri-o's graceful termination period
+      securityContext:
+        privileged: true
       args:
       - |
         echo -n "Fixing audit permissions."


haircommander:
PTAL @sttts @soltysh @deads2k

sttts:
/lgtm
/hold

@smarterclayton have we decided that we break the static pod API for customers?

bindata/v4.1.0/kube-apiserver/pod.yaml
```diff
@@ -17,6 +17,8 @@ spec:
         - mountPath: /var/log/kube-apiserver
           name: audit-dir
       command: ['/usr/bin/timeout', '105', '/bin/bash', '-ec'] # a bit more than 60s for graceful termination + 35s for minimum-termination-duration, 5s extra cri-o's graceful termination period
+      securityContext:
+        privileged: true
```
sttts: is this also needed for the bootstrap pot or don't we have init containers there?

sttts: Also don't forget kube-controller-manager and kube-scheduler.

haircommander: If a container in the pod wants to touch something on the host, then it should be privileged or be given the right label

haircommander: a quick check of kube-controller-manager and kube-scheduler seem fine. Their init containers only wait on ports.

haircommander: I don't know where the bootstrap pod you're wondering about lives

bindata/v4.1.0/kube-apiserver/pod.yaml
```diff
@@ -17,6 +17,8 @@ spec:
...
       args:
       - |
         echo -n "Fixing audit permissions."
```
cgwalters: That said...let's move this bit into the MCO?  We can install a `systemd-tmpfiles` unit to do so.  Having an init container do this chmod seems strange at best.  I mean, we could just as easily do the `chmod` inside the go code too right?

cgwalters:
See also https://github.com/openshift/cluster-openshift-apiserver-operator/pull/150 which was the change that was clearly relying on the crio bug/issue here.

bindata/v4.1.0/kube-apiserver/pod.yaml
sttts: https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/bindata/bootkube/bootstrap-manifests/kube-apiserver-pod.yaml

No init container.

PR #629, opened 2019-10-28
Properly set pod as privileged

haircommander:
There is no pod-level security policy to set privileged. Instead, explicitly set privileged for every container in the pod that needs it (the two that mount audit-dir)

Signed-off-by: Peter Hunt <pehunt@redhat.com>

### DIFF
--- a/bindata/v4.1.0/kube-apiserver/pod.yaml
+++ b/bindata/v4.1.0/kube-apiserver/pod.yaml
@@ -26,6 +26,8 @@ spec:
           echo -n "."
           sleep 1
         done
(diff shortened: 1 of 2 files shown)


umohnani8:
@deads2k @sttts PTAL

deads2k:
/lgtm

Last one on credit.  Next time prove it first

haircommander:
@deads2k deal

sdodson:
This is already in release-4.3 because this merged prior to the fork. Doing some BZ shuffling to appease the bots though.

PR #630, opened 2019-10-29
metrics: report the control plane architecture

crawford:
We are going to soon start supporting instruction set architectures
other than x86_64/AMD64. This new metric will help use understand how
widely these different architectures are used.

The control plane will very likely always be homogeneous, so this metric
assumes that only a single architecture will be reported. Additionally,
for the foreseeable future, OpenShift will only support homogeneous
clusters, so reporting the architecture of the control plane will be
enough to know the architectures in use by the entire cluster.

### DIFF
--- a/pkg/operator/configmetrics/configmetrics.go
+++ b/pkg/operator/configmetrics/configmetrics.go
@@ -1,6 +1,8 @@
 package configmetrics
 
 import (
+	"runtime"
+
 	"github.com/prometheus/client_golang/prometheus"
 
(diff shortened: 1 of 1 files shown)


smarterclayton:
We already capture node workloads for master nodes.  We should revert this.

smarterclayton:
We already capture this metric in `node_role_os_version_machine:cpu_capacity_cores:sum` as `label_kubernetes_io_arch` which has node role labels as well for master and is reported through telemetry.  Opened a revert.

PR #634, opened 2019-10-31
Revert "metrics: report the control plane architecture"

smarterclayton:
Reverts openshift/cluster-kube-apiserver-operator#630


smarterclayton:
I will never be angry at someone who adds metrics.  Ever.  Well, unless those metrics blow up the entire cluster.

PR #635, opened 2019-10-31
Add cert regeneration controller

tnozicka:
Adds a container into kube-apiserver static pod that will automatically refresh certs if those are already expired.

https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/auto-cert-recovery.md
https://jira.coreos.com/browse/WRKLDS-100

TODO:
 - [x] extract library-go changes if the CI goes green (https://github.com/openshift/library-go/pull/579)
 - [x] real bump of library-go when the changes are in
 - [x] needs long lived service account token https://github.com/openshift/cluster-kube-apiserver-operator/pull/663
 - [x] try it

/cc @deads2k @sttts @soltysh @mfojtik

### DIFF
--- a/pkg/cmd/certregenerationcontroller/cabundlesyncer.go
+++ b/pkg/cmd/certregenerationcontroller/cabundlesyncer.go
@@ -1,0 +1,145 @@
+package certregenerationcontroller
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog"
+
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/operatorclient"
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/targetconfigcontroller"
+)
+
+const workQueueKey = "key"
+
+// CABundleController composes individual certs into CA bundle that is used
+// by kube-apiserver to validate clients.
+// Cert recovery refreshes "kube-control-plane-signer-ca" and needs the containing
+// bundle regenerated so kube-controller-manager and kube-scheduler can connect
+// using client certs.
+type CABundleController struct {
+	configMapGetter corev1client.ConfigMapsGetter
+	configMapLister corev1listers.ConfigMapLister
+
+	eventRecorder events.Recorder
+
+	cachesToSync []cache.InformerSynced
+
+	// queue only ever has one item, but it has nice error handling backoff/retry semantics
+	queue workqueue.RateLimitingInterface
+}
+
+func NewCABundleController(
+	configMapGetter corev1client.ConfigMapsGetter,
+	kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces,
+	eventRecorder events.Recorder,
+) (*CABundleController, error) {
+	c := &CABundleController{
+		configMapGetter: configMapGetter,
+		configMapLister: kubeInformersForNamespaces.ConfigMapLister(),
+		eventRecorder:   eventRecorder.WithComponentSuffix("manage-client-ca-bundle-recovery-controller"),
+		queue:           workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "CABundleRecoveryController"),
+	}
+
+	handler := cache.ResourceEventHandlerFuncs{
+		AddFunc:    func(obj interface{}) { c.queue.Add(workQueueKey) },
+		UpdateFunc: func(old, new interface{}) { c.queue.Add(workQueueKey) },
+		DeleteFunc: func(obj interface{}) { c.queue.Add(workQueueKey) },
+	}
+
+	// we react to some config changes
+	namespaces := []string{
+		operatorclient.GlobalUserSpecifiedConfigNamespace,
+		operatorclient.GlobalMachineSpecifiedConfigNamespace,
+		operatorclient.OperatorNamespace,
+		operatorclient.TargetNamespace,
+	}
+	for _, namespace := range namespaces {
+		informers := kubeInformersForNamespaces.InformersFor(namespace)
+		informers.Core().V1().ConfigMaps().Informer().AddEventHandler(handler)
+		c.cachesToSync = append(c.cachesToSync, informers.Core().V1().ConfigMaps().Informer().HasSynced)
+	}
+
+	return c, nil
+}
+
+func (c *CABundleController) Run(ctx context.Context) {
+	defer utilruntime.HandleCrash()
+
+	// FIXME: These are missing a wait group to track goroutines and handle graceful termination
+	// (@deads2k wants time to think it through)
+
+	klog.Info("Starting CA bundle controller")
+	defer func() {
+		klog.Info("Shutting down CA bundle controller")
+		c.queue.ShutDown()
+		klog.Info("CA bundle controller shut down")
+	}()
+
+	if !cache.WaitForNamedCacheSync("CABundleController", ctx.Done(), c.cachesToSync...) {
+		return
+	}
+
+	go func() {
+		wait.UntilWithContext(ctx, c.runWorker, time.Second)
+	}()
+
+	<-ctx.Done()
+}
+
+func (c *CABundleController) runWorker(ctx context.Context) {
+	for c.processNextItem(ctx) {
+	}
+}
+
+func (c *CABundleController) processNextItem(ctx context.Context) bool {
+	key, quit := c.queue.Get()
+	if quit {
+		return false
+	}
+	defer c.queue.Done(key)
+
+	err := c.sync(ctx)
+
+	if err == nil {
+		c.queue.Forget(key)
+		return true
+	}
+
+	utilruntime.HandleError(fmt.Errorf("%v failed with : %w", key, err))
+	c.queue.AddRateLimited(key)
+
+	return true
+}
+
+func (c *CABundleController) sync(ctx context.Context) error {
+	// Always start 10 seconds later after a change occurred. Makes us less likely to steal work and logs from the operator.
+	timer := time.NewTimer(10 * time.Second)
+	defer timer.Stop()
+	select {
+	case <-timer.C:
+	case <-ctx.Done():
+		return nil
+	}
+
+	_, changed, err := targetconfigcontroller.ManageClientCABundle(c.configMapLister, c.configMapGetter, c.eventRecorder)
+	if err != nil {
+		return err
+	}
+
+	if changed {
+		klog.V(2).Info("Refreshed client CA bundle.")
+	}
+
+	return nil
+}
--- a/pkg/cmd/certregenerationcontroller/cmd.go
+++ b/pkg/cmd/certregenerationcontroller/cmd.go
@@ -1,0 +1,153 @@
+package certregenerationcontroller
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"k8s.io/client-go/kubernetes"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	configeversionedclient "github.com/openshift/client-go/config/clientset/versioned"
+	configexternalinformers "github.com/openshift/client-go/config/informers/externalversions"
+	"github.com/openshift/library-go/pkg/controller/controllercmd"
+	"github.com/openshift/library-go/pkg/operator/certrotation"
+	"github.com/openshift/library-go/pkg/operator/genericoperatorclient"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/certrotationcontroller"
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/operatorclient"
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/version"
+)
+
+type Options struct {
+	controllerContext *controllercmd.ControllerContext
+
+	TLSServerName string
+}
+
+func NewCertRegenerationControllerCommand(ctx context.Context) *cobra.Command {
+	o := &Options{
+		TLSServerName: "localhost-recovery",
+	}
+
+	cmd := controllercmd.
+		NewControllerCommandConfig("cert-regeneration-controller", version.Get(), func(ctx context.Context, controllerContext *controllercmd.ControllerContext) error {
+			o.controllerContext = controllerContext
+
+			err := o.Validate(ctx)
+			if err != nil {
+				return err
+			}
+
+			err = o.Complete(ctx)
+			if err != nil {
+				return err
+			}
+
+			err = o.Run(ctx)
+			if err != nil {
+				return err
+			}
+
+			return nil
+		}).NewCommandWithContext(ctx)
+	cmd.Use = "cert-regeneration-controller"
+	cmd.Short = "Start the Cluster Certificate Regeneration Controller"
+
+	cmd.PersistentFlags().StringVarP(&o.TLSServerName, "tls-server-name", "", o.TLSServerName, "The SNI hostname to set for the server in kubeconfig")
+
+	return cmd
+}
+
+func (o *Options) Validate(ctx context.Context) error {
+	return nil
+}
+
+func (o *Options) Complete(ctx context.Context) error {
+	return nil
+}
+
+func (o *Options) Run(ctx context.Context) error {
+	if len(o.TLSServerName) != 0 {
+		// TLSServerName chooses the SNI serving endpoint on the apiserver.
+		// Particularly useful when connecting to "localhost" and wanting to choose a special
+		// serving endpoint like "localhost-recovery" that has long-lived serving certs
+		// for localhost connections.
+		o.controllerContext.KubeConfig.TLSClientConfig.ServerName = o.TLSServerName
+		o.controllerContext.ProtoKubeConfig.TLSClientConfig.ServerName = o.TLSServerName
+	}
+
+	kubeClient, err := kubernetes.NewForConfig(o.controllerContext.ProtoKubeConfig)
+	if err != nil {
+		return fmt.Errorf("can't build kubernetes client: %w", err)
+	}
+
+	configClient, err := configeversionedclient.NewForConfig(o.controllerContext.KubeConfig)
+	if err != nil {
+		return fmt.Errorf("failed to create config client: %w", err)
+	}
+
+	configInformers := configexternalinformers.NewSharedInformerFactory(configClient, 10*time.Minute)
+
+	kubeAPIServerInformersForNamespaces := v1helpers.NewKubeInformersForNamespaces(
+		kubeClient,
+		operatorclient.GlobalMachineSpecifiedConfigNamespace,
+		operatorclient.GlobalUserSpecifiedConfigNamespace,
+		operatorclient.OperatorNamespace,
+		operatorclient.TargetNamespace,
+	)
+
+	operatorClient, dynamicInformers, err := genericoperatorclient.NewStaticPodOperatorClient(o.controllerContext.KubeConfig, operatorv1.GroupVersion.WithResource("kubeapiservers"))
+	if err != nil {
+		return err
+	}
+
+	certRotationScale, err := certrotation.GetCertRotationScale(kubeClient, operatorclient.GlobalUserSpecifiedConfigNamespace)
+	if err != nil {
+		return err
+	}
+
+	kubeAPIServerCertRotationController, err := certrotationcontroller.NewCertRotationControllerOnlyWhenExpired(
+		kubeClient,
+		operatorClient,
+		configInformers,
+		kubeAPIServerInformersForNamespaces,
+		o.controllerContext.EventRecorder,
+		certRotationScale,
+	)
+	if err != nil {
+		return err
+	}
+
+	caBundleController, err := NewCABundleController(
+		kubeClient.CoreV1(),
+		kubeAPIServerInformersForNamespaces,
+		o.controllerContext.EventRecorder,
+	)
+	if err != nil {
+		return err
+	}
+
+	// We can't start informers until after the resources have been requested. Now is the time.
+	configInformers.Start(ctx.Done())
+	kubeAPIServerInformersForNamespaces.Start(ctx.Done())
+	dynamicInformers.Start(ctx.Done())
+
+	// FIXME: These are missing a wait group to track goroutines and handle graceful termination
+	// (@deads2k wants time to think it through)
+
+	go func() {
+		kubeAPIServerCertRotationController.Run(ctx, 1)
+	}()
+
+	go func() {
+		caBundleController.Run(ctx)
+	}()
+
+	<-ctx.Done()
+
+	return nil
+}
--- a/pkg/operator/certrotationcontroller/certrotationcontroller.go
+++ b/pkg/operator/certrotationcontroller/certrotationcontroller.go
@@ -51,6 +51,45 @@ func NewCertRotationController(
 	eventRecorder events.Recorder,
 	day time.Duration,
 ) (*CertRotationController, error) {
+	return newCertRotationController(
+		kubeClient,
+		operatorClient,
+		configInformer,
+		kubeInformersForNamespaces,
+		eventRecorder,
+		day,
+		false,
+	)
+}
+
+func NewCertRotationControllerOnlyWhenExpired(
+	kubeClient kubernetes.Interface,
+	operatorClient v1helpers.StaticPodOperatorClient,
+	configInformer configinformers.SharedInformerFactory,
+	kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces,
+	eventRecorder events.Recorder,
+	day time.Duration,
+) (*CertRotationController, error) {
+	return newCertRotationController(
+		kubeClient,
+		operatorClient,
+		configInformer,
+		kubeInformersForNamespaces,
+		eventRecorder,
+		day,
+		true,
+	)
+}
+
+func newCertRotationController(
+	kubeClient kubernetes.Interface,
+	operatorClient v1helpers.StaticPodOperatorClient,
+	configInformer configinformers.SharedInformerFactory,
+	kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces,
+	eventRecorder events.Recorder,
+	day time.Duration,
+	refreshOnlyWhenExpired bool,
+) (*CertRotationController, error) {
 	ret := &CertRotationController{
 		networkLister:        configInformer.Config().V1().Networks().Lister(),
 		infrastructureLister: configInformer.Config().V1().Infrastructures().Lister(),
@@ -83,14 +122,15 @@ func NewCertRotationController(
 	certRotator, err := certrotation.NewCertRotationController(
 		"AggregatorProxyClientCert",
 		certrotation.SigningRotation{
-			Namespace:     operatorclient.OperatorNamespace,
-			Name:          "aggregator-client-signer",
-			Validity:      30 * defaultRotationDay,
-			Refresh:       15 * defaultRotationDay,
-			Informer:      kubeInformersForNamespaces.InformersFor(operatorclient.OperatorNamespace).Core().V1().Secrets(),
-			Lister:        kubeInformersForNamespaces.InformersFor(operatorclient.OperatorNamespace).Core().V1().Secrets().Lister(),
-			Client:        kubeClient.CoreV1(),
-			EventRecorder: eventRecorder,
+			Namespace:              operatorclient.OperatorNamespace,
+			Name:                   "aggregator-client-signer",
+			Validity:               30 * defaultRotationDay,
+			Refresh:                15 * defaultRotationDay,
+			RefreshOnlyWhenExpired: refreshOnlyWhenExpired,
+			Informer:               kubeInformersForNamespaces.InformersFor(operatorclient.OperatorNamespace).Core().V1().Secrets(),
+			Lister:                 kubeInformersForNamespaces.InformersFor(operatorclient.OperatorNamespace).Core().V1().Secrets().Lister(),
+			Client:                 kubeClient.CoreV1(),
+			EventRecorder:          eventRecorder,
 		},
 		certrotation.CABundleRotation{
 			Namespace:     operatorclient.GlobalMachineSpecifiedConfigNamespace,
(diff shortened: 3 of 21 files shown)


pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -82,14 +83,15 @@ func NewCertRotationController(
...
-			Validity:      30 * defaultRotationDay,
-			Refresh:       15 * defaultRotationDay,
-			Informer:      kubeInformersForNamespaces.InformersFor(operatorclient.OperatorNamespace).Core().V1().Secrets(),
-			Lister:        kubeInformersForNamespaces.InformersFor(operatorclient.OperatorNamespace).Core().V1().Secrets().Lister(),
-			Client:        kubeClient.CoreV1(),
-			EventRecorder: eventRecorder,
+			Namespace:              operatorclient.OperatorNamespace,
+			Name:                   "aggregator-client-signer",
+			Validity:               30 * defaultRotationDay,
+			Refresh:                30 * defaultRotationDay,
```
soltysh: Why you're changing the refresh rate?

tnozicka: this was just a test before I changed library-go with an explicit option - need to clean it up now

tnozicka:
/hold
requires https://github.com/openshift/cluster-kube-apiserver-operator/pull/656

pkg/cmd/certregenerationcontroller/certregenerationcontroller.go
```diff
@@ -0,0 +1,208 @@
...
+
+	if !cache.WaitForCacheSync(
+		ctx.Done(),
+		kubeAPIServerInformersForNamespaces.InformersFor(operatorclient.GlobalUserSpecifiedConfigNamespace).Core().V1().ConfigMaps().Informer().HasSynced,
+		kubeAPIServerInformersForNamespaces.InformersFor(operatorclient.OperatorNamespace).Core().V1().ConfigMaps().Informer().HasSynced,
+		kubeAPIServerInformersForNamespaces.InformersFor(operatorclient.TargetNamespace).Core().V1().ConfigMaps().Informer().HasSynced,
+	) {
+		return fmt.Errorf("caches did not sync")
+	}
+	go wait.UntilWithContext(ctx, func(ctx context.Context) {
```
tnozicka: @deads2k How real we want to get w.r.t. to these control loops? Running the full enclosing controller with queue seemed pretty heavy weight and it could steal work from the operator, this steals just a small loop which runs in parallel.

pkg/cmd/certregenerationcontroller/certregenerationcontroller.go
```diff
@@ -0,0 +1,208 @@
...
+	o.configClient, err = configeversionedclient.NewForConfig(o.restConfig)
+	if err != nil {
+		return fmt.Errorf("failed to create config client: %v", err)
+	}
+
+	return nil
+}
+
+func (o *Options) Run() error {
+	ctx, cancel := context.WithCancel(context.Background())
```
tnozicka: @deads2k do we want leader election between recovery controllers or just keeping the premise of acting only if the cluster has broken certs?

pkg/cmd/certregenerationcontroller/certregenerationcontroller.go
deads2k: You should create a different file with a `NewFooController` and a `Run(context) function` for its structural value.  I don't care if it has a queue or not or triggering functions, but the encapsulation of "this is my input, this is what I use for state, this is how I start" makes maintenance and testing easier.

tnozicka: done
(I wired the queue as well when I was making an object for it.)

pkg/cmd/certregenerationcontroller/certregenerationcontroller.go
tnozicka: leader election added

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/operatorclient"
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/targetconfigcontroller"
+)
+
+const workQueueKey = "key"
+
+type CABundleController struct {
```
mfojtik: nit: add godoc pls

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
```diff
@@ -0,0 +1,149 @@
...
+	kubeClient kubernetes.Interface
```
mfojtik: you only need `ConfigMapsGetter` right?

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
```diff
@@ -0,0 +1,149 @@
...
+	c := &CABundleController{
+		kubeClient:      kubeClient,
+		configMapLister: kubeInformersForNamespaces.InformersFor(metav1.NamespaceAll).Core().V1().ConfigMaps().Lister(),
+		eventRecorder:   eventRecorder.WithComponentSuffix("manage-client-ca-bundle-recovery-controller"),
+		queue:           workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "CABundleRecoveryController"),
+	}
+
+	c.cachesToSync = append(c.cachesToSync, kubeInformersForNamespaces.InformersFor(metav1.NamespaceAll).Core().V1().ConfigMaps().Informer().HasSynced)
+
+	handler := cache.ResourceEventHandlerFuncs{
```
mfojtik: nit: we usually extract this into func, but this is fine I guess (consistency)

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
```diff
@@ -0,0 +1,149 @@
...
+		AddFunc:    func(obj interface{}) { c.queue.Add(workQueueKey) },
+		UpdateFunc: func(old, new interface{}) { c.queue.Add(workQueueKey) },
+		DeleteFunc: func(obj interface{}) { c.queue.Add(workQueueKey) },
+	}
+
+	// we react to some config changes
+	namespaces := []string{
```
mfojtik: nit: i would make this global var with godoc on what this list means

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
```diff
@@ -0,0 +1,149 @@
...
+		wg.Wait()
+
+		klog.Info("CA bundle controller shut down")
+	}()
+
+	if !cache.WaitForNamedCacheSync("CABundleController", ctx.Done(), c.cachesToSync...) {
+		return
+	}
+
+	for i := 0; i < workers; i++ {
```
mfojtik: nit: we are never going to support more than 1 worker here right? i would make a check in beginning of this func and panic() and remove this loop.

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func (c *CABundleController) Run(ctx context.Context, workers int) {
+	defer utilruntime.HandleCrash()
+
+	var wg sync.WaitGroup
+	klog.Info("Starting CA bundle controller")
+	defer func() {
+		klog.Info("Shutting down CA bundle controller")
+
```
mfojtik: nit: remove the extra new lines

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
```diff
@@ -0,0 +1,149 @@
...
+	defer c.queue.Done(key)
+
+	err := c.sync(ctx)
+
+	if err == nil {
+		c.queue.Forget(key)
+		return true
+	}
+
+	utilruntime.HandleError(fmt.Errorf("%v failed with : %w", key, err))
```
mfojtik: what is `%w` ?

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
```diff
@@ -0,0 +1,149 @@
...
+		return nil
+	}
+
+	_, changed, err := targetconfigcontroller.ManageClientCABundle(c.configMapLister, c.kubeClient.CoreV1(), c.eventRecorder)
+	if err != nil {
+		return err
+	}
+
+	if changed {
+		klog.Info("Refreshed client CA bundle.")
```
mfojtik: how often will we see this in logs?

pkg/cmd/certregenerationcontroller/certregenerationcontroller.go
```diff
@@ -0,0 +1,302 @@
...
+	"github.com/openshift/library-go/pkg/operator/genericoperatorclient"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/certrotationcontroller"
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/operatorclient"
+)
+
+type Options struct {
+	Workers        int
+	KubeconfigPath string
```
mfojtik: nit: KubeConfigPath

pkg/cmd/certregenerationcontroller/certregenerationcontroller.go
```diff
@@ -0,0 +1,302 @@
...
+	TlsServerName  string
```
mfojtik: nit: TLSServerName

pkg/cmd/certregenerationcontroller/certregenerationcontroller.go
```diff
@@ -0,0 +1,302 @@
...
+
+	LeaderelectionLeaseDuration time.Duration
```
mfojtik: do we need all these? aren't we picking them in some default?

pkg/cmd/certregenerationcontroller/certregenerationcontroller.go
```diff
@@ -0,0 +1,302 @@
...
+			}
+
+			return nil
+		},
+
+		SilenceErrors: true,
+		SilenceUsage:  true,
+	}
+
+	cmd.PersistentFlags().IntVarP(&o.Workers, "workers", "", o.Workers, "Number of workers to run")
```
mfojtik: are we ever going to support more than 1 worker?

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
tnozicka: I think the applyConfig map required live client too

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
tnozicka: I suppose I can remove the param completely, given we have only one key

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
tnozicka: golang 1.13 proper way of wrapping errors

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
tnozicka: hopefully never

pkg/cmd/certregenerationcontroller/certregenerationcontroller.go
tnozicka: no, I think that pattern can be removed for 1 key

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
tnozicka: I wonder about the reasoning for creating separate lambdas for every informer having the same implementation? The function is not used anywhere else and I only see the informer using the callbacks, not modifying it.

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
tnozicka: these namespaces are conventionally defined when setting the handlers https://github.com/openshift/cluster-kube-apiserver-operator/blob/2a44f96452f248818ff4c2467d51eef50fd139f2/pkg/operator/targetconfigcontroller/targetconfigcontroller.go#L86-L89

I'd prefer not to create a layer of indirection obfuscation which namespaces are watched.

pkg/cmd/certregenerationcontroller/certregenerationcontroller.go
tnozicka: is it? https://github.com/openshift/cluster-kube-apiserver-operator/blob/c854bee148b1ec63a3dbfbb87237a832bc820677/vendor/k8s.io/apiserver/pkg/server/options/coreapi.go#L59-L60

pkg/cmd/certregenerationcontroller/certregenerationcontroller.go
tnozicka: they are defaulted in library-go, bundled to some controllers, the default values are matching

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
tnozicka: done

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
tnozicka: done

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
tnozicka: done

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
tnozicka: done

pkg/cmd/certregenerationcontroller/certregenerationcontroller.go
tnozicka: done

pkg/cmd/certregenerationcontroller/certregenerationcontroller.go
tnozicka: done

pkg/cmd/certregenerationcontroller/certregenerationcontroller.go
soltysh: Maybe we could just reuse `configv1.LeaderElection` type and especially `LeaderElectionDefaulting` to ensure they don't diverge.

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/operatorclient"
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/targetconfigcontroller"
+)
+
+const workQueueKey = "key"
+
+// CABundleController composes individual certs into CA bundles
```
deads2k: yeah, you'll need to tell me why we need it in the comments.  I *think* this is because we may affect the client-cert-signers from this process and if we do, we need to make sure that the changes ripple through the configmaps so the cert-syncer observes it so that the kube-apiserver recognizes the new identities.  But if so, you need to say so, because no one besides me and you will ever know that.

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
deads2k: this is event worthy.

pkg/cmd/certregenerationcontroller/certregenerationcontroller.go
```diff
@@ -0,0 +1,295 @@
...
+		},
+
+		SilenceErrors: true,
+		SilenceUsage:  true,
+	}
+
+	cmd.PersistentFlags().StringVarP(&o.KubeconfigPath, "kubeconfig", "", o.KubeconfigPath, "Path to the kubeconfig file")
+	cmd.PersistentFlags().StringVarP(&o.TLSServerName, "tls-server-name", "", o.TLSServerName, "The SNI hostname to set for the server in kubeconfig")
+
+	cmd.PersistentFlags().DurationVar(&o.LeaderelectionLeaseDuration, "leaderelection-lease-duration", o.LeaderelectionLeaseDuration, "LeaseDuration is the duration that non-leader candidates will wait to force acquire leadership.")
```
deads2k: I see no reason to expose this knob.

pkg/cmd/certregenerationcontroller/certregenerationcontroller.go
```diff
@@ -0,0 +1,295 @@
...
+	cmd.PersistentFlags().DurationVar(&o.LeaderelectionRenewDeadline, "leaderelection-renew-deadline", o.LeaderelectionRenewDeadline, "RenewDeadline is the duration that the acting master will retry refreshing leadership before giving up.")
```
deads2k: or any of these leader election ones.

pkg/cmd/certregenerationcontroller/certregenerationcontroller.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+func (o *Options) Validate() error {
+	return nil
+}
+
+func (o *Options) Complete() error {
+	var err error
+
+	if len(o.KubeconfigPath) != 0 {
```
deads2k: there should be a helper this that we use for the operator initialization.

pkg/cmd/certregenerationcontroller/certregenerationcontroller.go
```diff
@@ -0,0 +1,295 @@
...
+		}
+	} else {
+		klog.V(1).Infof("No kubeconfig specified, using InClusterConfig.")
+		o.restConfig, err = restclient.InClusterConfig()
+		if err != nil {
+			return fmt.Errorf("can't create InClusterConfig: %w", err)
+		}
+	}
+
+	if len(o.TLSServerName) != 0 {
```
deads2k: you need to explain this.  not everyone is familiar with SNI and why this matters.

pkg/cmd/certregenerationcontroller/certregenerationcontroller.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	return nil
+}
+
+func (o *Options) Run() error {
+	var leWg sync.WaitGroup
+	var wg sync.WaitGroup
+	leCtx, leCancel := context.WithCancel(context.Background())
+	defer func() {
```
deads2k: Seems fancier than required.  I wasn't expecting leader election at all.

pkg/cmd/certregenerationcontroller/certregenerationcontroller.go
```diff
@@ -0,0 +1,295 @@
...
+
+	hostname, err := os.Hostname()
+	if err != nil {
+		return err
+	}
+	// add a uniquifier so that two processes on the same host don't accidentally both become active
+	id := hostname + "_" + string(uuid.NewUUID())
+	klog.V(4).Infof("Leaderelection ID is %q", id)
+
+	lock := &resourcelock.ConfigMapLock{
```
deads2k: pretty sure have leader election boilerplate in the operator command stuff. Also, I don't think that this should require leader election.  Race and have one winner.

pkg/cmd/certregenerationcontroller/certregenerationcontroller.go
```diff
@@ -0,0 +1,295 @@
...
+	case <-stopCh:
+		klog.Info("Interrupted before leaderelection")
+		return nil
+	}
+
+	configInformers := configexternalinformers.NewSharedInformerFactory(o.configClient, 10*time.Minute)
+
+	kubeAPIServerInformersForNamespaces := v1helpers.NewKubeInformersForNamespaces(
+		o.kubeClient,
+		metav1.NamespaceAll,
```
deads2k: uh, why?

pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -50,6 +51,7 @@ func NewCertRotationController(
 	kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces,
 	eventRecorder events.Recorder,
 	day time.Duration,
+	refreshOnlyWhenExpired bool,
```
deads2k: make two `New***` method variants instead of this bool please. This method can then be internal with a bool arg. That way it's very obvious and intentional.

pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -484,16 +502,51 @@ func (c *CertRotationController) RunOnce() error {
...
 }
 
-func (c *CertRotationController) Run(ctx context.Context, workers int) {
+func (c *CertRotationController) Run(ctx context.Context) {
+	defer utilruntime.HandleCrash()
+
+	var wg sync.WaitGroup
 	klog.Infof("Starting CertRotation")
-	defer klog.Infof("Shutting down CertRotation")
+	defer func() {
```
deads2k: this change looks distinct.  What breaks if we don't have it and more importantly, get it merged separately.  It's buried wher eit is and this construct is clearly new.

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
tnozicka: isn't the existing event enough? https://github.com/openshift/cluster-kube-apiserver-operator/blob/2a44f96452f248818ff4c2467d51eef50fd139f2/pkg/operator/targetconfigcontroller/targetconfigcontroller.go#L348

pkg/cmd/certregenerationcontroller/certregenerationcontroller.go
tnozicka: I'd prefer recovery rotation logs to be consistent, not scattered around between pods, so it is easier to reason about what happened. Migrated to library-go based LE.

pkg/cmd/certregenerationcontroller/certregenerationcontroller.go
tnozicka: https://github.com/openshift/cluster-kube-apiserver-operator/blob/e3719e2d84a96379dd9d70574eebe8ee1456de81/pkg/cmd/certregenerationcontroller/cabundlesyncer.go#L47

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
tnozicka: true, added a comment

pkg/cmd/certregenerationcontroller/certregenerationcontroller.go
tnozicka: done

pkg/cmd/certregenerationcontroller/certregenerationcontroller.go
tnozicka: gone

pkg/operator/certrotationcontroller/certrotationcontroller.go
tnozicka: done

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
```diff
@@ -0,0 +1,148 @@
...
+		c.queue.ShutDown()
+		wg.Wait()
+		klog.Info("CA bundle controller shut down")
+	}()
+
+	if !cache.WaitForNamedCacheSync("CABundleController", ctx.Done(), c.cachesToSync...) {
+		return
+	}
+
+	wg.Add(1)
```
deads2k: remove wait group logic for this PR.

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
deads2k: > isn't the existing event enough?

if it is enough, you wouldn't have this log line.  YOu can reduce to v(2) if you wish.

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
```diff
@@ -0,0 +1,148 @@
...
+	utilruntime.HandleError(fmt.Errorf("%v failed with : %w", key, err))
+	c.queue.AddRateLimited(key)
+
+	return true
+}
+
+func (c *CABundleController) sync(ctx context.Context) error {
+	// Always start 10 seconds later after a change occurred. Makes us less likely to steal work and logs from the operator.
+	select {
+	case <-time.NewTimer(10 * time.Second).C:
```
deads2k: time.After please.  You forgot to stop this one.

pkg/cmd/certregenerationcontroller/cmd.go
```diff
@@ -0,0 +1,169 @@
...
+	configClient, err := configeversionedclient.NewForConfig(o.controllerContext.KubeConfig)
+	if err != nil {
+		return fmt.Errorf("failed to create config client: %w", err)
+	}
+
+	configInformers := configexternalinformers.NewSharedInformerFactory(configClient, 10*time.Minute)
+
+	kubeAPIServerInformersForNamespaces := v1helpers.NewKubeInformersForNamespaces(
+		kubeClient,
+		metav1.NamespaceAll,
```
deads2k: why this one.  Do you have cluster scoped resources?

pkg/cmd/certregenerationcontroller/cmd.go
```diff
@@ -0,0 +1,169 @@
...
+
+	operatorClient, dynamicInformers, err := genericoperatorclient.NewStaticPodOperatorClient(o.controllerContext.KubeConfig, operatorv1.GroupVersion.WithResource("kubeapiservers"))
+	if err != nil {
+		return err
+	}
+
+	eventRecorder := events.NewKubeRecorder(kubeClient.CoreV1().Events(""), "openshift-kube-apiserver-operator", &corev1.ObjectReference{
+		APIVersion: "v1",
+		Kind:       "namespace",
+		Name:       "openshift-kube-apiserver-operator",
```
deads2k: no, put a name unique the this binary in here.

deads2k: > no, put a name unique the this binary in here.

on line 113, where you're passing a name.

pkg/cmd/certregenerationcontroller/cmd.go
```diff
@@ -0,0 +1,169 @@
...
+	if err != nil {
+		return err
+	}
+
+	// We can't start informers until after the resources have been requested. Now is the time.
+	configInformers.Start(ctx.Done())
+	kubeAPIServerInformersForNamespaces.Start(ctx.Done())
+	dynamicInformers.Start(ctx.Done())
+
+	var wg sync.WaitGroup
```
deads2k: no wait group logic yet.

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
tnozicka: `time.After` seems equivalent, I think I should rather stop it

pkg/cmd/certregenerationcontroller/cmd.go
tnozicka: https://github.com/openshift/cluster-kube-apiserver-operator/pull/635#discussion_r369546475

pkg/cmd/certregenerationcontroller/cmd.go
tnozicka: wired it for the builder created recorder

pkg/cmd/certregenerationcontroller/cmd.go
tnozicka: done :(

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func NewCABundleController(
+	configMapGetter corev1client.ConfigMapsGetter,
+	kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces,
+	eventRecorder events.Recorder,
+) (*CABundleController, error) {
+	c := &CABundleController{
+		configMapGetter: configMapGetter,
+		configMapLister: kubeInformersForNamespaces.InformersFor(metav1.NamespaceAll).Core().V1().ConfigMaps().Lister(),
```
deads2k: you can't do namespace all.  That's too broad and will cache far too much.  You can do `kubeInformersForNamespaces.ConfigMapLister()` which will allow you to pull from any namespace you cache.

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
```diff
@@ -0,0 +1,148 @@
...
+
+	utilruntime.HandleError(fmt.Errorf("%v failed with : %w", key, err))
+	c.queue.AddRateLimited(key)
+
+	return true
+}
+
+func (c *CABundleController) sync(ctx context.Context) error {
+	// Always start 10 seconds later after a change occurred. Makes us less likely to steal work and logs from the operator.
+	timer := time.NewTimer(10 * time.Second)
```
deads2k: blech, but ok

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
```diff
@@ -0,0 +1,148 @@
...
+		return nil
+	}
+
+	_, changed, err := targetconfigcontroller.ManageClientCABundle(c.configMapLister, c.configMapGetter, c.eventRecorder)
+	if err != nil {
+		return err
+	}
+
+	if changed {
+		klog.Info("Refreshed client CA bundle.")
```
deads2k: this still needs to be fixed.  if it's top level info worth it should be an event. If it isn't (more likely) make it a V(2).

deads2k:
/lgtm

/hold

you need to fix that global lister before merge, you should update the info statement or someone will do it after.

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
tnozicka: sorry I missed adding v2 yesterday, will update it

pkg/cmd/certregenerationcontroller/cabundlesyncer.go
tnozicka: done

tnozicka:
addressed the global lister and add the v2

pkg/cmd/certregenerationcontroller/cmd.go
```diff
@@ -0,0 +1,153 @@
...
+func NewCertRegenerationControllerCommand(ctx context.Context) *cobra.Command {
+	o := &Options{
+		TLSServerName: "localhost-recovery",
+	}
+
+	cmd := controllercmd.
+		NewControllerCommandConfig("cert-regeneration-controller", version.Get(), func(ctx context.Context, controllerContext *controllercmd.ControllerContext) error {
+			o.controllerContext = controllerContext
+
+			err := o.Validate(ctx)
```
soltysh: nit: but:
```
if err := o.Validate(ctx); err != nil {
    return err
}
```
reads better imho :wink: but I'm the if-:policeman:

soltysh: Nit, but me being if-:policeman: would write it:
```
if err := o.Validate(ctx); err != nil {
    return err
}
````
:wink:

**Comment by tnozicka** [comment]
```

tnozicka:
```
level=error msg="Error: NoSuchBucket: The specified bucket does not exist"
level=error msg="\tstatus code: 404, request id: 70F131ED608434FD, host id: 0aXIZGWRmhJ3z51ZjcbCOPO2ONZ8bpUnL717B+O2/vRpIpYNsB1S13LbD8STe7yQsCExX1Ttzks="
level=error
level=error msg="  on ../tmp/openshift-install-854949451/bootstrap/main.tf line 5, in resource \"aws_s3_bucket\" \"ignition\":"
level=error msg="   5: resource \"aws_s3_bucket\" \"ignition\" {"
level=error
level=error
level=fatal msg="failed to fetch Cluster: failed to generate asset \"Cluster\": failed to create cluster: failed to apply using Terraform"
```
/retest

PR #637, opened 2019-11-01
encryption: switch to kube-storage-version-migrator

sttts:
Requires https://github.com/openshift/cluster-kube-storage-version-migrator-operator to be in the release.


sttts:
/hold 

until https://github.com/openshift/cluster-kube-storage-version-migrator-operator is ready.

PR #648, opened 2019-11-11
Bug 1769011: manifests: enable pruning

sttts:


soltysh:
/hold
the unit test is real

PR #649, opened 2019-11-11
metrics: fix service monitors

mfojtik:
This fixes metrics scraping for KAS (making it secure) and KAO (fixing the CA in operator service monitor).

### DIFF
--- a/manifests/0000_90_kube-apiserver-operator_03_servicemonitor.yaml
+++ b/manifests/0000_90_kube-apiserver-operator_03_servicemonitor.yaml
@@ -15,8 +15,7 @@ spec:
     port: https
     scheme: https
     tlsConfig:
-      # TODO where do you mount the service-ca.crt?
-      insecureSkipVerify: true
+      caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
       serverName: metrics.openshift-kube-apiserver-operator.svc
   jobLabel: component
   namespaceSelector:
@@ -25,21 +24,3 @@ spec:
   selector:
     matchLabels:
       app: kube-apiserver-operator
----
-apiVersion: monitoring.coreos.com/v1
-kind: PrometheusRule
-metadata:
-  name: kube-apiserver-operator
-  namespace: openshift-kube-apiserver-operator
-spec:
-  groups:
-  - name: cluster-version
-    rules:
-    - alert: TechPreviewNoUpgrade
-      annotations:
-        message: Cluster has enabled tech preview features that will prevent upgrades.
-      expr: |
-        cluster_feature_set{name!="", namespace="openshift-kube-apiserver-operator"} == 0
-      for: 10m
-      labels:
-        severity: warning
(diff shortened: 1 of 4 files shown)


manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -35,34 +35,3 @@ spec:
     matchLabels:
       component: apiserver
       provider: kubernetes
----
```
mfojtik: @sttts @tnozicka moving this to its own file because i'm not fan of lists

manifests/0000_90_kube-apiserver-operator_03_servicemonitor.yaml
```diff
@@ -15,8 +15,7 @@ spec:
     port: https
     scheme: https
     tlsConfig:
-      # TODO where do you mount the service-ca.crt?
-      insecureSkipVerify: true
+      caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
```
mfojtik: @sttts this seems to be monitoring team magic we have to follow

mfojtik:
@soltysh @ingvagabund we need similar for KCO and scheduler operator (they are insecure now)

mfojtik:
@sttts @sanchezl somebody has to make e2e test, so we won't break again... i suggest using some stupid `oc get -raw /metrics` command or `oc adm inspect` and grepping to check our metrics are registered....

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
sttts: Always put moves in their own commits. They are noise.

ingvagabund:
@mfojtik the same test is failing in both this PR and https://github.com/openshift/cluster-kube-scheduler-operator/pull/183:
```
fail [github.com/openshift/origin/test/extended/prometheus/prometheus_builds.go:134]: Expected
    &lt;map[string]error | len:1>: {
        "ALERTS{alertname!~\"Watchdog|ClusterMachineApproverDown|UsingDeprecatedAPIExtensionsV1Beta1\",alertstate=\"firing\"} >= 1": {
            s: "promQL query: ALERTS{alertname!~\"Watchdog|ClusterMachineApproverDown|UsingDeprecatedAPIExtensionsV1Beta1\",alertstate=\"firing\"} >= 1 had reported incorrect results: ALERTS{alertname=\"ClusterAutoscalerOperatorDown\", alertstate=\"firing\", severity=\"critical\"} => 1 @[1573556100.924]",
        },
    }
to be empty
```

The test does not expect `ClusterAutoscalerOperatorDown` to fire. Not related to this PR.

ingvagabund:
Cloud team is aware of the issue.

ingvagabund:
Produced by https://github.com/openshift/cluster-autoscaler-operator/pull/119/files#diff-f95096b7967c99b51373b423a243286c

manifests/0000_90_kube-apiserver-operator_03_servicemonitor.yaml
brancz: We're working on allowing people to do secret/configmap key selectors instead, we'll let people know once it's available and tested. This is an unfortunate artifact of early poor API decisions ...

PR #650, opened 2019-11-11
metrics: add operator metrics endpoint

mfojtik:


pkg/operator/starter.go
```diff
@@ -187,7 +189,17 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
...
+	// TODO: Move this to library-go and expose only registry
+	r := metrics.NewKubeRegistry()
+	// register process start time
+	if err := metrics.RegisterProcessStartTime(r.RawRegister); err != nil {
+		return err
+	}
+	// register config metrics
+	configmetrics.NewRegistry(r).Register(configInformers)
+	// add metrics endpoint
+	ctx.Server.Handler.GoRestfulContainer.Handle("/metrics/operator/v1alpha1", promhttp.HandlerFor(r, promhttp.HandlerOpts{ErrorHandling: promhttp.ContinueOnError}))
```
mfojtik: note, @smarterclayton does not like this and we will end up with 2 endpoints and our service monitor will suck because we will need relabeling :-(

pkg/operator/configmetrics/configmetrics.go
```diff
@@ -12,39 +15,45 @@ import (
...
+func (m *configMetrics) Create(version *semver.Version) bool {
+	return true
 }
 
 // Describe reports the metadata for metrics to the prometheus collector.
 func (m *configMetrics) Describe(ch chan<- *prometheus.Desc) {
-	ch <- m.cloudProvider.WithLabelValues("", "").Desc()
-	ch <- m.featureSet.WithLabelValues("").Desc()
+	ch <- m.cloudProvider.WithLabelValues("", "").(metrics.Metric).Desc()
+	ch <- m.featureSet.WithLabelValues("").(metrics.Metric).Desc()
```
sttts: this type conversion is really needed? How ugly.

mfojtik: @sttts actually it seems it is

mfojtik: @sttts if we use kubemetrics... i switched it back to prometheus.

PR #653, opened 2019-11-12
If the cluster is single-stack IPv6, bind to :: rather than 0.0.0.0.

danwinship:
This updates our kube-apiserver config to specify a bind address of `::` rather than `0.0.0.0` when running in a single-stack IPv6 cluster. Combined with https://github.com/kubernetes/kubernetes/pull/84727 this will then cause it to advertise an IPv6 address for `kubernetes.default`.

Originally I modified both `bindata/bootkube/config/bootstrap-config-overrides.yaml` and `bindata/bootkube/config/config-overrides.yaml`, but the changes to the latter had no effect; I'm guessing that the "real" (post-bootstrap) config is rendered without access to the cluster manifets and so doesn't pass `--cluster-config-file`, so the whole `if len(renderConfig.ClusterCIDR) > 0` section just gets skipped. So instead I added the code to the config observer.

There is, at the moment, no simple way to test this, but this is going to be one of the PRs that needs to merge before we can even have an `e2e-aws-ipv6`... (see https://github.com/openshift/enhancements/pull/87, https://github.com/openshift/installer/pull/2555)

### DIFF
--- a/pkg/cmd/render/render.go
+++ b/pkg/cmd/render/render.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"path/filepath"
 
 	"github.com/ghodss/yaml"
@@ -130,6 +131,12 @@ type TemplateData struct {
 
 	// ServiceClusterIPRange is the IP range for service IPs.
 	ServiceCIDR []string
+
+	// BindAddress is the IP address and port to bind to
+	BindAddress string
+
+	// BindNetwork is the network (tcp4 or tcp6) to bind to
+	BindNetwork string
 }
 
 // Run contains the logic of the render command.
@@ -138,6 +145,8 @@ func (r *renderOpts) Run() error {
 		LockHostPath:   r.lockHostPath,
 		EtcdServerURLs: r.etcdServerURLs,
 		EtcdServingCA:  r.etcdServingCA,
+		BindAddress:    "0.0.0.0:6443",
+		BindNetwork:    "tcp4",
 	}
 	if len(r.clusterConfigFile) > 0 {
 		clusterConfigFileData, err := ioutil.ReadFile(r.clusterConfigFile)
@@ -146,6 +155,24 @@ func (r *renderOpts) Run() error {
 		}
 		if err = discoverCIDRs(clusterConfigFileData, &renderConfig); err != nil {
 			return fmt.Errorf("unable to parse restricted CIDRs from config %q: %v", r.clusterConfigFile, err)
+		}
+	}
+	if len(renderConfig.ClusterCIDR) > 0 {
+		anyIPv4 := false
+		for _, cidr := range renderConfig.ClusterCIDR {
+			cidrBaseIP, _, err := net.ParseCIDR(cidr)
+			if err != nil {
+				return fmt.Errorf("invalid cluster CIDR %q: %v", cidr, err)
+			}
+			if cidrBaseIP.To4() != nil {
+				anyIPv4 = true
+				break
+			}
+		}
+		if !anyIPv4 {
+			// Single-stack IPv6 cluster, so listen on IPv6 not IPv4.
+			renderConfig.BindAddress = "[::]:6443"
+			renderConfig.BindNetwork = "tcp6"
 		}
 	}
 	if err := r.manifest.ApplyTo(&renderConfig.ManifestConfig); err != nil {
--- a/pkg/operator/configobservation/network/observe_network.go
+++ b/pkg/operator/configobservation/network/observe_network.go
@@ -1,6 +1,8 @@
 package network
 
 import (
+	"net"
+
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
 	"github.com/openshift/library-go/pkg/operator/configobserver"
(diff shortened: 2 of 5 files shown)


sttts:
Can't and shouldn't we just always listen on `::`?

pkg/operator/configobservation/network/observe_network.go
```diff
@@ -73,30 +75,56 @@ func ObserveRestrictedCIDRs(genericListers configobserver.Listers, recorder even
...
+	bindNetworkConfigPath := []string{"servingInfo", "bindNetwork"}
 
 	errs := []error{}
+
 	serviceCIDR, err := network.GetServiceCIDR(listers.NetworkLister, recorder)
 	if err != nil {
 		errs = append(errs, err)
+
+		// Pass through existingConfig
+		prev, ok, err2 := unstructured.NestedString(existingConfig, servicesSubnetConfigPath...)
```
sttts: why err2? Use err.

sttts:
Some nits. Looks good otherwise.

pkg/cmd/render/render.go
```diff
@@ -130,6 +131,12 @@ type TemplateData struct {
 
 	// ServiceClusterIPRange is the IP range for service IPs.
 	ServiceCIDR []string
+
+	// BindAddress is the IP address and port to bind to
+	BindAddress string
+
+	// BindNetwork is the network (tcp4 or tcp6) to bind to
+	BindNetwork string
```
squeed: Arguably this should be called BindFamily (tiny nit)

pkg/operator/configobservation/network/observe_network.go
squeed: stomping on the `err` above seems like a recipe for, well, errors.

pkg/cmd/render/render.go
squeed: Ah, nevermind, that's what it's called upstream. Ohwell.

pkg/cmd/render/render.go
```diff
@@ -148,6 +157,27 @@ func (r *renderOpts) Run() error {
...
+			if err != nil {
+				return fmt.Errorf("invalid cluster CIDR %q: %v", cidr, err)
+			}
+			if cidrBaseIP.To4() != nil {
+				anyIPv4 = true
+				break
+			}
+		}
+		if !anyIPv4 {
+			// Single-stack IPv6 cluster, so listen on IPv6 not IPv4. IPv6
```
squeed: Wouldn't it just be easier to put the quotes in the template? Extremely minor nit.

squeed:
:+1: lgtm

pkg/operator/configobservation/network/observe_network.go
sttts: err is the last command's error by convention. If you want to preserve one, rename it.</wednesday-bike-shedding>

danwinship:
> Can't and shouldn't we just always listen on `::`?

1. If someone disables IPv6 in the kernel then there won't be a `::`. (Though maybe that would cause other problems anyway?)
2. kubernetes/kubernetes#84727 makes apiserver treat `--bind-address ::` as additionally meaning "advertise an IPv6 address in `kubernetes.default.svc` by default" so we'd have to change that.

pkg/cmd/render/render.go
danwinship: ah, indeed. They're not required for IPv4 but of course they're not forbidden either

pkg/operator/configobservation/network/observe_network.go
```diff
@@ -73,30 +75,56 @@ func ObserveRestrictedCIDRs(genericListers configobserver.Listers, recorder even
...
+		}
+		prev, ok, err = unstructured.NestedString(existingConfig, bindAddressConfigPath...)
+		if ok && err == nil {
+			unstructured.SetNestedField(out, prev, bindAddressConfigPath...)
+		}
+		prev, ok, err = unstructured.NestedString(existingConfig, bindNetworkConfigPath...)
+		if ok && err == nil {
+			unstructured.SetNestedField(out, prev, bindNetworkConfigPath...)
+		}
+		return out, errs
```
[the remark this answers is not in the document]
soltysh: Alternatively fail fast, when non-nil error happens, to match what we used to do before.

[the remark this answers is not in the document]
danwinship: ok, I made the code more like `pkg/operator/configobservation/apiserver` (including copying a helper function from there)

danwinship:
> Can't and shouldn't we just always listen on `::`?

OK, so it turns out that golang will actually give you a dual-stack socket if you say `net.Listen("tcp", "0.0.0.0:6443")`, but since we were passing `"tcp4"` rather than `"tcp"` we got single-stack. So if we changed `BindNetwork`, we could get dual-stack behavior.

However, this is going to need to go into 4.3, and making hostNetwork servers in single-stack IPv4 clusters suddenly be dual-stack is scary as a last-minute (or z-stream) change. So I think for now we want to stick with this approach, but probably later in 4.4 we'll want to switch to `BindNetwork: tcp`. (For pod-network servers though, we can do the dual stack thing.)

danwinship:
/retest
@soltysh updated for your comments

mfojtik:
Per slack discussion yesterday, I understand that this is something we want to do for 4.3 (single-stack), but we don't want to do for 4.4 (dual-stack). But because of our "process", we have to merge this into master and then do a backport of this into 4.3.z, correct?

I'm not sure how this will work if somebody take 4.3.0 and install it on IPv6 cluster (which I suppose won't work) and then upgrade to whatever z-stream level has this fix. Upgrade does not seems to magically make his cluster work on IPv6 as this is only dealing with the z-stream.

Should this be a hard 4.3.0 blocker? /cc @smarterclayton @eparis

danwinship:
> I'm not sure how this will work if somebody take 4.3.0 and install it on IPv6 cluster (which I suppose won't work)

The 4.3.0 installer won't allow you to specify an IPv6 `machineCIDR` / `clusterNetwork` / `serviceNetwork`.

mfojtik:
@danwinship so one need to install 4.3.0 on ipv4, then get the right z-stream and switch to ipv6?

danwinship:
No, if you need IPv6, you have to wait for the 4.3.z release that supports IPv6, and then install that directly. You'd never run 4.3.0 on that cluster.

danwinship:
We are not worrying about supporting 4.3.0-with-IPv4 to 4.3.z-with-IPv6 upgrades

mfojtik:
@danwinship gotcha

PR #654, opened 2019-11-13
load test for migration

p0lyn0mial:


test/e2e-encryption/encryption_perf_test.go
```diff
@@ -0,0 +1,355 @@
...
+		end := time.Now()
+		t.Logf("Populating etcd took %v", end.Sub(start))
+	}()
+
+	r := newRunner()
+	loaderFun := repeat(1, true,
+		createNamespace,
+		waitUntilNamespaceActive,
+		// TODO: consider adding ConfigMaps
+		repeatParallel(9010, 50, false, createSecret, reportSecret))
```
p0lyn0mial: this is probably too much given the current speed of migration.

sttts: how much time do we permit for e2e?

p0lyn0mial: 4h for `e2e-encryption` tests and 1h for `standard-e2e` tests.

Do we want to have a separate job just for the load tests?

p0lyn0mial:
This pull is ready for review, the test fails because I wanted to print the logs.

I think that we will create a separate CI job just for the performance test. That would give us isolation, speed up tests execution (we wouldn't have to wait for other tests to finish) and some room for the potential future changes.


the last test run gave me:
```
Verified 27995 Secrets 
Verified 15299 ConfigMaps 
migration took 24m3.550645976s 
```

pkg/operator/starter.go
```diff
@@ -42,7 +42,10 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 	if err != nil {
 		return err
 	}
-	dynamicClient, err := dynamic.NewForConfig(ctx.KubeConfig)
+	dynamicClientConfig := dynamic.ConfigFor(ctx.KubeConfig)
+	dynamicClientConfig.Burst = 40
```
p0lyn0mial: we need to experiment and find good numbers for `QPS/Burst`

sttts: Please rename to `migrationClientConfig` and verify that it is only used for migration

sttts: Are those the numbers upstream migrator uses as well?

p0lyn0mial:
will be merge as part of https://github.com/openshift/cluster-kube-apiserver-operator/pull/670

PR #638, opened 2019-11-04
Add 10y localhost-recovery serving cert

sttts:
This will be used by the cert recovery sidecar.

PR #663, opened 2019-11-22
Wire certsyncer to use localhost-recovery-client SA

tnozicka:
https://github.com/openshift/cluster-kube-apiserver-operator/pull/638 wired long lived serving certs for _localhost-recovery_ SNI endpoint and partially wired certsyncer to use dedicated _localhost-recovery_ endpoint in https://github.com/openshift/cluster-kube-apiserver-operator/pull/638/commits/2622adeb0da9a7fa91d24baaf1fd3d3d14fd6b89#diff-79f70c8858100d23aa0da941b6136509R102 

We also need certsyncer to use long lived client credentials so as part of automatic recovery flow it can download new certs even if the regular trust chain is expired.

@deads2k has pointed out that using SA is strictly better then long lived client certs in this case as we always have it available (past bootstrap) and it is revokable.

@deads2k I don't think we can make the SA valid only for a particular SNI endpoint like localhost-recovery, can we? But you have mentioned previously that SA is at least revocable.

/cc @deads2k @sttts @mfojtik @soltysh


manifests/0000_20_kube-apiserver-operator_04_clusterrolebinding.yaml
```diff
@@ -9,3 +9,6 @@ subjects:
 - kind: ServiceAccount
   namespace: openshift-kube-apiserver-operator
   name: kube-apiserver-operator
+- kind: ServiceAccount
+  name: localhost-recovery-client
```
tnozicka: @deads2k the client certs were issued for group "system:masters", so binding it for cluster-admin seems equivalent, but I'd prefer if you sign it :)

tnozicka:
/hold
(to manually check cert-syncer logs before we let it merge since it is not tested)

tnozicka:
operator hit rate limits
/retest

bindata/v4.1.0/kube-apiserver/localhost-recovery-token.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: localhost-recovery-client-token
+  annotations:
+    kubernetes.io/service-account.name: localhost-recovery-client
+type: kubernetes.io/service-account-token
```
sttts: any reason this is not a simple CVO manifest?

bindata/v4.1.0/kube-apiserver/localhost-recovery-sa.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: localhost-recovery-client
```
sttts: any reason this is not a simple CVO manifest?

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -383,6 +387,68 @@ func ensureKubeAPIServerTrustedCA(client coreclientv1.CoreV1Interface, recorder
...
+			return err
+		}
+
+		serviceAccount, err = saClient.Create(requiredSA)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Create custom secret to be populated with SA token so we have a stable name
```
sttts: don't understand that comment

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -383,6 +387,68 @@ func ensureKubeAPIServerTrustedCA(client coreclientv1.CoreV1Interface, recorder
...
+		}
+
+		token, err = secretsClient.Create(requiredToken)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Token creation / injection for a SA is asynchronous.
+	// We will report and error if it's missing, go degraded and get re-queued when the SA token is updated.
```
sttts: when do we go degraded? On operator startup?

Again my upper questoin: why not move the token secret into the manifest dir. Then there is a time gap anyway between creation and code here.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -383,6 +387,68 @@ func ensureKubeAPIServerTrustedCA(client coreclientv1.CoreV1Interface, recorder
...
+		return fmt.Errorf("secret %s/%s hasn't been populated with SA token yet: missing SA UID", token.Namespace, token.Name)
+	}
+
+	if uid != string(serviceAccount.UID) {
+		return fmt.Errorf("secret %s/%s hasn't been populated with current SA token yet: SA UID mismatch", token.Namespace, token.Name)
+	}
+
+	if len(token.Data) == 0 {
+		return fmt.Errorf("secret %s/%s hasn't been populated with current SA data yet", token.Namespace, token.Name)
+	}
```
sttts: unnecessary check . The if below does the same.

pkg/operator/configobservation/apiserver/observe_apiserver.go
```diff
@@ -118,8 +118,8 @@ func observeNamedCertificates(apiServer *configv1.APIServer, recorder events.Rec
 		"certFile": "/etc/kubernetes/static-pod-certs/secrets/internal-loadbalancer-serving-certkey/tls.crt",
 		"keyFile":  "/etc/kubernetes/static-pod-certs/secrets/internal-loadbalancer-serving-certkey/tls.key"})
 	observedNamedCertificates = append(observedNamedCertificates, map[string]interface{}{
-		"certFile": "/etc/kubernetes/static-pod-certs/secrets/localhost-recovery-serving-certkey/tls.crt",
-		"keyFile":  "/etc/kubernetes/static-pod-certs/secrets/localhost-recovery-serving-certkey/tls.key"})
+		"certFile": "/etc/kubernetes/static-pod-resources/secrets/localhost-recovery-serving-certkey/tls.crt",
+		"keyFile":  "/etc/kubernetes/static-pod-resources/secrets/localhost-recovery-serving-certkey/tls.key"})
```
sttts: was the path wrong before?

pkg/operator/starter.go
```diff
@@ -227,11 +227,12 @@ var RevisionConfigMaps = []revision.RevisionResource{
 var RevisionSecrets = []revision.RevisionResource{
 	// these need to removed, but if we remove them now, the cluster will die because we don't reload them yet
 	{Name: "etcd-client"},
-	// this is needed so that the cert syncer itself can request certs.  It uses localhost
-	{Name: "kube-apiserver-cert-syncer-client-cert-key"},
 	{Name: "kubelet-client"},
 	// etcd encryption
 	{Name: "encryption-config", Optional: true},
+
+	{Name: "localhost-recovery-serving-certkey"},
```
sttts: why is this moved?

tnozicka: I don't think certsyncer kubeconfig is wired to be live reloaded, nor recovery so it needs to be revisioned.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
tnozicka: token secrets get random name, creating our own makes sure we have a stable name

sttts: write that

pkg/operator/starter.go
sttts: write that

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
tnozicka: checking if map is empty or if map has key "token" are not equivalent checks, but looks like I forgot to reword the error message when copy pasting

bindata/v4.1.0/kube-apiserver/localhost-recovery-token.yaml
tnozicka: It is operand's config, not operator's - I think these go here

If we ever need to change it or extend it with any additional data, we should be controlling that resource, not CVO. If we want to migrate away from it, we should be controlling it.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
tnozicka: That would just change the timing on the race (nor I think moving it there is correct), we would still have to handle the case.

Sounds like currently we go degraded immediately in all the other cases and this goes in line with that pattern, although we should probably adjust it in the future. That would likely be in a common place, so I don't think this PR would be blocked by that.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
tnozicka: not sure I want to be the one mirroring kube docs but ok

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
tnozicka: fixed the error message

tnozicka: (also this avoids indexing nil map)

pkg/operator/starter.go
tnozicka: done

tnozicka:
cert-syncer seems to be running fine in https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-kube-apiserver-operator/663/pull-ci-openshift-cluster-kube-apiserver-operator-master-e2e-aws/3077/artifacts/e2e-aws/must-gather/registry-svc-ci-openshift-org-ci-op-ff3sq68p-stable-sha256-51e78dd04992c2407dafe73e8adc58d3d9a4e7a40d24cefb12d8abebb04f9cf1/namespaces/openshift-kube-apiserver/pods/kube-apiserver-ip-10-0-131-207.ec2.internal/kube-apiserver-cert-syncer-7/kube-apiserver-cert-syncer-7/logs/current.log
/hold cancel

bindata/v4.1.0/kube-apiserver/localhost-recovery-token.yaml
sttts: Don't we have operand config as manifest too in many places? I don't think complexity for future potential needs makes sense here.

bindata/v4.1.0/kube-apiserver/localhost-recovery-sa.yaml
deads2k: > any reason this is not a simple CVO manifest?

CVO cannot remove resources.  It's merge logic is suspect (frequent failures from an old copy of library-go).  It has a 10 minute sync time.  And it doesn't logically manage this resource.

I would recommend using the CVO as little as possible.

manifests/0000_20_kube-apiserver-operator_04_clusterrolebinding.yaml
deads2k: > @deads2k the client certs were issued for group "system:masters", so binding it for cluster-admin seems equivalent, but I'd prefer if you sign it :)

It's not the same.  If you lose an etcd connection, `system:masters` will continue to work, this will not.

pkg/operator/starter.go
```diff
@@ -227,11 +227,13 @@ var RevisionConfigMaps = []revision.RevisionResource{
...
 	// these need to removed, but if we remove them now, the cluster will die because we don't reload them yet
 	{Name: "etcd-client"},
-	// this is needed so that the cert syncer itself can request certs.  It uses localhost
-	{Name: "kube-apiserver-cert-syncer-client-cert-key"},
 	{Name: "kubelet-client"},
 	// etcd encryption
 	{Name: "encryption-config", Optional: true},
+
+	// this needs to be revisioned as certsyncer's kubeconfig isn't wired to be live reloaded, nor will be autorecovery
+	{Name: "localhost-recovery-serving-certkey"},
```
deads2k: This is named like a serving cert.  If it's used for serving by the kube-apiserver, it should be live reloaded.

manifests/0000_20_kube-apiserver-operator_04_clusterrolebinding.yaml
deads2k: This binding logically belongs to the operator, not the CVO. Please move it so we can successfully change it in the future.

pkg/operator/starter.go
tnozicka: it is also used by the clients that don't support live reloading (cert syncer and regeneration controller)

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -383,6 +387,69 @@ func ensureKubeAPIServerTrustedCA(client coreclientv1.CoreV1Interface, recorder
...
+	requiredToken := resourceread.ReadSecretV1OrDie(v410_00_assets.MustAsset("v4.1.0/kube-apiserver/localhost-recovery-token.yaml"))
+
+	saClient := client.ServiceAccounts(operatorclient.TargetNamespace)
+	serviceAccount, err := saClient.Get(requiredSA.Name, metav1.GetOptions{})
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return err
+		}
+
+		serviceAccount, err = saClient.Create(requiredSA)
```
deads2k: This should be in the standard create block above

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -383,6 +387,69 @@ func ensureKubeAPIServerTrustedCA(client coreclientv1.CoreV1Interface, recorder
...
+	// The default token secrets get random names so we will create a custom secret
+	// to be populated with SA token so we have a stable name.
+	secretsClient := client.Secrets(operatorclient.TargetNamespace)
+	token, err := secretsClient.Get(requiredToken.Name, metav1.GetOptions{})
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return err
+		}
+
+		token, err = secretsClient.Create(requiredToken)
```
deads2k: this should also be in the standard create block above.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
tnozicka: done

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
tnozicka: done

manifests/0000_20_kube-apiserver-operator_04_clusterrolebinding.yaml
tnozicka: you are right. I've moved it to the operator's bindata

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
tnozicka: the reconciliation logic above, where I moved it, is nuking the injected token

tnozicka: It seem expected given how it is written
https://github.com/openshift/cluster-kube-apiserver-operator/blob/00bc23d7d3dcd0551c78df93249084584173ee45/vendor/github.com/openshift/library-go/pkg/operator/resource/resourceapply/core.go#L230-L239

tnozicka: do we want to compare only existing keys in .Data fixture and leave the others be?

tnozicka: but then you can't delete keys if you change the fixture...

tnozicka:
picked up https://github.com/openshift/library-go/pull/645

vendor/github.com/openshift/library-go/pkg/operator/resource/resourceapply/core.go
```diff
@@ -222,23 +222,38 @@ func ApplySecret(client coreclientv1.SecretsGetter, recorder events.Recorder, re
...
 	existingCopy := existing.DeepCopy()
 
-	resourcemerge.EnsureObjectMeta(modified, &existingCopy.ObjectMeta, required.ObjectMeta)
+	resourcemerge.EnsureObjectMeta(resourcemerge.BoolPtr(false), &existingCopy.ObjectMeta, required.ObjectMeta)
 
-	dataSame := equality.Semantic.DeepEqual(existingCopy.Data, required.Data)
-	if dataSame && !*modified {
-		return existingCopy, false, nil
+	switch required.Type {
+	case corev1.SecretTypeServiceAccountToken:
```
tnozicka: the fix

pkg/operator/starter.go
sttts: is it only used by the sidecars such that live-reload is not needed?

tnozicka: I'd phrase it so that the side cars can't handle live-reload so even the apiserver can't use live-reload even if it would support it as they share the file on disk.

sttts: There is no other actor who will fall over a missing reload, right?

tnozicka: the won't be missing reload, it will trigger new rollout

pkg/operator/configobservation/apiserver/observe_apiserver.go
sttts: all in static-pod-certs are live reloaded.

pkg/operator/starter.go
sttts: ok

bindata/v4.1.0/kube-apiserver/kubeconfig-cm.yaml
```diff
@@ -8,7 +8,7 @@ data:
     apiVersion: v1
     clusters:
       - cluster:
-          certificate-authority: /etc/kubernetes/static-pod-resources/configmaps/kube-apiserver-server-ca/ca-bundle.crt
+          certificate-authority: /etc/kubernetes/static-pod-resources/secrets/localhost-recovery-client-token/ca.crt
```
sttts, line 11: this CA must include the localhost-recovery serving cert CA. It's one created by the operator when the localhost-recovery serving cert secret is created. How do we make sure that this is known to kcm when creating the localhost-recovery service account. This smells like a race.

sttts: I wonder how this can ever work:
- bootstrap control plane is shut down if all control plane pods are up and ready on the master nodes
- the master kcm only takes over from bootstrap kcm when the latter is killed
- the localhost-recovery CA only shows up in the kube-apiserver-server-ca after the bootstrap has been started
- how can the localhost-recovery service-account ever include the localhost-recovery CA as that service-account is need to start a master kube-apiserver.

bindata/v4.1.0/kube-apiserver/kubeconfig-cm.yaml
```diff
@@ -8,7 +8,7 @@ data:
     apiVersion: v1
     clusters:
       - cluster:
-          certificate-authority: /etc/kubernetes/static-pod-resources/configmaps/kube-apiserver-server-ca/ca-bundle.crt
```
sttts, line 11: wasn't this correct before the change?

bindata/v4.1.0/kube-apiserver/kubeconfig-cm.yaml
sttts: The localhost-recovery-client-token is revisioned. So this might work after bootstrapping is over. But the bootstrapping step is still unclear to me.

pkg/operator/starter.go
```diff
@@ -243,11 +243,13 @@ var RevisionConfigMaps = []revision.RevisionResource{
...
+	{Name: "localhost-recovery-client-token"},
```
sttts, line 252: why is this revisioned? It never changes.

PR #664, opened 2019-11-22
Fix tests

tnozicka:
Encryption tests have side effects. It left operator in a state where it was rolling out several revisions for other tests and because it was a separate package it was running in parallel with other e2e.
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-kube-apiserver-operator/662/pull-ci-openshift-cluster-kube-apiserver-operator-master-e2e-aws-operator/1510/artifacts/e2e-aws-operator/must-gather/registry-svc-ci-openshift-org-ci-op-lwp8pzd2-stable-sha256-64c63eedf863406fbc6c7515026f909a7221472cf70283708fb7010dd5e6139e/namespaces/openshift-kube-apiserver-operator/pods/kube-apiserver-operator-5bb6778647-dt22w/kube-apiserver-operator/kube-apiserver-operator/logs/current.log

Tests need to run serially and clean after itself. Make sure operator is in stable state before finishing so it doesn't affect other tests. If not a separate suite is required.

/hold
for corresponding CI change

/cc @p0lyn0mial

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,13 +45,12 @@ $(call add-bindata,v4.1.0,./bindata/v4.1.0/...,bindata,v410_00_assets,pkg/operat
 $(call add-crd-gen,manifests,$(CRD_APIS),./manifests,./manifests)
 
 # these are extremely slow serial e2e encryption tests that modify the cluster's global state 
-.PHONY: test-e2e-encryption
 test-e2e-encryption: GO_TEST_PACKAGES :=./test/e2e-encryption/...
 test-e2e-encryption: GO_TEST_FLAGS += -v
 test-e2e-encryption: GO_TEST_FLAGS += -timeout 4h
-test-e2e-encryption: GO_TEST_FLAGS += -p 1
 test-e2e-encryption: GO_TEST_FLAGS += -parallel 1
 test-e2e-encryption: test-unit
+.PHONY: test-e2e-encryption
 
 update-codegen: update-codegen-crds
 .PHONY: update-codegen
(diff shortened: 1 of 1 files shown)


tnozicka:
CI PR https://github.com/openshift/release/pull/6044

sttts:
It's unclear to me what the goal here is? We will have two sets of encryption tests: short and long?

tnozicka:
> It's unclear to me what the goal here is? We will have two sets of encryption tests: short and long?

currently, one of the encryption test is hooked into the regular suite, but has side effects, leaves the operator in rolling state for the other test and because its a different package it runs in parallel with the other tests. This move this one test out of e2e suite into it's own.

p0lyn0mial:
Tomas noticed that there are multiple tests in `e2e-aws-operator` that require the API server config change. Each config change rolls out a new revision and potentially can influence other tests. To mitigate that we would like to uncouple them.

p0lyn0mial:
so we will have the following jobs:
- `e2e-aws-operator-encryption-serial-short` (always runs the basic e2e encryption scenario)
- `e2e-encryption-full`- (previously `e2e-aws-operator-encryption`)
-  `e2e-aws-operator-serial` (previously `e2e-aws-operator `) but after the changes it will run the tests in serial.

question is why we don't move the basic e2e scenario to  `e2e-aws-operator-serial` and remove `e2e-aws-operator-encryption-serial-short` ?

tnozicka:
/close
let's start with just making serial e2e tests to be really serial

tnozicka:
^ https://github.com/openshift/cluster-kube-apiserver-operator/pull/666

PR #666, opened 2019-11-25
Fix e2e to run serially

tnozicka:
Operator e2e have side effects and need to run serially. Golang does that by default for unit tests in the same package but with addition of etcd tests in a different package, the e2e now collide and flake.

/cc @p0lyn0mial @sttts

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,13 +45,12 @@ $(call add-bindata,v4.1.0,./bindata/v4.1.0/...,bindata,v410_00_assets,pkg/operat
 $(call add-crd-gen,manifests,$(CRD_APIS),./manifests,./manifests)
 
 # these are extremely slow serial e2e encryption tests that modify the cluster's global state 
-.PHONY: test-e2e-encryption
 test-e2e-encryption: GO_TEST_PACKAGES :=./test/e2e-encryption/...
 test-e2e-encryption: GO_TEST_FLAGS += -v
 test-e2e-encryption: GO_TEST_FLAGS += -timeout 4h
 test-e2e-encryption: GO_TEST_FLAGS += -p 1
-test-e2e-encryption: GO_TEST_FLAGS += -parallel 1
 test-e2e-encryption: test-unit
+.PHONY: test-e2e-encryption
 
 update-codegen: update-codegen-crds
 .PHONY: update-codegen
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -45,13 +45,12 @@ $(call add-bindata,v4.1.0,./bindata/v4.1.0/...,bindata,v410_00_assets,pkg/operat
 $(call add-crd-gen,manifests,$(CRD_APIS),./manifests,./manifests)
 
 # these are extremely slow serial e2e encryption tests that modify the cluster's global state 
-.PHONY: test-e2e-encryption
 test-e2e-encryption: GO_TEST_PACKAGES :=./test/e2e-encryption/...
 test-e2e-encryption: GO_TEST_FLAGS += -v
 test-e2e-encryption: GO_TEST_FLAGS += -timeout 4h
-test-e2e-encryption: GO_TEST_FLAGS += -p 1
```
p0lyn0mial: I think that `-p` flag ensures that tests are run in serial across packages and it's different than `-parallel` flag

tnozicka: it is not in the `go help test` so I assumed this is only a short version duplicated there

p0lyn0mial:
```
-parallel n
    Allow parallel execution of test functions that call t.Parallel.
    The value of this flag is the maximum number of tests to run
    simultaneously; by default, it is set to the value of GOMAXPROCS.
    Note that -parallel only applies within a single test binary.
    The 'go test' command may run tests for different packages
    in parallel as well, according to the setting of the -p flag
    (see 'go help build').
```

fun, does it mean that the tests within the same pkg are not run in `parallel` ? Does it mean that `e2e-aws-operator` (all tests in a single pkg) were run in serial anyway?

tnozicka: yeah, it's the other way around, we need only `-p`

tnozicka: yes, same package is always serial unless itself calls t.Parallel() - e.g. to avoid flakes on using the same global var in that package.

p0lyn0mial: we don't want to have both so that it doesn't have an effect even if someone would call `t.Parallel()` in a new test?

tnozicka: I'd say if you call t.Parallel manually, you are well aware that your test has to tolerate side effects, I'd not limit explicit choices

p0lyn0mial: okay :)

PR #673, opened 2019-11-26
configobservation/etcd: add exception for etcd-bootstrap to not use FQDN

alaypatel07:
This PR checks if the etcd-bootstrap host is configured with a real IP address
and uses it to connect to etcd instead of etcd-bootstrap FQDN. This will
relax the requirement of configuring A records in the installer for etcd-bootstrap


deads2k:
needs test

/assign @hexfusion

alaypatel07:
```
secret/support created
set -o pipefail; go test -race -v -timeout 1h -p 1 -json ./test/e2e/... | gotest2junit > /tmp/artifacts/unit_report.xml
FAIL: github.com/openshift/cluster-kube-apiserver-operator/test/e2e TestEncryptionTypeAESCBC 4m0.783s
make: *** [test-unit] Error 1
---
```
I do not understand the error, going to try again to see if it resurfaces

/test e2e-aws

pkg/operator/configobservation/etcd/observe_etcd.go
```diff
@@ -63,6 +63,13 @@ func ObserveStorageURLs(genericListers configobserver.Listers, recorder events.R
 				errs = append(errs, addressErr)
 				continue
 			}
+			// the installer uses dummy addresses in the subnet `192.0.2.` for host-etcd endpoints
+			// this check see if etcd-bootstrap is populated with a real ip address and uses it
+			// instead of FQDN
+			if address.Hostname == "etcd-bootstrap" && !strings.Contains(address.IP, "192.0.2") {
```
hexfusion: could we be more explicit using `strings.HasPrefix`

https://play.golang.org/p/zhN4gUkjV_3

hexfusion: We will need to rethink this if the endpoint address is IPv6 should we add TODO?

pkg/operator/configobservation/etcd/observe_etcd_test.go
```diff
@@ -0,0 +1,119 @@
...
+		wantErrs        []error
+		endpoint        *v1.Endpoints
+	}{
+		{
+			name:            "test etcd-bootstrap with dummy IP",
+			indexer:         cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc}),
+			currentConfig:   nil,
+			wantStorageURLs: []string{"https://etcd-bootstrap." + clusterFQDN + ":2379"},
+			wantErrs:        nil,
+			endpoint:        getEndpoint("etcd-bootstrap", "192.0.2.1"),
```
hexfusion: add testcase outlined above?

vendor/modules.txt
```diff
@@ -13,23 +13,23 @@ github.com/blang/semver
 # github.com/certifi/gocertifi v0.0.0-20190905060710-a5e0173ced67
 github.com/certifi/gocertifi
 # github.com/coreos/etcd v3.3.15+incompatible
-github.com/coreos/etcd/auth/authpb
 github.com/coreos/etcd/clientv3
+github.com/coreos/etcd/auth/authpb
 github.com/coreos/etcd/clientv3/balancer
```
hexfusion: Nit In general, I think it would be good for this file to only contain changes to modules you are adding. While this was done with automated tooling it is confusing to tell what actually changed here.

hexfusion:
No change to go.mod or go sum here? Can we verify the dep changes?

pkg/operator/configobservation/etcd/observe_etcd.go
hexfusion: I guess since the ep is static in the check populated by installer probably not a big deal.

pkg/operator/configobservation/etcd/observe_etcd_test.go
hexfusion: add ipv6 testcase?

pkg/operator/configobservation/etcd/observe_etcd.go
```diff
@@ -63,6 +63,13 @@ func ObserveStorageURLs(genericListers configobserver.Listers, recorder events.R
...
+				etcdURLs = append(etcdURLs, "https://"+address.IP+":2379")
```
hexfusion: should we also check IP is not an empty string?

alaypatel07:
@hexfusion since the deps were sub packages of client-go and we already have it in go.mod amd go.sum IMO `go mod vendor` just downloaded those and I committed them. The ci/prow/verify test passes which means we are good with deps?

alaypatel07:
@hexfusion sure I can add the test cases. In general I assumed that ip addresses would be good because we have it from the installer and hence tested only the change in this PR

hexfusion:
> @hexfusion sure I can add the test cases. In general I assumed that ip addresses would be good because we have it from the installer and hence tested only the change in this PR

I don't disagree but being overly defensive here isn't a bad thing.

pkg/operator/configobservation/etcd/observe_etcd.go
alaypatel07: done

pkg/operator/configobservation/etcd/observe_etcd_test.go
alaypatel07: added

pkg/operator/configobservation/etcd/observe_etcd.go
alaypatel07: added a check for IP validity, should cover empty string!

vendor/modules.txt
alaypatel07: dropping this change could very well be differences in tool versions

alaypatel07:
@hexfusion addressed all concerns, PTAL

alaypatel07:
infra flake
```
level=fatal msg="failed to fetch Cluster: failed to generate asset \"Cluster\": failed to create cluster: failed to apply using Terraform"
```
/test e2e-aws

alaypatel07:
/test e2e-aws

Marketplace flake

vendor/modules.txt
```diff
@@ -16,20 +16,20 @@ github.com/certifi/gocertifi
...
 github.com/coreos/etcd/etcdserver/etcdserverpb
 github.com/coreos/etcd/mvcc/mvccpb
 github.com/coreos/etcd/pkg/logutil
-github.com/coreos/etcd/pkg/systemd
-github.com/coreos/etcd/pkg/tlsutil
-github.com/coreos/etcd/pkg/transport
 github.com/coreos/etcd/pkg/types
+github.com/coreos/etcd/clientv3/balancer/connectivity
+github.com/coreos/etcd/pkg/systemd
 github.com/coreos/etcd/raft
```
hexfusion: @alaypatel07 if no additions are made here let’s drop this change?

hexfusion: Meaning drop the file if we are just rearranging.

pkg/operator/configobservation/etcd/observe_etcd.go
```diff
@@ -95,3 +101,10 @@ func ObserveStorageURLs(genericListers configobserver.Listers, recorder events.R
 
 	return
 }
+
+func isValidIP(ip string) bool {
```
sttts: this does really need a helper? It wraps one func in another with zero added value or abstraction.

pkg/operator/configobservation/etcd/observe_etcd.go
```diff
@@ -63,10 +64,15 @@ func ObserveStorageURLs(genericListers configobserver.Listers, recorder events.R
...
+			if !isValidIP(address.IP) {
+				ipErr := fmt.Errorf("endpoints %s/%s: subsets[%v]addresses[%v].IP is not a valid IP address", etcdEndpointName, etcdEndpointNamespace, subsetIndex, addressIndex)
+				errs = append(errs, ipErr)
+				continue
+			}
 			// the installer uses dummy addresses in the subnet `192.0.2.` for host-etcd endpoints
 			// this check see if etcd-bootstrap is populated with a real ip address and uses it
 			// instead of FQDN
-			if address.Hostname == "etcd-bootstrap" && !strings.Contains(address.IP, "192.0.2") {
+			if address.Hostname == "etcd-bootstrap" && !strings.HasPrefix(address.IP, "192.0.2") {
```
sttts: squash into previous commit where Contains was added.

pkg/operator/configobservation/etcd/observe_etcd.go
alaypatel07: valid point, I was not sure if that was the only validity check we wanted. Thought we would add more checks in future hence premature-abstraction if the is even a term :)

I will drop it for time being, thanks

pkg/operator/configobservation/etcd/observe_etcd.go
alaypatel07: done

alaypatel07:
@hexfusion I addressed the deps concern. FTR, I did a `go mod tidy` and `go mod vendor` initially that brought the other dependencies in. I reset the head to `HEAD~1` and just used the `go mod vendor` command and that fixed the concerns. PTAL

@sttts addressed your concerns as well PTAL

alaypatel07:
Infra flake

```
evel=fatal msg="failed to fetch Cluster: failed to generate asset \"Cluster\": failed to create cluster: failed to apply using Terraform"
```

/test e2e-aws

hexfusion:
thanks @alaypatel07 

/lgtm

PR #674, opened 2019-11-26
pkg/operator/configobservation/etcd/OWNERS: Drop wking

wking:
I was added here in 03b2b018d2 (#156), but I'm not sure what this package does, let alone how it should be maintained.

CC @abhinavdahiya in case he wants out too ;).


abhinavdahiya:
/lgtm

/cc @hexfusion @alaypatel07 

I think you should add yourself to the OWNERS here.

PR #677, opened 2019-11-27
starter: initialize statis pod controllers with service monitor

mfojtik:
Prevent panic in operator startup as when the dynamic client is not set the monitoring resource will crash.


pkg/operator/starter.go
```diff
@@ -111,6 +111,7 @@ func RunOperator(ctx *controllercmd.ControllerContext) error {
 		WithResources(operatorclient.TargetNamespace, "kube-apiserver", RevisionConfigMaps, RevisionSecrets).
 		WithCerts("kube-apiserver-certs", CertConfigMaps, CertSecrets).
 		WithVersioning(operatorclient.OperatorNamespace, "kube-apiserver", versionRecorder).
+		WithServiceMonitor(dynamicClient).
```
sttts: I believe I removed that ;-)

sttts:
As discussed in Slack, we have a manifest (highly customized for kube-apiserver) and this operator which only creates a standard service monitor. We have moved all our operators now to have this manifest with customized  content.

mfojtik:
https://github.com/openshift/library-go/pull/614 is driving deprecation of service monitor without panic

PR #680, opened 2019-12-02
Bug 1771665: file observer has not restarted process on serving cert change

sttts:


bparees:
cherry pick approving on behalf of @shawn-hurley

PR #682, opened 2019-12-02
termination observer controller

mfojtik:
The purpose of this controller is to watch the termination of API server static pods and record the termination events. This will help debugging issues with non-gracefully terminated API servers and also help to correlate the "api dropped out" situations.

For starters, this will log messages when any API server is recreated and it will watch for termination events and record them into operator log as well.

As a next step, we should add these as Prometheus metrics, so we can plot timeline for each API server termination for any revision at any time.

As for the implementation, static pods does not report "terminating" and does not update the mirrored pod (the pod is just instantly replaced with new pod)... The only reliable source of information here is the creation timestamps that this controller use to detect that the API server static pod was recreated (new revision, eviction, node reboot, etc...).


pkg/operator/terminationobserver/termination_observer.go
```diff
@@ -0,0 +1,249 @@
...
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog"
+
+	"github.com/openshift/library-go/pkg/operator/events"
+)
+
+var (
+	controllerWorkQueueKey = "key"
+
+	// terminationEventReasons lists all events that are observed when an API server is shutting down gracefully.
```
deads2k: do we have these for the openshift-apiserver or no?

pkg/operator/terminationobserver/termination_observer.go
```diff
@@ -0,0 +1,249 @@
...
+
+	c.cachesToSync = append(c.cachesToSync, kubeInformersForTargetNamespace.Core().V1().Pods().Informer().HasSynced)
+	c.cachesToSync = append(c.cachesToSync, kubeInformersForTargetNamespace.Core().V1().Events().Informer().HasSynced)
+
+	return c
+}
+
+// updateApiServerCreationTimestamp updates the creation timestamp for single API server
+func (c *TerminationObserver) updateApiServerCreationTimestamp(name string, lastCreationTimestamp time.Time) {
+	c.apiServersMutex.Lock()
```
deads2k: this makes me nervous.  Surely this isn't so high throughput that it needs to run multi-threaded

mfojtik: @deads2k i'm sure the unit tests when ran with `-race` will implode without this locking.

deads2k:
This is just dropping log messages that duplicate our events?  It seems a little weird, but I'll leave it @sttts discretion.  If openshift-apiserver does the same, be sure to promote this.

pkg/operator/terminationobserver/termination_observer.go
mfojtik: @sttts ^ i think every API server has them, if this proof to be useful, we can run this in openshift-apiserver operator as well and I will librarize it ;-)

mfojtik:
**UPDATE**: Added prometheus metrics, will play more tomorrow.

sttts:
> This is just dropping log messages that duplicate our events? It seems a little weird, but I'll leave it @sttts discretion. If openshift-apiserver does the same, be sure to promote this.

See it as an aggregation of everything around termination, not only the events, but also node restarts at some point or whatever helps us to paint a picture from an external view on the termination of our operand.

pkg/operator/terminationobserver/termination_metrics.go
```diff
@@ -0,0 +1,65 @@
+package terminationobserver
```
sttts: rename file to metrics.go

pkg/operator/terminationobserver/termination_observer.go
```diff
@@ -0,0 +1,272 @@
+package terminationobserver
```
sttts: rename file to observer.go or even controller.go

pkg/operator/terminationobserver/termination_observer.go
```diff
@@ -0,0 +1,272 @@
...
+
+func (c *TerminationObserver) sync() error {
+	podList, err := c.podsGetter.Pods(c.targetNamespace).List(metav1.ListOptions{LabelSelector: "app=openshift-kube-apiserver"})
+	if err != nil {
+		return fmt.Errorf("unable to list pods in %q namespace: %v", c.targetNamespace, err)
+	}
+
+	for _, pod := range podList.Items {
+		// get the last seen creation timestamp for this api server
+		lastSeenCreationTimestamp := c.getApiServerCreationTimestamp(pod.Name)
```
sttts: s/getApiServerCreationTimestamp/lastCreationTimestamp/

pkg/operator/terminationobserver/termination_observer.go
```diff
@@ -0,0 +1,272 @@
...
+		klog.Infof("Observed termination event %q for API server %q: %s", event.Reason, c.apiServers[i].serverName, event.Message)
+		break
+	}
+}
+
+func (c *TerminationObserver) getApiServerCreationTimestamp(name string) *time.Time {
+	c.apiServersMutex.RLock()
+	defer c.apiServersMutex.RUnlock()
+	for i := range c.apiServers {
+		if c.apiServers[i].serverName != name {
```
sttts: make the list a map and get rid of the loop

pkg/operator/terminationobserver/termination_observer.go
```diff
@@ -0,0 +1,272 @@
...
+		}
+		found = true
+		klog.Infof("Observed API server pod %q termination at %s", c.apiServers[i].serverName, currentCreationTimestamp)
+		if !isNew {
+			c.apiServers[i].terminatedTimestamp = &c.apiServers[i].currentCreatedTimestamp
+		}
+		c.apiServers[i].currentCreatedTimestamp = currentCreationTimestamp
+		break
+	}
+	if !found {
```
sttts: why `isNew` and `found` logic? Isn't that the inverse?

pkg/operator/terminationobserver/termination_observer.go
sttts: +1, this looks suspicious. Forbid more than one worker.

pkg/operator/terminationobserver/termination_observer.go
```diff
@@ -0,0 +1,272 @@
...
+	c.cachesToSync = append(c.cachesToSync, kubeInformersForTargetNamespace.Core().V1().Events().Informer().HasSynced)
+
+	return c
+}
+
+func (c *TerminationObserver) Register() {
+	registerMetrics(c.getApiServers)
+}
+
+func (c *TerminationObserver) getApiServers() []*apiServerStatus {
```
sttts: who needs this?

pkg/operator/terminationobserver/termination_metrics.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+type terminationMetrics struct {
+	getApiServersFn func() []*apiServerStatus
+	apiServerMetric *prometheus.GaugeVec
+}
+
+func registerMetrics(getApiServersFn func() []*apiServerStatus) {
+	legacyregistry.MustRegister(&terminationMetrics{
+		getApiServersFn: getApiServersFn,
```
sttts: this is super ugly to have a callback to an object from a global variable. This also keeps the object around during after tests. Can we change all this logic to avoid this callback to the controller?

pkg/operator/terminationobserver/termination_observer.go
```diff
@@ -0,0 +1,272 @@
...
+		"TerminationStart",
+		"TerminationPreShutdownHooksFinished",
+		"TerminationPreShutdownHooksFinished",
+		"TerminationMinimalShutdownDurationFinished",
+		"TerminationStoppedServing",
+		"TerminationGracefulTerminationFinished",
+	}
+)
+
+type TerminationObserver struct {
```
sttts: 5 liner go doc what this is and why it exists.

pkg/operator/terminationobserver/termination_observer.go
```diff
@@ -0,0 +1,272 @@
...
+	targetNamespace string
+
+	podsGetter corev1client.PodsGetter
+
+	cachesToSync  []cache.InformerSynced
+	queue         workqueue.RateLimitingInterface
+	eventRecorder events.Recorder
+
+	apiServers      []*apiServerStatus
```
sttts: make it a map

pkg/operator/terminationobserver/termination_observer.go
```diff
@@ -0,0 +1,272 @@
...
+
+type apiServerStatus struct {
+	serverName string
+	// currentCreatedTimestamp is creation timestamp of currently running API server
+	// this API server does not have any termination events, those events are only valid for
+	// "old" API server.
+	currentCreatedTimestamp time.Time
+	// terminatedTimestamp is timestamp that represent "old" API server that is currently being terminated.
+	// this timestamp can be nil if the API server was never terminated.
+	terminatedTimestamp *time.Time
```
sttts: previousTermination

pkg/operator/terminationobserver/termination_observer.go
```diff
@@ -0,0 +1,272 @@
...
+	apiServers      []*apiServerStatus
+	apiServersMutex sync.RWMutex
+}
+
+type apiServerStatus struct {
+	serverName string
+	// currentCreatedTimestamp is creation timestamp of currently running API server
+	// this API server does not have any termination events, those events are only valid for
+	// "old" API server.
+	currentCreatedTimestamp time.Time
```
sttts: lastCreation

pkg/operator/terminationobserver/termination_observer.go
```diff
@@ -0,0 +1,272 @@
...
+	cachesToSync  []cache.InformerSynced
+	queue         workqueue.RateLimitingInterface
+	eventRecorder events.Recorder
+
+	apiServers      []*apiServerStatus
+	apiServersMutex sync.RWMutex
+}
+
+type apiServerStatus struct {
+	serverName string
```
sttts: name

pkg/operator/terminationobserver/termination_metrics.go
mfojtik: why? i want to be able to grep logs from this file... klog messages from `metrics.go` are not very useful

pkg/operator/terminationobserver/termination_observer.go
mfojtik: same comment as above

pkg/operator/terminationobserver/termination_observer.go
mfojtik: no, this is "current" creation timestamp

pkg/operator/terminationobserver/termination_observer.go
mfojtik: sure, was thinking about the same

pkg/operator/terminationobserver/termination_observer.go
mfojtik: it is tricky... at the time when the creationTimestamp is updated on pod, the events you are receiving are for the "previous" pod that is now terminating... is `isNew` means this api server is "new" and was not running before (new master node?)

pkg/operator/terminationobserver/termination_observer.go
mfojtik: metrics...

pkg/operator/terminationobserver/termination_metrics.go
mfojtik: we can make some in-memory storage that metrics and observer will share, does that sound better?

pkg/operator/terminationobserver/storage.go
```diff
@@ -0,0 +1,147 @@
...
+	"sync"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// The store interface implementation keeps the in-memory state for each API server.
+// The state consists of current creationTimestamp and previous creationTimestamp as well as termination
+// events recorded by termination observer.
+type Store interface {
```
sttts: why public?

sttts: why an interface at all?

pkg/operator/terminationobserver/termination_metrics.go
```diff
@@ -0,0 +1,57 @@
...
+	"k8s.io/component-base/metrics/legacyregistry"
+)
+
+type terminationMetrics struct {
+	store           Store
+	apiServerMetric *prometheus.GaugeVec
+}
+
+// RegisterMetrics registers the termination metrics into legacy prometheus registry.
+func RegisterMetrics(store Store) {
```
sttts: this store does not make it better.

sttts: it's still a back reference. Why do we need it? Why not call out to the metric from the place where we actually collect the data originally?

sttts: in other words: make the call graph a tree. No need for any synchronization (locks...).

pkg/operator/terminationobserver/termination_metrics.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+}
+
+var _ metrics.Registerable = &terminationMetrics{}
+
+func (t *terminationMetrics) Describe(ch chan<- *prometheus.Desc) {
+	ch <- t.apiServerMetric.WithLabelValues("").Desc()
+}
+
+func (t *terminationMetrics) Collect(ch chan<- prometheus.Metric) {
```
sttts: metrics can be set imperatively. I don't see a reason to build this callback thing here which read an imperatively constructed state. Make it direct and let the prometheus library do the serving/collection.

pkg/operator/terminationobserver/storage.go
```diff
@@ -0,0 +1,147 @@
...
+	sync.RWMutex
+
+	state map[string]APIServerState
+}
+
+type APIServerState struct {
+	// currentCreatedTimestamp is creation timestamp of currently running API server
+	// this API server does not have any termination events, those events are only valid for
+	// "old" API server.
+	currentCreatedTimestamp time.Time
```
sttts: we also don't postfix other fields with their types. Why timestamps?

pkg/operator/terminationobserver/storage.go
mfojtik: because I'm used to wrap "storage" into interface (so you can mock them in unit tests), I can drop the interface, I just think it is an anti-pattern in go and it does not really simplify the code.

pkg/operator/terminationobserver/termination_metrics.go
mfojtik: We are not collecting data synchronously. We collect pods from one informer and we collect events for second informer and I use the store to represents that data and make them collectable by prometheus. I don't see any complexity here, there is just simple `map` with synchronized access to read/write (which will be needed regardless of merging the logic of observer and metrics collector).

pkg/operator/terminationobserver/termination_metrics.go
mfojtik: We use this pattern everywhere and I think this is the right way to expose metrics to prometheus. Having this baked into a controller and then determining when to send the data to prometheus seems like less error prone to me.

pkg/operator/terminationobserver/storage.go
mfojtik: because this store the 'creationTimestamp' field and I liked to have this matching the original field (or close to original field) if possible.

pkg/operator/terminationobserver/storage.go
```diff
@@ -0,0 +1,147 @@
...
+	m.Lock()
+	defer m.Unlock()
+	oldState, ok := m.state[name]
+	if !ok {
+		return ApiServerNotFoundError
+	}
+	newState := APIServerState{
+		currentCreatedTimestamp:      state.currentCreatedTimestamp,
+		previousTerminationTimestamp: &oldState.currentCreatedTimestamp,
+		previousTerminationEvents:    oldState.previousTerminationEvents,
```
sttts: what happens if previousTerminationEvents is set in `state`?

pkg/operator/terminationobserver/storage.go
```diff
@@ -0,0 +1,147 @@
...
+	GetEventWithReason(name, reason string) *corev1.Event
+
+	// ListNames return list of API server names (hostnames)
+	// Used by prometheus metrics collector
+	ListNames() []string
+
+	// Create add new API server state
+	Create(name string, state APIServerState)
+
+	// Update updates the existing API server state.
```
sttts: keeping the old events?

pkg/operator/terminationobserver/termination_observer.go
```diff
@@ -0,0 +1,200 @@
...
+			continue
+		}
+
+		// the pod was recreated and the creation timestamp changed, update the last seen creation timestamps
+		// and start watching for termination events.
+		if pod.CreationTimestamp.Time != currentServerState.currentCreatedTimestamp {
+			if err := c.store.Update(pod.Name, APIServerState{currentCreatedTimestamp: pod.CreationTimestamp.Time}); err != nil {
+				return fmt.Errorf("unexpected sync error: %v", err)
+			}
+		}
```
sttts: this whole loop would be a oneliner with a Upsert func in the store.

pkg/operator/terminationobserver/storage.go
```diff
@@ -0,0 +1,147 @@
...
+	Update(name string, state APIServerState) error
+
+	// UpdateEvents add new events into existing API server state
+	UpdateEvents(name string, event *corev1.Event) error
```
sttts: why is it not called AddEvents if it adds and not updates?

pkg/operator/terminationobserver/storage.go
```diff
@@ -0,0 +1,147 @@
...
+	if !ok {
+		return ApiServerNotFoundError
+	}
+	reconciledEvents := []*corev1.Event{}
+	for i := range state.previousTerminationEvents {
+		if state.previousTerminationEvents[i].LastTimestamp.Time.Before(*state.previousTerminationTimestamp) {
+			continue
+		}
+		reconciledEvents = append(reconciledEvents, state.previousTerminationEvents[i])
+	}
```
sttts: this is surprising in this func. Why is it not in `Update(..)`?

pkg/operator/terminationobserver/termination_metrics.go
```diff
@@ -0,0 +1,57 @@
...
+	legacyregistry.MustRegister(newTerminationMetrics(store))
+}
+
+func newTerminationMetrics(store Store) *terminationMetrics {
+	return &terminationMetrics{
+		store: store,
+		apiServerMetric: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "openshift_kube_apiserver_termination",
+			Help: "Report individual termination events and their timestamps for each Kubernetes API server",
+		}, []string{"name", "createdTimestamp", "eventName"}),
```
sttts: isn't a timestamp as label exploding memory consumption for a gauge?

sttts: verified with metrics team: timestamp labels are a big NO.

sttts: https://i.imgflip.com/3i9htp.jpg with regards from monitoring team.

mfojtik: We can use the unix timestamp as value and if it is == 0 then it means we haven't seen that event. I'm not sure how will the prom query looks like then to plot a timeline that shows the timeline of termination for particular API server

pkg/operator/terminationobserver/termination_observer.go
mfojtik: i simplified that loop now

PR #687, opened 2019-12-09
Bug 1781286: staticpod: deprecate service monitor controller

mfojtik:


deads2k:
wow, go mod is really inscrutable.  

/lgtm

PR #689, opened 2019-12-10
Bug 1779796: bump(*): get leader election TCP timeout

sttts:


sttts:
Change came in through https://github.com/openshift/cluster-kube-apiserver-operator/pull/691.

PR #700, opened 2019-12-18
Bump: for new library-go with go mod verify-deps

tnozicka:
`verify-deps` comes in https://github.com/openshift/release/pull/6472 but needs this first

/cc @soltysh @mfojtik


vendor/github.com/openshift/library-go/alpha-build-machinery/make/targets/openshift/deps-gomod.mk
```diff
@@ -0,0 +1,41 @@
+self_dir :=$(dir $(lastword $(MAKEFILE_LIST)))
```
tnozicka: here it goes

PR #658, opened 2019-11-14
The API server operator should include the SCTP feature gate

dougbtv:
Updates to latest openshift/api 

Method used to update:

```
# go get with commit @ HEAD to include sctp feature gate
go get github.com/openshift/api@4efd1a50a74e455e53d0f4478657e5c226fe9b24
# pick up generated version from a diff
git diff 
go checkout .
# modify the replace section to include generated version from go get above
vi go.mod
go mod vendor
go mod tidy
```

Happy to change that workflow -- the `replace` section was a curve for me. Unsure if this is the proper workflow.

PR #698, opened 2019-12-17
stomp secrets if types don't match

deads2k:
also update kube to 1.17

PR #702, opened 2019-12-19
Consume the latest openshift api so sctp is enabled in lowlatency featureset

fedepaol:
Fetched the latest pseudoversion, used it in the replace directive.
Ran `go mod tidy` and `go mod vendor`.
Not sure I need to update something in https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/bindata/bootkube/config/bootstrap-config-overrides.yaml the same way it was done in https://github.com/openshift/cluster-kube-apiserver-operator/pull/658

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/googleapis/gnostic v0.3.1 // indirect
 	github.com/imdario/mergo v0.3.8
 	github.com/jteeuwen/go-bindata v3.0.8-0.20151023091102-a0ff2567cfb7+incompatible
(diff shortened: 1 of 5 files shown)


deads2k:
/hold

hold for https://github.com/openshift/cluster-kube-apiserver-operator/pull/698 which harmonizes api and k8s versions.

fedepaol:
@deads2k I think this is good (if CI agrees) now that https://github.com/openshift/cluster-kube-apiserver-operator/pull/698 is in.

fedepaol:
@deads2k you should also remove the hold

PR #707, opened 2020-01-06
use tcp instead of tcp4 in default serving info

mfojtik:


deads2k:
/lgtm

if CI passes

mfojtik:
/hold

I think we need something more sophisticated, like https://github.com/openshift-kni/cluster-kube-apiserver-operator/commit/1a96ae7f6065704c568949c340a28cf7f706dc16

danwinship:
/hold

This would need to be backported to 4.3, and I've been arguing that it would be bad (securitywise, etc) to switch a hostNetwork service from single-stack to dual-stack in a z-stream release... So I had filed #653 to toggle between single-stack IPv6 and single-stack IPv4 as needed.

I'm not really convinced that there are any real-world deployments where suddenly unexpectedly making kube-apiserver dual-stack would actually cause any problems. But still... it seems a little sketchy. Anyway, feel free to take this PR over #653 if you want. (Or perhaps, merge and backport #653 and then commit this over it for 4.4 only.)

PR #708, opened 2020-01-07
bump(*): openshift/library-go

mfojtik:
* openshift/library-go#657 : config: default bind network to tcp instead of tcp4


mfojtik:
self-lgtm, forgot to run `go mod tidy`

PR #709, opened 2020-01-07
Enable the IPv6DualStack feature gate by default

smarterclayton:
This will allow the cluster to be configured for dual stack
for CI testing of IPv6 in cloud environments (clouds do not
fully support single stack, and dual stack is the next feature
set to be enabled).


go.mod
```diff
@@ -10,18 +10,20 @@ require (
 	github.com/davecgh/go-spew v1.1.1
 	github.com/getsentry/raven-go v0.2.0 // indirect
 	github.com/ghodss/yaml v1.0.0
+	github.com/gogo/protobuf v1.3.1 // indirect
```
mfojtik: interesting... why we suddenly need to import this?

mfojtik: hmm, I see https://github.com/openshift/api/blob/master/go.mod#L6

smarterclayton:
I’m fixing the broken kube, which I will then backport to 1.17

PR #713, opened 2020-01-09
encryption: switch to kube-storage-version-migrator

vareti:
continuing the work done in #637 

Requires https://github.com/openshift/cluster-kube-storage-version-migrator-operator to be in the release.


pkg/operator/starter.go
```diff
@@ -44,13 +45,6 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 	if err != nil {
 		return err
 	}
-	migrationClientConfig := dynamic.ConfigFor(controllerContext.KubeConfig)
-	migrationClientConfig.Burst = 40
-	migrationClientConfig.QPS = 30
-	dynamicClientForMigration, err := dynamic.NewForConfig(migrationClientConfig)
```
sttts: verify that kube-storage-version-migrator has similar values.

pkg/operator/starter.go
```diff
@@ -160,7 +154,10 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 	if err != nil {
 		return err
 	}
-	migrator := migrators.NewInProcessMigrator(dynamicClientForMigration, kubeClient.Discovery())
+	//migrator := migrators.NewInProcessMigrator(dynamicClientForMigration, kubeClient.Discovery())
```
sttts: this line can go

pkg/operator/starter.go
vareti: kube-storage-version migrator does not set anything specific. so, i restored these values in new commits.

pkg/operator/starter.go
vareti: yes. removed it.

pkg/operator/starter.go
```diff
@@ -160,7 +155,13 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 	if err != nil {
 		return err
 	}
-	migrator := migrators.NewInProcessMigrator(dynamicClientForMigration, kubeClient.Discovery())
+
+	migrationClientConfig := dynamic.ConfigFor(controllerContext.KubeConfig)
+	migrationClientConfig.Burst = 40
+	migrationClientConfig.QPS = 30
```
sttts: this client is just for the CRD. No need to tweak the values. Instead we have to verify that kube-storage-version-migrator does the same tweaking internally.

sttts:
/approve
/hold

Want to see the e2e-aws-operator-encryption test before merging.

p0lyn0mial:
> Want to see the e2e-aws-operator-encryption test before merging.

scheduled

pkg/operator/starter.go
```diff
@@ -160,7 +154,10 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 	if err != nil {
 		return err
 	}
-	migrator := migrators.NewInProcessMigrator(dynamicClientForMigration, kubeClient.Discovery())
+
+	migrationClient := kubemigratorclient.NewForConfigOrDie(controllerContext.KubeConfig)
+	migrationInformer := migrationv1alpha1informer.NewSharedInformerFactory(migrationClient, time.Minute*30)
```
p0lyn0mial: do we have to wait until `migrationInformer` has synced?

pkg/operator/starter.go
```diff
@@ -160,7 +154,10 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 	if err != nil {
 		return err
 	}
-	migrator := migrators.NewInProcessMigrator(dynamicClientForMigration, kubeClient.Discovery())
+
+	migrationClient := kubemigratorclient.NewForConfigOrDie(controllerContext.KubeConfig)
```
p0lyn0mial: based on the previous comment https://github.com/openshift/cluster-kube-apiserver-operator/pull/713/files#r366551501 we need to set QPS/Burst

sttts: This is just for the CRD, not the actual migrated objects. We have to verify qps/burst in the migrator.

p0lyn0mial: ic

p0lyn0mial:
`ci/prow/e2e-aws-operator-encryption-perf` failed because the migration didn't complete in the allotted time. Either we were too slow or we didn't make any progress. `kube-storage-version-migrator`'s log [1] looks empty.


```
All workers completed successfully
Created 26986 secrets
Created 14997 configmaps
Populating etcd took 53.279371803s
Starting encryption e2e test for "aescbc" mode
Waiting up to 1h0m0s for the condition "EncryptionMigrationControllerProgressing" with the reason "Migrating" to be set to true
Found configuration for host https://api.ci-op-5yphjqxk-4b168.origin-ci-int-aws.dev.rhcloud.com:6443.
Updating encryption type in the config file for APIServer to "aescbc"
Waiting up to 1h0m0s for the next key "encryption-key-openshift-kube-apiserver-1", previous key was "no previous key"
Observed key "encryption-key-openshift-kube-apiserver-1", waiting up to 1h0m0s until it will be used to migrate [secrets configmaps]
EncryptionMigrationControllerProgressing condition observed at 2020-01-20 12:36:37 +0000 UTC
panic: test timed out after 1h0m0s
```


[1]
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-kube-apiserver-operator/713/pull-ci-openshift-cluster-kube-apiserver-operator-master-e2e-aws-operator-encryption-perf/42/artifacts/e2e-aws-operator-encryption-perf/pods/openshift-kube-storage-version-migrator_migrator-7b5fdccffb-gxwbq_migrator.log

go.mod
```diff
@@ -15,6 +15,7 @@ require (
 	github.com/googleapis/gnostic v0.3.1 // indirect
 	github.com/imdario/mergo v0.3.8
 	github.com/jteeuwen/go-bindata v3.0.8-0.20151023091102-a0ff2567cfb7+incompatible
+	github.com/kubernetes-sigs/kube-storage-version-migrator v0.0.0-20191127225502-51849bc15f17
```
p0lyn0mial: I think that we should update https://github.com/openshift/cluster-kube-apiserver-operator/blob/4d26538c218b348a2e3376bfdb402bd8350d5485/go.mod#L36 instead

p0lyn0mial: unless we don't need our fork then we should remove an entry from `replace` directive.

p0lyn0mial: upstream seems to be using `1.17.0` - https://github.com/kubernetes-sigs/kube-storage-version-migrator/blob/master/Gopkg.toml

pkg/operator/starter.go
```diff
@@ -160,7 +154,10 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
...
+	migrationInformer := migrationv1alpha1informer.NewSharedInformerFactory(migrationClient, time.Minute*30)
+	migrator := migrators.NewKubeStorageVersionMigrator(migrationClient, migrationInformer.Migration().V1alpha1(), kubeClient.Discovery())
```
p0lyn0mial: @sttts we should switch to `dynamic-{informer,lister,client}` so that we don't depend on `kubernetes-sigs/kube-storage-version-migrator` in `cluster-kube-apiserver-operator`

sttts: Ugly. I see some risk that we might depend on upstream's kube version level though. Ugly, did I mention that before?

sttts: @vareti can you prototype that?

sttts: @p0lyn0mial not a blocker, is it?

p0lyn0mial: no, not a blocker.

sttts:
@vareti I think we only wait for our migrator fork to get the two flags?

go.mod
vareti: Yes, you are right! we can remove our fork if it is not needed for this release.

pkg/operator/starter.go
vareti: @sttts yes, I will try to prototype this tomorrow.

vareti:
@sttts not sure if we need our migrator fork to get the flags as upstream is merged now. I will sync up with @sanchezl tomorrow and come up with a plan.

pkg/operator/starter.go
vareti: I think yes, but I am not entirely sure(still learning and confused about informers and clients)

sttts: `KubeStorageVersionMigrator.AddEventHandler`  does that already, adding the `HasSynced` func to the list of sync funcs of the migration controller. Should be fine.

sttts:
Lgtm, when the perf test succeeds.

p0lyn0mial:
running `/test e2e-aws-operator-encryption-perf` because it keeps failing due to some infra issues.

vareti:
/test e2e-aws-operator-encryption-perf

It still hits the same AMI errors. not sure if something needs to be done from PR side.

go.mod
sttts: Upstream does not have the hardcoded qps/burst values, does it?

sanchezl: We import it here for the client. As I understand it, the qps/burst settings were hard coded for the actual migrator. So.. technically, yes we don't need the replace as we are only using the client here. The migrator image though, needs to be from our fork.

PR #714, opened 2020-01-13
Bug 1759182: Block all externalIPs by default

abhat:
Only allow external IPs to be configured by cluster administrators and block normal users to set external IPs on a Service.

### DIFF
--- a/pkg/operator/configobservation/network/observe_network.go
+++ b/pkg/operator/configobservation/network/observe_network.go
@@ -10,7 +10,8 @@ import (
 	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/configobservation"
 )
 
-// ObserveRestrictedCIDRs generates two configuration
+// ObserveRestrictedCIDRs watches the network configuration and updates the
+// RestrictedEndpointsAdmission controller config.
 func ObserveRestrictedCIDRs(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
 	listers := genericListers.(configobservation.Listers)
 
@@ -52,7 +53,7 @@ func ObserveRestrictedCIDRs(genericListers configobserver.Listers, recorder even
 	//      network.openshift.io/RestrictedEndpointsAdmission:
 	//        configuration:
 	//          version: network.openshift.io/v1
-	//          kind: RestrictedEndpointsAdmission
+	//          kind: RestrictedEndpointsAdmissionConfig
 	//          restrictedCIDRs:
 	//            - 10.3.0.0/16 # ServiceCIDR
 	//            - 10.2.0.0/16 # ClusterCIDR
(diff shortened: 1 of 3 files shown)


danwinship:
So what's the deal with this vs https://github.com/openshift/origin/pull/24390 ? Are they both needed?

abhat:
Yes, this one is the actual change that will create the ExternalIPRangerAdmissionConfig. The other origin PR only fixes the typo for the module on which the Install method is called. `restrictedendpoints` vs `externaliprange`

pkg/operator/configobservation/network/observe_network.go
```diff
@@ -101,3 +102,31 @@ func ObserveServicesSubnet(genericListers configobserver.Listers, recorder event
...
+
+	// For 4.1, we didn't expose the ExternalIP in the network spec so only need to block all IPs by default.
+
+	// Policy, synthesize config
+	// Simply by creating this configuration, the admission controller will
+	// be enabled and block all IPs by default.
+	admissionControllerConfig := unstructured.Unstructured{}
+	admissionControllerConfig.SetAPIVersion("network.openshift.io/v1")
+	admissionControllerConfig.SetKind("ExternalIPRangerAdmissionConfig")
+	unstructured.SetNestedField(admissionControllerConfig.Object, false, "allowIngressIP")
```
sttts: `allowIngressIP: false` is missing in the comment above, isn't it?

sttts: and `externalIPNetworkCIDRs` I don't see being set here.

pkg/operator/configobservation/network/observe_network_test.go
```diff
@@ -62,16 +52,18 @@ func TestObserveRestrictedCIDRs(t *testing.T) {
...
-		t.Error(restrictedCIDRs[0])
-	}
-	if restrictedCIDRs[1] != "serviceCIDR" {
-		t.Error(restrictedCIDRs[1])
-	}
+	assert.Empty(t, errors)
+	shouldMatchYaml(t, result, `
+admission:
+  pluginConfig:
+    network.openshift.io/RestrictedEndpointsAdmission:
```
sttts: ...Config ?

pkg/operator/configobservation/network/observe_network.go
abhat: Added the comment for `allowIngressIP: false`.
The `externalIPNetworkCIDRs` is going to be an empty slice in this case since there is no property in 4.1 under https://github.com/openshift/api/blob/release-4.1/config/v1/types_network.go#L28 to translate this to.

sttts:
Looks good to me. It needs some peer from sdn team to lgtm.

ashcrow:
https://bugzilla.redhat.com/show_bug.cgi?id=1759182#c12 indicates verification did not pass yet.

abhat:
Right, I am trying to root cause why the `ExternalIPRangerAdmissionConfig` is not correctly registered in the `scheme`.

bparees:
At this point we are only accepting critical bug fixes for 4.1.  This does not seem to meet that target, do you want to close it (and the BZ) or make a case for why it is critical?

abhat:
There is a security implication associated with this bug. In 4.1, we didn't have a field in the spec for ExternalIPCIDR, which meant that an arbitrary non-admin user can set external IP on services to any arbitrary IP. We plugged part of the hole by adding a custom RBAC checker. But for the RBAC checker to be "effective" we need this change. IMO, it's critical if we will continue to support 4.1.x. If 4.1.x will be EOL soon, we can close out the bug and the PR.

bparees:
>  If 4.1.x will be EOL soon, we can close out the bug and the PR.

https://access.redhat.com/support/policy/updates/openshift

4.1 ends maintenance support when 4.4 GAs.

jwforres:
@abhat please see https://bugzilla.redhat.com/show_bug.cgi?id=1759182#c13 I tend to agree with Scott. But the BZ has no severity assigned, so would like @danwinship and/or @knobunc to ack this is important enough to be backported at this time.

abhat:
Closing this since we don't see much value in porting/adding fixes for 4.1

PR #718, opened 2020-01-16
Enable bound SA tokens

marun:
Implements operator support for https://github.com/openshift/enhancements/pull/150

TODO
 - [x] - Add key management controller and test coverage of same
 - [x] - Ensure keys are configured in the apiserver pods
 - [x] - Ensure configuration of issuer and audience
 - [x] - Unskip test coverage for the `TokenRequest` API
 - [x] - Ensure e2e is passing consistently


bindata/v4.1.0/kube-apiserver/defaultconfig.yaml
```diff
@@ -106,6 +106,11 @@ servingInfo:
   requestTimeoutSeconds: 3600
 serviceAccountPublicKeyFiles:
   - /etc/kubernetes/static-pod-resources/configmaps/sa-token-signing-certs
+  - /etc/kubernetes/static-pod-certs/configmaps/bound-sa-token-signing-certs
+serviceAccountIssuer: auth.openshift.io
+apiAudiences:
+  - auth.openshift.io
+serviceAccountSigningKeyFile: /etc/kubernetes/static-pod-certs/secrets/service-account-signing-key/service-account.key
```
sttts: isn't there are need to override this for bootstrapping? Or do we get this key from the render step that early?

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,296 @@
...
+
+	// queue only ever has one item, but it has nice error handling backoff/retry semantics
+	queue workqueue.RateLimitingInterface
+}
+
+func NewBoundSATokenSignerController(
+	kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces,
+	kubeClient kubernetes.Interface,
+	eventRecorder events.Recorder,
+
```
sttts: nit: no empty line

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,296 @@
...
+		cachesSynced: []cache.InformerSynced{
+			kubeInformersForNamespaces.InformersFor(operatorNamespace).Core().V1().Secrets().Informer().HasSynced,
+			kubeInformersForNamespaces.InformersFor(targetNamespace).Core().V1().Secrets().Informer().HasSynced,
+			kubeInformersForNamespaces.InformersFor(targetNamespace).Core().V1().ConfigMaps().Informer().HasSynced,
+		},
+
+		queue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "BoundSATokenSignerController"),
+	}
+
```
sttts: nit: no double empty lines

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,296 @@
...
+	targetNamespace   = operatorclient.TargetNamespace
+
+	keySize                  = 2048
+	NextSigningKeySecretName = "next-service-account-signing-key"
+	SigningKeySecretName     = "service-account-signing-key"
+	PrivateKeyKey            = "service-account.key"
+	PublicKeyKey             = "service-account.pub"
+
+	TokenReadyAnnotation = "kube-apiserver.openshift.io/ready-to-use"
+	readyInterval        = 5 * time.Minute
```
sttts: what is this?

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,296 @@
...
+		newSecret, err := newSigningSecret()
+		if err != nil {
+			return err
+		}
+		signingSecret, _, err = resourceapply.ApplySecret(c.secretClient, c.eventRecorder, newSecret)
+		if err != nil {
+			return err
+		}
+		// requeue for after we should have recovered
+		c.queue.AddAfter(workQueueKey, readyInterval+10*time.Second)
```
sttts: why this line?

sttts: And if that line, why no immediate `return` ?

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+
+	certConfigMap, err := c.configMapClient.ConfigMaps(targetNamespace).Get(CertConfigMapName, metav1.GetOptions{})
```
sttts: what happens if the secret is changed, but the config map is not?

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,296 @@
...
+	currPublicKey := string(signingSecret.Data[PublicKeyKey])
+	hasThisKey := false
+	for _, value := range certConfigMap.Data {
+		if value == currPublicKey {
+			hasThisKey = true
+			break
+		}
+	}
+	if !hasThisKey {
+		// Increment until a unique name is found
```
sttts: this is no pruning?

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,296 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	// Check if next-sa-signing-key has been around long enough to be promoted.
+	// Giving time for apiserver instances to pick up the change in public keys before
+	// changing the private key minimizes the potential for one or more apiservers to
+	// issue tokens signed by the new private key that apiservers without the
+	// corresponding public key are unable to validate.
```
sttts: so we lack back-pressure again here, right? If rolling update is blocked for some reason, we might get into trouble.

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,296 @@
...
+	//
+	// TODO(marun) Find a more accurate indication that all apiservers are capable of
+	// validating tokens signed by the new private key.
```
sttts: In the etcd encryption code, we wait until all API servers have settled on the same revision, and that there is no new pending revision. Maybe that approach works here too?

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,296 @@
...
+		}
+	}
+
+	// If past promotion time, make the new private key available to the operand.
+	if readyToPromote {
+		_, _, err := resourceapply.SyncSecret(c.secretClient, c.eventRecorder,
+			operatorNamespace, NextSigningKeySecretName,
+			targetNamespace, SigningKeySecretName, []metav1.OwnerReference{})
+		return err
+	}
```
sttts: should we requeue the work queue key here? Or is line 110 exactly there for that reason? (some comment here would help).

pkg/operator/resourcesynccontroller/resourcesynccontroller.go
```diff
@@ -86,5 +86,7 @@ func NewResourceSyncController(
 		return nil, err
 	}
 
+	// TODO(marun) Sync the public keys used to validate bound tokens to GlobalMachineSpecifiedConfigNamespace for use by other agents
+
```
sttts: if we don't do that, will people pick the configmap from operand namespace? Do we promote that to an API then accidently?

sttts: Note that we have configmap sync controllers in the apiserver operators.

pkg/operator/boundsatokensignercontroller/controller.go
stlaz: +1 for having the CM revisioned.

Are the service account tokens private and pub keys read dynamically since we're just swapping them here without redeployment?

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,296 @@
...
+		for {
+			c.queue.Add(workQueueKey)
+			select {
+			case <-ticker.C:
+			case <-stopCh:
+				return
+			}
+		}
+
+	}, time.Minute, stopCh)
```
stlaz: It this correct? That inner `ticker` looks quite unnecessary given that `wait.Until` has its own internal timer which does the same, this basically runs `wait.Until` and freezes it on its first loop?

bindata/v4.1.0/kube-apiserver/defaultconfig.yaml
marun: The intent is not to enable bound tokens in the bootstrapping phase, since as per a [discussion](https://github.com/openshift/enhancements/pull/150/files#r357266863) on the enhancement there does not appear to be a need for that. Will it be necessary to set these values in code that can detect that bootstrapping is complete rather than here?

pkg/operator/boundsatokensignercontroller/controller.go
marun: Done.  I was just being consistent with [the controller this was derived from](https://github.com/openshift/cluster-kube-controller-manager-operator/blob/master/pkg/operator/certrotationcontroller/satokensigner_controller.go#L58), I guess that could use an update too.

pkg/operator/boundsatokensignercontroller/controller.go
marun: I'm not sure what you mean by 'double empty lines' in this instance, can you explain?

pkg/operator/boundsatokensignercontroller/controller.go
marun: Added comment, PTAL

pkg/operator/boundsatokensignercontroller/controller.go
marun: Added comment, PTAL

pkg/operator/boundsatokensignercontroller/controller.go
marun: If its the operator secret that is changing, that is not a problem because it is not used directly by the apiserver instances. In the case of the operand secret, notice that promotion occurs only after the configmap has been successfully updated.

pkg/operator/boundsatokensignercontroller/controller.go
marun: Why would you suggest pruning? And how would you suggest that the operator determine which public keys are safe to delete? As per the example of the equivalent kcmo operator, I don't think pruning is something we can safely do automatically.

pkg/operator/boundsatokensignercontroller/controller.go
marun: Requeue is automatic on error. (see `processNextWorkItem`)

pkg/operator/resourcesynccontroller/resourcesynccontroller.go
marun: What's the precedent here? When writing the enhancement I assumed that copying the config map containing public keys to the global config namespace would make it less likely that the configmap would be lost. Is that true, and still a desired goal? Or is it ok that the source of public keys be the operand configmap?

bindata/v4.1.0/kube-apiserver/defaultconfig.yaml
```diff
@@ -106,6 +106,11 @@ servingInfo:
   requestTimeoutSeconds: 3600
 serviceAccountPublicKeyFiles:
   - /etc/kubernetes/static-pod-resources/configmaps/sa-token-signing-certs
+  - /etc/kubernetes/static-pod-certs/configmaps/bound-sa-token-signing-certs
+serviceAccountIssuer: auth.openshift.io
```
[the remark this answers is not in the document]
marun: Ok, it wasn't clear to me that a value that wasn't set by default at the API level should be documented for the API type.  Done.

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,296 @@
...
+	keySize                  = 2048
+	NextSigningKeySecretName = "next-service-account-signing-key"
+	SigningKeySecretName     = "service-account-signing-key"
+	PrivateKeyKey            = "service-account.key"
+	PublicKeyKey             = "service-account.pub"
+
+	TokenReadyAnnotation = "kube-apiserver.openshift.io/ready-to-use"
+	readyInterval        = 5 * time.Minute
+
+	CertConfigMapName = "bound-sa-token-signing-certs"
```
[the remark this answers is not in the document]
marun: No, but I'm being consistent with the name of the controller-manager equivalent (`sa-tokens-signing-certs`) which has effectively the same content.

pkg/operator/boundsatokensignercontroller/controller.go
marun: I think you're right - it should be sufficient to call `func() {c.queue.Add(workQueueKey)}` as the argument to `wait.Util`. Updated.

bindata/v4.1.0/kube-apiserver/defaultconfig.yaml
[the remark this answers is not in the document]
marun: Done.

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+func (c *BoundSATokenSignerController) sync() error {
+	signingSecret, err := c.secretClient.Secrets(operatorNamespace).Get(NextSigningKeySecretName, metav1.GetOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+	needKeypair := errors.IsNotFound(err) || len(signingSecret.Data[PrivateKeyKey]) == 0 || len(signingSecret.Data[PublicKeyKey]) == 0
+	if needKeypair {
+		newSecret, err := newSigningSecret()
```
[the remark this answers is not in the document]
marun: Done

pkg/operator/boundsatokensignercontroller/controller.go
marun: Updated comment, PTAL

pkg/operator/boundsatokensignercontroller/controller.go
marun: Can you point me to the etcd code in question? And should there be a corresponding update to the kcmo token controller?

re: dynamic key reads - afaict it's enough to just update the resource. Any changes to the resources/config that influence the state of apiserver pods prompt a redeployment of a single pod and only if that redeployment is successful will the change be rolled out to all pods.

bindata/v4.1.0/kube-apiserver/defaultconfig.yaml
deads2k: please pass as a flag, not as a struct value.  Same with all these values.

deads2k: You're looking for `apiServerArguments` above and add comments to help future me who won't remember this as well.

bindata/v4.1.0/kube-apiserver/defaultconfig.yaml
deads2k: I think we're ok not having bound tokens available from the bootstrap kubeapiserver.  we asked in the enhancement.

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,303 @@
...
+}
+
+func (c *BoundSATokenSignerController) sync() error {
+	signingSecret, err := c.secretClient.Secrets(operatorNamespace).Get(NextSigningKeySecretName, metav1.GetOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+	needKeypair := errors.IsNotFound(err) || len(signingSecret.Data[PrivateKeyKey]) == 0 || len(signingSecret.Data[PublicKeyKey]) == 0
+	if needKeypair {
+		klog.Infof("Creating a new signing secret for bound service account tokens.")
```
deads2k: if it's important enough for an info message, it's important enough for an event. If it isn't important enough for an event, then it belongs at a lower level.

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,303 @@
...
+	// contents of the operator secret be promoted to the operand
+	// secret for use by the apiserver. This interval is intended to
+	// ensure that apiservers have a chance to update to the latest
+	// public key configmap that can validate tokens issued by the new
+	// private key. If the secret was promoted immediately, and new
+	// tokens were issued by its private key, apiserver instances that
+	// did not have the latest public key would not be able to
+	// validate those new tokens.
+	TokenReadyAnnotation = "kube-apiserver.openshift.io/ready-to-use"
+	readyInterval        = 5 * time.Minute
```
deads2k: You have the ability to inspect the revisions to figure out which ones contains the key in question.  You can then inspect the levels on the nodes (the kubeapiservers.operator.openshift.io) to know if the nodes actually have the levels required.  No need for time.

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,303 @@
...
+		}
+		signingSecret, _, err = resourceapply.ApplySecret(c.secretClient, c.eventRecorder, newSecret)
+		if err != nil {
+			return err
+		}
+		// Ensure that the new keypair will be promoted after the ready interval is
+		// passed (see comment above the ready interval constant). Do not return
+		// immediately to ensure that the new public key can be set in the configmap
+		// in advance of promotion.
+		c.queue.AddAfter(workQueueKey, readyInterval+10*time.Second)
```
deads2k: if you make your decision based on levels of the configmap actually on the nodes and trigger based on updates to kubeapiserver.operator.openshift.io, you don't need to have this delay.

pkg/operator/boundsatokensignercontroller/controller.go
deads2k: For this code, you're even ok to avoid demanding a stable level, you just need each revision on nodes to include the cert for your key.

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+
+	kubeInformersForNamespaces.InformersFor(operatorNamespace).Core().V1().Secrets().Informer().AddEventHandler(ret.eventHandler())
+	kubeInformersForNamespaces.InformersFor(targetNamespace).Core().V1().Secrets().Informer().AddEventHandler(ret.eventHandler())
+	kubeInformersForNamespaces.InformersFor(targetNamespace).Core().V1().ConfigMaps().Informer().AddEventHandler(ret.eventHandler())
+
+	return ret
+}
+
+func (c *BoundSATokenSignerController) sync() error {
```
deads2k: please break this function into logical bits with no data based between them.  I see 

1. create a secret if needed. depends on nothing outside the secrets
2. update configmap if needed.  This can retrieve the current secret.  A stale lister will always get another event notification, so staleness of a cache doesn't matter and this can then be contained.
3. promoting a key.  This is based on the current secret in the operator namespace, the current secret in the operand namespace, the revisions on the nodes, the content of the configmaps on those nodes.  All of which can be cached.

If any step has an error, the other steps should still be run.

pkg/operator/boundsatokensignercontroller/controller.go
marun: Which would you prefer - even or lower level log?

pkg/operator/boundsatokensignercontroller/controller.go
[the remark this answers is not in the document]
stlaz: interesting, ok

pkg/operator/boundsatokensignercontroller/controller.go
marun: Done

pkg/operator/boundsatokensignercontroller/controller.go
marun: Done

pkg/operator/boundsatokensignercontroller/controller.go
marun: Done

pkg/operator/boundsatokensignercontroller/controller.go
marun: Done

pkg/operator/boundsatokensignercontroller/controller.go
marun: Done

pkg/operator/resourcesynccontroller/resourcesynccontroller.go
marun: I've updated the enhancement to have the configmap written only to the operand namespace. Please comment [there](https://github.com/openshift/enhancements/pull/150/files#diff-93464a309fcc6be1fb0cca2c92d077b2R102).

pkg/operator/boundsatokensignercontroller/controller.go
marun: No longer relevant after the refactor into separate methods.

pkg/operator/boundsatokensignercontroller/controller.go
marun: Removed.

pkg/operator/boundsatokensignercontroller/controller.go
marun: Removed

pkg/operator/boundsatokensignercontroller/controller.go
marun: Hopefully the functional separation makes my intent clear?

pkg/operator/boundsatokensignercontroller/controller.go
marun: Is this still a problem now that actual state is considered rather than just waiting for a random amount of time?

pkg/operator/boundsatokensignercontroller/controller.go
marun: Done, PTAL.

pkg/operator/boundsatokensignercontroller/controller.go
marun: No longer relevent

bindata/v4.1.0/kube-apiserver/defaultconfig.yaml
marun: Done

bindata/v4.1.0/kube-apiserver/defaultconfig.yaml
sttts: My question was more whether this path is a valid one during bootstrap or whether the process dies if not.

pkg/operator/boundsatokensignercontroller/controller.go
sttts:
```
+
+
```

bindata/v4.1.0/kube-apiserver/defaultconfig.yaml
marun: Ah, ok.  Will test tomorrow. Maybe the operator will have to detect that it is past the bootstrap phase and compose the config accordingly in code?

pkg/operator/boundsatokensignercontroller/controller.go
marun: That doesn't appear now, though?  I didn't see it when you commented, and I'm not sure how it could have been committed given that my editor runs gofmt every time I save.

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,402 @@
...
+
+func configMapHasValue(configMap *corev1.ConfigMap, desiredValue string) bool {
+	hasValue := false
+	for _, value := range configMap.Data {
+		if value == desiredValue {
+			hasValue = true
+			break
+		}
+	}
+	return hasValue
```
stlaz: return false

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,402 @@
...
+		},
+	)
+	return keyinPem, nil
+}
+
+func configMapHasValue(configMap *corev1.ConfigMap, desiredValue string) bool {
+	hasValue := false
+	for _, value := range configMap.Data {
+		if value == desiredValue {
+			hasValue = true
```
stlaz: return true

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,402 @@
...
+	// current public key. If any configmap for any given revision is missing or does
+	// not contain the public key, assume the public key is not present on that node.
+	for _, revision := range uniqueRevisions {
+		configMapNameWithRevision := fmt.Sprintf("%s-%d", PublicKeyConfigMapName, revision)
+		configMap, err := c.configMapClient.ConfigMaps(operatorclient.TargetNamespace).Get(configMapNameWithRevision, metav1.GetOptions{})
+		if err != nil {
+			return false, err
+		}
+		found := false
+		for _, value := range configMap.Data {
```
stlaz: `configMapHasValue` ?

bindata/v4.1.0/kube-apiserver/defaultconfig.yaml
sttts: There are two override yaml files, one for the bootstrap phase, one for after. Just set sensible values for each.

pkg/operator/boundsatokensignercontroller/controller.go
sttts: Will read the new code. Looking at the deployed state should be enough.

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,402 @@
...
+	syncMethods := []func() error{
+		c.ensureOperatorSigningSecret,
+		c.ensurePublicKeyConfigMap,
+		c.ensureOperandSigningSecret,
+	}
+	for _, syncMethod := range syncMethods {
+		err := syncMethod()
+		if err != nil {
+			utilruntime.HandleError(err)
+			syncFailed = true
```
sttts: usually we aggregate the errors in a slice and then return `NewAggregate(errs)`.

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,402 @@
...
+			_, ok := configMap.Data[possibleKey]
+			if !ok {
+				nextKeyKey = possibleKey
+				break
+			}
+			nextKeyIndex += 1
+		}
+
+		// Ensure the configmap is updated with the current public key
+		configMap.Data[nextKeyKey] = currPublicKey
```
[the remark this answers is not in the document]
marun: Good catch. I have yet to internalize this new world where a 'Get' call is to the cache vs a direct API call.

Done.

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,402 @@
...
+		configMap, _, err = resourceapply.ApplyConfigMap(c.configMapClient, c.eventRecorder, configMap)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// ensureOperandSigningSecret ensures that the current signing key is
```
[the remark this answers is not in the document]
marun: Done

pkg/operator/boundsatokensignercontroller/controller.go
marun: Done

pkg/operator/boundsatokensignercontroller/controller.go
marun: Done

pkg/operator/boundsatokensignercontroller/controller.go
marun: Done

pkg/operator/boundsatokensignercontroller/controller.go
marun: I'd prefer to log the error raw to simplify troubleshooting. afaict wrapping errors in any way (including `NewAggregate`) changes the file+line that is logged. Is that ok, or do you require that I use `NewAggregate`?

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,402 @@
...
+
+	// Retrieve the operand signing secret
+	operandSecret, err := c.secretClient.Secrets(targetNamespace).Get(SigningKeySecretName, metav1.GetOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
+
+	// If operand secret matches the operator secret, all done
+	operandSecretUpToDate := (operandSecret != nil &&
+		bytes.Equal(operandSecret.Data[PublicKeyKey], operatorSecret.Data[PublicKeyKey]) &&
```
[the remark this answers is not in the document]
marun: The operator secret can never be nil since all errors on retrieval (including `NotFound`) result  in exit. The operand secret is checked for nil. Am I missing something?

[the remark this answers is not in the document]
stlaz: ah, my bad, I managed to overlook the statement on the above line. Sorry

marun:
Updated with fixup, PTAL

pkg/operator/resourcesynccontroller/resourcesynccontroller.go
marun: Done

stlaz:
this now looks good to me (commits need squashing). Leaving final LGTM to @deads2k / @sttts

marun:
Operator e2e is passing indicating that the bound tokens are working. Squashing in case it's ready for approval.

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,407 @@
...
+
+const (
+	workQueueKey = "key"
+
+	operatorNamespace = operatorclient.OperatorNamespace
+	targetNamespace   = operatorclient.TargetNamespace
+
+	keySize                  = 2048
+	NextSigningKeySecretName = "next-bound-service-account-signing-key"
+	SigningKeySecretName     = "bound-service-account-signing-key"
```
sttts: why do we need both, old and new?

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,407 @@
...
+	}
+
+	kubeInformersForNamespaces.InformersFor(operatorNamespace).Core().V1().Secrets().Informer().AddEventHandler(ret.eventHandler())
+	kubeInformersForNamespaces.InformersFor(targetNamespace).Core().V1().Secrets().Informer().AddEventHandler(ret.eventHandler())
+	kubeInformersForNamespaces.InformersFor(targetNamespace).Core().V1().ConfigMaps().Informer().AddEventHandler(ret.eventHandler())
+
+	return ret
+}
+
+func (c *BoundSATokenSignerController) sync() bool {
```
sttts: the inverse is more standard: true = ok

pkg/operator/boundsatokensignercontroller/controller.go
sttts: nvmd. It is because you have to deploy the public key first, i.e. the signing key is then promoted. Maybe a comment here helps to understand.

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,407 @@
...
+			syncFailed = true
+		}
+	}
+	return syncFailed
+}
+
+// ensureOperatorSigningSecret ensures the existence of a secret in the operator
+// namespace containing an RSA keypair used for signing and validating bound service
+// account tokens.
+func (c *BoundSATokenSignerController) ensureOperatorSigningSecret() error {
```
sttts: ensure**Next**OperatorSigningSecret

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,407 @@
...
+func (c *BoundSATokenSignerController) eventHandler() cache.ResourceEventHandler {
+	return cache.ResourceEventHandlerFuncs{
+		AddFunc:    func(obj interface{}) { c.queue.Add(workQueueKey) },
+		UpdateFunc: func(old, new interface{}) { c.queue.Add(workQueueKey) },
+		DeleteFunc: func(obj interface{}) { c.queue.Add(workQueueKey) },
+	}
+}
+
+// newSigningSecret creates a new secret populated with a new keypair.
+func newSigningSecret() (*corev1.Secret, error) {
```
sttts: new**Next**SigningSecret

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,407 @@
...
+	hasKey := configMapHasValue(configMap, currPublicKey)
+	if !hasKey {
+		// Increment until a unique name is found to ensure that the new public key
+		// does not overwrite an existing one. Except where key revocation is
+		// involved (which would require manual deletion of the verifying public
+		// key), existing public keys in the configmap should be maintained to
+		// minimize the potential for not being able to validate issued tokens.
+		nextKeyIndex := len(configMap.Data) + 1
+		nextKeyKey := ""
+		for len(nextKeyKey) == 0 {
```
sttts: is the len comparison even needed with the break?

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,407 @@
...
+			_, ok := configMap.Data[possibleKey]
+			if !ok {
+				nextKeyKey = possibleKey
+				break
+			}
+			nextKeyIndex += 1
+		}
+
+		// Ensure the configmap is updated with the current public key
+		configMap.Data[nextKeyKey] = currPublicKey
```
sttts: Can `configMap.Data` be nil?

pkg/operator/boundsatokensignercontroller/controller.go
```diff
@@ -0,0 +1,407 @@
...
+
+	if operandSecret == nil {
+		// If the operand secret is missing, it must be created to ensure the
+		// installer can proceed regardless of whether public keys have already been
+		// synced to the master nodes.
+		syncRequired = true
+	} else {
+		// Update the operand secret only if the current public key has been synced to
+		// all nodes.
+		syncRequired, err = c.publicKeySyncedToAllNodes(currPublicKey)
```
sttts: s/syncRequired/syncAllowed/

bindata/v4.1.0/kube-apiserver/boundsatokenconfig.yaml
```diff
@@ -0,0 +1,17 @@
...
+  - auth.openshift.io
+  api-audiences:
+  - auth.openshift.io
+  service-account-signing-key-file:
+  - /etc/kubernetes/static-pod-certs/secrets/bound-service-account-signing-key/service-account.key
+serviceAccountPublicKeyFiles:
+  - /etc/kubernetes/static-pod-resources/configmaps/sa-token-signing-certs
+  # The following path contains the public keys needed to verify bound sa
+  # tokens. This is only supported post-bootstrap.
+  - /etc/kubernetes/static-pod-resources/configmaps/bound-sa-token-signing-certs
```
sttts: just copy these into the overrides file

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -246,6 +247,7 @@ func manageKubeAPIServerConfig(client coreclientv1.ConfigMapsGetter, recorder ev
 		"config.yaml",
 		specialMergeRules,
 		defaultConfig,
+		boundSATokenConfig,
```
sttts: didn't we have the overrides yaml here as well?

marun: afaict the overrides yaml are only for bootstrap:

https://github.com/openshift/cluster-kube-apiserver-operator/tree/master/bindata/bootkube/config

I don't know why there are 2 files for bootstrap.  I do know that putting these options in the either of the bootstrap overrides will break bootstrapping because the bound token keypair is only created post-bootstrap.

bindata/v4.1.0/kube-apiserver/boundsatokenconfig.yaml
marun: Which overrides file?  afaict the only ones available are for bootstrap.  Would you prefer that I add another post-bootstrap override file rather than adding this feature-specific one?

pkg/operator/boundsatokensignercontroller/controller.go
marun: Done

pkg/operator/boundsatokensignercontroller/controller.go
marun: Done. Now always initialized if necessary.

pkg/operator/boundsatokensignercontroller/controller.go
marun: Done

pkg/operator/boundsatokensignercontroller/controller.go
marun: Done

pkg/operator/boundsatokensignercontroller/controller.go
marun: Done

pkg/operator/boundsatokensignercontroller/controller.go
marun: Done. All the more important given that the call site was assuming true = ok too.

pkg/operator/boundsatokensignercontroller/controller.go
marun: Done

marun:
@sttts Updated, PTAL

bindata/v4.1.0/kube-apiserver/boundsatokenconfig.yaml
sttts: https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/bindata/bootkube/config/config-overrides.yaml

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
sttts: https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/bindata/bootkube/config/config-overrides.yaml

bindata/v4.1.0/kube-apiserver/boundsatokenconfig.yaml
marun: This file has been renamed to config-overrides.yaml as requested.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
marun: I've posted a new PR to document the requirement to put post-bootstrap overrides in the renamed file: https://github.com/openshift/cluster-kube-apiserver-operator/pull/731

marun:
@sttts Updated, PTAL

marun:
@damemi I've added the `DO NOT MERGE Bump revision limit in TestRevisionLimit` commit to bump the apparently magic revision limit in the TestRevisionLimit. This PR is likely increasing the number of revisions in the process of coordinating the keypair configuration required to enable bound sa tokens. I'd appreciate your input as to whether changing the magic number is an acceptable workaround of if more involved changes are required.

test/e2e/operator_test.go
```diff
@@ -56,10 +56,10 @@ func TestRevisionLimits(t *testing.T) {
 
 	totalRevisionLimit := operatorSpec.SucceededRevisionLimit + operatorSpec.FailedRevisionLimit
 	if operatorSpec.SucceededRevisionLimit == 0 {
-		totalRevisionLimit += 5
+		totalRevisionLimit += 10
```
sttts: For a follow-up: @damemi you wrote this original test. Why was this 5? Magic number?

sttts: I now get the +=5 I believe (it's the default value?). But why does the loop below not wait for the revision pruning controller to do its work, but errors the test immediately?

test/e2e/operator_test.go
```diff
@@ -77,7 +77,9 @@ func TestRevisionLimits(t *testing.T) {
 		// are InProgress or Unknown (since these do not count toward failed or succeeded), which could indicate zombie revisions.
 		// Check total+1 to account for possibly a current new revision that just hasn't pruned off the oldest one yet.
 		if len(newRevisions) > int(totalRevisionLimit)+1 {
-			t.Errorf("more revisions (%v) than total allowed (%v): %+v", len(revisions), totalRevisionLimit, revisions)
+			// TODO(marun) If number of revisions has been exceeded, need to give time for the pruning controller to
```
sttts: cc @damemi

PR #725, opened 2020-01-22
Bug 1794309: Copy default SCC into release image

tkashem:
Currently the default SCC files in 'manifests' folder are not being
copied into the cvo payload. Copy the default SCC in .json files so
CVO can apply these on the cluster.


sttts:
@tkashem can we backport this to 4.2, compare https://bugzilla.redhat.com/show_bug.cgi?id=1803956 ?

PR #726, opened 2020-01-23
bump library-go and disable IPv6DualStack feature gate temporarily 

mfojtik:
We need to disable IPv6DualStack feature gate temporarily until https://github.com/kubernetes/kubernetes/pull/86895 land and is backported, otherwise the old clients will fail validation when they update Service objects.


go.mod
```diff
@@ -10,23 +10,26 @@ require (
...
 	github.com/spf13/cobra v0.0.5
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.4.0
-	k8s.io/api v0.17.0
-	k8s.io/apimachinery v0.17.1-beta.0
-	k8s.io/apiserver v0.17.0
-	k8s.io/client-go v0.17.0
-	k8s.io/component-base v0.17.0
+	golang.org/x/net v0.0.0-20200114155413-6afb5195e5aa // indirect
+	k8s.io/api v0.17.2
```
mfojtik: @deads2k @sttts this happened after `go get -u github.com/openshift/api@master` ... i wonder if we care.

PR #705, opened 2019-12-23
removes unused UsingDeprecatedAPIAppsV1Beta1 and UsingDeprecatedAPIAppsV1Beta2 alerts

p0lyn0mial:
Since `apps/v1beta1 and apps/v1beta2` are disabled in Kube `1.17` we can safely remove the mentioned alerts as any attempt to create a resource will immediately return an error.

I run the following commands on a quite recent cluster:
```
k get --raw /apis/apps/v1beta2
k get --raw /apis/apps/v1beta1

Error from server (NotFound): the server could not find the requested resource
```

Client/Server version:
```
k version                         
Client Version: version.Info{Major:"", Minor:"", GitVersion:"v0.0.0-master+$Format:%h$", GitCommit:"$Format:%H$", GitTreeState:"", BuildDate:"1970-01-01T00:00:00Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"17+", GitVersion:"v1.17.0", GitCommit:"8a9ae19", GitTreeState:"clean", BuildDate:"2019-12-21T07:03:31Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"linux/amd64"}
```

PR #730, opened 2020-01-23
Bug 1793850: Update UsingDeprecatedAPIExtensionsV1Beta1 alert

cblecker:
- Updated message that describes the issue more clearly
- Exclude velero-server client
- Change alert to use an increase in the last 24h
- Change severity to none

/assign @soltysh 
/cc @sttts @brancz @tnozicka

### DIFF
--- a/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
+++ b/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
@@ -110,8 +110,18 @@ spec:
     rules:
     - alert: UsingDeprecatedAPIExtensionsV1Beta1
       annotations:
-        message: A client in the cluster is using deprecated extensions/v1beta1 API that will be removed soon.
+        summary: "A client is using a deprecated {{"{{"}} $labels.group {{"}}"}}/{{"{{"}} $labels.version {{"}}"}} API version"
+        message: |
+          A client in the cluster has used a deprecated {{"{{"}} $labels.group {{"}}"}}/{{"{{"}} $labels.version {{"}}"}} API version of the {{"{{"}} $labels.resource {{"}}"}} resource in the last 24 hours.
+          This API version will be removed soon; please migrate to the current version of this API.
       expr: |
-        apiserver_request_count{group="extensions",version="v1beta1",resource!~"ingresses|",client!~"hyperkube/.*|cluster-policy-controller/.*"}
+        increase(
+          apiserver_request_total{
+            group="extensions",
+            version="v1beta1",
+            resource!~"ingresses|",
+            client!~"hyperkube/.*|cluster-policy-controller/.*|velero-server/.*"
+          }[24h]
+        ) > 0
       labels:
-        severity: warning
+        severity: info


cblecker:
This and #705 should be cherrypicked back to release-4.3

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -110,8 +110,18 @@ spec:
...
       annotations:
-        message: A client in the cluster is using deprecated extensions/v1beta1 API that will be removed soon.
+        summary: "A client is using a deprecated {{"{{"}} $labels.group {{"}}"}}/{{"{{"}} $labels.version {{"}}"}} API version"
+        message: |
+          A client in the cluster has used a deprecated {{"{{"}} $labels.group {{"}}"}}/{{"{{"}} $labels.version {{"}}"}} API version of the {{"{{"}} $labels.resource {{"}}"}} resource in the last 24 hours.
+          This API version will be removed soon; please migrate to the current version of this API.
       expr: |
-        apiserver_request_count{group="extensions",version="v1beta1",resource!~"ingresses|",client!~"hyperkube/.*|cluster-policy-controller/.*"}
+        increase(
+          apiserver_request_count{
```
LiliC: Please note that `apiserver_request_count` is a deprecated metric and we are dropping those metrics from openshift as well. We should be using `apiserver_request_total` instead.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -110,8 +110,18 @@ spec:
...
+            group="extensions",
+            version="v1beta1",
+            resource!~"ingresses|",
+            client!~"hyperkube/.*|cluster-policy-controller/.*|velero-server/.*"
+          }[24h]
+        ) > 0
       labels:
-        severity: warning
+        severity: none
```
LiliC: why severity none? 🤔

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -110,8 +110,18 @@ spec:
...
+          A client in the cluster has used a deprecated {{"{{"}} $labels.group {{"}}"}}/{{"{{"}} $labels.version {{"}}"}} API version of the {{"{{"}} $labels.resource {{"}}"}} resource in the last 24 hours.
+          This API version will be removed soon; please migrate to the current version of this API.
       expr: |
-        apiserver_request_count{group="extensions",version="v1beta1",resource!~"ingresses|",client!~"hyperkube/.*|cluster-policy-controller/.*"}
+        increase(
+          apiserver_request_count{
+            group="extensions",
+            version="v1beta1",
+            resource!~"ingresses|",
+            client!~"hyperkube/.*|cluster-policy-controller/.*|velero-server/.*"
```
tnozicka: why velero-server? the only reason for the others was using discovery to spin informers

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
soltysh: I remember that severity should've stayed, from the discussion we were to only change the metrics to be rate based.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
soltysh: @LiliC drop when? This PR is meant for 4.3 most importantly. If you're dropping this in 4.4 - fine change, if in the next we'll fix it then.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -110,8 +110,18 @@ spec:
...
+          }[24h]
```
soltysh: Didn't you want to have 5m here? Why going with 24h?

soltysh: Didn't you want to have 5m here? Why going with 24h?

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
LiliC: Yes we are dropping this metric start with 4.4. It has been deprecated since 1.14 k8s version and has a replacement which is `apiserver_request_total ` so just changing it to that should just work ™️ .

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
cblecker: In the OSD context, Critical and Warnings are fed into monitoring for cluster health. This alert isn't particularly actionable and doesn't actually impact the cluster's current health -- it's an "info" type alert that you should do something prior to the next upgrade, but the current cluster health isn't degraded. None in this case, I believe is appropriate.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
cblecker: Velero is a piece of backup software, and it scrapes every API type available on the system in order to back them up. This causes the alert to fire in OSD and never resolve. @sttts mentioned that ignoring this client may be a good idea.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
cblecker: So my understanding of this alert, is we want it to surface to the administrators of the cluster that some client is making a call that they will no longer be able to in 4.4+. Right now, this alert alarms and never resolves -- which on the one hand accomplishes this goal, but on the other hand doesn't allow for the alert to resolve if the condition stops.

I'm okay with 5m, but I think that lessons the usefulness of this alert, as the administrator would need to catch it within that 5 minute period. For clients that aren't calling the API that frequently (cron jobs, deploy pipelines, etc) it might be harder to catch this alert.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
cblecker: Note that with a severity of none, it still surfaces in the on-cluster UI

brancz: > alert isn't particularly actionable and doesn't actually impact the cluster's current health

Precisely why this alert shouldn't exist in the first place.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
cblecker: Made this change, and I verified the alert still fires as expected in a test cluster.

soltysh: :+1: @LiliC does the new metric will be working fine in 4.3 as well? Just ensuring before we switch it in the backport too, or the backport has to be different?

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
soltysh: The reasoning behind is that we need to yell loudly at people using this api, b/c otherwise after next upgrade this won't work. I can be convinced and I was originally going after an info level, but none such exists atm. That's why the warning, and it's important one to have.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
soltysh: yeah, that's a minor problem I can live with.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
soltysh: I don't have any strong preferences either way, I was just asking :wink:

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
cblecker: I'd also note that there is no action that we or velero need to take on this. When the API endpoint goes away in 4.4, it won't be there to scrape anymore.

cblecker:
@soltysh Updated to info. It's important that this *not* be a "warning" because the clusters health is not actually degraded.

cblecker:
Screenshot 2020-01-24 09 19 56
Example of how this looks in the on-cluster UI

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
cblecker: My test cluster is running 4.3.0 GA, so I can confirm that this works.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
LiliC: But in 4.4 its not firing, otherwise we would have had failed jobs which confirms @tnozicka:
> note that for 4.4 the serving is disabled so we should drop this check entirely there

Do we want our customers to react on this alert? And if yes how?

soltysh: > Do we want our customers to react on this alert? And if yes how?

If this fires in 4.3 it means you have clients accessing endpoints that will be removed in the next release and they should be updating that code.

soltysh:
Superseded with https://github.com/openshift/cluster-kube-apiserver-operator/pull/741.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
paulfantom: > I can be convinced and I was originally going after an info level, but none such exists atm.

They do exist. OLM uses `info` level alerts and we also (will) have special rules for handling this severity.

> If this fires in 4.3 it means you have clients accessing endpoints that will be removed in the next release and they should be updating that code.

This is not immediately actionable from operator point of view.

PR #731, opened 2020-01-24
Remove unused post-bootstrap overrides from bootstrap bindata

marun:
Post-bootstrap kube-apiserver config is rendered by TargetConfigController.manageKubeAPIServerConfig and does not consider the content of bindata/bootkube/config/config-overrides.yaml. That file still needs to exist, though, to maintain compatibility with the library-go type that renders the configuration.

/cc @sttts


sttts:
/close

in favor of https://github.com/openshift/cluster-kube-apiserver-operator/pull/718.

PR #733, opened 2020-01-27
bug 1797657: increase the max-requests limits

deads2k:
more in line with real values.

/assign @smarterclayton


sttts:
@deads2k What are real values? Why these?

smarterclayton:
The limits were chosen during 3.x pre etcd2 where we had demonstrated hard boundaries on CPU use.  Based on all data I have in etcd3 period, a 3x increase is not an issue and any bug we hit where load is not effectively transferred would be a bug in the platform, not due to this.

/approve

smarterclayton:
Basically 1k was chosen arbitrarily based on a CPU usage number in the early 3.x time period.  We have reclaimed somewhere between 5x and 10x CPU with etcd3, protobuf, improvements to internal efficiency, so a 3x bump gives us headroom without significantly increasing the risk of a failure due to excessive queries (we are likely to better respond with backpressure at multiple levels)

sttts:
/lgtm

This will help for reasonably sized cluster. It might still not be enough for very dense clusters.

PR #734, opened 2020-01-27
Revert 718+727: audience support because TokenReview API is broken with legacy SAs

sttts:
Legacy service accounts cannot be verified through the TokenReview API if audiences are in use. This is an upstream bug we have to fix first.


deads2k:
/lgtm
/hold

holding to see if this fixes e2e-aws-serial

marun:
The proposed fix for the issue prompting this reversion [appears to work](https://github.com/openshift/origin/pull/24461#issuecomment-579014066).

PR #737, opened 2020-01-27
Bug 1798214: Add a PrometheusRule to aggregate in-flight requests

tkashem:
- Add a PrometheusRule to aggregate the metric for in-flight requests into 4 time series per cluster - to monitor number of requests in flight for `kube-apiserver` and `openshift-apiserver` for `mutating` and `nonmutating` request type.

The new rule will provide us with insights on how loaded our api server(s) are globally.

### DIFF
--- a/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
+++ b/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
@@ -83,6 +83,10 @@ spec:
       regex: network_plugin_operations_latency_microseconds|sync_proxy_rules_latency_microseconds|rest_client_request_latency_seconds
       sourceLabels:
       - __name__
+    relabelings:
+    - action: replace
+      targetLabel: apiserver
+      replacement: kube-apiserver
     port: https
     scheme: https
     tlsConfig:
@@ -115,3 +119,13 @@ spec:
         apiserver_request_count{group="extensions",version="v1beta1",resource!~"ingresses|",client!~"hyperkube/.*|cluster-policy-controller/.*"}
       labels:
         severity: warning
+  - name: apiserver-requests-in-flight
+    rules:
+      # We want to capture requests in-flight metrics for kube-apiserver and openshift-apiserver.
+      # apiserver='kube-apiserver' indicates that the source is kubernetes apiserver.
+      # apiserver='openshift-apiserver' indicates that the source is openshift apiserver.
+      # The subquery aggregates by apiserver and request kind. requestKind is {mutating|readOnly}
+      # The following query gives us maximum peak of the apiserver concurrency over a 2-minute window.
+    - record: cluster:apiserver_current_inflight_requests:sum:max_over_time:2m
+      expr: |
+        max_over_time(sum(apiserver_current_inflight_requests{apiserver=~"openshift-apiserver|kube-apiserver"}) by (apiserver,requestKind)[2m:])


tkashem:
corresponding pr in telemeter - https://github.com/openshift/telemeter/pull/304

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -115,3 +115,8 @@ spec:
         apiserver_request_count{group="extensions",version="v1beta1",resource!~"ingresses|",client!~"hyperkube/.*|cluster-policy-controller/.*"}
       labels:
         severity: warning
+  - name: kube-apiserver-max-in-flight
+    interval: 5s
+    rules:
+    - record: cluster:apiserver_current_inflight_requests:sum
+      expr: 'sum(apiserver_current_inflight_requests{namespace=~"openshift-apiserver|default"}) by (requestKind,namespace)'
```
sttts: This namespace is probably the one of the requests, not of the apiserver container.

tkashem: @sttts I just checked the documentation - I think `namespace` points to the source, it's the namespace of the `Service`, `Endpoints` from which the metrics are scraped. In case of `kube-apiserver` it is the Endpoint `kubernetes` in `default` namespace.

s-urbaniak: `apiserver_current_inflight_requests` does not declare the `namespace` label itself, hence its value should indeed be the namespace location of the apiserver container.

That metric only sets `requestKind` as per https://github.com/kubernetes/kubernetes/blob/703a719557d8208f781c88c16816158cf0c98d22/staging/src/k8s.io/apiserver/pkg/endpoints/metrics/metrics.go#L143-L152 and https://github.com/kubernetes/kubernetes/blob/703a719557d8208f781c88c16816158cf0c98d22/staging/src/k8s.io/apiserver/pkg/endpoints/metrics/metrics.go#L203-L206

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -115,3 +115,11 @@ spec:
         apiserver_request_count{group="extensions",version="v1beta1",resource!~"ingresses|",client!~"hyperkube/.*|cluster-policy-controller/.*"}
       labels:
         severity: warning
+  - name: kube-apiserver-requests-in-flight
+    rules:
+    - record: cluster:apiserver_current_inflight_requests:sum
+      expr: |
+        label_replace(label_replace(sum(apiserver_current_inflight_requests{namespace=~"openshift-apiserver|default"}) by (requestKind,namespace), "apiserver", "openshift", "namespace", "openshift-apiserver"), "apiserver", "kubernetes", "namespace", "default")
```
sttts: can we add a comment explaining this magic?

sttts:
Lgtm.
/approve

Would like to have somebody with more Prometheus knowledge for a final look.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -115,3 +115,18 @@ spec:
         apiserver_request_count{group="extensions",version="v1beta1",resource!~"ingresses|",client!~"hyperkube/.*|cluster-policy-controller/.*"}
       labels:
         severity: warning
+  - name: apiserver-requests-in-flight
```
sttts: @smarterclayton are these approved now for telemetry?

tkashem:
> Would like to have somebody with more Prometheus knowledge for a final look.

I have asked in #forum-monitoring to review this pr.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
smarterclayton: Yes, you have my approval for 4 series for this

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -115,3 +115,18 @@ spec:
...
+      # We want to capture metrics for kube-apiserver and openshift-apiserver.
+      # namespace='default' indicates that the source is kube-apiserver.
+      # namespace='openshift-apiserver' indicates that the source is openshift-apiserver.
+      # After we aggregate, we add a new label 'apiserver' and map the values from existing 'namespace' label
+      # by chaining label_replace as follows:
+      #   'apiserver:kubernetes' if namespace = default
+      #   'apiserver:openshift' if namespace = openshift-apiserver
+    - record: cluster:apiserver_current_inflight_requests:sum
+      expr: |
+        label_replace(label_replace(sum(apiserver_current_inflight_requests{namespace=~"openshift-apiserver|default"}) by (requestKind,namespace), "apiserver", "openshift", "namespace", "openshift-apiserver"), "apiserver", "kubernetes", "namespace", "default")
```
brancz: If this labeling is confusing/bothersome, I would prefer to fix that at discovery/collection time not just for telemetry. These metrics should be just as useful and comprehensible for someone querying them in-cluster as when we're sending them and if it's confusing for us then certainly it will be for users.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -115,3 +115,18 @@ spec:
...
+    - record: cluster:apiserver_current_inflight_requests:sum:max_over_time
+      expr: |
+        max_over_time(cluster:apiserver_current_inflight_requests:sum[1m])
```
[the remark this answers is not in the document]
tkashem: @brancz Sure, will use `2m`. Can you explain why it would fail or point me to any doc that describes these internals (for my own education)?

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -115,3 +115,18 @@ spec:
...
+      # namespace='default' indicates that the source is kube-apiserver.
+      # namespace='openshift-apiserver' indicates that the source is openshift-apiserver.
+      # After we aggregate, we add a new label 'apiserver' and map the values from existing 'namespace' label
+      # by chaining label_replace as follows:
+      #   'apiserver:kubernetes' if namespace = default
+      #   'apiserver:openshift' if namespace = openshift-apiserver
+    - record: cluster:apiserver_current_inflight_requests:sum
+      expr: |
+        label_replace(label_replace(sum(apiserver_current_inflight_requests{namespace=~"openshift-apiserver|default"}) by (requestKind,namespace), "apiserver", "openshift", "namespace", "openshift-apiserver"), "apiserver", "kubernetes", "namespace", "default")
+    - record: cluster:apiserver_current_inflight_requests:sum:max_over_time
```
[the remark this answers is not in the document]
tkashem: Sure, you mean `cluster:apiserver_current_inflight_requests:sum:max_over_time:2m`?  I originally had it this way, but didn't see any naming similar to this in `metrics.jsonnet`

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
tkashem: I agree, so I have to change the corresponding `ServiceMonitor` object, correct? I will try this.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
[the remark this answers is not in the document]
brancz: You want to account for at least one scrape failure and still end up with at least two sample. Failures can happen for various reasons of network connectivity or other (intended) interruptions like evictions etc.

tkashem:
Related PR - https://github.com/openshift/cluster-openshift-apiserver-operator/pull/307. This adds the label `apiserver:openshift-apiserver` for all time series of `apiserver_current_inflight_requests` exposed by the kube-apiserver.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
brancz: Correct

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -83,6 +83,12 @@ spec:
       regex: network_plugin_operations_latency_microseconds|sync_proxy_rules_latency_microseconds|rest_client_request_latency_seconds
       sourceLabels:
       - __name__
+    - action: replace
```
brancz: This needs to be just a relabeling rule ("relabelings"), not a metric relabeling rule, then you also don't need to match on the metric label name at all, every metric will have the label.

tkashem: @brancz changed it to use `relabelings`.

PR #740, opened 2020-01-28
Drop extensions/v1beta1 alert

soltysh:
Replaces https://github.com/openshift/cluster-kube-apiserver-operator/pull/730 in master where the alert is not relevant any more.

/assign @tnozicka 
/cc @LiliC @cblecker


soltysh:
Wrong idea.
/close

PR #741, opened 2020-01-28
Bug 1795617:  Remove entirely deprecated alerts

soltysh:
This picks commit from https://github.com/openshift/cluster-kube-apiserver-operator/pull/730 and applies the same pattern to the remaining alerts.

/assign @tnozicka 

/cc @LiliC @cblecker

### DIFF
--- a/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
+++ b/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
@@ -68,36 +68,3 @@ spec:
     matchLabels:
       component: apiserver
       provider: kubernetes
----
-apiVersion: monitoring.coreos.com/v1
-kind: PrometheusRule
-metadata:
-  name: kube-apiserver
-  namespace: openshift-kube-apiserver
-  annotations:
-    exclude.release.openshift.io/internal-openshift-hosted: "true"
-spec:
-  groups:
-  - name: using-deprecated-apis
-    rules:
-    - alert: UsingDeprecatedAPIAppsV1Beta1
-      annotations:
-        message: A client in the cluster is using deprecated apps/v1beta1 API that will be removed soon.
-      expr: |
-        apiserver_request_count{group="apps",version="v1beta1"}
-      labels:
-        severity: warning
-    - alert: UsingDeprecatedAPIAppsV1Beta2
-      annotations:
-        message: A client in the cluster is using deprecated apps/v1beta2 API that will be removed soon.
-      expr: |
-        apiserver_request_count{group="apps",version="v1beta2"}
-      labels:
-        severity: warning
-    - alert: UsingDeprecatedAPIExtensionsV1Beta1
-      annotations:
-        message: A client in the cluster is using deprecated extensions/v1beta1 API that will be removed soon.
-      expr: |
-        apiserver_request_count{group="extensions",version="v1beta1",resource!~"ingresses|",client!~"hyperkube/.*|cluster-policy-controller/.*"}
-      labels:
-        severity: warning


manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -82,22 +82,48 @@ spec:
...
+        message: |
+          A client in the cluster has used a deprecated {{"{{"}} $labels.group {{"}}"}}/{{"{{"}} $labels.version {{"}}"}} API version of the {{"{{"}} $labels.resource {{"}}"}} resource in the last 24 hours.
+          This API version will be removed soon; please migrate to the current version of this API.
       expr: |
-        apiserver_request_count{group="extensions",version="v1beta1",resource!~"ingresses|",client!~"hyperkube/.*|cluster-policy-controller/.*"}
+        increase(
+          apiserver_request_total{
+            group="extensions",
+            version="v1beta1",
+            resource!~"ingresses|",
```
tnozicka: do we want to add ingress here as well, it is deprecated, just still served upstream

tnozicka:
pls clean the commits, modulo the ingress question it lgtm

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -82,22 +82,47 @@ spec:
...
-        apiserver_request_count{group="extensions",version="v1beta1",resource!~"ingresses|",client!~"hyperkube/.*|cluster-policy-controller/.*"}
+        increase(
+          apiserver_request_total{
+            group="extensions",
+            version="v1beta1",
+            client!~"hyperkube/.*|cluster-policy-controller/.*|velero-server/.*"
+          }[24h]
+        ) > 0
       labels:
         severity: warning
```
cblecker: severity level is hasn't changed (should be info, not warning)

soltysh: Nope, that won't change as stated before.
/hold cancel

cblecker: @soltysh I'm confused as here (https://github.com/openshift/cluster-kube-apiserver-operator/pull/730#pullrequestreview-348095535) you stated that "info" was okay, and I had updated my original PR to reflect this.

Warning alerts should be indicative of the cluster health being degraded in some way, and action is required. In the case of OSD, warning alerts are sent to SREs. This alert is **not** actionable in this way, and should not be at a warning level. It's a bad experience for RH SREs, and both OSD and OCP customers as it's a piece of information that they should take action before the next minor version upgrade.

jeremyeder: Seems like info would work just fine for all known use-cases.  Can we go with that, please?

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -82,22 +82,47 @@ spec:
     rules:
     - alert: UsingDeprecatedAPIAppsV1Beta1
       annotations:
-        message: A client in the cluster is using deprecated apps/v1beta1 API that will be removed soon.
+        summary: "A client is using a deprecated {{"{{"}} $labels.group {{"}}"}}/{{"{{"}} $labels.version {{"}}"}} API version"
```
LiliC: If this is meant for customers can we specify at least this is a "kubernetes client".

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
LiliC: Note that regardless of the severity level it will still show up in the main console view.

Also:
> This alert is not actionable in this way, and should not be at a warning level. 

Exactly why this is why I think we should reconsider this as it can induce alert fatigue, as it will just be there. Are these all components users can control and migrate away from this client?

crawford:
/hold

Something about this PR is causing the control plane to fail to come up. Let's stop retesting in the meantime.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
jeremyeder: @soltysh pinging again here - request to move to info level.

jeremyeder: @soltysh can you please set the severity of these alerts to info rather than warning?

cblecker: Should this now change to info, to be in line with the master PR (https://github.com/openshift/cluster-kube-apiserver-operator/pull/742/files#diff-9cc7eba97b17dc12b44afb453dc9e69cL126) ?

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -82,22 +82,50 @@ spec:
...
       expr: |
-        apiserver_request_count{group="apps",version="v1beta2"}
+        increase(
+          apiserver_request_total{
+            group="apps",
+            version="v1beta2",
+          }[24h]
+        ) > 0
       labels:
         severity: warning
```
cblecker: Should this now change to info, to be in line with the master PR (https://github.com/openshift/cluster-kube-apiserver-operator/pull/742/files#diff-9cc7eba97b17dc12b44afb453dc9e69cL126) ?

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -82,22 +82,50 @@ spec:
...
       expr: |
-        apiserver_request_count{group="apps",version="v1beta1"}
+        increase(
+          apiserver_request_total{
+            group="apps",
+            version="v1beta1",
+          }[24h]
+        ) > 0
       labels:
         severity: warning
```
cblecker: Should this now change to info, to be in line with the master PR (https://github.com/openshift/cluster-kube-apiserver-operator/pull/742/files#diff-9cc7eba97b17dc12b44afb453dc9e69cL126) ?

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
tnozicka: The master PR is about different resource that is still present in Kubernetes.

These ones are already removed in the Kubernetes version we are shipping and we'll drop the hack reenabling them in the next release - so warning seems pretty warranted to me. There should be action taken to hunt down the team using it. I think most of our customer use the clusters internally and care about apps that will stop working after the next upgrade. There is also the other case of clusters like Online where you provision for outside users and don't care about what they run, so I'd just suppress the particular warning there unless there is a better way.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -82,22 +82,50 @@ spec:
...
+        message: |
+          A client in the cluster has used a deprecated {{"{{"}} $labels.group {{"}}"}}/{{"{{"}} $labels.version {{"}}"}} API version of the {{"{{"}} $labels.resource {{"}}"}} resource in the last 24 hours.
+          This API version will be removed soon; please migrate to the current version of this API.
       expr: |
-        apiserver_request_count{group="extensions",version="v1beta1",resource!~"ingresses|",client!~"hyperkube/.*|cluster-policy-controller/.*"}
+        increase(
+          apiserver_request_total{
+            group="extensions",
+            version="v1beta1",
+            client!~"hyperkube/.*|cluster-policy-controller/.*|velero-server/.*"
```
LiliC: We need to add also:
openshift-controller-manager and kube-state-metrics to this list, as we have that firing in our cluster.

soltysh:
Update with the two other clients excluded too.
/hold cancel

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -82,22 +82,50 @@ spec:
...
+            group="apps",
+            version="v1beta2",
+          }[24h]
+        ) > 0
       labels:
         severity: warning
     - alert: UsingDeprecatedAPIExtensionsV1Beta1
       annotations:
-        message: A client in the cluster is using deprecated extensions/v1beta1 API that will be removed soon.
+        summary:
```
cblecker: Is this missing a `|`?

soltysh: good catch, thx!

bparees:
skipping this one for approval this week anyway due to the number of other changes already accepted.

cblecker:
@knobunc This is a low risk PR (removes a PrometheusRule), and is actively causing customer pain (unfixable, never resolving alert in all 4.3 clusters). Asking for consideration for this week's cherrypick batch per your e-mail.

lilic:
The upgrade job does not fail, there is a problem with the test platform, the logs say it passes:
> 2020/03/06 08:30:46 Container setup in pod e2e-aws-upgrade completed successfully
2020/03/06 09:06:30 Container test in pod e2e-aws-upgrade completed successfully
2020/03/06 09:14:49 Container teardown in pod e2e-aws-upgrade completed successfully
2020/03/06 09:14:49 Pod e2e-aws-upgrade succeeded after 1h18m32s

lilic:
/retest

now CI should be fixed

PR #743, opened 2020-01-28
Bug 1798609: Report a metric about proxy enablement

smarterclayton:
Like feature set and infrastructure, report whether the core apiserver
knows that it should enable proxy support for HTTP, HTTPS, or has a
trusted CA. This will allow additional information to be alerted on.


smarterclayton:
/assign @deads2k 

has been requested a few times now, this doesn't seem completely off base here (this is a metric this component cares about due to CA) and any future movement wouldn't be more difficult.

PR #746, opened 2020-01-30
Bug 1798450: alerts for aggregated API metrics

p0lyn0mial:
defines alerts for aggregated API as requested in https://bugzilla.redhat.com/show_bug.cgi?id=1772564

### DIFF
--- a/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
+++ b/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
@@ -111,25 +111,17 @@ spec:
     - alert: AggregatedAPIErrors
       annotations:
         message: "An aggregated API {{ $labels.name }}/{{ $labels.namespace }} have reported errors."
-        description: "The number of error ({{ $labels.reason }}) has increased for an aggregated API {{ $labels.name }}/{{ $labels.namespace }} in the last 2 minutes."
+        description: "The number of errors have increased for an aggregated API {{ $labels.name }}/{{ $labels.namespace }} in the last minute. Note that high values indicate that the availability of the services changes too often."
       expr: |
-        sum by(name, namespace, reason)(increase(aggregator_unavailable_apiservice_count[2m])) > 1
+        sum by(name, namespace)(increase(aggregator_unavailable_apiservice_count[1m])) > 2
       labels:
         severity: warning
     - alert: AggregatedAPIDown
       annotations:
-        message: "An aggregated API {{ $labels.name }}/{{ $labels.namespace }} down."
-        description: "An aggregated API {{ $labels.name }}/{{ $labels.namespace }} has not been available for 1 min."
+        message: "An aggregated API {{ $labels.name }}/{{ $labels.namespace }} is down."
+        description: "An aggregated API {{ $labels.name }}/{{ $labels.namespace }} has not been available at least for the last minute."
       expr: |
         sum by(name, namespace)(sum_over_time(aggregator_unavailable_apiservice[1m])) > 0
-      labels:
-        severity: warning
-    - alert: AggregatedAPIFlapping
-      annotations:
-        message: "An aggregated API {{ $labels.name }}/{{ $labels.namespace }} flapping."
-        description: "An aggregated API {{ $labels.name }}/{{ $labels.namespace }} availability changes too often. Seen at least 2 changes in the last 5 minutes."
-      expr: |
-        sum by(name, namespace)(changes(aggregator_unavailable_apiservice[5m])) > 2
       labels:
         severity: warning
   - name: using-deprecated-apis
--- a/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
+++ b/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
@@ -106,6 +106,32 @@ metadata:
     exclude.release.openshift.io/internal-openshift-hosted: "true"
 spec:
   groups:
+  - name: aggregated-apis
+    rules:
+    - alert: AggregatedAPIErrors
+      annotations:
+        message: "An aggregated API {{ $labels.name }}/{{ $labels.namespace }} have reported errors."
+        description: "The number of error ({{ $labels.reason }}) has increased for an aggregated API {{ $labels.name }}/{{ $labels.namespace }} in the last 2 minutes."
+      expr: |
+        sum by(name, namespace, reason)(increase(aggregator_unavailable_apiservice_count[2m])) > 1
+      labels:
+        severity: warning
+    - alert: AggregatedAPIDown
+      annotations:
+        message: "An aggregated API {{ $labels.name }}/{{ $labels.namespace }} down."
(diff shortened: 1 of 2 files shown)


manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -106,6 +106,32 @@ metadata:
...
+      expr: |
+        sum by(name, namespace)(sum_over_time(aggregator_unavailable_apiservice[1m])) > 0
+      labels:
+        severity: warning
+    - alert: AggregatedAPIFlapping
+      annotations:
+        message: "An aggregated API {{ $labels.name }}/{{ $labels.namespace }} flapping."
+        description: "An aggregated API {{ $labels.name }}/{{ $labels.namespace }} availability changes too often. Seen at least 2 changes in the last 5 minutes."
+      expr: |
+        sum by(name, namespace)(changes(aggregator_unavailable_apiservice[5m])) > 2
```
p0lyn0mial: it deserves an e2e test that would help find good values.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -106,6 +106,32 @@ metadata:
...
+      annotations:
+        message: "An aggregated API {{ $labels.name }}/{{ $labels.namespace }} have reported errors."
+        description: "The number of error ({{ $labels.reason }}) has increased for an aggregated API {{ $labels.name }}/{{ $labels.namespace }} in the last 2 minutes."
+      expr: |
+        sum by(name, namespace, reason)(increase(aggregator_unavailable_apiservice_count[2m])) > 1
+      labels:
+        severity: warning
+    - alert: AggregatedAPIDown
+      annotations:
+        message: "An aggregated API {{ $labels.name }}/{{ $labels.namespace }} down."
```
p0lyn0mial: At the moment this metric doesn't work. I suspect that the upstream code is broken. In an `HA` setup one instance can set its instance of the metric to `1` whereas the other instance might flip its own instance to `0` preventing the first instance from changing it to `0`(I need to investigate it). But definitely something fishy is going on because I saw some instances reported a service as unavailable whereas it was available.

@sttts FYI

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -106,6 +106,32 @@ metadata:
...
 spec:
   groups:
+  - name: aggregated-apis
+    rules:
+    - alert: AggregatedAPIErrors
+      annotations:
+        message: "An aggregated API {{ $labels.name }}/{{ $labels.namespace }} have reported errors."
+        description: "The number of error ({{ $labels.reason }}) has increased for an aggregated API {{ $labels.name }}/{{ $labels.namespace }} in the last 2 minutes."
+      expr: |
+        sum by(name, namespace, reason)(increase(aggregator_unavailable_apiservice_count[2m])) > 1
```
p0lyn0mial: this will only report errors that have occurred in the last 2 minutes.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
p0lyn0mial: also is not perfect, frequent changes can be unnoticeable  - value hasn't changed between the scrapes.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
p0lyn0mial: high values actually might indicate frequent changes in availability

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
p0lyn0mial: longer duration might give a better perspective.

p0lyn0mial: alternatively, we could use `sum by(name, namespace, reason)(changes(aggregator_unavailable_apiservice_count[5m])) > 1`

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
p0lyn0mial: note that values `> 1` indicate errors in the given timeframe.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
p0lyn0mial: suppose that we fix it upstream, will this alert be present on downgrade?

p0lyn0mial:
will something clean up the new alerts on downgrade? I am asking because the alert I am going to add will likely fire as the previous versions don't have https://github.com/openshift/origin/pull/24496.

p0lyn0mial:
requires https://github.com/openshift/origin/pull/24496

deads2k:
> will something clean up the new alerts on downgrade? I am asking because the alert I am going to add will likely fire as the previous versions don't have [openshift/origin#24496](https://github.com/openshift/origin/pull/24496).

backport it.

p0lyn0mial:
> backport it

how far back? I think up to `4.2`

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -106,6 +106,24 @@ metadata:
...
+  - name: aggregated-apis
+    rules:
+    - alert: AggregatedAPIErrors
+      annotations:
+        message: "An aggregated API {{ $labels.name }}/{{ $labels.namespace }} have reported errors."
+        description: "The number of errors have increased for an aggregated API {{ $labels.name }}/{{ $labels.namespace }} in the last minute. Note that high values indicate that the availability of the services changes too often."
+      expr: |
+        sum by(name, namespace)(increase(aggregator_unavailable_apiservice_count[1m])) > 2
+      labels:
+        severity: warning
```
sttts: @sichvoge we are kind of blindly impementing these alerts, without having some rule of thumbs about severity and how quick they should fire. Can you comment on this, here in this case about aggregated APIs being down for one minute (like service-catalog, openshift-apiserver or metrics apiserver) ?

lilic: 5 minute is more of a normal time range, so I would change it to that, for both of them.

I like to ask myself what can a user do if this fires, what would the rulebook/playbook say. How can they fix this?

p0lyn0mial:
moved the alerts `kubernetes-mixin` repo  https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/355

p0lyn0mial:
closing as it is replaced by https://github.com/openshift/cluster-monitoring-operator/pull/691

/close

PR #748, opened 2020-02-03
bug 1797601: force cert rotation every couple days for development

deads2k:
This will make DR easier to test and ensure rotation is working pretty well.

### DIFF
--- a/pkg/operator/certrotationcontroller/certrotationcontroller.go
+++ b/pkg/operator/certrotationcontroller/certrotationcontroller.go
@@ -117,6 +117,10 @@ func newCertRotationController(
 		rotationDay = day
 		klog.Warningf("!!! UNSUPPORTED VALUE SET !!!")
 		klog.Warningf("Certificate rotation base set to %q", rotationDay)
+	} else {
+		// for the development cycle, make the rotation 60 times faster (every twelve hours or so).
+		// This must be reverted before we ship
+		rotationDay = rotationDay / 60
 	}
 
 	certRotator, err := certrotation.NewCertRotationController(


pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -119,6 +119,10 @@ func newCertRotationController(
 		klog.Warningf("Certificate rotation base set to %q", rotationDay)
 	}
 
+	// for the development cycle, make the rotation 60 times faster (every twelve hours or so).
+	// This must be reverted before we ship
+	rotationDay = rotationDay / 60
```
tnozicka: should this be in an else block above or it break the configmap override?

pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -117,6 +117,10 @@ func newCertRotationController(
 		rotationDay = day
 		klog.Warningf("!!! UNSUPPORTED VALUE SET !!!")
 		klog.Warningf("Certificate rotation base set to %q", rotationDay)
+	} else {
+		// for the development cycle, make the rotation 60 times faster (every twelve hours or so).
+		// This must be reverted before we ship
```
xingxingxia: Can we design it to have a way to set in install-config.yaml to pass the validity duration?

wangke19: Agree with you, reducing the code change.

sttts: We don't want to see this accessible in any way to customers without tainting the cluster as unsupported.

PR #750, opened 2020-02-04
Bug 1798048: bindata: remove revision suffix from containers

mfojtik:
We don't need revision suffix as the pod already has revision in label. The original intent was to make it clear which revision we have in artifact logs, but having this suffix makes it harder to debug or to get logs when debugging broken clusters.

Combined with https://github.com/kubernetes/kubernetes/pull/87809 I think we will get into desirable state :-)

/cc @deads2k 
/cc @smarterclayton

### DIFF
--- a/bindata/v4.1.0/kube-apiserver/pod.yaml
+++ b/bindata/v4.1.0/kube-apiserver/pod.yaml
@@ -29,7 +29,7 @@ spec:
       securityContext:
         privileged: true
   containers:
-  - name: kube-apiserver-REVISION
+  - name: kube-apiserver
     image: ${IMAGE}
     imagePullPolicy: IfNotPresent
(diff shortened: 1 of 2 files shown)


mfojtik:
@sanchezl do you remember any logic built into must-gather/inspect that expects the `-REVISION` suffix for kube-apiserver containers?

mfojtik:
if somebody tag it, I will make BZ that will state that revision suffix is a defect and makes debugging harder :-)

sttts:
/lgtm

I want this if there is no real reason for those suffices.

deads2k:
> /lgtm
> 
> I want this if there is no real reason for those suffices.

I initially added them because we were failing in ways that required getting on hosts without kube-apiserver access and figuring out which revision was available. We fixed it later by adding information in the static pod on disk.

/hold cancel

PR #752, opened 2020-02-05
Bug 1801154: manifests/: Throw away unused high cardinality apiserver duration buckets

LiliC:
This shouldn't have any effect on our alerts or dashboards as the buckets are cumulative and for those purposes the remaining buckets should be sufficient.

Note: this change was made in "upstream" kube-promethues repo, see background for this https://github.com/coreos/kube-prometheus/pull/387


LiliC:
/retest

doesn't seem to be related.

PR #753, opened 2020-02-05
bindata: preselect container name for oc

mfojtik:
This is cosmetic change to kube-apiserver pod that preselect kube-apiserver container when calling `kubectl logs`. Users don't have to specify the container name anymore and api server container will be used by default.

It require `oc` from master, there is no change in behavior with older `oc`.


mfojtik:
This is just cosmetic change and this don't need a bug.

PR #754, opened 2020-02-05
bug 1798725: static pod operator cannot be composed

deads2k:
Update to fix the operator.  This cannot be done in the static pod code itself because some consumers like etcd still rely on those directly set conditions and removing them would break it.

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -6,33 +6,12 @@ import (
 	"os"
 	"time"
 
-	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/v410_00_assets"
-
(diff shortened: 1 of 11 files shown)


pkg/operator/starter.go
```diff
@@ -224,6 +221,16 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		controllerContext.EventRecorder,
 	)
 
+	staleConditionsController := staleconditions.NewRemoveStaleConditions(
+		[]string{
+			// the static pod operator used to directly set these. this removes those conditions since the static pod operator was updated.
+			// these can be removed in 4.5
```
sttts: `// TODO: remove in 4.5`?

PR #757, opened 2020-02-10
Migrate to build-machinery-go

tnozicka:
Migrating to openshift/build-machinery-go and also getting min version fix from https://github.com/openshift/build-machinery-go/pull/7

/cc @mfojtik @soltysh


tnozicka:
/close
in favour of https://github.com/openshift/cluster-kube-apiserver-operator/pull/758 picking up these commits

PR #758, opened 2020-02-10
Bug 1794376: Remove IPv6 dualstack feature gate

mfojtik:
Builds on https://github.com/openshift/cluster-kube-apiserver-operator/pull/757


pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -106,7 +105,7 @@ func NewConfigObserver(
 				[]string{"apiServerArguments", "cloud-config"}),
 			featuregates.NewObserveFeatureFlagsFunc(
 				nil,
-				sets.NewString("IPv6DualStack"), // IPv6DualStack is bugged, don't turn it on
+				nil,
```
mfojtik: cc @deads2k ... we removed that feature gate, so we don't have to black list this anymore...

mfojtik:
/hold cancel

@soltysh this does not need any pre-reqs

PR #760, opened 2020-02-10
baremetal: Explicitly advertise the host IP chosen by kubelet

russellb:
When using the baremetal infrastructure platform, the networking architecture
includes 3 Virtual IP addresses (VIPs).  For more information about the use
of these VIPs, see this document:

https://github.com/openshift/installer/blob/master/docs/design/baremetal/networking-infrastructure.md

By default, the kube-apiserver will advertise an IP address that it determines
from the default network interface.  When running in IPv4, we have been getting
lucky that the VIPs were never chosen.  They are always listed after the primary
address when you observe with "ip", and the same order is reflected in golang
when it iterates available IP addresses.

When we starting running IPv6 clusters, we observed kube-apiserver often
advertising one of the VIPs instead of the permanent IP address for the
interface.  This is bad because the VIP can move between servers.

These IPv6 VIPs are "deprecated" addresses, and we observed them listed
first in the list of addresses for an interface.  For more info on "deprecated" 
IPv6 addresses, see:

http://www.davidc.net/networking/ipv6-source-address-selection-linux

Unfortunately, this attribute is not accessible using golang's net package.
Accessing this attribute from golang would require rewriting the relevant
code for Linux to use the netlink interface directly.  This is far from a
trivial change, so we seek a simpler workaround in the meantime.

This same problem exists with kubelet and which IP address it chooses for
a master Node.  That was resolved with some scripts that filter out VIPs
and then specify the correct IP address directly as a command line argument.
For more information on the related kubelet workaround, see:

https://github.com/openshift/machine-config-operator/pull/1444

This PR builds on what was done for kubelet, and explicitly tells
kube-apiserver to advertise the address that we already determined was
correct for kubelet.

### DIFF
--- a/bindata/v4.1.0/kube-apiserver/pod.yaml
+++ b/bindata/v4.1.0/kube-apiserver/pod.yaml
@@ -40,7 +40,7 @@ spec:
             echo "Copying system trust bundle"
             cp -f /etc/kubernetes/static-pod-certs/configmaps/trusted-ca-bundle/ca-bundle.crt /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
           fi
-          exec hyperkube kube-apiserver --openshift-config=/etc/kubernetes/static-pod-resources/configmaps/config/config.yaml
+          exec hyperkube kube-apiserver --openshift-config=/etc/kubernetes/static-pod-resources/configmaps/config/config.yaml --advertise-address=${HOST_IP}
     resources:
       requests:
         memory: 1Gi
@@ -79,6 +79,10 @@ spec:
             fieldPath: metadata.namespace
       - name: STATIC_POD_VERSION # Avoid using 'REVISION' here otherwise it will be substituted
         value: REVISION
+      - name: HOST_IP
+        valueFrom:
+          fieldRef:
+            fieldPath: status.hostIP
     securityContext:
       privileged: true
   - name: kube-apiserver-cert-syncer-REVISION
--- a/pkg/operator/v410_00_assets/bindata.go
+++ b/pkg/operator/v410_00_assets/bindata.go
@@ -349,7 +349,7 @@ spec:
             echo "Copying system trust bundle"
             cp -f /etc/kubernetes/static-pod-certs/configmaps/trusted-ca-bundle/ca-bundle.crt /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
           fi
-          exec hyperkube kube-apiserver --openshift-config=/etc/kubernetes/static-pod-resources/configmaps/config/config.yaml
+          exec hyperkube kube-apiserver --openshift-config=/etc/kubernetes/static-pod-resources/configmaps/config/config.yaml --advertise-address=${HOST_IP}
     resources:
       requests:
         memory: 1Gi
@@ -388,6 +388,10 @@ spec:
             fieldPath: metadata.namespace
       - name: STATIC_POD_VERSION # Avoid using 'REVISION' here otherwise it will be substituted
         value: REVISION
+      - name: HOST_IP
+        valueFrom:
+          fieldRef:
+            fieldPath: status.hostIP
     securityContext:
       privileged: true
   - name: kube-apiserver-cert-syncer-REVISION


russellb:
WIP because I haven't figured out how to make this specific to the `baremetal` platform yet

sttts:
Sgtm.

danwinship:
> The reason this is marked as WIP is because we probably only want this
> workaround for the baremetal infrastructure platform, and some guidance is
> requested on how the operator maintainers would like to see that done.

I don't think it needs to be conditional. All kube-apiservers in all OCP clusters should be advertising the IP address that their node declares as its primary IP. This PR should be a no-op on any non-ipv6-bare-metal OCP system.

russellb:
@

> > The reason this is marked as WIP is because we probably only want this
> > workaround for the baremetal infrastructure platform, and some guidance is
> > requested on how the operator maintainers would like to see that done.
> 
> I don't think it needs to be conditional. All kube-apiservers in all OCP clusters should be advertising the IP address that their node declares as its primary IP. This PR should be a no-op on any non-ipv6-bare-metal OCP system.

@sttts you ok with this?  It did pass e2e-aws, at least ...

deads2k:
/hold

this approach will break other operators which need to create a pod definition to run on multiple nodes that requires those pods communicate with each other. The canonical example is etcd using IP based communication.

deads2k:
A solution that determines which value is the correct value based on the node object and creates an observed configuration will allow all the use-cases we know we have to function.

russellb:
To clarify - are you suggesting that we may hit this problem in other places?  and not that this particular change is going to break something?

deads2k:
> To clarify - are you suggesting that we may hit this problem in other places? and not that this particular change is going to break something?

This change is bypassing a symptom of a general problem with a specific solution that isn't suited to all operators.  If you don't have a general solution for determining which IP address is correct, then you will break in other operators like etcd.

russellb:
> > To clarify - are you suggesting that we may hit this problem in other places? and not that this particular change is going to break something?
> 
> This change is bypassing a symptom of a general problem with a specific solution that isn't suited to all operators. If you don't have a general solution for determining which IP address is correct, then you will break in other operators like etcd.

Our solution right now was to fix it to ensure that the Node address is correct.  Anything that uses the address from the Node will be fine.  It's only things that inspect the host directly that may or may not get it right.

If the etcd operator is getting the IPs from the Node, we're fine.  If it isn't, we may have to fix something there too.

deads2k:
> Our solution right now was to fix it to ensure that the Node address is correct. Anything that uses the address from the Node will be fine. It's only things that inspect the host directly that may or may not get it right.
> 
> If the etcd operator is getting the IPs from the Node, we're fine. If it isn't, we may have to fix something there too.

Will the node have a single internal IP address for us to choose?  If not, are we safe to choose the first?  If so, then I'm ok releasing this PR because the information is available externally via the resource.

russellb:
> > Our solution right now was to fix it to ensure that the Node address is correct. Anything that uses the address from the Node will be fine. It's only things that inspect the host directly that may or may not get it right.
> > If the etcd operator is getting the IPs from the Node, we're fine. If it isn't, we may have to fix something there too.
> 
> Will the node have a single internal IP address for us to choose? If not, are we safe to choose the first? If so, then I'm ok releasing this PR because the information is available externally via the resource.

Yes, the Node has a single InternalIP on this platform now.  Here's a sample from a running cluster:

```
  status:
    addresses:
    - address: fd2e:6f44:5dd8:c956::14
      type: InternalIP
    - address: master-0.ostest.test.metalkube.org
      type: Hostname
```

danwinship:
> Will the node have a single internal IP address for us to choose? If not, are we safe to choose the first?

In the bare-metal case, there will always be a single Internal IP. (kubelet enforces that.) In the cloud case there may be more than one but we already take steps to ensure that the first one is the one you want.

deads2k:
> > Will the node have a single internal IP address for us to choose? If not, are we safe to choose the first?
> 
> In the bare-metal case, there will always be a single Internal IP. (kubelet enforces that.) In the cloud case there may be more than one but we already take steps to ensure that the first one is the one you want.

thanks
/hold cancel

bindata/v4.1.0/kube-apiserver/pod.yaml
```diff
@@ -40,7 +40,8 @@ spec:
             echo "Copying system trust bundle"
             cp -f /etc/kubernetes/static-pod-certs/configmaps/trusted-ca-bundle/ca-bundle.crt /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
           fi
-          exec hyperkube kube-apiserver --openshift-config=/etc/kubernetes/static-pod-resources/configmaps/config/config.yaml
+          echo "HOST_IP is ${HOST_IP}"
```
danwinship: is this really needed?

russellb: No, I can drop it.

russellb: done.

PR #761, opened 2020-02-10
Bug 1785498: regenerate-certificates command blocked by error illegal base64 data at input byte 3

p0lyn0mial:
when starting a recovery API server we must also mount the encryption config otherwise we won't be able to decrypt certain resources for example `secrets and configmaps` and we won't be able to create new certificates.

### DIFF
--- a/bindata/v4.1.0/kube-apiserver/recovery-encryption-config.yaml
+++ b/bindata/v4.1.0/kube-apiserver/recovery-encryption-config.yaml
@@ -1,0 +1,3 @@
+apiVersion: apiserver.config.k8s.io/v1
+kind: EncryptionConfiguration
+resources:
--- a/pkg/recovery/apiserver.go
+++ b/pkg/recovery/apiserver.go
@@ -28,13 +28,15 @@ import (
 )
 
 const (
-	KubeApiserverStaticPodFileName = "kube-apiserver-pod.yaml"
-	RecoveryPodFileName            = "recovery-kube-apiserver-pod.yaml"
-	RecoveryCofigFileName          = "config.yaml"
-	AdminKubeconfigFileName        = "admin.kubeconfig"
+	KubeApiserverStaticPodFileName  = "kube-apiserver-pod.yaml"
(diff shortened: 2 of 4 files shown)


bindata/v4.1.0/kube-apiserver/recovery-encryption-config.yaml
```diff
@@ -0,0 +1,3 @@
+apiVersion: apiserver.config.k8s.io/v1
+kind: EncryptionConfiguration
+resources:
```
tnozicka: \n

pkg/recovery/apiserver.go
```diff
@@ -238,6 +240,30 @@ func (s *Apiserver) Create() error {
 		return fmt.Errorf("failed to write recovery config %q: %v", recoveryConfigPath, err)
 	}
 
+	// always create a default (empty) encryption config to simplify the deployment
+	// only copy the real one if it exists
+	encryptionConfigExists, err := fileExists(filepath.Join(s.ResourceDirPath, "secrets/encryption-config/encryption-config"))
+	if err != nil {
+		return fmt.Errorf("failed to check if encryption config exists due to %v", err)
```
tnozicka: nit: `s/exists due to/exists:/`

sttts:
/lgtm
/approve

/hold
for manual tests.

sttts:
Do we have a BZ for this? We should.

p0lyn0mial:
applied Tomas's review comments, please retag.

p0lyn0mial:
/hold
for manual tests.

I haven't gone through the whole recovery procedure described at https://docs.openshift.com/container-platform/4.2/backup_and_restore/disaster_recovery/scenario-3-expired-certs.html

I only executed the step from `1 to 6` to make sure that the recovery API server can successfully decrypt data. If this is not enough I can go through all the steps.

tnozicka:
/hold cancel
step 6. validates reading etcd using recovery-apiserver

PR #763, opened 2020-02-12
Bug 1802247: Check operator readiness before each bound token controller test

marun:
Previously operator readiness was only checked at the start of the set of bound token controller tests. This introduced the possibility of the configmap deletion test triggering rollout of a new apiserver pod revision that the subsequent operand secret deletion test wouldn't wait for before starting.


marun:
Updated to check for operator readiness for each bound token controller test rather than before any are executed. I suspect that not checking operator readiness prior to deletion of the operator signing secret was the reason the test was flaking.

PR #766, opened 2020-02-17
Bug 1803750: [release-4.2] Improve node controller condition message

mfojtik:
This is needed to determine why 50+ 4.2.z clusters are reporting NodeControllerDegraded: The master node(s) "xyz" not ready as this change will append the reason to why are they not ready.

Not ready master nodes might prevent/break future upgrades, so this is critical fix.

/cc @eparis
/cc @bparees


bparees:
why do we have both this and https://github.com/openshift/cluster-kube-controller-manager-operator/pull/348

?

i'd expect a single PR given that they are both for the same bug and both bumping deps?

PR #770, opened 2020-02-20
certrotationcontroller: 10y validity of KubeAPIServerToKubeletClientCert signer

sttts:
Was this 1y intentional or typo? What are the consequences?


sttts:
/hold

until we are sure whether this is intentional or not.

deads2k:
> Was this 1y intentional or typo? What are the consequences?

You'll want to see what the validity in the installer is.  That is a cert used by the kubelet to confirm the kube-apiserver

sttts:
https://github.com/openshift/installer/blob/7b184740d7eefc84196dc6236a9dbd9f38f30660/pkg/asset/tls/apiserver.go#L30

/close

PR #784, opened 2020-02-26
Bug 1808602: kube-apiserver-operator should set upgradeable=false if default SCC mutates

tkashem:
In 4.4, the SCC is maintained by the CVO.  This means that the ratchet logic of "reconcile-scc" has been lost.  Overall this is good (we documented that nothing should change the bootstrap SCC), but practically, anyone who has changed needs to handle the situation before they upgrade and have the resources stomped.

We can make a change to 4.3.z which prevents upgrades (upgradeable=false) on fairly complicated predicates and provide links and help for dealing with the situation *before* it goes bad.


pkg/operator/defaultscccontroller/syncer.go
```diff
@@ -10,21 +10,39 @@ import (
...
+	// If it's not to do with the default SCC, we don't care.
+	if !s.defaultSCCSet.IsDefault(key.Name) {
+		return nil
+	}
+
+	mutated := make([]string, 0)
+	for _, original := range s.defaultSCCSet.List() {
+		current, err := s.lister.Get(original.GetName())
+		if err != nil {
+			if k8serrors.IsNotFound(err) {
```
tkashem: If a default scc has been removed, is it okay to not set upgradeable=false? I am not sure what the state of a cluster is when a default scc is removed.

pkg/operator/defaultscccontroller/syncer.go
```diff
@@ -1,16 +1,71 @@
...
+	for _, name := range s.cache.DefaultSCCNames() {
+		original := s.cache.GetDefaultSCC(name)
+		if original == nil {
+			return fmt.Errorf("name=%s default scc not found in default scc cache", name)
+		}
+
+		current, err := s.lister.Get(name)
+		if err != nil {
+			if k8serrors.IsNotFound(err) {
+				klog.Infof("[%s] name=%s scc has been deleted - %s", ControllerName, name, err.Error())
```
tkashem: Is it okay to ignore if a default SCC is deleted? I am thinking the upgrade will automatically recreate it. What if the admin has intentionally deleted the SCC for a reason?

pkg/operator/defaultscccontroller/syncer.go
```diff
@@ -1,16 +1,71 @@
...
+
+	if len(mutated) > 0 {
+		klog.Infof("[%s] default scc has mutated %s", ControllerName, mutated)
+	}
+
+	condition := NewCondition(mutated)
+	return s.updater.UpdateStatus(condition)
+}
+
+func withNoMeta(obj1, obj2 *securityv1.SecurityContextConstraints) (copy1, copy2 *securityv1.SecurityContextConstraints) {
```
tkashem: this works, but I am wondering if there is a better way of doing it.

pkg/operator/defaultscccontroller/defaultscccontroller.go
```diff
@@ -0,0 +1,131 @@
...
+		syncer: &Syncer{
+			lister:   lister,
+			recorder: recorder,
+		},
+	}
+
+	return
+}
+
+type DefaultSCCController struct {
```
sttts: doc what it does

pkg/operator/defaultscccontroller/resourcehandler.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog"
+)
+
+// NewResourceHandler returns a cache.ResourceEventHandler appropriate for
+// reconciliation of SecurityContextConstraints object(s).
+func NewResourceHandler(queue workqueue.RateLimitingInterface) ResourceHandler {
```
sttts: does this really need this61 line file instead of an inline event handler in the main controller?

pkg/operator/defaultscccontroller/defaultscccontroller.go
```diff
@@ -0,0 +1,131 @@
...
+	if !ok {
+		// As the item in the work queue is actually invalid, we call Forget here else
+		// we'd go into a loop of attempting to process a work item that is invalid.
+		c.queue.Forget(key)
+
+		utilruntime.HandleError(fmt.Errorf("[%s] expected types.NamespacedName in workqueue but got %#v", ControllerName, key))
+		return true
+	}
+
+	if err := c.syncer.Sync(request); err != nil {
```
sttts: do we have this style with a Syncer abstraction elsewhere? Would try to avoid to introduce new one-off patterns for this controller.

pkg/operator/defaultscccontroller/defaultscc.go
```diff
@@ -0,0 +1,103 @@
...
+	securityv1 "github.com/openshift/api/security/v1"
+
+	assets "github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/defaultscc_assets"
+)
+
+// DefaultSCCCache maintains the list of default SecurityContextConstraints objects
+// shipped wth an OpenShift cluster.
+type DefaultSCCCache struct {
+	// When we iterate through the default SCC we want to maintain a distinct order.
+	names []string
```
sttts: call it orderedNames and get rid of the comment.

pkg/operator/defaultscccontroller/defaultscc.go
```diff
@@ -0,0 +1,103 @@
...
+
+func (d *DefaultSCCCache) IsDefaultSCC(name string) bool {
+	_, exists := d.set[name]
+	return exists
+}
+
+// GetDefault returns the SecurityContextConstraints object associated with the
+// specified name. The returned SecurityContextConstraints is a deep copy and so
+// it can be modified at will without impacting the integrity of the underlying cache.
+func (d *DefaultSCCCache) GetDefaultSCC(name string) *securityv1.SecurityContextConstraints {
```
sttts: s/GetDefaultSCC/Get/

pkg/operator/defaultscccontroller/defaultscc.go
```diff
@@ -0,0 +1,103 @@
...
+
+// DefaultSCCCache maintains the list of default SecurityContextConstraints objects
+// shipped wth an OpenShift cluster.
+type DefaultSCCCache struct {
+	// When we iterate through the default SCC we want to maintain a distinct order.
+	names []string
+	set   map[string]*securityv1.SecurityContextConstraints
+}
+
+func (d *DefaultSCCCache) IsDefaultSCC(name string) bool {
```
sttts: merge this and the next func in the more common `Get(string) (*SCC, bool)` interface.

pkg/operator/defaultscccontroller/defaultscc.go
```diff
@@ -0,0 +1,103 @@
...
+	return scc.DeepCopy()
+}
+
+func (d *DefaultSCCCache) DefaultSCCNames() []string {
+	return d.names
+}
+
+// Render renders the assets associated with the default set of SCC and
+// loads it into a cache (DefaultSCCCache)
+func Render() (cache *DefaultSCCCache, err error) {
```
sttts: why is it called `Render`? `NewDefaultSCCCache` would be a better name, no?

pkg/operator/defaultscccontroller/syncer.go
sttts: having a func for two objects is strange. Call it twice with one.

sttts: make the deepcopies internally here. `withNoMeta` with that signature does not suggest it has side-effects-

pkg/operator/defaultscccontroller/defaultscccontroller.go
tkashem: each controller I see in ckao has a `Sync` function.

pkg/operator/defaultscccontroller/resourcehandler.go
tkashem: are you referring to this?
```
func defaultSCCControllerEventHandler(queue workqueue.RateLimitingInterface) cache.ResourceEventHandler {
	addFunc := func(key string, queue workqueue.RateLimitingInterface) {
		namespace, name, err := cache.SplitMetaNamespaceKey(key)
		if err != nil {
			return
		}

		queue.Add(types.NamespacedName{
			Namespace: namespace,
			Name:      name,
		})
	}

	return cache.ResourceEventHandlerFuncs{
		AddFunc: func(obj interface{}) {
			key, err := cache.MetaNamespaceKeyFunc(obj)
			if err != nil {
				klog.Errorf("[%s] OnAdd: could not extract key, type=%T- %s", ControllerName, obj, err.Error())
				return
			}

			addFunc(key, queue)
		},

		UpdateFunc: func(old, new interface{}) {
			key, err := cache.MetaNamespaceKeyFunc(new)
			if err != nil {
				klog.Errorf("[%s] OnUpdate: could not extract key, type=%T- %s", ControllerName, new, err.Error())
				return
			}

			addFunc(key, queue)
		},
		DeleteFunc: func(obj interface{}) {
			key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
			if err != nil {
				klog.Errorf("[%s] OnDelete: could not extract key, type=%T - %s", ControllerName, obj, err.Error())
				return
			}

			addFunc(key, queue)
		},
	}
}
```
or are you referring to an existing resource handler that I can use off the shelf?

pkg/operator/defaultscccontroller/syncer.go
sttts: Isn't openshift-apiserver recreating it anyway?

pkg/operator/defaultscccontroller/syncer.go
sttts: As the objects are CRD based, there is no defaulting going on. So those manifests shoulds be equal modulo metadata. So yes, I think it is fine.

pkg/operator/defaultscccontroller/defaultscc.go
sttts: Do we need this func anymore?

pkg/operator/defaultscccontroller/defaultscccontroller.go
```diff
@@ -0,0 +1,141 @@
...
+	Config         *rest.Config
+	OperatorClient v1helpers.OperatorClient
+	Recorder       events.Recorder
+	ResyncPeriod   time.Duration
+}
+
+func NewDefaultSCCController(options *Options) (controller *DefaultSCCController, err error) {
+	client, clientErr := securityv1client.NewForConfig(options.Config)
+	if clientErr != nil {
+		err = fmt.Errorf("[%s] failed to create client for security/v1 - %s", ControllerName, clientErr.Error())
```
sttts: this style of named error is pretty uncommon, and IMO hard to read.

pkg/operator/defaultscccontroller/defaultscccontroller.go
```diff
@@ -0,0 +1,141 @@
...
+		return
+	}
+
+	defaultSCCSet, renderErr := NewDefaultSCCCache()
```
sttts: why these special error var names? Any shadowing logic going on here?

pkg/operator/defaultscccontroller/defaultscccontroller.go
sttts: Sync yes, but a `Syncer` abstraction?

pkg/operator/defaultscccontroller/resourcehandler.go
sttts: yes, this. ERROVERABSTRACTION.

pkg/operator/defaultscccontroller/defaultscccontroller.go
tkashem: I don't have any strong opinion either way, but what drove me to this style is:
- avoid re-purposing or reassigning `err`, wanted to avoid this 
```
err = fmt.Errorf("message - %s", err.Error())
```
The error returned from a function invocation and the error we return to the caller are separate.

- with separate names it is explicit to the reader how the error is being chained, also it is explicit that there is no shadowing. 

But again, readability is in the eye of the reader :), so if this is confusing or hard to read I can use `err` across, let me know.

pkg/operator/defaultscccontroller/defaultscccontroller.go
tkashem: No shadowing, the same explanation I gave above.

pkg/operator/defaultscccontroller/defaultscccontroller.go
sttts: Reassignment of `err` is the usual pattern. When code does not follow it, it's suspicious as there must be a reason, making the reading flow worse.

pkg/operator/defaultscccontroller/syncer.go
tkashem: Yes, when `openshift-apiserver` restarts, if I am not mistaken. if one is missing during an upgrade to 4.4 the ckao will recreate it. So we should be good. Just wanted to confirm.

pkg/operator/defaultscccontroller/defaultscccontroller.go
tkashem: "a controller `has` a syncer" is more composable and testable than "a controller `is` a syncer". I tried to keep the concerns separate as they should be but it is not consistent with other controllers at the same time.

I removed it to be consistent for now.

pkg/operator/defaultscccontroller/defaultscc.go
tkashem: removed it

pkg/operator/defaultscccontroller/condition.go
```diff
@@ -0,0 +1,84 @@
...
+
+import (
+	"fmt"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+// OperatorStatusUpdater is responsible for updating the status block of the operator.
+type OperatorStatusUpdater func(condition operatorv1.OperatorCondition) error
```
sttts: nit: it updates a condition, not the status. Misnamed?

sttts: side effects as promise 😱

pkg/operator/defaultscccontroller/condition.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func NewCondition(mutated []string) operatorv1.OperatorCondition {
+	status := operatorv1.ConditionTrue
+	message := ""
+	reason := "AsExpected"
+	if len(mutated) > 0 {
+		reason = "Mutated"
+		status = operatorv1.ConditionFalse
+		message = fmt.Sprintf("Default SecurityContextConstraints object(s) have mutated %s", mutated)
```
sttts: nit: have been mutated?

pkg/operator/defaultscccontroller/condition.go
```diff
@@ -0,0 +1,84 @@
...
+
+	return nil
+}
+
+func isConditionEqual(desired, current *operatorv1.OperatorCondition) bool {
+	if desired.Type == current.Type &&
+		desired.Reason == current.Reason &&
+		desired.Status == current.Status &&
+		desired.Message == current.Message {
+		return true
```
sttts:
```
return desired.Type == current.Type &&
		desired.Reason == current.Reason &&
		desired.Status == current.Status &&
		desired.Message == current.Message
```

pkg/operator/defaultscccontroller/defaultscc.go
```diff
@@ -0,0 +1,98 @@
...
+// Get returns the SecurityContextConstraints object associated with the
+// specified name. The returned SecurityContextConstraints is a deep copy and so
+// it can be modified at will without impacting the integrity of the underlying cache.
+func (d *DefaultSCCCache) Get(name string) (scc *securityv1.SecurityContextConstraints, exists bool) {
+	obj, exists := d.set[name]
+	if !exists {
+		return
+	}
+
+	scc = obj.DeepCopy()
```
sttts: would move this to the place where you mutate the SCC. Defensive DeepCopies are an anti-pattern.

pkg/operator/defaultscccontroller/defaultscc.go
```diff
@@ -0,0 +1,98 @@
...
+func NewDefaultSCCCache() (cache *DefaultSCCCache, err error) {
+	decoder, decoderErr := decoder()
+	if decoderErr != nil {
+		err = fmt.Errorf("failed to create decoder - %s", decoderErr.Error())
+		return
+	}
+
+	names := make([]string, 0)
+	set := make(map[string]*securityv1.SecurityContextConstraints)
+	for _, name := range assets.AssetNames() {
```
sttts: are asset names ordered? Can a new and old operator with different bindata race?

pkg/operator/defaultscccontroller/defaultscc.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	return
+}
+
+func decoder() (decoder runtime.Decoder, err error) {
+	scheme := runtime.NewScheme()
+	if err = api.InstallKube(scheme); err != nil {
+		return
+	}
+	if err = api.Install(scheme); err != nil {
```
sttts: installing all into the scheme is not necessary. Install the security group only.

pkg/operator/defaultscccontroller/defaultscccontroller.go
```diff
@@ -0,0 +1,232 @@
...
+		},
+
+		WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+			return client.SecurityContextConstraints().Watch(options)
+		},
+	}
+
+	queue := workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "DefaultSCCController")
+
+	indexer, informer := cache.NewIndexerInformer(watcher, &securityv1.SecurityContextConstraints{}, options.ResyncPeriod, defaultSCCControllerEventHandler(queue), cache.Indexers{})
```
sttts: use the shared informer factory, not independent informers.

pkg/operator/defaultscccontroller/condition.go
tkashem: yes, the intent and scope should e clear.

pkg/operator/defaultscccontroller/defaultscc.go
tkashem: done.

pkg/operator/defaultscccontroller/defaultscc.go
tkashem: we don't depend on the order in which `assets.AssetNames()` returns the names. The objects are stored in a map and since iterating a map does not guarantee any particular order, we store the names in a slice and use it to iterate.

The reason we want a consistent order (be it any) is so that the error message (that ends up in status. conditions) `default scc has been mutated [anyuid hostaccess]` is consistent as well.

sttts: I thought the asset name order determines the order in those error messages. If this is not constant, condition will change just because the operator has been updated. Not super critical, yet little ugly.

tkashem: yes, asset name order determines the order of the scc name in the error message. `assets.AssetNames()` returns from a map and thus the order is indeterminate. I am now sorting the `names` slice. Also, I have added a unit test test verifies that the order is fixed. So, we cover the following cases:
- If the operator process restarts we will see the same ordering now.
- If we make change to the code and the order is violated, the unit test will fail to alert us.

pkg/operator/defaultscccontroller/defaultscccontroller.go
tkashem: done.

pkg/operator/defaultscccontroller/defaultscccontroller.go
```diff
@@ -0,0 +1,211 @@
...
+type Options struct {
+	Factory        security1informers.SharedInformerFactory
+	OperatorClient v1helpers.OperatorClient
+	Recorder       events.Recorder
+}
+
+func NewDefaultSCCController(options *Options) (controller *DefaultSCCController, err error) {
+	defaultSCCSet, err := NewDefaultSCCCache()
+	if err != nil {
+		err = fmt.Errorf("[%s] failed to render default SCC assets - %s", ControllerName, err.Error())
```
sttts: nit: it's still a little odd style. Usually you see `return fmt.Errorf(...)` here. And usually you don't name the error arg at all.

PR #787, opened 2020-03-04
Bug 1810133: connect to etcd using IPs

sanchezl:
* Use IPs instead of DNS names to generate storage URLs.
* Always add local host storage URLs.


deads2k:
/lgtm

@sanchezl how did you break verify-deps?

PR #793, opened 2020-03-11
Bug 1812583: Normalize CPU requests on masters

smarterclayton:
The kube-apiserver uses approximately 33% of master CPU in a
reasonable medium sized workload. Given a 1 core per master baseline
(since CPU is compressible and shared), assign the kube-apiserver
roughly 33% of that core on each master.


smarterclayton:
4.4 bug is 1812709

smarterclayton:
/retest

Is what I meant...

PR #799, opened 2020-03-16
WIP: Move public-info-viewer ClusterRole and Binding here

sttts:
Note: this will race with kube-apiserver for now.


sttts:
/hold

We should maybe apply changes from the operator, to preserve the labels and avoid races.

PR #801, opened 2020-03-18
Reduce graceful termination window

tkashem:
Reduce graceful termination window to speed up kube-apiserver roll out
- set shutdown-delay-duration to 30s (close to worst case scenario by  which an LB should detect /readyz failure) and react. We are saving 40s (previous value=70s)
- consequently, set terminationGracePeriodSeconds to 95 seconds, bit more than 30s (shutdown-delay-duration) + 60s (kube-apiserver graceful termination)

### DIFF
--- a/bindata/v4.1.0/config/defaultconfig.yaml
+++ b/bindata/v4.1.0/config/defaultconfig.yaml
@@ -22,7 +22,7 @@ apiServerArguments:
   enable-aggregator-routing:
   - "true"
   shutdown-delay-duration:
-  - 70s # give SDN some time to converge: 30s for iptable lock contention, 25s for the second try and some seconds for AWS to update ELBs
+  - 30s # give SDN some time to converge: 30s for iptable lock contention, 25s for the second try and some seconds for AWS to update ELBs
   http2-max-streams-per-connection:
   - "2000"  # recommended is 1000, but we need to mitigate https://github.com/kubernetes/kubernetes/issues/74412
   kubelet-preferred-address-types:
--- a/bindata/bootkube/bootstrap-manifests/kube-apiserver-pod.yaml
+++ b/bindata/bootkube/bootstrap-manifests/kube-apiserver-pod.yaml
@@ -64,7 +64,7 @@ spec:
     ports:
     - containerPort: 6080
 {{end}}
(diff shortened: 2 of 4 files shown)


tkashem:
/hold
want to run some tests before we merge it.

bindata/v4.1.0/config/defaultconfig.yaml
```diff
@@ -22,7 +22,7 @@ apiServerArguments:
   enable-aggregator-routing:
   - "true"
   shutdown-delay-duration:
-  - 70s # give SDN some time to converge: 30s for iptable lock contention, 25s for the second try and some seconds for AWS to update ELBs
+  - 30s # give SDN some time to converge: 30s for iptable lock contention, 25s for the second try and some seconds for AWS to update ELBs
```
tkashem: @sttts does this still apply `give SDN some time to converge: 30s for iptable lock contention, 25s for the second try`?

gcp-route-controller takes around 50 seconds (`FailureThreshold=10` * `Interval=5`) before it stops the gcp-route service. I still don't know what it achieves, but want to make sure our change has no adverse effect.
https://github.com/openshift/machine-config-operator/blob/master/cmd/gcp-routes-controller/run.go#L87-L103

and the gcp-service processes routes every 30s - https://github.com/openshift/installer/pull/3067/files#diff-f3a509446e9615909e1407b8d19b3dcdR81. I think the bash script has moved, I saw a PR that converts the script to go lang program but can't remember where it is.

tkashem:
/hold cancel

I have run test against an aws cluster. If I reduce `ShutDownDelayDuration` to 10s I see `unexpected EOF` error.  With 30s, I see no error as the `kube-apiserver` rolls out.

bindata/v4.1.0/config/defaultconfig.yaml
sttts: @squeed once told me about 30s retry duration for iptables.

p0lyn0mial:
I've just come across https://docs.google.com/document/d/1BUmtdTth49V02UZ5EjRvJ92A5vjF8wMJMSdPb1Wz3wQ/edit# which gives some tips for the amount of time a graceful shutdown should take.

tkashem:
waiting for gcp route fix

/hold

PR #803, opened 2020-03-19
configobservers: fix existingConfig error handling

sttts:
The observers get the complete previously observed config as `existingConfig`. Some tried to filter that down to their paths, others just returned everything (even paths from other observers).

This PR adds systematic return value pruning and unifies the error handling through out the observers:

- on read errors from existing config, keep going
- on set errors of `observedConfig` return the existing config.


pkg/operator/configobservation/network/observe_network.go
```diff
@@ -14,39 +14,32 @@ import (
 
 // ObserveRestrictedCIDRs watches the network configuration and updates the
 // RestrictedEndpointsAdmission controller config.
-func ObserveRestrictedCIDRs(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
-	listers := genericListers.(configobservation.Listers)
+func ObserveRestrictedCIDRs(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, _ []error) {
```
sttts: @squeed @rphillips please review

pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -106,7 +106,7 @@ func NewConfigObserver(
 			//apiserver.ObserveDefaultUserServingCertificate,
 			apiserver.ObserveNamedCertificates,
 			apiserver.ObserveUserClientCABundle,
-			apiserver.ObserveAdditionalCORSAllowedOrigins,
+			libgoapiserver.ObserveAdditionalCORSAllowedOrigins,
```
sttts: @stlaz do you remember why we had a copy?

stlaz: it hasn't always been in library-go, I had a PR to move to library-go but I never pushed anyone hard enough to review it

pkg/operator/configobservation/images/observe_images.go
```diff
@@ -41,14 +34,19 @@ func ObserveInternalRegistryHostname(genericListers configobserver.Listers, reco
 		return observedConfig, errs
```
p0lyn0mial: shouldn't we return `existingConfig` in this case ?

pkg/operator/configobservation/images/observe_images.go
```diff
@@ -41,14 +34,19 @@ func ObserveInternalRegistryHostname(genericListers configobserver.Listers, reco
...
 	}
 	if err != nil {
-		return prevObservedConfig, errs
+		return existingConfig, append(errs, err)
 	}
 
 	internalRegistryHostName := configImage.Status.InternalRegistryHostname
 	if len(internalRegistryHostName) > 0 {
 		if err := unstructured.SetNestedField(observedConfig, internalRegistryHostName, internalRegistryHostnamePath...); err != nil {
 			errs = append(errs, err)
```
p0lyn0mial: in what state is `observedConfig` at this point ?

pkg/operator/configobservation/images/observe_images.go
sttts: it's not the normal error case here, but just the default value: empty observedConfig.

pkg/operator/configobservation/network/observe_network.go
```diff
@@ -65,35 +58,38 @@ func ObserveRestrictedCIDRs(genericListers configobserver.Listers, recorder even
...
 
 	if err := unstructured.SetNestedStringSlice(admissionControllerConfig.Object, restrictedCIDRs, "restrictedCIDRs"); err != nil {
-		errs = append(errs, err)
+		return existingConfig, append(errs, err)
 	}
 
 	observedConfig := map[string]interface{}{}
-	unstructured.SetNestedMap(observedConfig, admissionControllerConfig.Object, configPath...)
+	if err := unstructured.SetNestedMap(observedConfig, admissionControllerConfig.Object, configPath...); err != nil {
+		return existingConfig, append(errs, err)
```
p0lyn0mial: we should behave consistently, for example, won't return an error in a similar case https://github.com/openshift/cluster-kube-apiserver-operator/blob/423e7c85c9aafa12a4e5981baa2970f85b46ed22/pkg/operator/configobservation/images/observe_images.go#L43

pkg/operator/configobservation/images/observe_images.go
p0lyn0mial: consider adding a comment.

pkg/operator/configobservation/images/observe_images.go
p0lyn0mial: from what I've seen in cases like that we return `existingConfig`

pkg/operator/configobservation/auth/auth_metadata.go
```diff
@@ -45,8 +38,7 @@ func ObserveAuthMetadata(genericListers configobserver.Listers, recorder events.
 		return observedConfig, errs
```
vareti, line 38: Is the intent here is to overwrite `existingConfig` with empty config?

vareti: I see that same pattern is used at couple of other places too. I am suspecting all these are intentional then?

pkg/operator/configobservation/network/observe_network.go
squeed: Wait, should we be returning the pruned config (i.e. only keys we care about), or the full Existing on failure?

pkg/operator/configobservation/network/observe_network.go
```diff
@@ -14,39 +14,32 @@ import (
...
-	if ok && err == nil {
-		// If we ever bump the API version, we'll have to do conversion here.
-		admissionControllerConfig.SetUnstructuredContent(prev)
-	}
 	admissionControllerConfig.SetAPIVersion("network.openshift.io/v1")
 	admissionControllerConfig.SetKind("RestrictedEndpointsAdmissionConfig")
 
 	clusterCIDRs, err := network.GetClusterCIDRs(listers.NetworkLister, recorder)
 	if err != nil {
-		errs = append(errs, err)
```
squeed, line 37: While I do prefer to sum errors, the reality is that a cluster with a missing ServiceCIDR and ClusterCIDR is unlikely, and not a user error, so I'll let this one go :-)

pkg/operator/configobservation/auth/auth_metadata.go
sttts: that depends maybe on the config. A non-existing auth config means not to configure auth at all. Also note that this is preexisting. I didn't touch that logic.

pkg/operator/configobservation/images/observe_images.go
sttts: it's empty. I.e. this will never fail.

pkg/operator/configobservation/network/observe_network.go
sttts: only pruned. We forgot that everywhere. Hence, if some observer errored, other observers potentially stopped working (their config was overwritten by the one with the error).

stlaz:
/lgtm cancel
the unit test is failing

PR #802, opened 2020-03-19
Bug 1810181: Components using globs in Dockerfile COPY commands may break on OCP 4

p0lyn0mial:

PR #804, opened 2020-03-19
Remove wildcard copies from docker

stlaz:
Alternative to #802


p0lyn0mial:
doesn't a directory like `bootkube/manifests/manifests` look strange :) ?

Dockerfile.rhel7
```diff
@@ -6,7 +6,7 @@ RUN make build --warn-undefined-variables
 
 FROM registry.svc.ci.openshift.org/ocp/4.4:base
 RUN mkdir -p /usr/share/bootkube/manifests /manifests
-COPY --from=builder /go/src/github.com/openshift/cluster-kube-apiserver-operator/bindata/bootkube/* /usr/share/bootkube/manifests/
+COPY --from=builder /go/src/github.com/openshift/cluster-kube-apiserver-operator/bindata/bootkube/manifests /usr/share/bootkube/
 COPY --from=builder /go/src/github.com/openshift/cluster-kube-apiserver-operator/cluster-kube-apiserver-operator /usr/bin/
 COPY manifests/*.yaml /manifests/
```
p0lyn0mial: I think that those two lines could be simplified to just a single command `COPY manifests /manifests`
```
COPY manifests/*.yaml /manifests/
COPY manifests/image-references /manifests/
```

stlaz: that's a good observation, we are not storing anything else in there

p0lyn0mial:
/assign @sttts 

it seems like we have competing approaches on how to fix the issue. We need to choose one :)

p0lyn0mial:
/close

in favour of https://github.com/openshift/cluster-kube-apiserver-operator/pull/802

PR #807, opened 2020-03-26
kubelet-client secret dynamic reload

vareti:
With https://github.com/openshift/origin/pull/24646 merged,  a new revision need not be created when `kubelet-client` certificate changes

### DIFF
--- a/bindata/v4.1.0/config/defaultconfig.yaml
+++ b/bindata/v4.1.0/config/defaultconfig.yaml
@@ -88,8 +88,8 @@ authConfig:
 consolePublicURL: ""
 kubeletClientInfo:
   ca: /etc/kubernetes/static-pod-resources/configmaps/kubelet-serving-ca/ca-bundle.crt
-  certFile: /etc/kubernetes/static-pod-resources/secrets/kubelet-client/tls.crt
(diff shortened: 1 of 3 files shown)


bindata/v4.1.0/config/defaultconfig.yaml
```diff
@@ -88,8 +88,8 @@ authConfig:
 consolePublicURL: ""
 kubeletClientInfo:
   ca: /etc/kubernetes/static-pod-resources/configmaps/kubelet-serving-ca/ca-bundle.crt
-  certFile: /etc/kubernetes/static-pod-resources/secrets/kubelet-client/tls.crt
-  keyFile: /etc/kubernetes/static-pod-resources/secrets/kubelet-client/tls.key
```
sttts: what does this mean? We never wired the kubelet client cert correctly? How did it work like that?

sttts: Please check 4.4 and older.

deads2k: it was located in both.  this is just updating us to point at a maintained file.  this looks correct.

vareti: Thanks @deads2k.

I figured out that this was needed after my first try to install failed at bootstrap because I did not change the wiring.

PR #809, opened 2020-04-02
Bug 1819246: Set service account issuer to internal dns name of kube apiserver

marun:
The default issuer for bound service account tokens is updated to point to the internal dns name of the kube apiserver to ensure compatibility with the `ServiceAccountIssuerDiscovery` feature [1] added in kube 1.18. This fix is intended to be backported to 4.4 (the release that added support for bound service account otkens) before GA to ensure that the default will be compatible from introduction with the discovery feature without having to update the issuer (which would involve reissuing all existing bound tokens).

1: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-issuer-discovery

Related change: https://github.com/openshift/api/pull/617

### DIFF
--- a/test/e2e/bound_sa_token_test.go
+++ b/test/e2e/bound_sa_token_test.go
@@ -206,7 +206,8 @@ func TestTokenRequestAndReview(t *testing.T) {
 		sa.Name,
 		&authenticationv1.TokenRequest{
 			Spec: authenticationv1.TokenRequestSpec{
-				Audiences: []string{"auth.openshift.io"},
+				// Avoid specifying any audiences so that the token will be
+				// issued for the default audience of the issuer.
 			},
 		},
 		metav1.CreateOptions{})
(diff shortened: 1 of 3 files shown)


test/e2e/bound_sa_token_test.go
```diff
@@ -237,7 +238,7 @@ func TestChangeServiceAccountIssuer(t *testing.T) {
 	// Wait for operator readiness
 	test.WaitForKubeAPIServerClusterOperatorAvailableNotProgressingNotDegraded(t, configClient)
 
-	defaultIssuer := "auth.openshift.io"
+	defaultIssuer := "https://apiserver.openshift-kube-apiserver.svc"
```
stlaz, line 241: In order for the SA OIDC discovery to work, this must be reachable from the outside of the cluster

stlaz: /hold
see https://github.com/openshift/api/pull/617#discussion_r404755006 first, since this PR somewhat depends on the api, I think we should add the hold to this PR until the discussion there is resolved

stlaz: As per the above discussion, set to `https://kubernetes.default.svc` please

marun: Updated, PTAL.

marun:
Also updated the enhancement for consistency: https://github.com/openshift/enhancements/pull/281

PR #815, opened 2020-04-02
Bug 1820255: Fix race for localhost-recovery-client-token snapshotting

tnozicka:
When `localhost-recovery-client-token` secret got created there is a race between kube-controller-manager to inject a token into it and revision controller that snapshots it. If it snapshots it before the token is injected we deploy without a token. 

/cc @deads2k @soltysh @sttts


vendor/github.com/openshift/library-go/pkg/operator/resource/resourceapply/core.go
```diff
@@ -348,6 +348,24 @@ func SyncSecret(client coreclientv1.SecretsGetter, recorder events.Recorder, sou
 	case err != nil:
 		return nil, false, err
 	default:
+		if source.Type == corev1.SecretTypeServiceAccountToken {
```
tnozicka, line 351: fix is here

pkg/operator/starter.go
```diff
@@ -41,6 +42,18 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+type encryptionProvider []schema.GroupResource
```
tnozicka, line 45: @p0lyn0mial I suppose this could have been in library-go next to the interface as `SimpleProvider`

pkg/operator/starter.go
```diff
@@ -41,6 +42,18 @@ import (
...
+
+var _ controllers.Provider = encryptionProvider{}
+
+func (p encryptionProvider) EncryptedGRs() []schema.GroupResource {
```
soltysh, line 49: @p0lyn0mial I missed this in library-go, but I think we should try to keep interfaces more verbose rather than cryptic, iow. this should've been `EncryptedGroupResources` just to follow the API patterns we have everywhere. It's ok to name local variables this way but interfaces not necessarily, if in doubt look at k8s always :wink:

pkg/operator/starter.go
p0lyn0mial: yes, thx :)

pkg/operator/starter.go
p0lyn0mial: yes, thx :)

PR #816, opened 2020-04-03
Bug 1810184: Components using globs in Dockerfile COPY commands may break on OCP 4

p0lyn0mial:


p0lyn0mial:
should mirror https://github.com/openshift/cluster-kube-apiserver-operator/pull/808

Dockerfile.rhel7
```diff
@@ -5,8 +5,9 @@ ENV GO_PACKAGE github.com/openshift/cluster-kube-apiserver-operator
 RUN make build --warn-undefined-variables
 
 FROM registry.svc.ci.openshift.org/ocp/4.3:base
-RUN mkdir -p /usr/share/bootkube/manifests /manifests
-COPY --from=builder /go/src/github.com/openshift/cluster-kube-apiserver-operator/bindata/bootkube/* /usr/share/bootkube/manifests/
+COPY --from=builder /go/src/github.com/openshift/cluster-kube-apiserver-operator/bindata/bootkube/bootstrap-manifests /usr/share/bootkube/manifests/bootstrap-manifests/
+COPY --from=builder /go/src/github.com/openshift/cluster-kube-apiserver-operator/bindata/bootkube/config /usr/share/bootkube/manifests/config/
+COPY --from=builder /go/src/github.com/openshift/cluster-kube-apiserver-operator/bindata/bootkube/manifests /usr/share/bootkube/manifests/manifests/
 COPY --from=builder /go/src/github.com/openshift/cluster-kube-apiserver-operator/cluster-kube-apiserver-operator /usr/bin/
 COPY manifests/*.yaml /manifests/
```
p0lyn0mial, line 12: surprise, we can't copy the whole directory as we did in `master and 4.4` (`COPY manifests /manifests`) because it contains some `JSON` files to.

bparees:
BZ overridden:  covered by automated tests, low risk, needed for api.ci migration.  

will let team review/lgtm.

Dockerfile.rhel7
```diff
@@ -5,8 +5,9 @@ ENV GO_PACKAGE github.com/openshift/cluster-kube-apiserver-operator
 RUN make build --warn-undefined-variables
 
 FROM registry.svc.ci.openshift.org/ocp/4.3:base
-RUN mkdir -p /usr/share/bootkube/manifests /manifests
-COPY --from=builder /go/src/github.com/openshift/cluster-kube-apiserver-operator/bindata/bootkube/* /usr/share/bootkube/manifests/
```
p0lyn0mial, line 9: @bparees out of curiosity does it mean that in the past the directory structure for `bootkube` was incorrect?

bparees: no, you were just reliant on how "docker" handles `/*` which is not the same as how buildah does.  So it was fine as long as you were being built on docker.  As we're switching CI to build things using buildah, your dockerfile would have stopped building the right thing.

bparees: sorry, you were reliant on imagebuilder behavior, not docker.  docker+buildah work the same way.

bparees: see https://docs.google.com/document/d/1TJwlRj9C9HU-OhxKWnmd2O0FtYuXAZ3G0aFsD9UZPOA/edit#heading=h.tkhqbleqgny

bparees: imagebuilder preserved the subdirs when doing a cp from `foo/*` whereas docker+buildah flatten the subdir when you use *

p0lyn0mial: okay, I was just wondering why CI jobs were passing despite the destination dir being different. Thanks for explaining!

PR #822, opened 2020-04-08
Bug 1821448: Provide better messaging when Upgradeable is False

tkashem:
If Upgradeable is False due to default SCC mutation, we should provide better messaging.

### DIFF
--- a/pkg/operator/defaultscccontroller/defaultscccontroller.go
+++ b/pkg/operator/defaultscccontroller/defaultscccontroller.go
@@ -150,7 +150,7 @@ func (c *DefaultSCCController) Sync(key types.NamespacedName) error {
 	}
 
 	if len(mutated) > 0 {
-		klog.Infof("[%s] default scc has been mutated %s", ControllerName, mutated)
+		klog.Infof("[%s] default scc has been mutated %s, please visit https://bugzilla.redhat.com/show_bug.cgi?id=1821905#c22 to resolve the issue", ControllerName, mutated)
 	}
 
 	condition := NewCondition(mutated)


pkg/operator/defaultscccontroller/defaultscccontroller.go
```diff
@@ -150,7 +150,7 @@ func (c *DefaultSCCController) Sync(key types.NamespacedName) error {
 	}
 
 	if len(mutated) > 0 {
-		klog.Infof("[%s] default scc has been mutated %s", ControllerName, mutated)
+		klog.Infof("[%s] default scc has been mutated %s, reset the changes by a) delete the mutated SCC object(s) using oc b) delete any of the `apiserver-*` Pods from the `openshift-apiserver` namespace", ControllerName, mutated)
```
sttts, line 153: do we need b) ?

How about: has been mutated. Revert change by deleting the mutated SCCs and they will be recreated automatically.

tkashem: @sttts  Yes, I think so, we need to restart the openshift-apiserver which will create the missing SCC in a post-startup hook.

sttts: was reading this as "a or b". It's fine.

deads2k:
/hold

you need to describe how they actually grant the permissions too, or a pointer to docs.

derekwaynecarr:
note for 4/9: will review when hold is removed by team.

pkg/operator/defaultscccontroller/defaultscccontroller.go
tkashem: > /hold
> 
> you need to describe how they actually grant the permissions too, or a pointer to docs.

I don't think we can fit it in a message. All official documentation on SCC dates back to 3.9 and it basically asks to add user/group to the default SCC. 

I opened a BZ to fix the doc - https://bugzilla.redhat.com/show_bug.cgi?id=1822324

tkashem:
> /hold
> 
> you need to describe how they actually grant the permissions too, or a pointer to docs.

I don't think we can fit it in a message. All official documentation on SCC dates back to 3.9 and it basically asks to add user/group to the default SCC. 

I opened a BZ to fix the doc - https://bugzilla.redhat.com/show_bug.cgi?id=1822324

tkashem:
As an FYI, we are going to make some changes:
- OpenShift 4.3: Revert DefaultSecurityContextConstraints_Mutated in 4.3. We have a PR open for this - https://github.com/openshift/cluster-kube-apiserver-operator/pull/830. It will go into 4.3.z.

- OpenShift 4.4: Mark the CVO manifests for the default SCCs as `create-only`. CVO will create/recreate if any default SCCs are deleted but will tolerate changed made to any default SCC. 
https://github.com/openshift/cluster-kube-apiserver-operator/pull/831 (will be back ported to 4.4)

Hence, we no longer need this PR, closing it.

/close

PR #830, opened 2020-04-14
Bug 1823921: Revert SCC Upgradeable False

tkashem:


tkashem:
good to review
/hold cancel

PR #832, opened 2020-04-15
Fix kube-apiserver-to-kubelet-signer's refresh date

tnozicka:
[We always rotate after 4/5 of the life time](https://github.com/openshift/cluster-kube-apiserver-operator/blob/d62bbd3dce5400e28968447668b68f54f00dd685/vendor/github.com/openshift/library-go/pkg/operator/certrotation/signer.go#L83). Fixing the refresh date to be explicit helps others to better understand the life cycle.

/cc @deads2k @pweil-

### DIFF
--- a/pkg/operator/certrotationcontroller/certrotationcontroller.go
+++ b/pkg/operator/certrotationcontroller/certrotationcontroller.go
@@ -174,7 +174,7 @@ func newCertRotationController(
 			Namespace:              operatorclient.OperatorNamespace,
 			Name:                   "kube-apiserver-to-kubelet-signer",
 			Validity:               1 * 365 * defaultRotationDay, // this comes from the installer
-			Refresh:                8 * 365 * defaultRotationDay, // this means we effectively do not rotate
+			Refresh:                292 * defaultRotationDay,
 			RefreshOnlyWhenExpired: refreshOnlyWhenExpired,
 			Informer:               kubeInformersForNamespaces.InformersFor(operatorclient.OperatorNamespace).Core().V1().Secrets(),
 			Lister:                 kubeInformersForNamespaces.InformersFor(operatorclient.OperatorNamespace).Core().V1().Secrets().Lister(),


sttts:
/hold

This needs wider discussion: moving from "basically never" to "once a year".

tnozicka:
> This needs wider discussion: moving from "basically never" to "once a year".

@sttts We rotate when only 1/5 of the lifetime remains no matter what is set in refresh, so this PR should be noop. See the link I've provided in the PR description https://github.com/openshift/cluster-kube-apiserver-operator/blob/d62bbd3dce5400e28968447668b68f54f00dd685/vendor/github.com/openshift/library-go/pkg/operator/certrotation/signer.go#L83

pweil-:
@sttts part of the motivation here is exactly what you just stated: by just reading this code it appears we have a cert with 1 year validity that we will never rotate 😄 That's confusing to readers unless they happen to also know the code in `signer.go`.  Though a comment could suffice if folks object to putting a more realistic date in there.

tnozicka:
I think the 8 years was there initially because of a bug when this was set to 10 years validity. I've fixed that to 1 year initially to match the date from installer, but I missed the refresh. So it was at 4/5th of the lifetime which means we don't rotate unless we have to. This is adjusting it back to the 4/5th again.

sttts:
/hold cancel

Thanks for the explanation. Sounds reasonable.

pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -174,7 +174,7 @@ func newCertRotationController(
 			Namespace:              operatorclient.OperatorNamespace,
 			Name:                   "kube-apiserver-to-kubelet-signer",
 			Validity:               1 * 365 * defaultRotationDay, // this comes from the installer
-			Refresh:                8 * 365 * defaultRotationDay, // this means we effectively do not rotate
+			Refresh:                292 * defaultRotationDay,
```
deads2k, line 177: we refresh others in  halves

tnozicka: I am not sure how gracefully kubelet handles it, so I was just explicitly codifying existing behaviour so this PR is noop. If you sign up for 1/2 I can adjust that.

tnozicka:
bump @deads2k

tnozicka:
/retest

@deads2k ptal

tnozicka:
@deads2k ptal

PR #838, opened 2020-04-23
Bug 1827337: Remove stale condition DefaultSecurityContextConstraintsUpgradeable

tkashem:
We removed the controller that sets the condition `DefaultSecurityContextConstraintsUpgradeable` if any default SCC has been mutated. A cluster that already has user-modified default SCCs will have the above condition set. We need to remove the stale condition.


russellb:
As discussed elsewhere, we are manually bypassing some process on this one, since this 4.3 fix needs to get in as soon as possible.  I've manually adjusted the bug label for this special case.

PR #839, opened 2020-04-23
Bug 1827336: Remove stale condition

tkashem:
We removed the controller that sets the condition `DefaultSecurityContextConstraintsUpgradeable` if any default SCC has been mutated. A cluster that already has user-modified default SCCs will have the above condition set. We need to remove the stale condition.

Some user may do force upgrade to OpenShift 4.4, so we need this fix to go in 4.4 to clean the stale condition.


mfojtik:
/bugzilla refresh

The labels were applied manually as exception approved by @eparis as this bug is blocking upgrades from 4.3.z to 4.4.0.

PR #844, opened 2020-04-29
Copy openshift-tests binary in order to run resourcewatch

damemi:
This is needed to make the `openshift-tests` binary available during the operator e2e (see https://github.com/openshift/release/pull/8368)

### DIFF
--- a/Dockerfile.rhel7
+++ b/Dockerfile.rhel7
@@ -5,6 +5,7 @@ ENV GO_PACKAGE github.com/openshift/cluster-kube-apiserver-operator
 RUN make build --warn-undefined-variables
 
 FROM registry.svc.ci.openshift.org/ocp/4.4:base
+COPY --from=registry.svc.ci.openshift.org/ocp/4.5:tests /usr/bin/openshift-tests /usr/bin/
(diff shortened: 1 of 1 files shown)


damemi:
cc @mfojtik

sttts:
/assign @tnozicka 

Does this technically work?

Dockerfile.rhel7
```diff
@@ -5,6 +5,7 @@ ENV GO_PACKAGE github.com/openshift/cluster-kube-apiserver-operator
 RUN make build --warn-undefined-variables
 
 FROM registry.svc.ci.openshift.org/ocp/4.4:base
+COPY --from=registry.svc.ci.openshift.org/ocp/4.5:tests /usr/bin/openshift-tests /usr/bin/
```
soltysh, line 8: Is it possible to have this line conditional, I don't want to end up with production images containing that binary?

damemi: I think there are certain env flags that are set in CI now to indicate it's a test environment, but I need to find what those are and we can use them

damemi: So, there is an `${OPENSHIFT_CI}` env var we could get: https://steps.svc.ci.openshift.org/help#env

But from a quick search it doesn't seem like you can really conditionally COPY in Dockerfiles based on env vars. My docker-fu might just be weak, but if we can figure that out we could do this

PR #845, opened 2020-04-29
Bug 1829580: Add e2e test that ensures stale condition is removed

tkashem:
Add an e2e test to validates that stale condition DefaultSecurityContextConstraintsUpgradeable is removed from the status block.

### DIFF
--- a/test/e2e/clean_stale_condition_test.go
+++ b/test/e2e/clean_stale_condition_test.go
@@ -1,0 +1,75 @@
+package e2e
+
+import (
+	"testing"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/util/retry"
+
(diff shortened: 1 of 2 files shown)


test/e2e/clean_stale_condition_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"testing"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/util/retry"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	operatorclient "github.com/openshift/client-go/operator/clientset/versioned/typed/operator/v1"
+	test "github.com/openshift/cluster-kube-apiserver-operator/test/library"
+
+	"github.com/stretchr/testify/require"
```
sttts, line 13: nit: move into second block

PR #850, opened 2020-05-08
Set flags instead of different config format

deads2k:
Switch all the "easy" settings to flags instead of config.  There is a little bit more that isn't too hard, but requires config observation changes.  Let's clear the early stuff first.  This makes us more kube-like and will help answer questions about "how do we do benchmark X" and it helps focus our patches.

/assign @sttts

### DIFF
--- a/bindata/v4.1.0/config/config-overrides.yaml
+++ b/bindata/v4.1.0/config/config-overrides.yaml
@@ -10,8 +10,11 @@ apiServerArguments:
     - https://kubernetes.default.svc
   service-account-signing-key-file:
     - /etc/kubernetes/static-pod-certs/secrets/bound-service-account-signing-key/service-account.key
-  service-account-key-file:
-    - /etc/kubernetes/static-pod-resources/configmaps/sa-token-signing-certs
-      # The following path contains the public keys needed to verify bound sa
-      # tokens. This is only supported post-bootstrap.
-    - /etc/kubernetes/static-pod-resources/configmaps/bound-sa-token-signing-certs
+serviceAccountPublicKeyFiles:
+  # this being a directory means we cannot directly use the upstream flags.
+  # TODO make a configobserver that writes the individual values that we need.
+  - /etc/kubernetes/static-pod-resources/configmaps/sa-token-signing-certs
+  # The following path contains the public keys needed to verify bound sa
+  # tokens. This is only supported post-bootstrap.
+  - /etc/kubernetes/static-pod-resources/configmaps/bound-sa-token-signing-certs
+
--- a/bindata/v4.1.0/config/config-overrides.yaml
+++ b/bindata/v4.1.0/config/config-overrides.yaml
@@ -5,13 +5,13 @@ apiServerArguments:
   # tokens. This is only supported post-bootstrap so these
   # values must not appear in defaultconfig.yaml.
   service-account-issuer:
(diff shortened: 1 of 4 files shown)


deads2k:
see https://github.com/openshift/origin/pull/24960 for comparison

p0lyn0mial:
lgtm

sttts:
lgtm

Would like to have another pair of eyes on it.

pkg/operator/v410_00_assets/bindata.go
```diff
@@ -179,49 +251,16 @@ auditConfig:
...
-    - X-Remote-Group
-    usernameHeaders:
-    - X-Remote-User
-  webhookTokenAuthenticators: null
 consolePublicURL: ""
-kubeletClientInfo:
-  ca: /etc/kubernetes/static-pod-resources/configmaps/kubelet-serving-ca/ca-bundle.crt
-  certFile: /etc/kubernetes/static-pod-resources/secrets/kubelet-client/tls.crt
-  keyFile: /etc/kubernetes/static-pod-resources/secrets/kubelet-client/tls.key
-  port: 10250
```
p0lyn0mial, line 200: this was removed, I think it's okay because it's the default port anyway.

pkg/operator/v410_00_assets/bindata.go
```diff
@@ -113,39 +116,108 @@ admission:
...
+    - "0"
+  kubernetes-service-node-port:
+    - "0"
   # value should logically scale with max-requests-inflight
   max-mutating-requests-inflight:
-  - "1000"
+    - "1000"
+  # value needed to be bumped for scale tests.  The kube-apiserver did ok here
+  max-requests-inflight:
+    - "3000"
```
p0lyn0mial, line 183: previously it was set to `1200`

deads2k: it lied.  it was hard coded in origin.  it should be at least 3k

pkg/operator/v410_00_assets/bindata.go
deads2k: > this was removed, I think it's okay because it's the default port anyway.

and there's no obvious flag for me to set.

bindata/v4.1.0/config/config-overrides.yaml
```diff
@@ -5,13 +5,16 @@ apiServerArguments:
   # tokens. This is only supported post-bootstrap so these
   # values must not appear in defaultconfig.yaml.
   service-account-issuer:
-  - https://kubernetes.default.svc
+    - https://kubernetes.default.svc
```
sttts, line 8: indented yaml lists hurt my eyes

bindata/v4.1.0/config/defaultconfig.yaml
```diff
@@ -59,36 +76,88 @@ apiServerArguments:
...
   # switch to direct pod IP routing for aggregated apiservers to avoid service IPs as on source of instability
   enable-aggregator-routing:
-  - "true"
-  shutdown-delay-duration:
-  - 70s # give SDN some time to converge: 30s for iptable lock contention, 25s for the second try and some seconds for AWS to update ELBs
+    - "true"
+  enable-logs-handler:
+    - "false"
+  enable-swagger-ui:
+    - "true"
```
sttts, line 85: is this still a thing?

PR #852, opened 2020-05-11
switch etcd storage urls to be kube-apiserver args

deads2k:
Brings us closer in-line with kube-apiserver.  There's no value to a separate config option here.


ironcladlou:
Deletes a bunch of code, what's not to love?

/lgtm

PR #853, opened 2020-05-11
pass admission plugins to be enabled

deads2k:
This keeps the list of admission separate from the kube patches that we maintain for inspectability and reliability.

### DIFF
--- a/bindata/v4.1.0/config/defaultconfig.yaml
+++ b/bindata/v4.1.0/config/defaultconfig.yaml
@@ -14,6 +14,51 @@ aggregatorConfig:
     certFile: /etc/kubernetes/static-pod-certs/secrets/aggregator-client/tls.crt
     keyFile: /etc/kubernetes/static-pod-certs/secrets/aggregator-client/tls.key
 apiServerArguments:
+  enable-admission-plugins:
+    - CertificateApproval
+    - CertificateSigning
+    - CertificateSubjectRestriction
+    - DefaultIngressClass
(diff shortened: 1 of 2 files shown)


deads2k:
@sttts my manual diff says the admission plugin lists look the same before and after

bindata/v4.1.0/config/defaultconfig.yaml
```diff
@@ -14,6 +14,51 @@ aggregatorConfig:
     certFile: /etc/kubernetes/static-pod-certs/secrets/aggregator-client/tls.crt
     keyFile: /etc/kubernetes/static-pod-certs/secrets/aggregator-client/tls.key
 apiServerArguments:
+  enable-admission-plugins:
```
sttts, line 17: as discussed, would like to see some test that fails when this list diverges from upstream+static-diff. Some simple origin carry patch should be able to do that. Or I believe you can even call the kube-apiserver binary. The command line help text contains the list.

deads2k: > as discussed, would like to see some test that fails when this list diverges from upstream+static-diff. Some simple origin carry patch should be able to do that. Or I believe you can even call the kube-apiserver binary. The command line help text contains the list.

This is no worse than today with regard to in and out.  In fact, it has better failure modes on registration failures.  I suggest dealing with the e2e test separately.

mfojtik:
@deads2k please make a BZ about e2e test so we won't forget to add one.

PR #859, opened 2020-05-15
Bug 1836927: Prefer the new etcd endpoints configmap for storage URL discovery

ironcladlou:
xref https://github.com/openshift/cluster-etcd-operator/pull/354


deads2k:
it looks plausible to me, but ask @sttts or @p0lyn0mial @sanchezl

ironcladlou:
Now that I'm looking at it again, I'd rather keep them orthogonal if possible, thinking...

ironcladlou:
Latest commit tries to keep the observers orthogonal by making the `endpoint` observer no-op when the configmap exists.

pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -14,15 +14,17 @@ import (
 	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 
+	"github.com/openshift/library-go/pkg/controller/factory"
+
```
deads2k, line 18: no spaces

pkg/operator/configobservation/etcd/observe_etcd.go
```diff
@@ -23,6 +23,14 @@ const (
 // then the previously observed storage config URLs will be re-used.
 func ObserveStorageURLs(genericListers configobserver.Listers, recorder events.Recorder, currentConfig map[string]interface{}) (map[string]interface{}, []error) {
 	listers := genericListers.(configobservation.Listers)
+
+	// If the cluster has already been migrated to the endpoints configmap, no-op
+	if _, err := listers.ConfigmapLister.ConfigMaps(etcdEndpointNamespace).Get("etcd-endpoints"); err == nil {
```
deads2k, line 28: This is racy and risks breakage if a trigger happens for a different reason and one observer sees the configmap and the other doesn't

ironcladlou: Do you have some suggestion for how to implement this safely?

ironcladlou: I'm not sure I understand the impact of the race. If the endpoints version doesn't see the configmap and continues to try and read the endpoint, the controller will:

1. Fail to find the endpoint because this is a new cluster where the endpoint never existed and will never exist, and on retry the cm should eventually be observed and this func should begin correctly no-op'ing
2. Fail to find the endpoint because this is an upgrade and the cm either hasn't been created yet or again the cache is stale (in which case either read the old endpoint which is still valid, and eventually start no-op'ing once the cm is available)

Help me understand the race impact?

deads2k: @ironcladlou if the endpoint is found and the configmap is found, then two configobservers will try to set the same fields and we go degraded on that condition.  If it doesn't happen for very long, this may be acceptable.

ironcladlou: Two ways I can think of to address this.

The original code I had (https://github.com/openshift/cluster-kube-apiserver-operator/pull/859/commits/582e221c1368bb5536bb08b2e19f981753247a50) registers only the configmap observer, which itself delegates to the endpoints observer if the configmap isn't found.

Another possible way is to refactor the endpoints observer code to internally handle the existence of either the endpoints or configmap. I have a commit started for this but it's pretty invasive to the current endpoints handling code.

If https://github.com/openshift/cluster-kube-apiserver-operator/pull/859/commits/582e221c1368bb5536bb08b2e19f981753247a50 is okay I can revert to that.

ironcladlou: Alright, I refactored to an approach which uses a delegation model to ensure only one implementation is ever used at a time depending on whether the configmap exists during the observation attempt

pkg/operator/configobservation/etcdendpoints/observe_etcd_endpoints.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/configobservation"
+	endpointsobserver "github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/configobservation/etcd"
+)
+
+const (
+	etcdEndpointNamespace = "openshift-etcd"
+	etcdEndpointName      = "etcd-endpoints"
+)
+
+func BackwardsCompatibleObserveStorageURLs(genericListers configobserver.Listers, recorder events.Recorder, currentConfig map[string]interface{}) (map[string]interface{}, []error) {
```
ironcladlou, line 23: @sanchezl PTAL at this approach, I think it's a decent middle ground

ironcladlou:
e2e test failures seem unrelated.

ironcladlou:
@deads2k @sanchezl @hexfusion @retroflexer this is ready for review.

pkg/operator/configobservation/etcdendpoints/observe_etcd_endpoints.go
```diff
@@ -0,0 +1,110 @@
...
+
+const (
+	etcdEndpointNamespace = "openshift-etcd"
+	etcdEndpointName      = "etcd-endpoints"
+)
+
+// BackwardsCompatibleObserveStorageURLs will prefer to observe the new
+// etcd-endpoints configmap, and will fall back to reading the endpoint
+// resource only if the configmap doesn't yet exist.
+func BackwardsCompatibleObserveStorageURLs(genericListers configobserver.Listers, recorder events.Recorder, currentConfig map[string]interface{}) (map[string]interface{}, []error) {
```
deads2k, line 26: interesting. These used to be different, but have we collapsed to the same as kas-o now?  If so, we should consider generalizing at some point.

pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -115,7 +115,7 @@ func NewConfigObserver(
 				// static path at which we expect to find the encryption config secret
 				"/etc/kubernetes/static-pod-resources/secrets/encryption-config/encryption-config",
 			),
-			etcd.ObserveStorageURLs,
+			etcdendpoints.BackwardsCompatibleObserveStorageURLs,
```
deads2k, line 118: interesting namespaces don't include etcd, so you won't be triggered I don't think

deads2k, line 118: also needs to be in preruncachesync

ironcladlou: I also just realized my configmap lister setup isn't scoped to the etcd namespace

ironcladlou: Fixed

ironcladlou: Fixed

pkg/operator/configobservation/etcdendpoints/observe_etcd_endpoints.go
```diff
@@ -0,0 +1,109 @@
...
+	listers := genericListers.(configobservation.Listers)
+	if _, err := listers.ConfigmapLister.ConfigMaps(etcdEndpointNamespace).Get(etcdEndpointName); err == nil {
+		return ObserveStorageURLs(genericListers, recorder, currentConfig)
+	} else if !errors.IsNotFound(err) {
+		return currentConfig, []error{err}
```
deads2k, line 31: `	previouslyObservedConfig := map[string]interface{}{}` as done a few lines down.  returning `currentConfig` means a different thing and will result in conflicts with other observers.

pkg/operator/configobservation/etcdendpoints/observe_etcd_endpoints.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	var etcdURLs []string
+	etcdEndpoints, err := listers.ConfigmapLister.ConfigMaps(etcdEndpointNamespace).Get(etcdEndpointName)
+	if errors.IsNotFound(err) {
+		// In clusters prior to 4.5, fall back to reading the old host-etcd-2 endpoint
+		// resource, if possible. In 4.6 we can assume consumers have been updated to
+		// use the configmap, delete the fallback code, and throw an error if the
+		// configmap doesn't exist.
+		observedConfig, fallbackErrors := fallbackObserver(listers, recorder, currentConfig)
```
sanchezl, line 50: Consider adding an event here letting us know that we are using the fallback. It would be useful for debugging.

ironcladlou: Would that bring us back to the same event spam situation already discussed in https://github.com/openshift/cluster-openshift-apiserver-operator/pull/364#discussion_r426928928?

ironcladlou: Based on prior convo I decided not to add an event here. If it turns out there's some unforeseen problem in this piece, we could do a followup.

ironcladlou:
@hexfusion @deads2k @sanchezl this is once again ready for review, PTAL

deads2k:
let's go!

/lgtm

ironcladlou:
I haven't had a chance to deep dive into the latest CI runs to explain them away as systemic flakes vs. regressions in this PR.

ironcladlou:
e2e is flaking and the problems are obscured by https://bugzilla.redhat.com/show_bug.cgi?id=1837992. Added the fix from https://github.com/openshift/cluster-kube-apiserver-operator/pull/864 to see if that helps understand or resolve the problem.

ironcladlou:
Both the endpoint and configmap observer were missing a sort of the final etcd URL array, causing endless churn rolling out new apiserver revisions.

PR #863, opened 2020-05-19
Bug 1806980: remove override for fast cert rotation

deads2k:
we need to do this and then revert it for 4.6.  Ideally, we only have this for 4.5


sttts:
/hold

until release has branched.

PR #864, opened 2020-05-20
Bug 1837992: wait for port 6443 to be open in the kube-apiserver container; use ss isntead of lsof

squeed:
There's already an InitContainer that does this, but it doesn't handle container restarts, only pod restarts.

### DIFF
--- a/bindata/v4.1.0/kube-apiserver/pod.yaml
+++ b/bindata/v4.1.0/kube-apiserver/pod.yaml
@@ -24,7 +24,7 @@ spec:
         echo -n "Fixing audit permissions."
         chmod 0700 /var/log/kube-apiserver
         echo -n "Waiting for port :6443 and :6080 to be released."
-        while [ -n "$(lsof -ni :6443)$(lsof -ni :6080)" ]; do
+        while [ -n "$(ss -Htan '( sport = 6443 or sport = 6080 )')" ]; do
           echo -n "."
           sleep 1
         done
@@ -42,6 +42,18 @@ spec:
             echo "Copying system trust bundle"
             cp -f /etc/kubernetes/static-pod-certs/configmaps/trusted-ca-bundle/ca-bundle.crt /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
           fi
+          echo -n "Waiting for port :6443 and :6080 to be released."
+          tries=0
+          while [ -n "$(ss -Htan '( sport = 6443 or sport = 6080 )')" ]; do
+            echo -n "."
+            sleep 1
+            (( tries += 1 ))
+            if [[ "${tries}" -gt 105 ]]; then
+              echo "timed out waiting for port :6443 and :6080 to be released"
+              exit 1
+            fi
+          done
+          echo
           exec hyperkube kube-apiserver --openshift-config=/etc/kubernetes/static-pod-resources/configmaps/config/config.yaml --advertise-address=${HOST_IP}
     resources:
       requests:
--- a/pkg/operator/v410_00_assets/bindata.go
+++ b/pkg/operator/v410_00_assets/bindata.go
@@ -461,7 +461,7 @@ spec:
         echo -n "Fixing audit permissions."
         chmod 0700 /var/log/kube-apiserver
         echo -n "Waiting for port :6443 and :6080 to be released."
-        while [ -n "$(lsof -ni :6443)$(lsof -ni :6080)" ]; do
+        while [ -n "$(ss -Htan '( sport = 6443 or sport = 6080 )')" ]; do
           echo -n "."
           sleep 1
         done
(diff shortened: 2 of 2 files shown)


squeed:
More background in the bug, but in https://bugzilla.redhat.com/show_bug.cgi?id=1834908 we see some connection timeouts that are actually because the apiserver restarted. However, we don't get the true reason for the restart, because the apiserver container crashloops until the port is available.

sttts:
/hold

This hides another issue of broken graceful termination.

squeed:
> This hides another issue of broken graceful termination.

True, but it also makes it possible to debug it.

Right now, the container logs for the kube-apisever process that failed to gracefully shut down are gone forever, thanks to the restart loop. With this in, at least we'll be able to figure out what's causing the restart.

sttts:
/lgtm

Let's unhold when we know that graceful termination is fixed. Tracked in https://bugzilla.redhat.com/show_bug.cgi?id=1828606.

bindata/v4.1.0/kube-apiserver/pod.yaml
```diff
@@ -42,6 +42,17 @@ spec:
             echo "Copying system trust bundle"
             cp -f /etc/kubernetes/static-pod-certs/configmaps/trusted-ca-bundle/ca-bundle.crt /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
           fi
+          echo -n "Waiting for port :6443 and :6080 to be released."
```
sttts, line 45: Look in line 27. Why is that not enough?

squeed: That handles this case on pod restarts, not container restarts. The InitContainer only runs once.

squeed:
Interesting; this doesn't fix the problem. Check out https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-kube-apiserver-operator/864/pull-ci-openshift-cluster-kube-apiserver-operator-master-e2e-aws/3746/artifacts/e2e-aws/gather-extra/pods/openshift-kube-apiserver_kube-apiserver-ip-10-0-228-116.ec2.internal_kube-apiserver_previous.log

It appears the port is free, according to `lsof`, but kube-apiserver fails to start. Investigating.

squeed:
Update: we should be using `ss`, not `lsof`. Filed https://github.com/openshift/origin/pull/25002 to add it to the image.

squeed:
`ss` is now available in hyperkube image; PR updated.

soltysh:
> Update: we should be using `ss`, not `lsof`. Filed [openshift/origin#25002](https://github.com/openshift/origin/pull/25002) to add it to the image.

Any more details why? We'll need to move this to kcm and ks, too.

squeed:
> Any more details why? We'll need to move this to kcm and ks, too.

In my testing, I couldn't get `lsof` to show me sockets in `TIME_WAIT` status. Now I'm not so sure about that, and I certainly can't find any documentation around it one way or another


In any case, the important fix is to wait in kube-apiserver's container spec, not just the init container.

squeed:
aws blip (terraform error)
/retest

bindata/v4.1.0/kube-apiserver/pod.yaml
```diff
@@ -24,7 +24,7 @@ spec:
         echo -n "Fixing audit permissions."
         chmod 0700 /var/log/kube-apiserver
         echo -n "Waiting for port :6443 and :6080 to be released."
-        while [ -n "$(lsof -ni :6443)$(lsof -ni :6080)" ]; do
+        while [ -n "$(ss -Htan '( sport = 6443 or sport = 6080 )')" ]; do
```
sttts, line 27: can you also update the bootstrap container? I think we have it there as well (some sibling directory).

bindata/v4.1.0/kube-apiserver/pod.yaml
sttts: Should we remove it from the init container?

bindata/v4.1.0/kube-apiserver/pod.yaml
squeed: the bootstrap doesn't have any port waiting - should I add it in?

sttts: Let me check. I thought we do have it there.

sttts: We don't. So nvmd. No need for bootstrapping then.

bindata/v4.1.0/kube-apiserver/pod.yaml
squeed: I was thinking about that. I decided against it, mostly because the init-container provides a useful bit of feedback w.r.t. pod lifecycle. But I don't hold this opinion particularly strongly.

squeed:
I suspect the two failing operator tests need to be tweaked; this code doesn't touch anything in that area.

squeed:
And the tests don't have timestamps so it's hard to correlate :-/

bindata/v4.1.0/kube-apiserver/pod.yaml
deads2k: > I was thinking about that. I decided against it, mostly because the init-container provides a useful bit of feedback w.r.t. pod lifecycle. But I don't hold this opinion particularly strongly.

The init container also prevents the time spent waiting from counting against a health check timer.

squeed:
all green! just needs lgtm.

sttts:
CI flakes

/retest

sttts:
Go CI, go!

PR #868, opened 2020-05-27
wait for Kube API rollout while checking the token

p0lyn0mial:


test/e2e/bound_sa_token_test.go
```diff
@@ -124,14 +124,22 @@ func TestBoundTokenSignerController(t *testing.T) {
...
+		}, promotionTimeout, newOperatorSecret.Data)
 	})
 }
 
 // checkBoundTokenOperandSecret checks that the operand secret is
 // populated with the expected data.
-func checkBoundTokenOperandSecret(t *testing.T, kubeClient *clientcorev1.CoreV1Client, timeout time.Duration, expectedData map[string][]byte) {
+func checkBoundTokenOperandSecret(t *testing.T, kubeClient *clientcorev1.CoreV1Client, additionalCheckFn func(), timeout time.Duration, expectedData map[string][]byte) {
 	err := wait.PollImmediate(interval, timeout, func() (done bool, err error) {
+
```
sttts, line 138: nit: no empty line

test/e2e/bound_sa_token_test.go
```diff
@@ -124,14 +124,22 @@ func TestBoundTokenSignerController(t *testing.T) {
...
+				t.Log("WaitForKubeAPIServerClusterOperatorAvailableNotProgressingNotDegraded")
+				test.WaitForKubeAPIServerClusterOperatorAvailableNotProgressingNotDegraded(t, configClient)
+		}, promotionTimeout, newOperatorSecret.Data)
 	})
 }
 
 // checkBoundTokenOperandSecret checks that the operand secret is
 // populated with the expected data.
-func checkBoundTokenOperandSecret(t *testing.T, kubeClient *clientcorev1.CoreV1Client, timeout time.Duration, expectedData map[string][]byte) {
+func checkBoundTokenOperandSecret(t *testing.T, kubeClient *clientcorev1.CoreV1Client, additionalCheckFn func(), timeout time.Duration, expectedData map[string][]byte) {
```
sttts, line 136: doc additionalCheckFn

sttts:
Needs a PR description what is actually done.

p0lyn0mial:
I think that the issue is that the test doesn't know how to detect that a new version of the server has been rolled out and it's okay to go.

Rolling out a new version is an expensive process that takes quite a long time (30 min?).
Likewise triggering a new release - it's not instant - I suspect it takes a few minutes between "I want I new version" and "OK, I'm starting rolling out a new version"

Like this particular line - I suspect it almost always yields immediately which will cause the test to think it's okay to test . . . 

https://github.com/openshift/cluster-kube-apiserver-operator/blob/31f3997aceeb8b239d0b469b12361b99e3533423/test/e2e/bound_sa_token_test.go#L84

I increased the timeout and added an additional check that will wait for the operator (AvailableNotProgressingNotDegraded) while waiting for the token and it seems to improve the test.

p0lyn0mial:
I think that we could change `WaitForKubeAPIServerClusterOperatorAvailableNotProgressingNotDegraded` function. 

It could read revisions and wait until all replicas are on the same revision. It could then report `OK` only when the same revision has been seen for `X minutes`

p0lyn0mial:
I think that increasing only the timeout will fix the issue - trying now.

p0lyn0mial:
I would also remove the usage of `WaitForKubeAPIServerClusterOperatorAvailableNotProgressingNotDegraded`from:

 - https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/test/e2e/bound_sa_token_test.go#L62
- https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/test/e2e/bound_sa_token_test.go#L71

p0lyn0mial:
I think it also has a few more issues, for example, `TestEncryptionTypeAESCBC` can start rolling out a new version and fail before completion. That will definitely impact the next tests - as they expect a stable environment and will very likely fail

squeed:
If you wind up force-pushing again, I'd suggest changing the `fmt.Printf` to `t.Logf`, so you get timestamps, making this a lot easier to debug in the future.

mfojtik:
retest

PR #870, opened 2020-05-28
make SA Token tests to wait for API Servers to be on the same revision

p0lyn0mial:
this PR improves the SA tokens tests by identifying tests that roll out a new version and waiting until all APIs converge onto the same revision.

### DIFF
--- a/test/e2e/bound_sa_token_test.go
+++ b/test/e2e/bound_sa_token_test.go
@@ -30,21 +30,17 @@ import (
 const (
 	interval       = 1 * time.Second
 	regularTimeout = 30 * time.Second
-
-	// Need a long time for promotion to account for the delay in
-	// nodes being updated to a revision of the configmap that
-	// contains the latest public key.
-	promotionTimeout = 40 * time.Minute
 )
 
 // TestBoundTokenSignerController verifies the expected behavior of the controller
 // with respect to the resources it manages.
+//
+// Note: this test will roll out a new version - multiple times
 func TestBoundTokenSignerController(t *testing.T) {
 	kubeConfig, err := testlibrary.NewClientConfigForTest()
 	require.NoError(t, err)
 	kubeClient, err := clientcorev1.NewForConfig(kubeConfig)
 	require.NoError(t, err)
-	configClient, err := configclient.NewForConfig(kubeConfig)
 	require.NoError(t, err)
 
 	targetNamespace := operatorclient.TargetNamespace
@@ -59,7 +55,6 @@ func TestBoundTokenSignerController(t *testing.T) {
 
 	// The operand secret should be recreated after deletion.
 	t.Run("operand-secret-deletion", func(t *testing.T) {
-		test.WaitForKubeAPIServerClusterOperatorAvailableNotProgressingNotDegraded(t, configClient)
 
 		err := kubeClient.Secrets(targetNamespace).Delete(context.TODO(), tokenctl.SigningKeySecretName, metav1.DeleteOptions{})
 		require.NoError(t, err)
@@ -67,25 +62,30 @@ func TestBoundTokenSignerController(t *testing.T) {
 	})
 
 	// The operand config map should be recreated after deletion.
+	// Note: it will roll out a new version
 	t.Run("configmap-deletion", func(t *testing.T) {
-		test.WaitForKubeAPIServerClusterOperatorAvailableNotProgressingNotDegraded(t, configClient)
-
 		err := kubeClient.ConfigMaps(targetNamespace).Delete(context.TODO(), tokenctl.PublicKeyConfigMapName, metav1.DeleteOptions{})
 		require.NoError(t, err)
 		checkCertConfigMap(t, kubeClient, map[string]string{
 			"service-account-001.pub": string(operatorPublicKey),
 		})
+
+		// deletion triggers a roll-out - wait until a new version has been rolled out
+		test.WaitForAPIServerToStabilizeOnTheSameRevision(t, kubeClient.Pods(operatorclient.TargetNamespace))
 	})
 
 	// The secret in the operator namespace should be recreated with a new keypair
 	// after deletion. The configmap in the operand namespace should be updated
 	// immediately, and the secret once the configmap is present on all nodes.
+	//
+	// Note: it will roll out a new version
 	t.Run("operator-secret-deletion", func(t *testing.T) {
-		test.WaitForKubeAPIServerClusterOperatorAvailableNotProgressingNotDegraded(t, configClient)
-
 		// Delete the operator secret
 		err := kubeClient.Secrets(operatorNamespace).Delete(context.TODO(), tokenctl.NextSigningKeySecretName, metav1.DeleteOptions{})
 		require.NoError(t, err)
+
+		// deletion triggers a roll-out - wait until a new version has been rolled out
+		test.WaitForAPIServerToStabilizeOnTheSameRevision(t, kubeClient.Pods(operatorclient.TargetNamespace))
 
 		// Ensure that the cert configmap is always removed at the end of the test
 		// to ensure it will contain only the current public key. This property is
@@ -124,7 +124,7 @@ func TestBoundTokenSignerController(t *testing.T) {
 		})
 
 		// Check that the operand secret is updated within the promotion timeout
-		checkBoundTokenOperandSecret(t, kubeClient, promotionTimeout, newOperatorSecret.Data)
+		checkBoundTokenOperandSecret(t, kubeClient, regularTimeout, newOperatorSecret.Data)
 	})
 }
 
@@ -221,38 +221,6 @@ func TestTokenRequestAndReview(t *testing.T) {
 	require.NoError(t, err)
 	require.Empty(t, trev.Status.Error)
 	require.True(t, trev.Status.Authenticated)
-}
-
-// TestChangeServiceAccountIssuer checks that the operator considers
-// the value set for Authentication.ServiceAccountIssuer when setting
-// the configuration configmap for the apiserver pods.
-func TestChangeServiceAccountIssuer(t *testing.T) {
-	kubeConfig, err := testlibrary.NewClientConfigForTest()
-	require.NoError(t, err)
-	kubeClient, err := kubernetes.NewForConfig(kubeConfig)
-	require.NoError(t, err)
-	coreClient := kubeClient.CoreV1()
-	configClient, err := configclient.NewForConfig(kubeConfig)
-	require.NoError(t, err)
-
-	// Wait for operator readiness
-	test.WaitForKubeAPIServerClusterOperatorAvailableNotProgressingNotDegraded(t, configClient)
-
-	defaultIssuer := "https://kubernetes.default.svc"
-
-	// Check that the default issuer is set in the operand config
-	require.NoError(t, pollForOperandIssuer(t, coreClient, defaultIssuer))
-
-	nonDefaultIssuer := "https://my-custom-issuer.com"
-	// Update the issuer to a valid value (corner cases are unit tested)
-	setServiceAccountIssuer(t, configClient, nonDefaultIssuer)
-	// Check that the issuer has changed to the non-default value
-	require.NoError(t, pollForOperandIssuer(t, coreClient, nonDefaultIssuer))
-
-	// Clear the issuer
-	setServiceAccountIssuer(t, configClient, "")
-	// Check that the issuer has changed back to the default
-	require.NoError(t, pollForOperandIssuer(t, coreClient, defaultIssuer))
 }
 
 func pollForOperandIssuer(t *testing.T, client clientcorev1.CoreV1Interface, expectedIssuer string) error {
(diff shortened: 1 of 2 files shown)


test/e2e/bound_sa_token_test.go
```diff
@@ -226,6 +227,8 @@ func TestTokenRequestAndReview(t *testing.T) {
 // TestChangeServiceAccountIssuer checks that the operator considers
 // the value set for Authentication.ServiceAccountIssuer when setting
 // the configuration configmap for the apiserver pods.
+//
+// Note: this test will roll out a new version
 func TestChangeServiceAccountIssuer(t *testing.T) {
```
p0lyn0mial, line 232: @marun  It seems that this test simply sets and check if a flag `apiServerArguments/service-account-issuer` appears in the config map. It will also trigger a deployment twice. So it would take ~20 minutes to complete. What's the real value of this test?

test/library/library.go
```diff
@@ -32,3 +47,117 @@ func GenerateNameForTest(t *testing.T, prefix string) string {
...
+	})); err != nil {
+		t.Fatal(err)
+	}
+}
+
+// areAPIServersOnTheSameRevision tries to find the current revision that the API servers are running at.
+// The number of instances is calculated based on the number of running pods in a namespace.
+// This should be okay because this function is meant to be used by WaitForAPIServerToBeOnTheSameRevision which will wait at least waitForAPIRevisionSuccessThreshold * waitForAPIRevisionSuccessInterval
+// The number of pods should stabilize in that period of time.
+func areAPIServersOnTheSameRevision(podClient corev1client.PodInterface) (bool, error) {
```
p0lyn0mial, line 70: I took that code from https://github.com/openshift/library-go/blob/master/pkg/operator/encryption/deployer/revisionedpod.go#L162 and simplified it a bit.

test/library/library.go
```diff
@@ -32,3 +47,117 @@ func GenerateNameForTest(t *testing.T, prefix string) string {
...
 }
+
+// WaitForAPIServerToBeOnTheSameRevision waits until all API Servers are running at the same revision.
+// The API Servers must stay on the same revision for at least waitForAPIRevisionSuccessThreshold * waitForAPIRevisionSuccessInterval
+//
+// Observations:
+//  rolling out a new version is not instant you need to account for a propagation time (~1/2 minutes)
+//  rolling out a new version can take ~10 minutes
+func WaitForAPIServerToBeOnTheSameRevision(t *testing.T, podClient corev1client.PodInterface) {
+	if err := wait.Poll(waitForAPIRevisionPollInterval, waitForAPIRevisionTimeout, mustSucceedMultipleTimes(waitForAPIRevisionSuccessThreshold, waitForAPIRevisionSuccessInterval, func() (bool, error) {
```
p0lyn0mial, line 58: it has to succeed multiple times (stable revision for ~1/2 minutes)

test/library/library.go
```diff
@@ -32,3 +47,117 @@ func GenerateNameForTest(t *testing.T, prefix string) string {
...
+// The number of instances is calculated based on the number of running pods in a namespace.
+// This should be okay because this function is meant to be used by WaitForAPIServerToBeOnTheSameRevision which will wait at least waitForAPIRevisionSuccessThreshold * waitForAPIRevisionSuccessInterval
+// The number of pods should stabilize in that period of time.
+func areAPIServersOnTheSameRevision(podClient corev1client.PodInterface) (bool, error) {
+	revisionLabel := "revision"
+
+	// do a live list so we never get confused about what revision we are on
+	apiServerPods, err := podClient.List(context.TODO(), metav1.ListOptions{LabelSelector: "apiserver=true"})
+	if err != nil {
+		return false, err
```
marun, line 76: Maybe ensure this error is logged rather than returning it to fail the test?  Operations subject to  transient infrastructure failures may be recoverable, and attempting to do so is likely to reduce the potential for flakes.

test/library/library.go
```diff
@@ -32,3 +47,117 @@ func GenerateNameForTest(t *testing.T, prefix string) string {
...
+
+func revisions(revisionLabel string, pods []*corev1.Pod) sets.String {
+	ret := sets.NewString()
+	for _, p := range pods {
+		ret.Insert(p.Labels[revisionLabel])
+	}
+	return ret
+}
+
+func categorizePods(pods []corev1.Pod) (good []*corev1.Pod, bad []*corev1.Pod, progressing bool, err error) {
```
marun, line 113: I didn't clue in to the majority of these additions being copied from library-go until after I'd reviewed this method in detail. Given that this function has to be maintained separately, though, I have some suggestions for simplification/cleanup:

- Rather than use a mix of named and literal return values, maybe pick one approach or the other and stick with it?
- Maybe collect sets of good and bad revisions directly rather than returning pods? The original implementation returned pods because it needed both revisions and node names but that's not a requirement here.
- Maybe treat both sets of revisions being empty as an implicit indication of 'progressing' and skip the explicit return value?

test/library/library.go
```diff
@@ -32,3 +47,117 @@ func GenerateNameForTest(t *testing.T, prefix string) string {
...
+	}
+
+	goodRevisions := revisions(revisionLabel, good)
+	failingRevisions := revisions(revisionLabel, bad)
+
+	if len(goodRevisions) != 1 {
+		return false, nil // api servers have not converged onto a single revision
+	}
+	revision, _ := goodRevisions.PopAny()
+	if len(revision) == 0 {
```
marun, line 94: Will this actually be an empty string or is it a legacy of the implementation being copied?

test/library/library.go
```diff
@@ -32,3 +47,117 @@ func GenerateNameForTest(t *testing.T, prefix string) string {
...
+		case corev1.PodRunning:
+			if !podReady(apiServerPod) {
+				return nil, nil, true, nil // pods are not fully ready
+			}
+			goodPod := apiServerPod // shallow copy because apiServerPod is bound loop var
+			good = append(good, &goodPod)
+		case corev1.PodPending:
+			return nil, nil, true, nil // pods are not fully ready
+		case corev1.PodUnknown:
+			return nil, nil, false, fmt.Errorf("api server pod %s in unknown phase", apiServerPod.Name)
```
marun, line 128: Is this condition unrecoverable?

test/library/library.go
```diff
@@ -32,3 +47,117 @@ func GenerateNameForTest(t *testing.T, prefix string) string {
...
+		case corev1.PodSucceeded, corev1.PodFailed:
+			// handle failed pods carefully to make sure things are healthy
+			// since the API server should never exit, a succeeded pod is considered as failed
+			badPod := apiServerPod // shallow copy because apiServerPod is bound loop var
+			bad = append(bad, &badPod)
+		default:
+			// error in case new unexpected phases get added
+			return nil, nil, false, fmt.Errorf("api server pod %s has unexpected phase %v", apiServerPod.Name, phase)
```
marun, line 136: Is this condition unrecoverable?

test/library/library.go
```diff
@@ -32,3 +47,117 @@ func GenerateNameForTest(t *testing.T, prefix string) string {
...
+
+// WaitForAPIServerToBeOnTheSameRevision waits until all API Servers are running at the same revision.
+// The API Servers must stay on the same revision for at least waitForAPIRevisionSuccessThreshold * waitForAPIRevisionSuccessInterval
+//
+// Observations:
+//  rolling out a new version is not instant you need to account for a propagation time (~1/2 minutes)
+//  rolling out a new version can take ~10 minutes
+func WaitForAPIServerToBeOnTheSameRevision(t *testing.T, podClient corev1client.PodInterface) {
+	if err := wait.Poll(waitForAPIRevisionPollInterval, waitForAPIRevisionTimeout, mustSucceedMultipleTimes(waitForAPIRevisionSuccessThreshold, waitForAPIRevisionSuccessInterval, func() (bool, error) {
+		converged, err := areAPIServersOnTheSameRevision(podClient)
```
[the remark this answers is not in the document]
p0lyn0mial: yes

test/library/library.go
p0lyn0mial: actually why not - sgtm

test/library/library.go
p0lyn0mial: I don't think so. In a case like that, I would rather return to fail early and let a human fix the issue - possibly by adding a new condition.

test/library/library.go
p0lyn0mial: I don't think so.

test/library/library.go
p0lyn0mial: oh, I think you are right - it was a copy/paste. I'll remove it.

test/library/library.go
p0lyn0mial: I like all your suggestion except for the last one. I think we might have a mix of pods in `PodRunning` and `PodPending` state which indicate that we are progressing.

test/library/library.go
p0lyn0mial: updated the comment.

test/library/library.go
```diff
@@ -32,3 +47,105 @@ func GenerateNameForTest(t *testing.T, prefix string) string {
...
+	if progressing {
+		return false, nil
+	}
+
+	if len(goodRevisions) != 1 {
+		return false, nil // api servers have not converged onto a single revision
+	}
+	revision, _ := goodRevisions.PopAny()
+	if failingRevisions.Has(revision) {
+		return false, fmt.Errorf("api server revision %s has both running and failed pods", revision)
```
marun, line 94: Is this error potentially recoverable?

test/library/library.go
p0lyn0mial: > if err is nil and both good and bad revision sets are nil -> keep trying

Mhm, I don't see when `getRevisions` could return the above . . .

test/e2e/bound_sa_token_test.go
```diff
@@ -59,34 +55,38 @@ func TestBoundTokenSignerController(t *testing.T) {
...
-		test.WaitForKubeAPIServerClusterOperatorAvailableNotProgressingNotDegraded(t, configClient)
-
 		err := kubeClient.ConfigMaps(targetNamespace).Delete(context.TODO(), tokenctl.PublicKeyConfigMapName, metav1.DeleteOptions{})
 		require.NoError(t, err)
 		checkCertConfigMap(t, kubeClient, map[string]string{
 			"service-account-001.pub": string(operatorPublicKey),
 		})
+
+		// deletion triggers a roll-out - wait until a new version has been rolled out
+		test.WaitForAPIServerToBeOnTheSameRevision(t, kubeClient.Pods(operatorclient.TargetNamespace))
```
sttts, line 74: don't we have to wait first that the operator has started progressing?

test/library/library.go
```diff
@@ -32,3 +47,105 @@ func GenerateNameForTest(t *testing.T, prefix string) string {
...
 }
+
+// WaitForAPIServerToBeOnTheSameRevision waits until all API Servers are running at the same revision.
+// The API Servers must stay on the same revision for at least waitForAPIRevisionSuccessThreshold * waitForAPIRevisionSuccessInterval.
+// Mainly because of the difference between the propagation time of triggering a new release and the actual roll-out.
+//
+// Observations:
+//  rolling out a new version is not instant you need to account for a propagation time (~1/2 minutes)
+//  rolling out a new version can take ~10 minutes
+func WaitForAPIServerToBeOnTheSameRevision(t *testing.T, podClient corev1client.PodInterface) {
```
sttts, line 58: WaitForAPIServerTo*Stabilize*OnTheSameRevision ? This is not a no-op if there is no rollout started yet. So the current name is misleading. It waits at least 3 min, giving the operator time to deploy.

test/library/library.go
sttts: I don't think it is. Once we have set the last failed revision, it is set forever. It needs another revision to reach a good state.

test/library/library.go
```diff
@@ -32,3 +47,105 @@ func GenerateNameForTest(t *testing.T, prefix string) string {
...
+	}
+	revision, _ := goodRevisions.PopAny()
+	if failingRevisions.Has(revision) {
+		return false, fmt.Errorf("api server revision %s has both running and failed pods", revision)
+	}
+
+	return true, nil
+}
+
+func getRevisions(revisionLabel string, pods []corev1.Pod) (sets.String, sets.String, bool, error) {
```
sttts, line 100: why do we need a copy of library-go/pkg/operator/encryption/deployer/revisionedpod.go's `categorizePods`? Why don't we make it available at a central place?

test/library/library.go
p0lyn0mial: done

test/e2e/bound_sa_token_test.go
p0lyn0mial: nope, see - https://github.com/openshift/cluster-kube-apiserver-operator/pull/870#discussion_r435882090

test/library/library.go
sttts: follow-up

p0lyn0mial: ok - I'll open a new PR.

test/library/library.go
marun: This is exactly the case where `progressing=true` is returned. i.e this function is returning a boolean that could be inferred.

PR #875, opened 2020-06-04
Bug 1843752: static pod: don't wait for 6080 in apiserver container

sttts:
6080 is occupied by readyz container which does not free the port when the main container restarts (obviously).

Contrary, we probably don't have to wait for 6080 to free up because the readyz container terminate immediately on SIGTERM. Hence, the port has like 70s+ to become available which is more than enough, plus the init container already waits for the port.


sttts:
cc @squeed

sttts:
Blocked by https://github.com/openshift/cluster-kube-apiserver-operator/pull/870.

PR #876, opened 2020-06-04
Log non-graceful termination to /var/log/kube-apiserver/termination.log and stdout

sttts:
This 
- adds a /var/log/kube-apiserver/termination.log file to masters (sitting next to the audit logs we already have). It lists non-graceful terminations for easier CI and customer cluster analysis.
- lets the watch-termination binary create `NonGracefulTermination` events in the openshift-kube-apiserver namespace on next launch.

Container logs alone are not enough because logging during termination is broken, and we lose logs from old pods. So we have to persist the data somewhere on disk.

Etcd is also no option as etcd struggles with the same termination issues and is not reliable.

Depends on https://github.com/openshift/origin/pull/25192.

### DIFF
--- a/bindata/v4.1.0/kube-apiserver/pod.yaml
+++ b/bindata/v4.1.0/kube-apiserver/pod.yaml
@@ -54,7 +54,7 @@ spec:
             fi
           done
           echo
-          exec hyperkube kube-apiserver --openshift-config=/etc/kubernetes/static-pod-resources/configmaps/config/config.yaml --advertise-address=${HOST_IP}
+          exec watch-termination --termination-touch-file=/var/log/kube-apiserver/.terminating --termination-log-file=/var/log/kube-apiserver/termination.log --kubeconfig=/etc/kubernetes/static-pod-resources/configmaps/kube-apiserver-cert-syncer-kubeconfig/kubeconfig -- hyperkube kube-apiserver --openshift-config=/etc/kubernetes/static-pod-resources/configmaps/config/config.yaml --advertise-address=${HOST_IP} ${VERBOSITY}
     resources:
       requests:
         memory: 1Gi
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -256,24 +256,24 @@ func managePod(client coreclientv1.ConfigMapsGetter, recorder events.Recorder, o
 	if argsCount := len(containerArgsWithLoglevel); argsCount > 1 {
 		return nil, false, fmt.Errorf("expected only one container argument, got %d", argsCount)
 	}
-	if !strings.Contains(containerArgsWithLoglevel[0], "exec hyperkube kube-apiserver") {
-		return nil, false, fmt.Errorf("exec hyperkube kube-apiserver not found in first argument %q", containerArgsWithLoglevel[0])
+	if !strings.Contains(containerArgsWithLoglevel[0], "hyperkube kube-apiserver") {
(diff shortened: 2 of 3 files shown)


bindata/v4.1.0/kube-apiserver/pod.yaml
```diff
@@ -54,7 +62,16 @@ spec:
             fi
           done
           echo
-          exec hyperkube kube-apiserver --openshift-config=/etc/kubernetes/static-pod-resources/configmaps/config/config.yaml --advertise-address=${HOST_IP}
+          trap 'sig=$(($? - 128)); echo "Received signal $(kill -l ${sig})"; touch "${TERM_FILE}"; kill -TERM ${PID}' TERM INT
+          hyperkube kube-apiserver --openshift-config=/etc/kubernetes/static-pod-resources/configmaps/config/config.yaml --advertise-address=${HOST_IP} &
+          PID=$!
+          wait $PID
+          trap - TERM INT
+          wait $PID
```
sttts, line 70: following http://veithen.io/2014/11/16/sigterm-propagation.html here

bindata/v4.1.0/kube-apiserver/pod.yaml
```diff
@@ -38,6 +38,11 @@ spec:
     command: ["/bin/bash", "-ec"]
```
tnozicka, line 38: "-euxo" , "pipefail", "-c"?

bindata/v4.1.0/kube-apiserver/pod.yaml
```diff
@@ -54,7 +59,16 @@ spec:
             fi
           done
           echo
-          exec hyperkube kube-apiserver --openshift-config=/etc/kubernetes/static-pod-resources/configmaps/config/config.yaml --advertise-address=${HOST_IP}
+          trap 'sig=$(($? - 128)); echo "Received signal $(kill -l ${sig})"; touch "${TERM_FILE}"; kill -TERM ${PID}' TERM INT
```
tnozicka, line 62: do we want to make it numeric with list from 1..64 to make sure we aren't loosing signals? not sure if there is easier way in bash to forward all

bindata/v4.1.0/kube-apiserver/pod.yaml
sttts: it's preexisting. Don't want to change more than necessary.

sttts:
Flakes.

/retest

sttts:
> I really don't like nested signal termination in this case either - why wouldn't kube-apiserver summarize failure better on exit to logs? That's why brian and I added fallbacktologsonerror - so that our infra components could log better failures on exit?

Because we don't have logs. Logging of termination is broken in kubelet since 4.1. Our BZ is open since 4.1 too. Am happy to remove this again as soon as

a) kubelet and cri-o start working and doing their job
b) we have termination logs e.g. through loki.

We are spending half of our time hunting issues because we are blind. Waiting is no option.

sttts:
Dependency merged.

/retest

sttts:
ci-operator flake

/retest

p0lyn0mial:
/hold

failing `ci/prow/e2e-aws` is real, it looks like the must-gather test doesn't know how to handle `terminating.gz` files

`
STEP: /tmp/test.oc-adm-must-gather.044028853/registry-svc-ci-openshift-org-ci-op-6nby4y5y-stable-sha256-90e78122d7d240f2b1a0eaa05507303205292cd745fbfa5b36298942dab81fe4/audit_logs/kube-apiserver/ip-10-0-163-93.us-west-2.compute.internal-.terminating.gz`

sttts:
https://github.com/openshift/origin/pull/25282 merged.

/retest

PR #881, opened 2020-06-08
audit when `/healthz` is called from anything else than kubelet

p0lyn0mial:
this PR allow us to see what else is using `/healhtz` endpoint in the system. We think that only kubelet should actually use it. 

Hopeful it will help us to prevent issues like  https://bugzilla.redhat.com/show_bug.cgi?id=1844387

### DIFF
--- a/bindata/v4.1.0/config/defaultconfig.yaml
+++ b/bindata/v4.1.0/config/defaultconfig.yaml
@@ -61,13 +61,16 @@ auditConfig:
       resources:
       - group: "oauth.openshift.io"
         resources: ["oauthaccesstokens", "oauthauthorizetokens"]
-    # Don't log authenticated requests to certain non-resource URL paths.
+    # Don't log requests to /healthz from kubelets
     - level: None
-      userGroups: ["system:authenticated", "system:unauthenticated"]
+      userGroups: ["system:nodes"]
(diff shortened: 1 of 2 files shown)


bindata/v4.1.0/config/defaultconfig.yaml
```diff
@@ -61,13 +61,16 @@ auditConfig:
...
+    # Don't log requests to /healthz from kubelets
+    - level: None
+      userGroups: ["system:nodes"]
+      nonResourceURLs:
+        - "/healthz"
+    # Don't log requests to certain non-resource URL paths.
     - level: None
-      userGroups: ["system:authenticated", "system:unauthenticated"]
       nonResourceURLs:
       - "/api*" # Wildcard matching.
```
p0lyn0mial, line 72: how it is possible that it doesn't match `/apis/group/version`?

sttts: because of nonResourceURLs. The actual APIs are resource URLs.

p0lyn0mial:
it turned out that `"userAgent": "kube-probe/1.18+"` is actually kubelet. I'm closing this PR in favor of https://github.com/openshift/origin/pull/25076

PR #887, opened 2020-06-17
Wire event watcher for LateConnections events

mfojtik:
This wire event watcher for `LateConnections` event that increase the `openshift_kube_apiserver_lateconnections_count` prometheus metric.

A follow up would be to add Prometheus Alert.

Blocked on https://github.com/openshift/origin/pull/25126
/hold

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -24,6 +24,8 @@ import (
 	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/targetconfigcontroller"
 	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/terminationobserver"
 	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/v410_00_assets"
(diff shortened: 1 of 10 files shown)


mfojtik:
(we don't have to wait for origin, this will be no-op until that PR merge and this can get some testing cycles meanwhile ;-)

pkg/operator/starter.go
```diff
@@ -104,6 +106,10 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		controllerContext.EventRecorder,
 	)
 
+	eventWatcher := eventwatch.New().
+		WithEventHandler(operatorclient.TargetNamespace, "LateConnections", terminationobserver.ProcessLateConnectionEvents).
```
soltysh, line 110: @mfojtik who's producing this event, asking for a friend :wink:

mfojtik: https://github.com/openshift/origin/pull/25126

PR #890, opened 2020-06-22
add an observer for webhookTokenAuthenticator

stlaz:
Adds an observer that's supposed to observe the `authentication.Spec.WebhookTokenAuthenticator` field and sync a potential referenced secret to `openshift-kube-apiserver` NS


pkg/operator/configobservation/auth/webhook_authenticator.go
```diff
@@ -0,0 +1,212 @@
...
+		configobserver.Pruned(ret, webhookTokenAuthenticatorPath)
+	}()
+
+	listers := genericListers.(configobservation.Listers)
+	resourceSyncer := genericListers.ResourceSyncer()
+
+	errs := []error{}
+	existingWebhookAuthenticator, _, err := unstructured.NestedSlice(existingConfig, webhookTokenAuthenticatorPath...)
+	if err != nil {
+		errs = append(errs, err)
```
sttts, line 38: we had this "keep going" comment elsewhere I think. This is intentional and we should not return or anything like that. So better write a comment like that.

pkg/operator/configobservation/auth/webhook_authenticator.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	existingWebhookConfigured := len(existingWebhookAuthenticator) > 0
+
+	observedConfig := map[string]interface{}{}
+
+	auth, err := listers.AuthConfigLister.Get("cluster")
+	if errors.IsNotFound(err) {
+		return observedConfig, nil
+	} else if err != nil {
+		return existingConfig, []error{err}
```
sttts, line 48: please compare with the existing observers and https://github.com/openshift/cluster-kube-apiserver-operator/pull/803. I believe we return `append(errs, err)` elsewhere.

pkg/operator/configobservation/auth/webhook_authenticator.go
```diff
@@ -0,0 +1,212 @@
...
+	if auth.Spec.WebhookTokenAuthenticator != nil {
+		webhookSecretName = auth.Spec.WebhookTokenAuthenticator.KubeConfig.Name
+	}
+
+	observedWebhookConfigured := len(webhookSecretName) > 0
+	if observedWebhookConfigured {
+		// retrieve the secret from config and validate it, don't proceed on failure
+		kubeconfigSecret, err := listers.ConfigSecretLister().Secrets("openshift-config").Get(webhookSecretName)
+		if err != nil {
+			return existingConfig, []error{fmt.Errorf("secret openshift-config/%s: %w", webhookSecretName, err)}
```
sttts, line 61: same here, reuse `errs` with append

pkg/operator/configobservation/auth/webhook_authenticator.go
```diff
@@ -0,0 +1,212 @@
...
+		}
+
+		if secretErrors := validateKubeconfigSecret(kubeconfigSecret); len(secretErrors) > 0 {
+			return existingConfig, append(errs,
+				fmt.Errorf("secret openshift-config/%s: %w", webhookSecretName, utilerrors.NewAggregate(secretErrors)))
```
sttts, line 66: `secret openshift-config/%s` is not a real error message. Am not sure that appending on aggregate error will look good at the end.

pkg/operator/configobservation/auth/webhook_authenticator.go
```diff
@@ -0,0 +1,212 @@
...
+		}
+
+		if err := unstructured.SetNestedField(observedConfig, webhookTokenAuthenticatorFile, webhookTokenAuthenticatorPath...); err != nil {
+			return existingConfig, append(errs, err)
+		}
+
+		resourceSyncer.SyncSecret(
+			resourcesynccontroller.ResourceLocation{Namespace: operatorclient.TargetNamespace, Name: "webhook-authenticator"},
+			resourcesynccontroller.ResourceLocation{Namespace: operatorclient.GlobalUserSpecifiedConfigNamespace, Name: webhookSecretName},
+		)
```
sttts, line 76: syncing from an observer is an unexpected side-effect. How do we do this elsewhere?

stlaz: This is being done in other observers as well:
https://github.com/openshift/cluster-kube-apiserver-operator/blob/f73bebb6361c3649dab5305d8c7d1cd9753e61aa/pkg/operator/configobservation/apiserver/observe_apiserver.go#L226
https://github.com/openshift/cluster-kube-apiserver-operator/blob/f73bebb6361c3649dab5305d8c7d1cd9753e61aa/pkg/operator/configobservation/auth/auth_metadata.go#L84-L93
We're handling the field that's a reference to a secret here, it makes sense to me to handle the sync as well.

pkg/operator/configobservation/auth/webhook_authenticator.go
stlaz: I turned it into an actual error message

pkg/operator/configobservation/auth/webhook_authenticator.go
sttts: This is backed by the resource sync controller, which then essential does this:
```
	c.configMapSyncRules[destination] = source
```

So this sync is ongoing, the call only registers a rule, does not do the a synchronous sync. And a new, different source overrides the old source.

So, I think this is fine.

pkg/operator/configobservation/auth/webhook_authenticator.go
```diff
@@ -0,0 +1,213 @@
...
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
+
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/configobservation"
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/operatorclient"
+)
+
+var (
+	webhookTokenAuthenticatorPath = []string{"apiServerArguments", "authentication-token-webhook-config-file"}
+	webhookTokenAuthenticatorFile = []interface{}{"/etc/kubernetes/static-pod-resources/secrets/webhook-authenticator/kubeConfig"}
```
mfojtik, line 24: is `kubeConfig` right? (usually it is `kubeconfig`)

pkg/operator/configobservation/auth/webhook_authenticator.go
```diff
@@ -0,0 +1,213 @@
...
+)
+
+func ObserveWebhookTokenAuthenticator(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, _ []error) {
```
mfojtik, line 27: godoc pls.

pkg/operator/configobservation/auth/webhook_authenticator.go
stlaz: as per API doc, this is expected

pkg/operator/configobservation/auth/webhook_authenticator.go
```diff
@@ -0,0 +1,213 @@
...
+		)
+	} else {
+		// don't sync anything and remove whatever we synced
+		resourceSyncer.SyncSecret(
+			resourcesynccontroller.ResourceLocation{Namespace: operatorclient.TargetNamespace, Name: "webhook-authenticator"},
+			resourcesynccontroller.ResourceLocation{Namespace: "", Name: ""},
+		)
+	}
+
+	if observedWebhookConfigured != existingWebhookConfigured {
```
mfojtik, line 86: we usually use reflect.DeepEqual()

mfojtik:
couple nits, LGTM otherwise

pkg/operator/configobservation/auth/webhook_authenticator.go
stlaz: this is a comparison of two booleans

PR #894, opened 2020-06-30
Support configuring ServicesNodePortRange via network config

abhat:
This PR adds support for configuring the servicesNodePortRange of the KubeAPIServerConfig via the network config API. This is necessary to handle cases where the default range (30000-32767) is not sufficient for users who have more services in their cluster than the range can handle.

### DIFF
--- a/pkg/operator/configobservation/network/observe_network.go
+++ b/pkg/operator/configobservation/network/observe_network.go
@@ -190,6 +190,42 @@ func ObserveExternalIPPolicy(genericListers configobserver.Listers, recorder eve
 	return observedConfig, errs
 }
 
+// ObserveServiceNodePortRange watches the network configuration and generates the
+// serviceNodePortRange
+func ObserveServicesNodePortRange(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
+	listers := genericListers.(configobservation.Listers)
+
+	out := map[string]interface{}{}
+	servicesNodePortRangeConfigPath := []string{"apiServerArguments", "service-node-port-range"}
+
+	previouslyObservedConfig, errs := extractPreviouslyObservedConfig(existingConfig, servicesNodePortRangeConfigPath)
+
+	serviceNodePortRange, err := network.GetServiceNodePortRange(listers.NetworkLister, recorder)
+	if err != nil {
+		errs = append(errs, err)
+		return previouslyObservedConfig, errs
+	}
+	// We want to make sure that a range user sets doesn't shrink. This
+	// is handled by a custom resource validator in openshift-api-server code.
+	// If the user setting for this field goes from
+	//      1) not being set to
+	//      2) set to 30000-36000 to
+	//      3) not being set
+	// The third change should fail customresourcevalidation. So once the field
+	// is set in the yaml, it should never come here for anything other than the
+	// default case of never being set.
+
+	if serviceNodePortRange == "" {
+		return map[string]interface{}{}, errs
+	}
+	snpr := []string{serviceNodePortRange}
+	if err := unstructured.SetNestedStringSlice(out, snpr, servicesNodePortRangeConfigPath...); err != nil {
+		errs = append(errs, err)
+		return previouslyObservedConfig, errs
+	}
+	return out, errs
+}
+
 // extractPreviouslyObservedConfig extracts the previously observed config from the existing config.
 func extractPreviouslyObservedConfig(existing map[string]interface{}, paths ...[]string) (map[string]interface{}, []error) {
 	var errs []error
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/openshift/api v0.0.0-20200714125145-93040c6967eb
 	github.com/openshift/build-machinery-go v0.0.0-20200713135615-1f43d26dccc7
 	github.com/openshift/client-go v0.0.0-20200608144219-584632b8fc73
-	github.com/openshift/library-go v0.0.0-20200617100204-823c70af3f78
+	github.com/openshift/library-go v0.0.0-20200630145007-34ebc8778b33
 	github.com/prometheus/client_golang v1.1.0
 	github.com/spf13/cobra v0.0.5
 	github.com/spf13/pflag v1.0.5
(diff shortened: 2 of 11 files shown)


pkg/operator/configobservation/network/observe_network.go
```diff
@@ -190,6 +190,35 @@ func ObserveExternalIPPolicy(genericListers configobserver.Listers, recorder eve
...
+func ObserveServicesNodePortRange(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
+	listers := genericListers.(configobservation.Listers)
+
+	out := map[string]interface{}{}
+	servicesNodePortRangeConfigPath := []string{"apiServerArguments", "service-node-port-range"}
+
+	previouslyObservedConfig, errs := extractPreviouslyObservedConfig(existingConfig, servicesNodePortRangeConfigPath)
+
+	serviceNodePortRange, err := network.GetServiceNodePortRange(listers.NetworkLister, recorder)
+
```
sttts, line 204: no empty line

pkg/operator/configobservation/network/observe_network.go
```diff
@@ -190,6 +190,35 @@ func ObserveExternalIPPolicy(genericListers configobserver.Listers, recorder eve
...
+	if err != nil {
+		errs = append(errs, err)
+		return previouslyObservedConfig, errs
+	}
+
+	if serviceNodePortRange == "" {
+		return previouslyObservedConfig, errs
```
sttts, line 211: do all old cluster have the range set?

I would also expect that the value is switched back to the default range if there is nothing specified in the config. This way here the old custom range would be kept.

pkg/operator/configobservation/network/observe_network.go
abhat: ack

pkg/operator/configobservation/network/observe_network.go
abhat: all old clusters wont have anything set in the range, the `previouslyObservedConfig` will be an empty map. If the new `serviceNodePortRange` is empty string, do we not want to keep the previously configured and instead reset it to default range?

sttts: There usually is either a real API default (defined via `// kubebuilder:default=....` and automatically applied of the field is not set) or some implicit default (the one we had before the field was introduced). I would expect that an empty string would lead to this default, i.e. to a change in configuration if a different range had been set before.

abhat: I am unsure what the cluster-kubeapiserver-operator will do in case we actually return an "" in such a case. Does it smartly merge the config from the [default config](https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/bindata/v4.1.0/config/defaultconfig.yaml) and the observed config every time?

sttts: I remerges everything, i.e. calls every observer and merges the results into a new config based on defaultconfig.yaml as base. I.e. if you return `map[string]interface{}{}` here, there won't be a range set (or the one from defaultconfig.yaml; haven't check whether there is a value) and the default will be used.

sttts: Check the default config. You would get `https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/bindata/v4.1.0/config/defaultconfig.yaml#L148` by returning `map[string]interface{}{}` here.

abhat: I am thinking about this now in context of operational impact. @knobunc and I talked about allowing the range to not shrink so that existing services don't get affected. In that sense, if we set the service node port range to "", and user has previously set it to a range that is > than the default, some of the services may get affected.

sttts: Why is `""` special. I could put a smaller range there and might get the same problem. If you cannot allow that, add a check to https://github.com/openshift/kubernetes/tree/master/openshift-kube-apiserver/admission/customresourcevalidation. I don't think that ignoring the user intent for `""` is the right thing to do.

abhat: I uploaded a version that validates and prevents the shrinking of the range. Also, I think the [API](https://github.com/abhat/api/blob/8a9a8bed32537335a3bdb17658a5cc3344b2358a/config/v1/0000_10_config-operator_01_network.crd.yaml#L116) validates that the user cannot just set it to empty string. Does this suffice?

sttts: Shrinking validation belongs into the place I linked above.

sttts: > Also, I think the API validates that the user cannot just set it to empty string

It doesn't. It's an optional field, even marked omitempty. If you want to enforce that it is set, make it required.

abhat: Thanks, I will move the validation to `customresourcevalidation` location you pointed me to above. As for the user explicitly setting the field to an empty string, I tried this out and the CRD validation does kick in (screen-shot below) and prevents user from setting it. Omitting the field completely is obviously allowed as well. And that just will retain the last known good value.

CR field validation error

sttts: Not to an empty string. Just leave the field out, i.e. remove the field from the JSON. Then https://github.com/openshift/library-go/blob/34ebc8778b33c1b2ddefd5477444e170770942d6/pkg/operator/configobserver/network/observe_network.go#L143 returns the empty string.

abhat: With the customresourcevalidation preventing the range from shrinking, what should the expected behavior be if someone takes the field out completely? Is it ok to retain the last known good value?

sttts: The customresourcevalidation has to forbid it. Then we don't care much what we do here as that code is never run.

abhat: Do you mean forbid removing the field completely from the spec once set?

sttts: My suggestion here was to return empty config for empty or missing field. The case that the range shrinks never happens. So we don't have to care about it here.

abhat: Yes. agreed.

PR #896, opened 2020-07-03
check-endpoints: missing openshift apiserver service endpoints endpoints

sanchezl:
* `api` endpoints in the openshift-apiserver namespace were not being picked up correctly.
* ipv6 endpoint addresses incorrectly formatted.
* handle storage urls when specified as kube-apiserver args

### DIFF
--- a/pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
+++ b/pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
@@ -2,9 +2,10 @@ package connectivitycheckcontroller
 
 import (
 	"context"
-	"fmt"
+	"net"
 	"net/url"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/ghodss/yaml"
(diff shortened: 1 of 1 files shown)


pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
```diff
@@ -145,10 +146,12 @@ func listAddressesForOpenShiftAPIServerService(serviceLister corev1listers.Servi
...
 	}
-	if len(service.Spec.Ports) == 0 {
-		return []string{fmt.Sprintf("%s:443", service.Spec.ClusterIP)}
+	for _, port := range service.Spec.Ports {
+		if port.TargetPort.IntValue() == 6443 {
+			return []string{net.JoinHostPort(service.Spec.ClusterIP, strconv.Itoa(int(port.Port)))}
+		}
 	}
-	return []string{fmt.Sprintf("%s:%d", service.Spec.ClusterIP, service.Spec.Ports[0].Port)}
+	return []string{net.JoinHostPort(service.Spec.ClusterIP, "443")}
```
deads2k, line 154: will this ever actually happen?  Don't we know it will always be 6443?

deads2k:
heh, oops

/lgtm

pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
sanchezl: We know the `.TargetPort` will be 6443. And the `.Port` will be 443, I do my best to figure it out "correctly", but worst case just return what it should be anyway.

PR #900, opened 2020-07-13
[release-4.2] Bug 1799079: bump(library-go): Revision/Pruner pod resource requests

damemi:
This bumps library-go to ensure that revision and pruner pods set resource
requests and limits.


damemi:
/close
4.2 drops support today, we don't need to backport this and the associated BZ has closed

PR #901, opened 2020-07-13
point-to-point network check: use tls dialer

sanchezl:
Using the check-endpoints tool results in log spam when the endpoint uses tls. 

The golang tls implementation cannot handle a clean close (tcp close notify) without first completing the tls handshake. Simply using the tls dialer (set to not verify server certs) over comes this issue.

If the server is also set to require a client cert, a client cert must be provided. This can't be just any self-signed cert (even if no verification will be performed) since the server will indicate which signers it supports and a well behaved tls client will not send a client cert that does not have a chain leading to a supported signer.


sttts:
We want new tcp connections for each ping, don't we? How expensive are those if we use tls? My gut feeling is that 10ish per second don't harm. But be aware that tls costs.

pkg/cmd/checkendpoints/controller/connection_checker.go
```diff
@@ -25,18 +28,21 @@ type ConnectionChecker interface {
...
+		recorder:    recorder,
+		stop:        make(chan interface{}),
 	}
+	return c
 }
 
 type connectionChecker struct {
 	check       *operatorcontrolplanev1alpha1.PodNetworkConnectivityCheck
 	client      v1alpha1helpers.PodNetworkConnectivityCheckClient
+	clientCerts []tls.Certificate
```
deads2k, line 45: make this a function that produces tls.Certificate.  We'll probably drive it based on a lister.

pkg/cmd/checkendpoints/controller/connection_checker.go
```diff
@@ -122,18 +128,95 @@ func (c *connectionChecker) checkEndpoint(ctx context.Context, check *operatorco
...
+		})
+		defer timer.Stop()
+	}
+
+	rawConn, err := dialer.DialContext(ctx, network, addr)
+	if err != nil {
+		return nil, err
+	}
+
+	colonPos := strings.LastIndex(addr, ":")
```
deads2k, line 182: net.SplitHostPort

pkg/cmd/checkendpoints/controller/connection_checker.go
```diff
@@ -122,18 +128,95 @@ func (c *connectionChecker) checkEndpoint(ctx context.Context, check *operatorco
...
+	conn := tls.Client(rawConn, config)
+
+	if timeout == 0 {
+		err = conn.Handshake()
+	} else {
+		go func() {
+			errChannel <- conn.Handshake()
+		}()
+
+		err = <-errChannel
```
deads2k, line 209: comment to clarify the two possible ways we exit.

pkg/cmd/checkendpoints/controller/connection_checker.go
```diff
@@ -122,18 +128,95 @@ func (c *connectionChecker) checkEndpoint(ctx context.Context, check *operatorco
...
+	if !dialer.Deadline.IsZero() {
+		deadlineTimeout := time.Until(dialer.Deadline)
+		if timeout == 0 || deadlineTimeout < timeout {
+			timeout = deadlineTimeout
+		}
+	}
+
+	var errChannel chan error
+
+	if timeout != 0 {
```
deads2k, line 169: why should our code allow no timeout. I prefer forcing a default timeout if none is specified.

pkg/cmd/checkendpoints/controller/pod_network_connectivity_check_controller.go
```diff
@@ -97,6 +99,29 @@ func (c *controller) Sync(ctx context.Context, syncContext factory.SyncContext)
 	return nil
 }
 
+// getClientCerts returns the client cert specified in the secret specified in the PodNetworkConnectivityCheck
+// or nil if not specified or there is an error retrieving the certs.
+func (c *controller) getClientCerts(check *operatorcontrolplanev1alpha1.PodNetworkConnectivityCheck) []tls.Certificate {
+	// TODO hack until we can specify client certs in the PodNetworkConnectivityCheck spec
```
deads2k, line 105: wow, this wins a prize for ugly

pkg/cmd/checkendpoints/controller/pod_network_connectivity_check_controller.go
```diff
@@ -97,6 +99,29 @@ func (c *controller) Sync(ctx context.Context, syncContext factory.SyncContext)
...
+		cert, err := tls.LoadX509KeyPair(certFile, keyFile)
+		if err != nil {
+			return nil
+		}
+		return []tls.Certificate{cert}
+	}
+	return nil
+}
+
+func getEtcdClientCertsFromMountedStaticPodResourcesVolume() tls.Certificate {
```
deads2k, line 118: dead?

pkg/cmd/checkendpoints/controller/connection_checker.go
deads2k: > why should our code allow no timeout. I prefer forcing a default timeout if none is specified.

we should prevent this. Always default the timeout to something so we don't need this if block.

pkg/cmd/checkendpoints/controller/pod_network_connectivity_check_controller.go
```diff
@@ -97,6 +107,27 @@ func (c *controller) Sync(ctx context.Context, syncContext factory.SyncContext)
 	return nil
 }
 
+// getClientCerts returns the client cert specified in the secret specified in the PodNetworkConnectivityCheck
+// or nil if not specified or there is an error retrieving the certs.
+func (c *controller) getClientCerts(check *operatorcontrolplanev1alpha1.PodNetworkConnectivityCheck) CertificatesGetter {
+	return func() []tls.Certificate {
```
deads2k, line 113: my advice about this signature was bad, wasn't it?  Should have been `func (check *operatorcontrolplanev1alpha1.PodNetworkConnectivityCheck) []tls.Certificate` and you could just directly reference eh?

Can be fixed up afterwards.

pkg/cmd/checkendpoints/controller/pod_network_connectivity_check_controller.go
```diff
@@ -97,6 +107,27 @@ func (c *controller) Sync(ctx context.Context, syncContext factory.SyncContext)
...
+		if len(check.Spec.TLSClientCert.Name) > 0 {
+			secret, err := c.secretLister.Secrets(c.podNamespace).Get(check.Spec.TLSClientCert.Name)
+			if err != nil {
+				klog.V(2).Infof("secret/%s: %v", check.Spec.TLSClientCert.Name, err)
+				return nil
+			}
+			cert, err := tls.X509KeyPair(secret.Data["tls.crt"], secret.Data["tls.key"])
+			if err != nil {
+				klog.V(2).Infof("error loading tls client key pair: %v", err)
```
deads2k, line 122: event worthy?  this shouldn't happen. Of course, writing events would be a big event write load.  

//TODO write a rate limited event every minute.

deads2k:
/lgtm

here's to less log spam

@sanchezl does this also fix the logspam in openshift-apiserver?

hexfusion:
etcd appreciates your efforts, thank you.

https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-kube-apiserver-operator/901/pull-ci-openshift-cluster-kube-apiserver-operator-master-e2e-aws/1283207912335872000/artifacts/e2e-aws/gather-extra/pods/openshift-etcd_etcd-ip-10-0-140-75.us-west-1.compute.internal_etcd.log

sanchezl:
> @sanchezl does this also fix the logspam in openshift-apiserver?

@deads2k yes, the tls handshake fixes openshift-apiserver, tls handshake + client cert fixes etcd.

ironcladlou:
/retitle Bug 1855284: point-to-point network check: use tls dialer

PR #905, opened 2020-07-17
kube-apiserver-cert-syncer-kubeconfig: point to serving cert

sttts:
The service account gets the service-account CA that is valid at the point of creation. So if the localhost-recovery CA is not yet part of the kube-apiserver serving CA (which is the case during bootstrapping), this service-account won't work with the localhost-recovery endpoints.

The serving CA bundle though is always correct as it changes with time.

Fixes https://github.com/openshift/cluster-kube-apiserver-operator/pull/663/commits/2a44f96452f248818ff4c2467d51eef50fd139f2#r456357777

Partly reverts https://github.com/openshift/cluster-kube-apiserver-operator/pull/663.

### DIFF
--- a/bindata/v4.1.0/kube-apiserver/kubeconfig-cm.yaml
+++ b/bindata/v4.1.0/kube-apiserver/kubeconfig-cm.yaml
@@ -8,7 +8,7 @@ data:
     apiVersion: v1
     clusters:
       - cluster:
-          certificate-authority: /etc/kubernetes/static-pod-resources/secrets/localhost-recovery-client-token/ca.crt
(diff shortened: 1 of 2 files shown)


sttts:
Flake.

/retest

sttts:
/hold

@p0lyn0mial noticed something that might be the root cause. Though my theory here stands.

sttts:
/hold cancel

Seems to be different.

soltysh:
Unrelated failure, and we'd like to see if this unblocks us
/override ci/prow/e2e-aws-serial

PR #858, opened 2020-05-14
Bug 1710766: Add kubeconfigs to masters

tnozicka:
When we ssh to a master there is currently no kubeconfig to use to talk to the apiserver. When we are debugging clusters and have only ssh access, building the kubeconfig manually and creating client certs is extremely painful and time consuming. Also when admin has an accident and looses system:admin kubeconfig, this can be used as temporary replacement. (He should still own the user client CA to sign new certs.) The kubeconfigs have intentionally inlined certificates for easy transfer.

The kubeconfigs being placed into `/etc/kubernetes/static-pod-resources/kube-apiserver-certs/secrets/node-kubeconfigs` are 
 - `localhost.kubeconfig`
 - `lb-ext.kubeconfig`
 - `lb-int.kubeconfig`
 - `localhost-recovery.kubeconfig`

Currently none of them is linked as default to oc. That would probably need MCO to copy it over and somehow handle updates otherwise the `oc` writing (e.g. current context) would fight with cert-syncer.

Requires:
- [x] manual testing
- [x] e2e making sure kubeconfigs work https://github.com/openshift/origin/pull/25453

/cc @deads2k @sttts 
@openshift/sig-master fyi, you'll need it at some point if you haven't already

PR #907, opened 2020-07-20
add /etc/kubernetes/static-pod-certs/configmaps/control-plane-node-ku…

deads2k:
…beconfig/kubeconfig

This is a system:masters identity which is in known location to make it easy to always have cluster-admin credentials.  Recall that if you can read the contents of this directory on disk, you already have etcd credentials, so this is not an escalation

once it is present, we can use the client-cert to wire up https://github.com/openshift/kubernetes/pull/271 to pass credentials along.

/assign @sanchezl


tnozicka:
there is already this one https://github.com/openshift/cluster-kube-apiserver-operator/pull/858

PR #911, opened 2020-07-21
add start/end logs to outage entries

sanchezl:
Add log entries that affect the state of an outage to the outage entry.


pkg/cmd/checkendpoints/controller/connection_checker.go
```diff
@@ -233,16 +233,30 @@ func manageStatusOutage(recorder events.Recorder) v1alpha1helpers.UpdateStatusFu
...
-		} else {
-			if latestSuccess.Start.After(latestFailure.Start.Time) {
-				currentOutage.End = latestSuccess.Start
-				recorder.Eventf("ConnectivityRestored", "Connectivity restored after %v: %s", currentOutage.End.Sub(currentOutage.Start.Time), latestSuccess.Message)
+			status.Outages = append([]operatorcontrolplanev1alpha1.OutageEntry{newOutage}, status.Outages...)
+			recorder.Warningf("ConnectivityOutageDetected", "Connectivity outage detected: %s", latestFailure.Message)
+		case currentOutage != nil && latestFailure.Start.After(latestSuccess.Start.Time):
+			// outage ongoing
+			if len(currentOutage.StartLogs) == 0 || currentOutage.StartLogs[0].Message != latestFailure.Message {
+				currentOutage.StartLogs = append([]operatorcontrolplanev1alpha1.LogEntry{latestFailure}, currentOutage.StartLogs...)
```
deads2k, line 249: this will produce infinite startlogs, right?  We want a cap.

Also, we want `EndLogs` to always have the last five-ish failures.

pkg/cmd/checkendpoints/controller/connection_checker.go
```diff
@@ -233,16 +233,30 @@ func manageStatusOutage(recorder events.Recorder) v1alpha1helpers.UpdateStatusFu
...
 			}
+		case currentOutage != nil && latestSuccess.Start.After(latestFailure.Start.Time):
+			// outage ended
+			currentOutage.End = latestSuccess.Start
+			outageDuration := currentOutage.End.Sub(currentOutage.Start.Time)
+			currentOutage.EndLogs = []operatorcontrolplanev1alpha1.LogEntry{latestSuccess}
```
deads2k, line 255: this should be an append.

deads2k:
I'm not really happy with the logic, but I need something in CI now. Please update the endlogs and startlogs.

1. StartLogs should have only five entries ever, after that, nothing more gets added
2. EndLogs should always have  at least one entry (the first failure  is the first entry). After that, it should have five entries max, but it should always be the most recent five.

/lgtm

pkg/cmd/checkendpoints/controller/connection_checker.go
sanchezl: Done.

pkg/cmd/checkendpoints/controller/connection_checker.go
sanchezl: Done.

deads2k:
/lgtm

probably want to factor the append logic out and keep using it.  I suspect we may want the lists sorted in opposite directions, but I guess we'll see.

PR #913, opened 2020-07-23
check-endpoints: use control-plane-node-kubeconfig

sanchezl:
Use https://github.com/openshift/cluster-kube-apiserver-operator/pull/907


deads2k:
before we ship, make one with narrower permissions.

/close

PR #914, opened 2020-07-23
allow the check-endpoints target and pod to mutate over time

deads2k:
This will allow us to have one podendpointconnectivity check for each kas-on-node to each oas-on-node instead of one for each oas that is ever created.  This will make it easier to inspect the results.

/assign @sanchezl

### DIFF
--- a/pkg/cmd/checkendpoints/controller/connection_checker.go
+++ b/pkg/cmd/checkendpoints/controller/connection_checker.go
@@ -25,10 +25,14 @@ type ConnectionChecker interface {
 	Stop()
 }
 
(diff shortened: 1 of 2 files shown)


deads2k:
> We'll probably want to [sync on PodNetworkConnectivityChecks](https://github.com/openshift/cluster-kube-apiserver-operator/blob/4443436c7814ab52a1cfacdb5db4abf35d5aa2a7/pkg/cmd/checkendpoints/controller/pod_network_connectivity_check_controller.go#L62)

yes will will.  Adding update.

PR #915, opened 2020-07-24
connectivitycheckcontroller: name endpoints using node names

sanchezl:
Create node "scoped" connectivity checks who's source pod and endpoints can change over time. Requires https://github.com/openshift/cluster-kube-apiserver-operator/pull/914.

### DIFF
--- a/pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
+++ b/pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
@@ -2,7 +2,7 @@ package connectivitycheckcontroller
 
 import (
 	"context"
-	"regexp"
+	"strings"
 
 	v1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
@@ -13,6 +13,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourcehelper"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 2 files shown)


pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
```diff
@@ -82,10 +83,20 @@ func (c *connectivityCheckController) Sync(ctx context.Context, syncContext fact
...
+				continue
+			}
+			updated := existing.DeepCopy()
+			updated.Spec = *check.Spec.DeepCopy()
+			_, err := pnccClient.Update(ctx, updated, metav1.UpdateOptions{})
+			if err != nil {
+				syncContext.Recorder().Warningf("EndpointDetectionFailure", "%s: %v", resourcehelper.FormatResourceForCLIWithNamespace(check), err)
+				continue
+			}
+			syncContext.Recorder().Eventf("EndpointCheckCreated", "Updated %s because it changed.", resourcehelper.FormatResourceForCLIWithNamespace(check))
```
deads2k, line 99: EndpointCheckUpdated

deads2k, line 99: these events are going to the wrong namespace.  We want them in the `TargetNamespace`

deads2k: > these events are going to the wrong namespace. We want them in the `TargetNamespace`

will take a bug to fix it instead of fixing it here

deads2k:
some comments, but this is a strict improvement

/lgtm

pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
sanchezl: > these events are going to the wrong namespace. We want them in the `TargetNamespace`

@deads2k  Are you sure? These are events from the operator, managing the `Spec` of the **PodNetworkConnectivityChecks** (similar to events creating other resources in target namespace).  Compared this to the events coming from the **check-endpoints** tool that runs in the target namespace and updates the `Status` of the **PodNetworkConnectivityChecks**.

sanchezl: @deads2k https://github.com/openshift/cluster-kube-apiserver-operator/pull/919

deads2k: it wasn't where I looked for them, but I guess I can live with it now that I know.

PR #917, opened 2020-07-24
check-endpoints: handle out of order results

sanchezl:
When a check results in a latency longer than the check period (1s), the result of a check is reported after the result of subsequent checks. This PR introduces a delay to give the long running checks a chance to be processed in the correct order.

Further enhancements in the followup PR: https://github.com/openshift/cluster-kube-apiserver-operator/pull/924

### DIFF
--- a/pkg/cmd/checkendpoints/controller/connection_checker.go
+++ b/pkg/cmd/checkendpoints/controller/connection_checker.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"net"
 	"regexp"
-	"sync"
 	"time"
 
 	operatorcontrolplanev1alpha1 "github.com/openshift/api/operatorcontrolplane/v1alpha1"
@@ -19,10 +18,15 @@ import (
 	"github.com/openshift/cluster-kube-apiserver-operator/pkg/cmd/checkendpoints/trace"
 )
 
+const (
+	checkPeriod  = 1 * time.Second
+	checkTimeout = 10 * time.Second
+)
+
 // ConnectionChecker checks a single connection and updates status when appropriate
 type ConnectionChecker interface {
 	Run(ctx context.Context)
-	Stop()
+	Stop(ctx context.Context)
 }
 
 type GetCheckFunc func() *operatorcontrolplanev1alpha1.PodNetworkConnectivityCheck
@@ -36,7 +40,15 @@ func NewConnectionChecker(name, podName string, getCheck GetCheckFunc, client v1
 		client:           client,
 		clientCertGetter: clientCertGetter,
 		recorder:         recorder,
+		updates:          NewUpdatesManager(checkTimeout+checkPeriod, newUpdatesProcessor(client, name)),
 		stop:             make(chan interface{}),
+	}
+}
+
+func newUpdatesProcessor(client v1alpha1helpers.PodNetworkConnectivityCheckClient, name string) UpdatesProcessor {
+	return func(ctx context.Context, updates ...v1alpha1helpers.UpdateStatusFunc) error {
+		_, _, err := v1alpha1helpers.UpdateStatus(ctx, client, name, updates...)
+		return err
 	}
 }
 
@@ -50,21 +62,13 @@ type connectionChecker struct {
 	client           v1alpha1helpers.PodNetworkConnectivityCheckClient
 	clientCertGetter CertificatesGetter
 	recorder         events.Recorder
-	updatesLock      sync.Mutex
-	updates          []v1alpha1helpers.UpdateStatusFunc
+	updates          UpdatesManager
 	stop             chan interface{}
 }
 
-// add queues status updates in a queue.
-func (c *connectionChecker) add(updates ...v1alpha1helpers.UpdateStatusFunc) {
-	c.updatesLock.Lock()
-	defer c.updatesLock.Unlock()
-	c.updates = append(c.updates, updates...)
-}
-
-// checkConnection checks the connection every second, updating status as needed
+// checkConnection checks the connection periodically, updating status as needed
 func (c *connectionChecker) checkConnection(ctx context.Context) {
-	ticker := time.NewTicker(1 * time.Second)
+	ticker := time.NewTicker(checkPeriod)
 	defer ticker.Stop()
 	defer klog.V(1).Infof("Stopped connectivity check %s.", c.name)
 	for {
@@ -79,11 +83,11 @@ func (c *connectionChecker) checkConnection(ctx context.Context) {
 				currCheck := c.getCheck()
 				// if we have no check or the check isn't for us or the check has no target, report status if needed, but nothing else
 				if currCheck == nil || currCheck.Spec.SourcePod != c.podName || len(currCheck.Spec.TargetEndpoint) == 0 {
-					c.updateStatus(ctx)
+					c.updateStatus(ctx, false)
 					return
 				}
 				c.checkEndpoint(ctx, currCheck)
-				c.updateStatus(ctx)
+				c.updateStatus(ctx, false)
 			}()
 		}
 	}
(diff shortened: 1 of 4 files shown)


pkg/cmd/checkendpoints/controller/connection_checker.go
```diff
@@ -186,7 +185,7 @@ func isDNSError(err error) bool {
 
 // manageStatusLogs returns a status update function that updates the PodNetworkConnectivityCheck.Status's
 // Successes/Failures logs reflect the results of the check.
-func manageStatusLogs(check *operatorcontrolplanev1alpha1.PodNetworkConnectivityCheck, checkErr error, latency *trace.LatencyInfo) []v1alpha1helpers.UpdateStatusFunc {
+func manageStatusLogs(check *operatorcontrolplanev1alpha1.PodNetworkConnectivityCheck, checkErr error, latency *trace.LatencyInfo) ([]v1alpha1helpers.UpdateStatusFunc, time.Time) {
```
deads2k, line 188: update doc. This time is the time the check started?

pkg/cmd/checkendpoints/controller/updates_manager.go
```diff
@@ -0,0 +1,75 @@
...
+	return &updatesManager{
+		window:       window,
+		sortingQueue: map[time.Time][]v1alpha1helpers.UpdateStatusFunc{},
+	}
+}
+
+// UpdatesManager manages a queue of updates. The lock must be obtained before
+// invoking any of the methods.
+type UpdatesManager interface {
+	sync.Locker
```
deads2k, line 22: no direct embedding.

pkg/cmd/checkendpoints/controller/updates_manager.go
```diff
@@ -0,0 +1,75 @@
...
+	window          time.Duration
+	sortingQueue    map[time.Time][]v1alpha1helpers.UpdateStatusFunc
+	processingQueue []v1alpha1helpers.UpdateStatusFunc
+	timestamps      []time.Time
+}
+
+// Add an update to the queue. There is a delay equal to the size of the sorting window before
+// updates are made available on the queue to allow for updates submitted out of order within
+// the sorting window to be sorted by timestamp.
+func (u *updatesManager) Add(timestamp time.Time, updates ...v1alpha1helpers.UpdateStatusFunc) {
```
deads2k, line 44: why are you managing the lock outside of the method? I'd rather manage the lock locally here.

pkg/cmd/checkendpoints/controller/updates_manager.go
```diff
@@ -0,0 +1,75 @@
...
+	Queue() []v1alpha1helpers.UpdateStatusFunc
+	// Delete the contents of the queue.
+	Clear()
+}
+
+// updateManage implements an UpdateManager sorts the incoming updates and queues updates
+// outside of a sorting window, anchored on one end by the latest update, for processing.
+type updatesManager struct {
+	sync.Mutex
+	window          time.Duration
```
deads2k, line 35: you need to doc these.  They aren't obvious.

pkg/cmd/checkendpoints/controller/updates_manager.go
deads2k, line 44: needs test

pkg/cmd/checkendpoints/controller/connection_checker.go
```diff
@@ -115,29 +112,31 @@ func (c *connectionChecker) Stop() {
 // updateStatus applies updates. If an error occurs applying an update,
 // it remain on the queue and retried on the next call to updateStatus.
 func (c *connectionChecker) updateStatus(ctx context.Context) {
-	c.updatesLock.Lock()
-	defer c.updatesLock.Unlock()
-	if len(c.updates) > 20 {
-		_, _, err := v1alpha1helpers.UpdateStatus(ctx, c.client, c.name, c.updates...)
+	c.updates.Lock()
+	defer c.updates.Unlock()
+	if len(c.updates.Queue()) > 20 {
```
deads2k, line 117: make this function return a copy so that you don't need to manage the lock here.

deads2k: > make this function return a copy so that you don't need to manage the lock here.

oh, blech. The `Clear` is being caught up in here too.  

How about just making `UpdateStatus` native on the updatemanager.

pkg/cmd/checkendpoints/controller/updates_manager.go
```diff
@@ -0,0 +1,75 @@
...
+func (u *updatesManager) Add(timestamp time.Time, updates ...v1alpha1helpers.UpdateStatusFunc) {
+	u.sortingQueue[timestamp] = updates
+
+	u.timestamps = append(u.timestamps, timestamp)
+	sort.Slice(u.timestamps, func(i, j int) bool {
+		return u.timestamps[i].Before(u.timestamps[j])
+	})
+
+	latestTimestamp := u.timestamps[len(u.timestamps)-1]
+	tmp := u.timestamps[:0]
```
deads2k, line 53: what does this do?

pkg/cmd/checkendpoints/controller/updates_manager.go
```diff
@@ -0,0 +1,75 @@
...
+	for _, timestamp := range u.timestamps {
```
deads2k, line 54: yeah, you need to comment this.  Looks like this an attempt to having a sliding window of results based on time. your window needs to be at least one second larger than the delay though.

Do you need to delay the events too?

pkg/cmd/checkendpoints/controller/updates_manager.go
sanchezl: I'm re-using the array backing the u.timstamps slice.

pkg/cmd/checkendpoints/controller/updates_manager.go
sanchezl: The window is initialized as `delay + check period`, so in this case 10s (conn timeout) + 1s (check period).

pkg/cmd/checkendpoints/controller/updates_manager.go
deads2k: > I'm re-using the array backing the u.timstamps slice.

let's just make another and burn th e memory

pkg/cmd/checkendpoints/controller/connection_checker.go
sanchezl: Updated.

pkg/cmd/checkendpoints/controller/updates_manager.go
sanchezl: Updated.

pkg/cmd/checkendpoints/controller/updates_manager.go
sanchezl: Refactored.

pkg/cmd/checkendpoints/controller/updates_manager.go
sanchezl: Done.

pkg/cmd/checkendpoints/controller/updates_manager.go
sanchezl: Done.

pkg/cmd/checkendpoints/controller/updates_manager.go
sanchezl: Added in followup https://github.com/openshift/cluster-kube-apiserver-operator/pull/924.

pkg/cmd/checkendpoints/controller/updates_manager.go
sanchezl: See followup PR where I've tried to improve this further: https://github.com/openshift/cluster-kube-apiserver-operator/pull/924.

PR #918, opened 2020-07-27
Bump to kubernetes 1.19.0-rc2

mfojtik:


mfojtik:
@deads2k please tag if green, want to test drive the bump here and if it is good, do the same for other repos.

PR #920, opened 2020-07-30
Bug 1861431: LatencySensitive feature gate allows upgrades

MarSik:
LatencySensitive FeatureGate should allow upgrades according to
https://github.com/openshift/api/blob/7192180f496aab1f7659d8660fc360498bab498b/config/v1/types_feature.go#L38

### DIFF
--- a/pkg/operator/featureupgradablecontroller/feature_upgradeable_controller.go
+++ b/pkg/operator/featureupgradablecontroller/feature_upgradeable_controller.go
@@ -15,7 +15,7 @@ import (
 )
 
 var (
-	featureGatesAllowingUpgrade = sets.NewString("")
+	featureGatesAllowingUpgrade = sets.NewString("", string(configv1.LatencySensitive))
 )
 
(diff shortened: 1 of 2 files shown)


pkg/operator/featureupgradablecontroller/feature_upgradeable_controller.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 var (
-	featureGatesAllowingUpgrade = sets.NewString("")
+	featureGatesAllowingUpgrade = sets.NewString("", string(configv1.LatencySensitive))
```
cynepco3hahue, line 18: Can we change it to `featureGatesAllowingUpgrade = sets.NewString(string(configv1.LatencySensitive))`

MarSik: No, that actually fails one of the tests (I tried). The empty string seems to represent a situation with no feature gates.

cynepco3hahue: I had such feelings, thanks for the clarification.

PR #923, opened 2020-08-03
[release-4.4] Bug 1863076: LatencySensitive feature gate allows upgrades

fromanirh:
LatencySensitive FeatureGate should allow upgrades according to
https://github.com/openshift/api/blob/7192180f496aab1f7659d8660fc360498bab498b/config/v1/types_feature.go#L38

This is a manual cherrypick from the release-4.5 branch

Signed-off-by: Francesco Romani <fromani@redhat.com>

### DIFF
--- a/pkg/operator/featureupgradablecontroller/feature_upgradeable_controller.go
+++ b/pkg/operator/featureupgradablecontroller/feature_upgradeable_controller.go
@@ -22,7 +22,7 @@ import (
 var (
 	featureUpgradeableControllerWorkQueueKey = "key"
 
-	featureGatesAllowingUpgrade = sets.NewString("")
+	featureGatesAllowingUpgrade = sets.NewString("", string(configv1.LatencySensitive))
 )
 
 // FeatureUpgradeableController is a controller that sets upgradeable=false if anything outside the whitelist is the specified featuregates.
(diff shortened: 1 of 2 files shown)


fromanirh:
This is a manual cherry-pick because the automation failed: https://github.com/openshift/cluster-kube-apiserver-operator/pull/921

sdodson:
Does this change clear a stale condition once it's been set?

fromanirh:
> Does this change clear a stale condition once it's been set?

The intention is to enable upgrades, not to clear existing conditions

sdodson:
> The intention is to enable upgrades, not to clear existing conditions

Right but if the cluster had already set the condition then they upgrade to this version which no longer sets the condition, is there something that clears the stale condition or will they be blocked from upgrading forever?

fromanirh:
> > The intention is to enable upgrades, not to clear existing conditions
> 
> Right but if the cluster had already set the condition then they upgrade to this version which no longer sets the condition, is there something that clears the stale condition or will they be blocked from upgrading forever?

AFAIU (will get in touch with teammate which analysed the bug) 
In general I believe that there is a elaborate workaround which involves disable the feature gate, upgrade, enable again. But I will doublecheck and report. Will also check about clearing the condition and report.

fromanirh:
Will retest later:
```
level=fatal msg="failed to fetch Common Manifests: failed to fetch dependency of \"Common Manifests\": failed to generate asset \"DNS Config\": getting public zone for \"...\": listing hosted zones: Throttling: Rate exceeded\n\tstatus code: 400, request id: 3985bd8d-555d-4aca-8d4e-5c904e64d78b"
```
:\

PR #924, opened 2020-08-04
check-endpoints: minimize sorting delay

sanchezl:
When delaying the processing of updates for the purposes of sorting updates made out of order, sometimes delaying for the shorter check period is enough vs. the longer check timeout.


deads2k:
I suspect we end up looking at this code again, but it's got unit tests now and I'll give it a try.

/lgtm

PR #932, opened 2020-08-13
bump to delay version update until operand updated

deads2k:
proof of https://github.com/openshift/library-go/pull/865


deads2k:
proof was green, this is updating library-go to match and it will clean up our upgrade logging and events

PR #933, opened 2020-08-18
Bug 1874521: connectivity check metrics cannot be aggregated across components 

sanchezl:
Implement changes to the metric names and labels as discovered during reviews for better usability.

* Names changed to no longer be specific to the component check-endpoints is being run in. They will always be:
```
pod_network_connectivity_check_count
pod_network_connectivity_check_tcp_connect_latency_gauge
pod_network_connectivity_check_dns_resolve_latency_gauge
```
* `endpoint` labels renamed to `targetEndpoint` to avoid collisions with default label of the same name.
* `component` label added (e.g. openshift-kube-apiserver or openshift-apiserver) 
* `checkName` label added

### DIFF
--- a/pkg/cmd/checkendpoints/controller/metrics.go
+++ b/pkg/cmd/checkendpoints/controller/metrics.go
@@ -16,44 +16,63 @@ var (
 	dnsResolveLatencyGauge *metrics.GaugeVec
 )
 
-func RegisterMetrics(prefix string) {
+// RegisterMetrics in the global registry
+func RegisterMetrics() {
 	registerMetrics.Do(func() {
 		endpointCheckCounter = metrics.NewCounterVec(&metrics.CounterOpts{
-			Name: prefix + "endpoint_check_count",
-			Help: "Report status of endpoint checks for each pod over time.",
-		}, []string{"endpoint", "tcpConnect", "dnsResolve"})
+			Name: "pod_network_connectivity_check_count",
+			Help: "Report status of pod network connectivity checks over time.",
+		}, []string{"component", "checkName", "targetEndpoint", "tcpConnect", "dnsResolve"})
 
 		tcpConnectLatencyGauge = metrics.NewGaugeVec(&metrics.GaugeOpts{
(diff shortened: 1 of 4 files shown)


tkashem:
> 
> pod_network_connectivity_check_count
> pod_network_connectivity_check_tcp_connect_latency_gauge
> pod_network_connectivity_check_dns_resolve_latency_gauge
> 

The metric naming convention followed by k8s is `{namespace}_{sub system}_name
{namespace} : this is the component name, for example: apiserver (kube-apiserver), etcd, kubelet etc.

what do you think about renaming the metrics to `connectivity_pod_network_check_*`. In future you could have something like `connectivity_sdn_*` or  `connectivity_lb_*`.

 
> component label added (e.g. openshift-kube-apiserver or openshift-apiserver)

We are already using the label value `kube-apiserver` to refer to kube-apiserver.  I would recommend using the same value to keep it consistent.

pkg/cmd/checkendpoints/controller/metrics.go
```diff
@@ -16,44 +16,63 @@ var (
...
-func getCounterMetricLabels(address string, latency *trace.LatencyInfo, checkErr error) map[string]string {
-	labels := map[string]string{
-		"endpoint":   address,
-		"dnsResolve": "",
-		"tcpConnect": "",
-	}
+func (m *metricsContext) getCounterMetricLabels(targetEndpoint string, latency *trace.LatencyInfo, checkErr error) map[string]string {
+	labels := m.getMetricLabels(targetEndpoint)
+	labels["dnsResolve"] = ""
+	labels["tcpConnect"] = ""
```
tkashem, line 75: we can think about choosing a non-empty value to indicate the default/initial condition? In general, it's safer to avoid an empty string - no need to worry about how prometheus handles non existing and empty label.

sanchezl: From https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels:
> A label with an empty label value is considered equivalent to a label that does not exist.

PR #935, opened 2020-08-19
Bug 1869913: connectivity outage events duplicated when status updates are retried

sanchezl:
- check-endpoints: don't duplicate events on update error
- check-endpoints: force flush ignored


deads2k:
I think we may need something different. Probably a separate ratelimiter or different correlators for these events.   I think the event recorder has options.

sanchezl:
/test e2e-aws-upgrade
--

deads2k:
/close

this doesn't practically happen and the existing de-duping logic applies.

PR #936, opened 2020-08-19
bug 1871890: add current profile annotations to CVO manifests

deads2k:
This is matches https://github.com/openshift/enhancements/pull/414 and doesn't change existing behavior


manifests/0000_20_kube-apiserver-operator_00_cr-anyuid.yaml
```diff
@@ -2,6 +2,8 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
```
derekwaynecarr, line 5: i dont really like this term as i would anticipate hypershift-installer working on aws and not needing new manifests.

retroflexer:
Opened a new PR https://github.com/openshift/cluster-kube-apiserver-operator/pull/957 to rebase and to supersede this PR.

/close

PR #946, opened 2020-09-09
Bug 1877234: Drop recovery apiserver

tnozicka:
The recovery apiserver is obsolete since we have cert auto-recovery.

For accessing API in situations when certs are expired KAS has localhost-recovery endpoint with long lived certs for localhost and local kubeconfigs on the master.

Requires:
 - [x] https://github.com/openshift/cluster-kube-apiserver-operator/pull/858

/hold

/cc @sttts @deads2k


tnozicka:
/hold cancel
https://github.com/openshift/cluster-kube-apiserver-operator/pull/858 merged

PR #949, opened 2020-09-11
Fix resync periods

tnozicka:
Resync period for targetconfigcontroller should be higher then 1s. I suspect this is the reason for a lot off throttle messages in the logs.

/cc @deads2k

### DIFF
--- a/pkg/operator/certrotationtimeupgradeablecontroller/certrotationtime_upgradeable.go
+++ b/pkg/operator/certrotationtimeupgradeablecontroller/certrotationtime_upgradeable.go
@@ -3,7 +3,6 @@ package certrotationtimeupgradeablecontroller
 import (
 	"context"
 	"fmt"
-	"time"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/library-go/pkg/controller/factory"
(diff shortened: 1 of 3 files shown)


pkg/operator/certrotationtimeupgradeablecontroller/certrotationtime_upgradeable.go
```diff
@@ -38,7 +38,7 @@ func NewCertRotationTimeUpgradeableController(
 	return factory.New().WithInformers(
 		operatorClient.Informer(),
 		configMapInformer.Informer(),
-	).WithSync(c.sync).ResyncEvery(time.Second).ToController("CertRotationTimeUpgradeableController", eventRecorder.WithComponentSuffix("certRotationTime-upgradeable"))
+	).WithSync(c.sync).ResyncEvery(1*time.Second).ToController("CertRotationTimeUpgradeableController", eventRecorder.WithComponentSuffix("certRotationTime-upgradeable"))
```
deads2k, line 41: seems way to tight

tnozicka: It does seem tight. Although I though it was previously set to react to cert life cycle with 1s accuracy - I we want to test shortened rotation periods of say 10 or 5 minutes this needs to granular enough not to miss critical rotation.

What's the biggest lag you think we can afford? I was fine with the 1s as it doesn't result in external action or using live client afaik

sttts: how is 30 sec or so to granular to catch something for rotation of 5 min?

tnozicka: I was only saying it's a choice and it doesn't seem to do any live calls unless a change is desired. We may be fine with rotating at 3m instead of 2:30. If we claim it as tolerable lag, 30 seconds seem reasonable.

tnozicka: Apologies, I was actually talking about another controller. This doesn't actually need a resync at all. It's purely based on object state and both informers are hooked.

PR #954, opened 2020-09-18
Bug 1880322: fix bug in reflector not recovering from "Too large resource version" 

p0lyn0mial:


vendor/k8s.io/client-go/tools/cache/reflector.go
```diff
@@ -256,13 +256,14 @@ func (r *Reflector) ListAndWatch(stopCh <-chan struct{}) error {
 			}
 
 			list, paginatedResult, err = pager.List(context.Background(), options)
-			if isExpiredError(err) {
-				r.setIsLastSyncResourceVersionExpired(true)
-				// Retry immediately if the resource version used to list is expired.
+			if isExpiredError(err) || isTooLargeResourceVersionError(err) {
```
p0lyn0mial, line 259: this is what we need.

PR #955, opened 2020-09-22
Bug 1882737: check-endpoints should gracefully handle missing podnetworkconnectivitycheck crd

sanchezl:
First step in being able to enable/disable check-endpoints. 

- check-endpoints: block if no podnetworkconnectivitycheck crd established
- check-endpoints-kubeconfig: add crd permissions


pkg/cmd/checkendpoints/controller/pod_network_connectivity_check_controller.go
```diff
@@ -67,6 +74,16 @@ func NewPodNetworkConnectivityCheckController(podName, podNamespace string,
 // Sync ensures that the status updaters for each PodNetworkConnectivityCheck is started
 // and then performs each check.
 func (c *controller) Sync(ctx context.Context, syncContext factory.SyncContext) error {
+	// if PNCC CRD is not installed and established, cleanup any existing checkers, block
+	established, err := c.podNetworkConnectivityCheckResourceDefinitionEstablished()
```
deads2k, line 78: make this "past the post".  Once you successfully see the CRD created, just assume it never goes away.  Separately, write a control loop that watches for the CRD to disappear and then exit(0)

deads2k, line 78: Also, this is in the wrong spot.  Placing it here means we'll have reflectors throwing perma-fails in our logs.  Do this check well in advance of starting the controllers.

pkg/cmd/checkendpoints/cmd.go
```diff
@@ -24,9 +26,11 @@ func NewCheckEndpointsCommand() *cobra.Command {
 		podName := os.Getenv("POD_NAME")
 		namespace := os.Getenv("POD_NAMESPACE")
 		kubeClient := kubernetes.NewForConfigOrDie(cctx.ProtoKubeConfig)
+		apiextensionsClient := apiextensionsclient.NewForConfigOrDie(cctx.KubeConfig)
```
deads2k, line 29: check goes here-ish.

pkg/cmd/checkendpoints/controller/pod_network_connectivity_check_controller.go
sanchezl: Updated.

sanchezl: Updated.

pkg/cmd/checkendpoints/cmd.go
sanchezl: Updated.

bindata/v4.1.0/kube-apiserver/check-endpoints-clusterrole-node-crd-reader.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: system:openshift:controller:check-endpoints-node-crd-reader
```
sttts, line 4: after this we have the old and th e new object, with overlap. Why not add one for CRDs only?

pkg/cmd/checkendpoints/cmd.go
```diff
@@ -60,10 +65,39 @@ func NewCheckEndpointsCommand() *cobra.Command {
...
+			apiextensionsInformers.Apiextensions().V1().CustomResourceDefinitions(),
+			recorder,
+		)
+
 		controller.RegisterMetrics()
+
+		// block until the PodNetworkConnectivityCheck CRD has been established
+		apiextensionsInformers.Start(ctx.Done())
+		ttsContext, ttsCancel := context.WithCancel(ctx)
+		go timeToStart.Run(ttsContext, 1)
```
sttts, line 84: this feels super overengineered. Why not a simple wait.Until func that gets the CRD every 30 sec and blocks?

pkg/cmd/checkendpoints/start_stop_controllers.go
```diff
@@ -0,0 +1,90 @@
...
+		ToController("CheckEndpointsTimeToStart", recorder)
+	return c
+}
+
+func (c *timeToStartController) Ready() <-chan error {
+	return c.C
+}
+
+// stopController stops the process using os.Exit(1) if the podnetworkconnectivitychecks crd
+// stops being established after the process has started.
```
sttts, line 50: why not `os.Exit(0)`? Does look like this is an error.

pkg/cmd/checkendpoints/start_stop_controllers.go
```diff
@@ -0,0 +1,90 @@
...
+		crdLister: crdInformer.Lister(),
+	}
+	c.Controller = factory.New().
+		WithInformers(crdInformer.Informer()).
+		WithSync(func(context.Context, factory.SyncContext) error {
+			established, err := podNetworkConnectivityCheckResourceDefinitionEstablished(c.crdLister)
+			if err != nil {
+				return err
+			}
+			if !established {
```
sttts, line 67: it can never go unestablished. Check for 404.

bindata/v4.1.0/kube-apiserver/check-endpoints-clusterrole-node-crd-reader.yaml
sanchezl: These are new for v4.6.

pkg/cmd/checkendpoints/start_stop_controllers.go
sanchezl: Done.

pkg/cmd/checkendpoints/start_stop_controllers.go
sanchezl: Done.

pkg/cmd/checkendpoints/cmd.go
sanchezl: I was under the impression that, overall, using Informer/Lister was a better practice than polling w/ get. The stop controller needed the informer, and our factory.Controller framework made it easy also use the informer in a "start controller" with minimal effort.

pkg/cmd/checkendpoints/start_stop_controllers.go
```diff
@@ -0,0 +1,86 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/klog/v2"
+)
+
+// timeToStartController will either close its ready chan, when the podnetworkconnectivitychecks
+// type exists, or send an error to the ready chan if an error has occurred.
+type timeToStartController struct {
+	factory.Controller
+	crdLister apiextensionslistersv1.CustomResourceDefinitionLister
+	C         chan error
```
deads2k, line 20: we don't charge you by the letter.

pkg/cmd/checkendpoints/start_stop_controllers.go
```diff
@@ -0,0 +1,86 @@
...
+	c := &timeToStartController{
+		crdLister: crdInformer.Lister(),
+		C:         make(chan error, 1),
+	}
+	c.Controller = factory.New().
+		WithInformers(crdInformer.Informer()).
+		WithSync(func(context.Context, factory.SyncContext) error {
+			ok, err := podNetworkConnectivityCheckTypeExists(c.crdLister)
+			if err != nil {
+				c.C <- err
```
deads2k, line 33: if this happens after a close, you write to a nil channel.

deads2k: > if this happens after a close, you write to a nil channel.

though I can admit would be nearly unheard of and an unlikely race

deads2k:
we made controllers too easy....


/lgtm

sanchezl:
@deads2k rebased

PR #958, opened 2020-09-23
Bug 1880786: PodNetworkConnectivityCheck is missing description for `oc explain`

sttts:
CRD update from api repo was missing.


vendor/github.com/openshift/api/operatorcontrolplane/v1alpha1/0000_10-pod-network-connectivity-check.crd.yaml
```diff
@@ -13,254 +12,247 @@ spec:
     plural: podnetworkconnectivitychecks
     singular: podnetworkconnectivitycheck
   scope: ""
```
sanchezl, line 14: I'm seeing this when trying to create:
```
The CustomResourceDefinition "podnetworkconnectivitychecks.controlplane.operator.openshift.io" is invalid: spec.scope: Required value
```

sanchezl (changes requested):

PR #962, opened 2020-09-29
Bug 1883698: Bug 1883603: Bump openshift/api

dmage:
openshift/api#736 added restrictions for logLevel values. This PR vendors new openshift/api to get new CRDs.


dmage:
/retitle Bug 1883698: Bug 1883603: Bump openshift/api

PR #966, opened 2020-10-04
protect openshift traffic by using dedicated flowschema

tkashem:
Define dedicated flowschema and priority configuration that will protect openshift specific traffic.
- `SAR` and `tokenreviews` from both `oas` and `oauth` are very important
- kcm, other `oas` and `oauth` requests, `/metrics` requests from openshift-monitoring is pretty important
- openshift controller manager is as important as `kcm`.
- control plane operators are important (kas-o, oas-o, auth operator, etcd operator)
- `workloads-low` goes below the traffic defined above.

Question:
- do we want to include `cvo` in the control plane operators, or are there traffic from any other critical operators that we should protect?

New configuration:
image

After applying the above configuration, we can see that the number of requests from `workload-low` drops. This implies that traffic from oas, `/metrics` call from openshift monitoring, traffic from control plane operators are not being throttled anymore.
image

### DIFF
--- a/manifests/0000_20_kube-apiserver-operator_08_flowschema.yaml
+++ b/manifests/0000_20_kube-apiserver-operator_08_flowschema.yaml
@@ -1,0 +1,78 @@
+apiVersion: flowcontrol.apiserver.k8s.io/v1alpha1
+kind: PriorityLevelConfiguration
+metadata:
+  name: openshift-aggregated-api-delegated-auth
+spec:
+  limited:
+    assuredConcurrencyShares: 20
+    limitResponse:
+      queuing:
+        handSize: 6
+        queueLengthLimit: 50
+        queues: 16
+      type: Queue
+  type: Limited
+---
+apiVersion: flowcontrol.apiserver.k8s.io/v1alpha1
+kind: PriorityLevelConfiguration
+metadata:
+  name: openshift-control-plane-operators
+spec:
+  limited:
+    assuredConcurrencyShares: 10
+    limitResponse:
+      queuing:
+        handSize: 6
+        queueLengthLimit: 50
+        queues: 128
+      type: Queue
+  type: Limited
+---
+apiVersion: flowcontrol.apiserver.k8s.io/v1alpha1
+kind: FlowSchema
+metadata:
+  name: openshift-monitoring-metrics
+spec:
+  distinguisherMethod:
+    type: ByUser
+  matchingPrecedence: 2000
+  priorityLevelConfiguration:
+    name: workload-high
+  rules:
+  - nonResourceRules:
+    - verbs:
+      - '*'
+      nonResourceURLs:
+      - "/metrics"
+    subjects:
+    - kind: ServiceAccount
+      serviceAccount:
+        name: prometheus-k8s
+        namespace: openshift-monitoring
+---
+apiVersion: flowcontrol.apiserver.k8s.io/v1alpha1
+kind: FlowSchema
+metadata:
+  name: openshift-kube-apiserver-operator
+spec:
+  distinguisherMethod:
+    type: ByUser
+  matchingPrecedence: 2000
+  priorityLevelConfiguration:
+    name: openshift-control-plane-operators
+  rules:
+  - resourceRules:
+    - apiGroups:
+      - '*'
+      clusterScope: true
+      namespaces:
+      - '*'
+      resources:
+      - '*'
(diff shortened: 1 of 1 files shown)


manifests/0000_20_kube-apiserver-operator_08_flowschema.yaml
```diff
@@ -0,0 +1,144 @@
...
+    - apiGroups:
+      - '*'
+      clusterScope: true
+      namespaces:
+      - '*'
+      resources:
+      - '*'
+      verbs:
+      - '*'
+    subjects:
```
deads2k, line 128: etcd, kas, kcm, ks, auth, oas

manifests/0000_20_kube-apiserver-operator_08_flowschema.yaml
```diff
@@ -0,0 +1,144 @@
+apiVersion: flowcontrol.apiserver.k8s.io/v1alpha1
+kind: PriorityLevelConfiguration
+metadata:
+  name: openshift-system
```
deads2k, line 4: This should be allowed more requests than `openshift-control-plane-operators`.  Even though it's one server, really don't want to run out.

Also, rename `aggregated-api-delegated-auth` I think.  We'll want to tie the oauth-apiserver to it too, right?

manifests/0000_20_kube-apiserver-operator_08_flowschema.yaml
```diff
@@ -0,0 +1,144 @@
...
+    assuredConcurrencyShares: 20
+    limitResponse:
+      queuing:
+        handSize: 6
+        queueLengthLimit: 50
+        queues: 128
+      type: Queue
+  type: Limited
+---
+apiVersion: flowcontrol.apiserver.k8s.io/v1alpha1
```
deads2k, line 31: I'd rather keep rules like this local to operators that use them.

manifests/0000_20_kube-apiserver-operator_08_flowschema.yaml
tkashem: > Also, rename aggregated-api-delegated-auth
sounds good, I am adding a prefix `openshift` to it. 

> We'll want to tie the oauth-apiserver to it too, right?
I have added oauth apiserver. 

> This should be allowed more requests than openshift-control-plane-operators. Even though it's one server, really don't want to run out.
Yes, makes sense. we expect the operators to have a steady call rate (unless there is a bug that causes a hot loop or something) and openshift api servers are tied to the user traffic. I doubled the concurrency share to 20 and reduced for the operators to 10.

manifests/0000_20_kube-apiserver-operator_08_flowschema.yaml
tkashem: @deads2k Yes, I thought about it. the only concern I have is the `matchingPrecedence` field. This defines the matching order, so we need to keep in mind all the other values to make sure the right order is maintained.

deads2k: > @deads2k Yes, I thought about it. the only concern I have is the `matchingPrecedence` field. This defines the matching order, so we need to keep in mind all the other values to make sure the right order is maintained.

We did this with conventions in the CVO.  Our payloads are local.  I'm ok moving these in 4.7 instead of 4.6.

manifests/0000_20_kube-apiserver-operator_08_flowschema.yaml
deads2k: found https://github.com/openshift/cluster-openshift-apiserver-operator/pull/398 I like it.

manifests/0000_20_kube-apiserver-operator_08_flowschema.yaml
```diff
@@ -0,0 +1,78 @@
...
+kind: PriorityLevelConfiguration
+metadata:
+  name: openshift-aggregated-api-delegated-auth
+spec:
+  limited:
+    assuredConcurrencyShares: 20
```
deads2k, line 7: bump this to 100 for me?  It should only matter when we start hitting pressure.

deads2k: > bump this to 100 for me? It should only matter when we start hitting pressure

actually, when looking at it, this is an ok starting point.

PR #968, opened 2020-10-05
BUG 1885353: protect openshift traffic by using dedicated flowschema

tkashem:
Define dedicated flowschema and priority configuration that will protect openshift specific traffic.
- `SAR` from `oas` is very important (exempt)
- kcm, other `oas` requests, `oauth server` requests, metrics requests from openshift-monitoring is pretty important
- control plane operators are important (kas-o, oas-o, auth operator, etcd operator)
- `workloads-low` goes below the traffic defined above.

Question:
- is `tokenreviews` from `oas` as important as `SAR`?
- do we want to include `cvo` in the control plane operators, or are there traffic from any other critical operators that we should protect?

After applying the above configuration, we can see that the number of requests from `workload-low` drops. This implies that traffic from oas, `/metrics` call from openshift monitoring, traffic from control plane operators are not being throttled anymore.
image

### DIFF
--- a/manifests/0000_20_kube-apiserver-operator_08_flowschema.yaml
+++ b/manifests/0000_20_kube-apiserver-operator_08_flowschema.yaml
@@ -1,0 +1,261 @@
+apiVersion: flowcontrol.apiserver.k8s.io/v1alpha1
+kind: PriorityLevelConfiguration
+metadata:
+  name: openshift-control-plane-operators
+spec:
+  limited:
+    assuredConcurrencyShares: 10
+    limitResponse:
+      queuing:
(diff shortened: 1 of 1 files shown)


tkashem:
/hold

doing some research whether we need to make SAR from oas/oauth exempt.

tkashem:
we are making SAR from oas/oauth exempt since borrowing is not supported in p&f

/hold cancel

wangke19:
Launched one 4.5 cluster with this PR successfully, verification steps see https://bugzilla.redhat.com/show_bug.cgi?id=1885353#c3.

PR #979, opened 2020-10-12
fakebump: set UID in static pods so kubelet honors grace period

deads2k:
proof of https://github.com/openshift/library-go/pull/920


deads2k:
/close

we have an actual PR now.

PR #981, opened 2020-10-13
bug 1888015: bump library-go to pick up static pod graceful timeout workaround

deads2k:


deads2k:
/cherrypick release-4.6

this will auto-make the right bugs

PR #984, opened 2020-10-14
BUG 1888309: p&f: make oas/oauth SAR/TokenReview traffic exempt

tkashem:


deads2k:
/lgtm
/hold

holding to get confirmation that we do not borrow from shares with more concurrency.

tkashem:
we are making SAR from oas/oauth exempt since borrowing is not supported in p&f

/hold cancel

PR #985, opened 2020-10-14
Bug 1888862: [4.6] set servicesSubnet correctly in dual-stack clusters

danwinship:
In a dual-stack cluster, kube-apiserver needs to know about both service CIDR ranges.

Backport of #986. Rather than also backporting the library-go portion this just imports the new 4.6 library-go function.

### DIFF
--- a/pkg/operator/configobservation/network/observe_network.go
+++ b/pkg/operator/configobservation/network/observe_network.go
@@ -1,10 +1,15 @@
 package network
 
 import (
-	"net"
+	"fmt"
+	"strings"
 
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	utilnet "k8s.io/utils/net"
 
+	configv1 "github.com/openshift/api/config/v1"
+	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
 	"github.com/openshift/library-go/pkg/operator/configobserver"
 	"github.com/openshift/library-go/pkg/operator/configobserver/network"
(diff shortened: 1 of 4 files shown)


pkg/operator/configobservation/network/observe_network.go
```diff
@@ -86,26 +90,21 @@ func ObserveServicesSubnet(genericListers configobserver.Listers, recorder event
...
+	if err := unstructured.SetNestedField(out, servicesSubnet, servicesSubnetConfigPath...); err != nil {
 		errs = append(errs, err)
 	}
 	bindAddress := "0.0.0.0:6443"
 	bindNetwork := "tcp4"
-	// TODO: only do this in the single-stack IPv6 case once network.GetServiceCIDR()
-	// supports dual-stack
-	if serviceCIDR != "" {
-		if ip, _, err := net.ParseCIDR(serviceCIDR); err != nil {
-			errs = append(errs, err)
```
aojea, line 104: this was validating that the service CIDR was valid and now we no longer validate here.
 I'm not familiar with this code but I assume that was validated before, right?

danwinship: Yes, it is known to already be valid. The code wasn't calling `net.ParseCIDR()` to validate it, it was calling it so it could then call `ip.To4()`, and the author didn't feel comfortable ignoring the error case, even though it "can't happen". The new code _does_ ignore parsing errors (`utilnet.IsIPv6CIDRString` returns `false` if the string doesn't parse, since a string which is not a CIDR is not an IPv6 CIDR). But again, "can't happen" (and would result in more obvious error messages elsewhere anyway if it did happen).

PR #989, opened 2020-10-21
connectivitycheckcontroller: disable by default

sanchezl:
Turning off again as it is preventing 4.6.0 -> 4.7 from completing successfully.


deads2k:
not awesome, but we need a payload

/lgtm

PR #986, opened 2020-10-15
Bug 1888861: set servicesSubnet correctly in dual-stack clusters

danwinship:
In a dual-stack cluster, kube-apiserver needs to know about both service CIDR ranges.

I considered just ignoring the bootstrap config, since we don't currently ever need dual-stack services to work at bootstrap time, but maybe we will in the future? (Note that for the runtime config we require the user to enable the feature gate explicitly via an install manifest, since it's still alpha, but it's difficult to figure out if they have done that from the bootstrap code so I just enabled it unconditionally in the bootstrap config if they specified dual-stack service CIDRs.)

/hold
depends on https://github.com/openshift/library-go/pull/925

PR #990, opened 2020-10-23
Bug 1890724: Reduce OperatorStatusChanged event noise for staticpod controller

sttts:
Bumping to get https://github.com/openshift/library-go/pull/930


sttts:
Depends on https://github.com/openshift/cluster-kube-apiserver-operator/pull/986

PR #995, opened 2020-10-27
Restrict log permissions

JAORMX:
While the current log directory permissions are restrictive and correct,
the file permissions are too permissive and it raises flags on
evaluations of the deployment. Let's instead change the permissions to
0600 which is more appropriate for these types of logs.


JAORMX:
This is similar to the fix proposed in https://github.com/openshift/cluster-openshift-apiserver-operator/pull/397

PR #998, opened 2020-10-29
add prometheus alert for deprecated API use

vareti:
Adds a rule to create an alert every time deprecated API is accessed.

Needs https://github.com/openshift/kubernetes/pull/421

### DIFF
--- a/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
+++ b/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
@@ -131,3 +131,26 @@ spec:
     - record: cluster:apiserver_current_inflight_requests:sum:max_over_time:2m
       expr: |
         max_over_time(sum(apiserver_current_inflight_requests{apiserver=~"openshift-apiserver|kube-apiserver"}) by (apiserver,requestKind)[2m:])
+---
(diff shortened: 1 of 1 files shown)


manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -131,3 +131,26 @@ spec:
...
+  - name: using-deprecated-apis
+    rules:
+    - alert: DeprecatedAPIUse
+      annotations:
+        message: |
+           "{{ $labels.client }}" has used a deprecated "{{ $labels.group }}"/"{{ $labels.version }}" API version of the "{{ $labels.resource }}" resource. This API version will be removed {{ $labels.removed_release }}; please migrate to the current version of this API.
+        summary: |
+            "{{ $labels.client }}" is using a deprecated "{{ $labels.group }}/{{ $labels.version }}" API version
+      expr: |
+        apiserver_requested_deprecated_apis == 1
```
sttts, line 154: I wonder whether we can restrict this to the APIs that go away in the next release.

vareti: I agree we need to refine this further. Also, message in the alert reports Kubernetes releases but not OpenShift releases. Even that needs to be looked at.

PR #1003, opened 2020-11-05
set more kube options directly via flags

deads2k:
continues the existing effort of trying to set all kube options via flags to limit our patches and make our configuration more transparent and comparable.


deads2k:
/hold

this now has an illegal carry to see if that's what's needed to fix it

pkg/operator/configobservation/network/observe_network.go
```diff
@@ -80,11 +80,10 @@ func ObserveServicesSubnet(genericListers configobserver.Listers, recorder event
 	listers := genericListers.(configobservation.Listers)
 
 	out := map[string]interface{}{}
-	servicesSubnetConfigPath := []string{"servicesSubnet"}
-	bindAddressConfigPath := []string{"servingInfo", "bindAddress"}
-	bindNetworkConfigPath := []string{"servingInfo", "bindNetwork"}
+	servicesSubnetConfigPath := []string{"apiServerArguments", "service-cluster-ip-range"}
+	bindAddressConfigPath := []string{"apiServerArguments", "bind-address"}
```
p0lyn0mial, line 84: on the downgrade, the old operator will just replace it with `servingInfo.bindAddress`,  right?

squeed:
/uncc
y'all got this under control :-)

pkg/operator/configobservation/network/observe_network.go
deads2k: > on the downgrade, the old operator will just replace it with `servingInfo.bindAddress`, right?

yeah, no impact to the old operator.

PR #1004, opened 2020-11-09
token webhook auth: set tokenreview API to v1

stlaz:
required in order for https://github.com/openshift/cluster-authentication-operator/pull/304 and https://github.com/openshift/oauth-apiserver/pull/9 to work

### DIFF
--- a/pkg/operator/configobservation/auth/webhook_authenticator.go
+++ b/pkg/operator/configobservation/auth/webhook_authenticator.go
@@ -20,8 +20,10 @@ import (
 )
 
 var (
-	webhookTokenAuthenticatorPath = []string{"apiServerArguments", "authentication-token-webhook-config-file"}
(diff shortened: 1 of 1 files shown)


deads2k:
`webhookSecretName` appears to be set from spec, which makes sense for user specified options, but wouldn't we expect the auth-operator to set a separate status field to avoid conflicts?  We would honor the spec value first and the status value second.

This does punt the problem of "deconflict with users" to any other operator that wants to directly set it, but they can be explicit opt-ins from the user.  Ours has to be on by default and disable-able.

stlaz:
> webhookSecretName appears to be set from spec, which makes sense for user specified options, but wouldn't we expect the auth-operator to set a separate status field to avoid conflicts? We would honor the spec value first and the status value second.

The original behavior of the observer was implemented according to the enhancement: https://github.com/openshift/enhancements/pull/142/files and it depends on the `Type` field in `authentication.config/cluster`. However, your suggestion is closer to what we're doing with the oauth metadata configmap and I wouldn't mind changing it that way to gain some more consistency here.

PR #1007, opened 2020-11-13
remove priority config openshift-aggregated-api-delegated-auth

tkashem:
The `SAR` traffic is being allocated to `exempt` now, so `openshift-aggregated-api-delegated-auth` has been orphaned.

### DIFF
--- a/manifests/0000_20_kube-apiserver-operator_08_flowschema.yaml
+++ b/manifests/0000_20_kube-apiserver-operator_08_flowschema.yaml
@@ -1,25 +1,6 @@
 apiVersion: flowcontrol.apiserver.k8s.io/v1alpha1
 kind: PriorityLevelConfiguration
 metadata:
-  name: openshift-aggregated-api-delegated-auth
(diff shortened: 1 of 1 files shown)


deads2k:
/lgtm
/approve

however, I'd like a new PR opened use https://github.com/openshift/enhancements/blob/master/enhancements/update/object-removal-manifest-annotation.md to remove it in 4.8 once @jottofar lands the impl.

sttts:
So we need some code to remove it, don't we?

tkashem:
> So we need some code to remove it, don't we?

@sttts I think the plan is to add a new PR with removal annotation once https://github.com/openshift/enhancements/blob/master/enhancements/update/object-removal-manifest-annotation.md lands. 

I am not aware if the cvo feature will be backported, otherwise we have to come up with a way to remove it. I am also thinking of opening a PR upstream where an orphaned priority-level is not taking into account so it does not eat away from the concurrency pool.

PR #1012, opened 2020-11-23
Support user provided service-account-signing-key and issuer

abhinavdahiya:
Revert the revert https://github.com/openshift/cluster-kube-apiserver-operator/pull/1011 therefore re-adding the changes from https://github.com/openshift/cluster-kube-apiserver-operator/pull/1006

And to fix the cause of the revert:

### boundsatokensignercontroller: enure that empty public key is not syned to bound-sa-token-signing-certs

in case the next-bound-service-account-signing-key has empty contents, for example
- when user want to rotate the keys and emptied the keys of the secret for operator to recreate them.
- when on bootstrap the render creates an empty secret as no user-provided keys exist, prompting the operator to create
  then in-cluster.

the bound-sa-token-signing-certs must wait for the public key to exist/non-empty before adding it to the ConfigMap, otherwise
there is failure starting kube-apiserver,

```
Error: error reading public key file /etc/kubernetes/static-pod-resources/configmaps/bound-sa-token-signing-certs/service-account-001.pub: data does not contain any valid RSA or ECDSA public keys
```

because the ConfigMap has a key with empty data like,
> https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-monitoring-operator/981/pull-ci-openshift-cluster-monitoring-operator-master-e2e-agnostic/1330864689164324864/artifacts/e2e-agnostic/gather-extra/configmaps.json

```
        {
            "apiVersion": "v1",
            "data": {
                "service-account-001.pub": "",
                "service-account-002.pub": "-----BEGIN RSA PUBLIC KEY-----\nMIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAqlOait7Q4okCBEZJ1NNn\nHtNQTC/i2bL9p7cTwbfDNbUJ/5rfYGPSCpyREckIfQb5qlMQa9vc8XI0tC5LGjAK\nTwUEfC/Z3I0kxcx71sDDt/qmtmuDtsauaOtFAyuNG28lULI58Jou6MsYAtADtoia\n9h+rpyQhCBIwmsNvAQBrijwv+eIRQXPEhnCFpVbB4sh2TpH7P0LjcRIDD6ZD2HmV\nRKAuKsWqwR2tRXbLqSvdMRGgOzGmqbA1IkG2xaOQmHuYg5GUkec6lRcYErE4I8MG\nYxFbtbGI/r6eNn+mkktNPk6od4Cz83zd+Z2+tSbvxBzkpnIOxd413FwGNgAOc1WO\nIwIDAQAB\n-----END RSA PUBLIC KEY-----\n"
            },
            "kind": "ConfigMap",
            "metadata": {
                "name": "bound-sa-token-signing-certs",
                "namespace": "openshift-config-managed",
                "resourceVersion": "7023",
                "selfLink": "/api/v1/namespaces/openshift-config-managed/configmaps/bound-sa-token-signing-certs",
                "uid": "0f34f86e-3b3a-4a9c-8ed2-1eaae3dcd243"
            }
        },
```

xref: https://issues.redhat.com/browse/CO-1266

### DIFF
--- a/pkg/cmd/render/render.go
+++ b/pkg/cmd/render/render.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net"
+	"os"
 	"path/filepath"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -38,6 +39,7 @@ type renderOpts struct {
 	etcdServerURLs    []string
 	etcdServingCA     string
 	clusterConfigFile string
+	clusterAuthFile   string
 }
 
 // NewRenderCommand creates a render command.
@@ -79,6 +81,7 @@ func (r *renderOpts) AddFlags(fs *pflag.FlagSet) {
 	fs.StringArrayVar(&r.etcdServerURLs, "manifest-etcd-server-urls", r.etcdServerURLs, "The etcd server URL, comma separated.")
 	fs.StringVar(&r.etcdServingCA, "manifest-etcd-serving-ca", r.etcdServingCA, "The etcd serving CA.")
 	fs.StringVar(&r.clusterConfigFile, "cluster-config-file", r.clusterConfigFile, "Openshift Cluster API Config file.")
+	fs.StringVar(&r.clusterAuthFile, "cluster-auth-file", r.clusterAuthFile, "Openshift Cluster Authentication API Config file.")
 }
 
 // Validate verifies the inputs.
@@ -142,6 +145,10 @@ type TemplateData struct {
 
 	// BindNetwork is the network (tcp4 or tcp6) to bind to
 	BindNetwork string
+
+	ServiceAccountIssuer string
+
+	UserProvidedBoundSASigningKey bool
 }
 
 // Run contains the logic of the render command.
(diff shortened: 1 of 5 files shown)


abhinavdahiya:
/assign @sttts @marun 

as you helped with the original PR in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1006 _thanks!_

pkg/cmd/render/render.go
```diff
@@ -162,6 +169,22 @@ func (r *renderOpts) Run() error {
 			return fmt.Errorf("unable to parse restricted CIDRs from config %q: %v", r.clusterConfigFile, err)
 		}
 	}
+	if len(r.clusterAuthFile) > 0 {
+		clusterAuthFileData, err := ioutil.ReadFile(r.clusterAuthFile)
+		if err != nil && !os.IsNotExist(err) {
```
tnozicka, line 174: ignoring a not found for a flag that is explicitly set should at least log what happened

PR #1013, opened 2020-11-25
Always set service-account-issuer flag

tnozicka:
The flag is required since kube 1.20

To be follower by a BZ to wire the proper key to the installer but we need to unblock the rebase.

/cc @sttts @marun

### DIFF
--- a/bindata/bootkube/config/bootstrap-config-overrides.yaml
+++ b/bindata/bootkube/config/bootstrap-config-overrides.yaml
@@ -37,6 +37,8 @@ admission:
         - {{.}}{{end}}
     {{end}}
 apiServerArguments:
+  api-audiences:
+  - {{ or .ServiceAccountIssuer "https://kubernetes.default.svc" }}
   client-ca-file:
     - /etc/kubernetes/secrets/kube-apiserver-complete-client-ca-bundle.crt
   etcd-cafile:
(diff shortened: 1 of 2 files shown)


marun:
@tnozicka Wait, how would this allow the apiserver to start? Doesn't the apiserver need a valid keyfile? It wouldn't make sense that it just accepts an empty signing key.

/lgtm cancel

marun:
Aha, it's just reusing the legacy token key.

/lgtm

bindata/bootkube/config/bootstrap-config-overrides.yaml
```diff
@@ -75,10 +77,14 @@ apiServerArguments:
...
-    - {{.ServiceAccountIssuer}}{{end}}
-  service-account-signing-key-file: {{if .UserProvidedBoundSASigningKey}}
-    - /etc/kubernetes/secrets/bound-service-account-signing-key.key{{end}}
+  service-account-issuer:
+    - {{ or .ServiceAccountIssuer "https://kubernetes.default.svc" }}
+  service-account-signing-key-file:
+  {{- if .UserProvidedBoundSASigningKey }}
+    - /etc/kubernetes/secrets/bound-service-account-signing-key.key
+  {{- else }}
+    - /etc/kubernetes/secrets/service-account.key # FIXME: wire a real default key
```
tnozicka, line 86: https://bugzilla.redhat.com/show_bug.cgi?id=1902059

PR #742, opened 2020-01-28
Bug 1793850: Update UsingDeprecatedAPIExtensionsV1Beta1 alert

soltysh:
/assign @tnozicka

PR #1018, opened 2020-12-08
add DeprecatedAPIInUse alert

sanchezl:
Warn via an `info` alert that an API that will be removed in the next OpenShift version is currently in use.


manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -135,3 +135,14 @@ spec:
...
       expr: |
         max_over_time(sum(apiserver_current_inflight_requests{apiserver=~"openshift-apiserver|kube-apiserver"}) by (apiserver,requestKind)[2m:])
+  - name: pre-release-lifecycle
+    rules:
+      - alert: DeprecratedAPIUsed
+        annotations:
+          message: |
+            Deprecated API that will be removed in the next version is being used: {{"{{$labels.group}}"}}.{{"{{$labels.version}}"}}/{{"{{$labels.resource}}"}}
+        expr: |
+          group(apiserver_requested_deprecated_apis{removed_release="1.20"}) by (group,version,resource,subresource) * on(group,version,resource,subresource) group_right() apiserver_request_total
```
wking, line 145: master is 1.20, right, so shouldn't this be `removed_release="1.21"` here?  And then changed to `1.20` if/when this gets backported to 4.6?

simonpasquier (commented):
the problem with such alert is that it will never clear by itself. That's already been discussed in https://bugzilla.redhat.com/show_bug.cgi?id=1793850 and #742 with the `UsingDeprecatedAPIExtensionsV1Beta1` alert. Eventually this alert was removed in #764. Are we sure that we want this again? A test in origin e2e might be better suited?

cc @lilic @s-urbaniak

lilic:
Agree with Simon, this was a huge pain for both our SREs and customers (info is in the Bugzilla linked by Simon). We need all alerts to be able to be resolvable by our customers, but our customers cannot control openshift operators only their own workload.

Adding an e2e test based on these metrics makes perfect sense. There was also talk about adding notifications in the console to warn users of deprecated APIs.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
sanchezl: @wking, you are correct. We weren't on 1.20 yet when I originally started this.

deads2k:
@lilic @simonpasquier  How should a customer be warned that a deprecated API, which is going to be removed is being used in their cluster and will result in a disruption for their particular workload?

Not all workloads come in OCP payloads, but any workload using the deprecated APIs will be impacted when they are removed in a following release.

sttts:
> those are more appropriate for this

@lilic technically both are feasible obviously. Sounds like we are moving into product manager or architect territory, how to best present that information to users. Are we sure that with console notifications we won't get "nobody warned me about the broken workloads after upgrade. I never look into the UI." sort of complains?

To repeat what @deads2k said between the lines: **we need this before 1.22 is upgraded to. So in 4.8 the latest.** If console team (@jhadvig) is not finished with notifications like that (metrics based notification in console), this is going to be too late.

Instead of being too late and breaking customers, I would prefer a little less ideal user experience (to be decided that it actually is) of metric based alerts in 4.8.

cc @mcurry-rh @derekwaynecarr @jhadvig

lilic:
> Are we sure that with console notifications we won't get "nobody warned me about the broken workloads after upgrade. I never look into the UI." sort of complains?

Some of our users also do not use alertmanager and never configure it, so same can be said for alerts really.

In any case, I believe we discussed offline and sort of agreed on info level alert for this is fine if there are no other alternatives. And @simonpasquier provided with a better alerti2ng rule, so that users can actually resolve this alert. And that is should be just for non openshift workloads IIRC? cc @s-urbaniak

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -135,3 +135,14 @@ spec:
...
+        for: 1h
+        labels:
+          severity: warning
```
RiRa12621, line 148: This is barely actionable and neither urgent and shouldn't be higher severity than "info"

RiRa12621 (changes requested):

s-urbaniak:
I agree that we are leveraging alerting for use cases it has not been envisioned for and am +1 on using info level only for this kind of workaround as discussed OOB.

RiRa12621:
@s-urbaniak can probably clarify but iirc all alerts are presented in the console. That should include info level.
So we'd get the same effect but no pages that are giving admins the noisy feelings.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -135,3 +135,14 @@ spec:
     - record: cluster:apiserver_current_inflight_requests:sum:max_over_time:2m
       expr: |
         max_over_time(sum(apiserver_current_inflight_requests{apiserver=~"openshift-apiserver|kube-apiserver"}) by (apiserver,requestKind)[2m:])
+  - name: pre-release-lifecycle
+    rules:
+      - alert: DeprecratedAPIUsed
+        annotations:
+          message: |
+            Deprecated API that will be removed in the next version is being used: {{"{{$labels.group}}"}}.{{"{{$labels.version}}"}}/{{"{{$labels.resource}}"}}
```
lilic, line 143: We mentioned to also say how to fix this, e.g. "Remove the workload that is using this API, otherwise <insert reason what happens>" anything that is helpful to end users, who receive this page.

Also is it not possible to get a namespace for this to display?

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -135,3 +135,14 @@ spec:
...
+        expr: |
+          group(apiserver_requested_deprecated_apis{removed_release="1.21"}) by (group,version,resource,subresource) * on(group,version,resource,subresource) group_right() rate(apiserver_request_total[10m]) > 0
+        for: 1m
```
lilic, line 146: Any reason why its 1minute?

lilic (commented):
Couple of comments.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
sanchezl: Fixed typo.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
sanchezl: Updated to:
> Deprecated API that will be removed in the next version<sup>1</sup> is being used. Removing the workload that is using the `group.version/resource` API might<sup>2</sup> be necessary for a successful upgrade to the next cluster version.
---
<sup>1</sup> by _next version_ here we mean next y-stream version. Is there a more exact, customer-facing terminology to use here to make it clear that a z-stream version would not remove an API? For example, _next minor version_, but the term _minor version_ is not well defined in the documentation.
<sup>2</sup> I say _might_ here because the deprecated API usage might be in the kube control plane itself, which of course would be updated in the next version to not use the deprecated APIs, but I can't easily filter that use in the metric/alert.

lilic (commented):
lgtm, thanks so much for addressing our comments!

@simonpasquier leaving up to Simon in case he has any more comments.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -135,3 +135,16 @@ spec:
...
+  - name: pre-release-lifecycle
+    rules:
+      - alert: DeprecratedAPIUsed
+        annotations:
+          message: >-
+            Deprecated API that will be removed in the next version is being used. Removing the workload that is using
+            the {{"{{$labels.group}}"}}.{{"{{$labels.version}}"}}/{{"{{$labels.resource}}"}} API might be necessary for
+            a successful upgrade to the next cluster version.
+        expr: |
+          group(apiserver_requested_deprecated_apis{removed_release="1.21"}) by (group,version,resource,subresource) * on(group,version,resource,subresource) group_right() rate(apiserver_request_total[10m]) > 0
```
sttts, line 147: how do we make sure this version is updated each release?

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -135,3 +135,16 @@ spec:
...
       expr: |
         max_over_time(sum(apiserver_current_inflight_requests{apiserver=~"openshift-apiserver|kube-apiserver"}) by (apiserver,requestKind)[2m:])
+  - name: pre-release-lifecycle
+    rules:
+      - alert: DeprecratedAPIUsed
+        annotations:
+          message: >-
+            Deprecated API that will be removed in the next version is being used. Removing the workload that is using
+            the {{"{{$labels.group}}"}}.{{"{{$labels.version}}"}}/{{"{{$labels.resource}}"}} API might be necessary for
+            a successful upgrade to the next cluster version.
```
sttts, line 145: Please refer to audit logs to find out the actor.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
lilic: If you want to you can either edit for the next release so in 4.8 you would replace or add an or for 1.22, or use a regex matcher here that would match against multiple versions, see -> https://prometheus.io/docs/prometheus/latest/querying/basics/#time-series-selectors

sttts: my question was more about how we remember this in practice. We could add CI test that compares the value with the running server version.

lilic: Yes that is a good idea, or something along the lines of creating one of the deprecated APIs and evaluating if the alert is firing.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
sanchezl: > Also is it not possible to get a namespace for this to display?
@lilic The namespace is not available.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
sanchezl: Updated.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
sanchezl: @sttts Added an e2e test.

simonpasquier: Suggestion to get rid of the `on(...) group_right()` part

```suggestion
          group(apiserver_requested_deprecated_apis{removed_release="1.21"}) by (group,version,resource,subresource) and (sum by(group,version,resource,subresource) (rate(apiserver_request_total[10m]))) > 0
```

You can even remove the `subresource` label from the by() clauses since it's not used in the annotations.

sanchezl: @simonpasquier Thank you. Updated.

simonpasquier:
lgtm

test/e2e/deprecated_api_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"time"
+
+	test "github.com/openshift/cluster-kube-apiserver-operator/test/library"
+	monitoringclient "github.com/prometheus-operator/prometheus-operator/pkg/client/versioned"
+	"github.com/stretchr/testify/require"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/discovery"
+)
+
+func TestDeprecatedAPIInUse(t *testing.T) {
```
deads2k, line 18: good test.  It will break every kube bump.  I find this acceptable since it won't block the kube bump.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -135,3 +135,16 @@ spec:
...
+    rules:
+      - alert: DeprecatedAPIInUse
+        annotations:
+          message: >-
+            Deprecated API that will be removed in the next version is being used. Removing the workload that is using
+            the {{"{{$labels.group}}"}}.{{"{{$labels.version}}"}}/{{"{{$labels.resource}}"}} API might be necessary for
+            a successful upgrade to the next cluster version. Refer to the audit logs to identify the workload.
+        expr: |
+          group(apiserver_requested_deprecated_apis{removed_release="1.21"}) by (group,version,resource) and (sum by(group,version,resource) (rate(apiserver_request_total[10m]))) > 0
+        for: 1h
```
deads2k, line 148: if we use 4h, will that cover the entire timeframe of an e2e test?  Is this the spot or the rate the spot?  Probably here I guess.

deads2k:
/lgtm
/hold

holding so we can decide about the exact timeframe.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
sanchezl: Followed up at https://github.com/openshift/cluster-kube-apiserver-operator/pull/1044.

test/e2e/deprecated_api_test.go
sttts: :+1: @sanchezl ignore my ask for a test that breaks elsewhere. This is it I guess.

PR #1019, opened 2020-12-10
Bug 1902059: bootstrap: always wire bound SA keys

stlaz:
/assign @tnozicka
cc @marun 

depends on https://github.com/openshift/installer/pull/4469

### DIFF
--- a/pkg/cmd/render/render.go
+++ b/pkg/cmd/render/render.go
@@ -2,7 +2,11 @@ package render
 
 import (
 	"bytes"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
 	"encoding/json"
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"io/ioutil"
@@ -107,6 +111,10 @@ func (r *renderOpts) Validate() error {
 		return errors.New("missing etcd serving CA: --manifest-etcd-serving-ca")
 	}
 
+	if err := validateBoundSATokensSigningKeys(r.generic.AssetInputDir); err != nil {
+		return err
+	}
+
 	return nil
 }
 
@@ -147,8 +155,6 @@ type TemplateData struct {
 	BindNetwork string
 
 	ServiceAccountIssuer string
-
-	UserProvidedBoundSASigningKey bool
 }
 
 // Run contains the logic of the render command.
@@ -180,11 +186,30 @@ func (r *renderOpts) Run() error {
 			}
 		}
 	}
-	if _, err := os.Stat(filepath.Join(r.generic.AssetInputDir, "bound-service-account-signing-key.key")); err == nil {
-		if _, err := os.Stat(filepath.Join(r.generic.AssetInputDir, "bound-service-account-signing-key.pub")); err == nil {
-			renderConfig.UserProvidedBoundSASigningKey = true
+
+	boundSAPublicPath := filepath.Join(r.generic.AssetInputDir, "bound-service-account-signing-key.pub")
+	boundSAPrivatePath := filepath.Join(r.generic.AssetInputDir, "bound-service-account-signing-key.key")
+	_, privStatErr := os.Stat(boundSAPrivatePath)
+	if privStatErr != nil {
+		if !os.IsNotExist(privStatErr) {
+			return fmt.Errorf("failed to access %s: %v", boundSAPrivatePath, privStatErr)
+		}
+
+		// the private key is missing => generate the keypair
+		pubPEM, privPEM, err := generateKeyPairPEM()
+		if err != nil {
+			return fmt.Errorf("failed to generate an RSA keypair for bound SA token signing: %v", err)
+		}
+
+		if err := ioutil.WriteFile(boundSAPrivatePath, privPEM, os.FileMode(0600)); err != nil {
+			return fmt.Errorf("failed to write private key for bound SA token signing: %v", err)
+		}
+
+		if err := ioutil.WriteFile(boundSAPublicPath, pubPEM, os.FileMode(0644)); err != nil {
+			return fmt.Errorf("failed to write public key for bound SA token verification: %v", err)
 		}
 	}
+
 	if len(renderConfig.ClusterCIDR) > 0 {
 		anyIPv4 := false
 		for _, cidr := range renderConfig.ClusterCIDR {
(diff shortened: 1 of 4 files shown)


stlaz:
@marun @tnozicka @abhinavdahiya I updated the PR. The lack of "Validate()" unit testing is concerning.

In my ideal world, the signing key files would have been provided by flags and validated as such.

pkg/cmd/render/render_test.go
```diff
@@ -408,75 +408,69 @@ spec:
...
-				issuerKeyFile := cfg.APIServerArguments["service-account-signing-key-file"]
-				// FIXME: Update when we wire a real default key
-				expectedIssuerKeyFile := kubecontrolplanev1.Arguments{"/etc/kubernetes/secrets/service-account.key"}
-				if !reflect.DeepEqual(issuerKeyFile, expectedIssuerKeyFile) {
-					return fmt.Errorf("expected the service-account-signing-key-file to be %q, but it was %q", expectedIssuerKeyFile, issuerKeyFile)
-				}
-				return nil
-			},
 		},
+		//
```
marun, line 419: Merging commented-out code seems like an anti-pattern to me, are you intending to address before merge or track resolution with bz etc?

marun:
> @marun @tnozicka @abhinavdahiya I updated the PR. The lack of "Validate()" unit testing is concerning.
> 
> In my ideal world, the signing key files would have been provided by flags and validated as such.

The installer is validating the keys iirc. Presumably to ensure fail-fast well before rendering?

pkg/cmd/render/render.go
```diff
@@ -180,11 +186,30 @@ func (r *renderOpts) Run() error {
...
+		}
+
+		// the private key is missing => generate the keypair
+		pubPEM, privPEM, err := generateKeyPairPEM()
+		if err != nil {
+			return fmt.Errorf("failed to generate an RSA private key for bound SA tokens singing: %v", err)
+		}
+
+		if err := ioutil.WriteFile(boundSAPrivatePath, privPEM, os.FileMode(0600)); err != nil {
+			return fmt.Errorf("failed to write private key for bound SA tokens signing: %v", err)
```
marun, line 205: nit: s/tokens/token/

pkg/cmd/render/render.go
```diff
@@ -180,11 +186,30 @@ func (r *renderOpts) Run() error {
...
+		}
+
+		if err := ioutil.WriteFile(boundSAPublicPath, pubPEM, os.FileMode(0644)); err != nil {
+			return fmt.Errorf("failed to write public key for bound SA tokens verification: %v", err)
```
marun, line 209: nit: s/tokens/token/

pkg/cmd/render/render.go
```diff
@@ -180,11 +186,30 @@ func (r *renderOpts) Run() error {
...
 		}
 	}
-	if _, err := os.Stat(filepath.Join(r.generic.AssetInputDir, "bound-service-account-signing-key.key")); err == nil {
-		if _, err := os.Stat(filepath.Join(r.generic.AssetInputDir, "bound-service-account-signing-key.pub")); err == nil {
-			renderConfig.UserProvidedBoundSASigningKey = true
+
+	boundSAPublicPath := filepath.Join(r.generic.AssetInputDir, "bound-service-account-signing-key.pub")
+	boundSAPrivatePath := filepath.Join(r.generic.AssetInputDir, "bound-service-account-signing-key.key")
+	_, privStatErr := os.Stat(boundSAPrivatePath)
+	if privStatErr != nil {
```
marun, line 193: nit: Maybe skip a level of indentation?

```golang
if os.IsNotExist(...) {
  // create and write key
} else if err != nil {
  // return error
}

File: pkg/cmd/render/render.go [hunk 5]
```diff
@@ -107,6 +111,10 @@ func (r *renderOpts) Validate() error {
 		return errors.New("missing etcd serving CA: --manifest-etcd-serving-ca")
 	}
 
+	if err := validateBoundSATokensSigningKeys(r.generic.AssetInputDir); err != nil {
```
marun, line 114: Maybe set a render opt here indicating whether generation is required?

```golang
ok, err := validateBoundSATokensSigningKeys(...)
if err != nil {
  return err
}
r.BoundTokenSigningKeysProvided = ok
```

That would ensure that validation of a provided keypair could be limited to the `Validate` function, and `Run` could just respond to the option.

pkg/cmd/render/render_test.go
```diff
@@ -218,7 +218,7 @@ func TestRenderCommand(t *testing.T) {
 		testFunction  func(cfg *kubecontrolplanev1.KubeAPIServerConfig) error
 	}{
 		{
-			name: "scenario 1 checks feature gates",
+			name: "checks feature gates",
```
marun, line 221: For my info, why is it desirable to remove scenario numbering? This seems like an unrelated non-functional change.

pkg/cmd/render/render_test.go
```diff
@@ -408,75 +408,69 @@ spec:
 			},
 		},
 		{
-			name: "scenario 8 no user provided bound-sa-signing-keys",
+			name: "no user provided bound-sa-signing-keys", // makes sure there are no errors generating the keys
```
marun, line 411: nit: Maybe update the name of the test to reflect the intent rather than adding a comment?

pkg/cmd/render/render_test.go
stlaz: You are getting the name of the test in a fail report anyway, so you have something to look up the test by. If you remove a test somewhere in the middle, or it feels only right to add a test in the middle of the array of the tests, you would have to to renumber. I do not see the value of test numbers.

pkg/cmd/render/render.go
```diff
@@ -180,11 +186,30 @@ func (r *renderOpts) Run() error {
...
+	_, privStatErr := os.Stat(boundSAPrivatePath)
+	if privStatErr != nil {
+		if !os.IsNotExist(privStatErr) {
+			return fmt.Errorf("failed to access %s: %v", boundSAPrivatePath, privStatErr)
+		}
+
+		// the private key is missing => generate the keypair
+		pubPEM, privPEM, err := generateKeyPairPEM()
+		if err != nil {
+			return fmt.Errorf("failed to generate an RSA private key for bound SA tokens singing: %v", err)
```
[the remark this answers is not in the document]
stlaz: > s/singing/signing/
I might have gotten a bit too festive back then :thinking:

pkg/cmd/render/render_test.go
stlaz: I think I wanted to make this mergable before the Holidays, but since I failed, I might just give it a try and address it before merge.

pkg/cmd/render/render.go
stlaz: I like the "something happened, what is it" approach maybe a little bit better.

pkg/cmd/render/render.go
stlaz: passing validation does not mean the keys were provided

stlaz: but I understand your point, I think I dropped the idea of validation setting the option because it should merely validate and not provide further guidance of what to do with the provided options/files

PR #1020, opened 2020-12-10
Bug 1906808: configobservation: override service-account-jwks-uri to use LB address

stlaz:
Without the override, the KAS would point to its own IP which is not
covered by the default serving certificate.

/assign @marun 
/cc @tnozicka

### DIFF
--- a/pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
+++ b/pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
@@ -5,14 +5,16 @@ import (
 	"net/url"
 	"strings"
 
-	v1 "github.com/openshift/api/config/v1"
-	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/configobservation"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/klog/v2"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/library-go/pkg/operator/configobserver"
 	"github.com/openshift/library-go/pkg/operator/events"
+
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/configobservation"
 )
 
 // ObserveServiceAccountIssuer changes apiServerArguments.service-account-issuer from
@@ -25,7 +27,7 @@ func ObserveServiceAccountIssuer(
 ) (map[string]interface{}, []error) {
 
 	listers := genericListers.(configobservation.Listers)
-	return observedConfig(existingConfig, listers.AuthConfigLister.Get)
+	return observedConfig(existingConfig, listers.AuthConfigLister.Get, listers.InfrastructureLister().Get)
 }
 
 // observedConfig returns an unstructured fragment of KubeAPIServerConfig that may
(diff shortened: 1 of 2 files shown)


pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
```diff
@@ -116,6 +129,11 @@ func unstructuredConfigForIssuer(issuer string) map[string]interface{} {
 			"service-account-issuer": []interface{}{
 				issuer,
 			},
+			"service-account-jwks-uri": []interface{}{
+				// We need to set the URL so that it points to our LB and doesn't
+				// default to KAS IP which isn't included in the serving certificate.
+				internalLBURL + "/openid/v1/jwks",
```
marun, line 135: If sts or cco were to override the default issuer, would they need to be able to set this value or at least have it be cleared?

Also, maybe an update to the [enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/bound-sa-tokens.md) is suggested?

stlaz: You're right. If I understand the logic of this observer, this would have to be moved to the condition above when no issuer is set, correct?

stlaz:
Updated to only set the jwks uri when the issuer is unset, lets' see whether it can pass CI

pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
```diff
@@ -25,19 +27,30 @@ func ObserveServiceAccountIssuer(
...
 	existingConfig map[string]interface{},
-	authConfigAccessor func(string) (*v1.Authentication, error),
+	authConfigAccessor func(string) (*configv1.Authentication, error),
+	infrastructureConfigAccessor func(string) (*configv1.Infrastructure, error),
 ) (map[string]interface{}, []error) {
 
 	issuer, errs := observedIssuer(existingConfig, authConfigAccessor)
-	config := unstructuredConfigForIssuer(issuer)
+
+	infrastructureConfig, err := infrastructureConfigAccessor("cluster")
```
marun, line 44: Maybe only perform this step when the issuer length is zero? Otherwise an error could occur trying to retrieve something that isn't needed.

stlaz:
@marun I fixed the logic to only grab the infra config when an issuer is not set. PTAL

pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
marun: What you've done looks fine, but maybe could have been accomplished with less effort?

```golang
internalURL := ""
if len(issuer) == 0 {
  ...
  internalURL = ...
}
```

pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
```diff
@@ -25,20 +27,51 @@ func ObserveServiceAccountIssuer(
...
+	return observedConfig(existingConfig, listers.AuthConfigLister.Get, listers.InfrastructureLister().Get)
 }
 
 // observedConfig returns an unstructured fragment of KubeAPIServerConfig that may
 // include an override of the default service account issuer if one was set in the
 // Authentication resource.
 func observedConfig(
 	existingConfig map[string]interface{},
-	authConfigAccessor func(string) (*v1.Authentication, error),
+	authConfigAccessor func(string) (*configv1.Authentication, error),
```
sttts, line 38: nit: accessor is a fixed term in apimachinery. Better: getAuthConfig, getInfrastructureConfig

pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
```diff
@@ -25,20 +27,51 @@ func ObserveServiceAccountIssuer(
...
+	infrastructureConfigAccessor func(string) (*configv1.Infrastructure, error),
 ) (map[string]interface{}, []error) {
 
 	issuer, errs := observedIssuer(existingConfig, authConfigAccessor)
```
sttts, line 42: what can errs be? Keep going with all the logic is a little confusing. At least doc some rational why.

pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
```diff
@@ -25,20 +27,51 @@ func ObserveServiceAccountIssuer(
...
+	}
+
+	return map[string]interface{}{
+		"apiServerArguments": map[string]interface{}{
+			"service-account-jwks-uri": []interface{}{
+				apiServerInternalURL + "/openid/v1/jwks",
+			},
+		},
+	}, errs
+
```
sttts, line 74: drop empty line

sttts:
nits

sttts:
/approve

Follow-up for nits.

PR #1032, opened 2021-01-27
Bug 1920159: kube-apiservers overstate steady-state CPU needs slightly

smarterclayton:
The kube-apiserver is one of our most dynamic scaling components,
but the majority of clusters have a base load that is fairly
predictable of around 300 millicore. In order to get more headroom
for smaller clusters, adjust the out of the box request to a more
accurate default. All of the sidecars were measured in heavily
loaded production clusters as taking less than 5 millicore each,
and the core kube-apiserver is tweaked slightly.  CPU is burstable
so the apiserver will grow as necessary into slack capacity, while
still yielding to etcd on saturated masters.

/hold

Testing multiple changes together

### DIFF
--- a/bindata/v4.1.0/kube-apiserver/pod.yaml
+++ b/bindata/v4.1.0/kube-apiserver/pod.yaml
@@ -58,7 +58,7 @@ spec:
     resources:
       requests:
         memory: 1Gi
-        cpu: 300m
+        cpu: 265m
     ports:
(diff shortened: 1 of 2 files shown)


smarterclayton:
/retest

Not related to resource reduction.

/hold cancel

Combined tests looked ok.  Team should review and see the comments in the bug.

PR #1034, opened 2021-02-01
Remove tnozicka and add soltysh to OWNERS

soltysh:
/assign @mfojtik


soltysh:
Adding valid-label tag since this is OWNERS change and not code change.

PR #1038, opened 2021-02-05
add rolebinding for delegated in-cluster authentication

deads2k:
Add a rolebinding to gain access to the configmap which holds CA bundles for client authentication.  This should alllow delegated-authentication to terminate mTLS even if they cannot run subjectaccessreviews.

@sttts @stlaz @s-urbaniak Please inspect carefully to ensure the configmap is correct and matches the content expected.


stlaz:
/lgtm
the cm appears to only contain the CAs singing client certs for various components + HTTP header configuration for requestheader authn, I don't think there's a reason to be afraid to make these public.

s-urbaniak:
lgtm as well, but might be a bit confusing for operator/proxy writers to inspect a configmap meant for api server aggregation. i.e. it contains the request header CA as well.

I peeked in k8s code, but didn't another place where the client-ca is persisted either so i think we're good here :+1:

deads2k:
requires https://github.com/openshift/origin/pull/25881

PR #1040, opened 2021-02-09
Bug 1926465: bootstrap: add --advertise-address=pod.status.hostIP

sttts:


aojea:
the jobs fail trying to access the apiserver endpoint, it seems related to the chane :thinking:

aojea:
@sttts seems the variable is not being expanded. from the logs here https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-kube-apiserver-operator/1040/pull-ci-openshift-cluster-kube-apiserver-operator-master-e2e-aws/1359135276831608832/artifacts/e2e-aws/ipi-install-install/artifacts/

> Error: invalid argument "${HOST_IP}" for "--advertise-address" flag: failed to parse IP: "${HOST_IP}"

bindata/bootkube/bootstrap-manifests/kube-apiserver-pod.yaml
```diff
@@ -39,6 +39,7 @@ spec:
     - --alsologtostderr
     - --v=2
     - --log-file=/var/log/bootstrap-control-plane/kube-apiserver.log
+    - --advertise-address=${HOST_IP}
```
aojea, line 42: it seems this is not expanded, if is terraform it seems we have to use a double $$ https://github.com/hashicorp/terraform/issues/1942#issuecomment-251298610

sttts: this is shell syntax, expanded dynamically.

aojea: yeah, I think so, but these logs shows that is not expanding https://github.com/openshift/cluster-kube-apiserver-operator/pull/1040#issuecomment-776008908 so I assumed that the file may be preprocessed, but I´m not very familiar with this process :smile:

sttts: kubelet is not passing the command through bash. That's why it is not expanded. I pushed another version that wraps the call with bash, similarly as in the non-bootstrap kube-apiserverr command.

aojea:
this time went through
/lgtm

PR #1046, opened 2021-02-15
Speed up port reuse via SO_REUSEADDR

sttts:


p0lyn0mial:
/lgtm

setting `SO_REUSEPORT` is not required. Setting `SO_REUSEADDR` on the new socket is enough and should work.

PR #1049, opened 2021-02-16
render initial SCC in addition to having the CVO reconcile

deads2k:
By rendering the SCCs and their clusterroles in addition to having the CVO create them, we can get them created during the bootkube phase, possibly even before the CVO starts running and applying its manifests.  This may tighten up the race between creation and usage further.


deads2k:
Found a theory for why bootstrapping never finishes

We messed with discovery to serve security.openshift.io discovery from the openshift-apiserver instead of the kube-apiserver.  this means the bootstrap node cannot reply to the discovery call.  this means bootkube never finishes.

@sttts I think I want to serve a hardcoded security.openshift.io

deads2k:
this relies on https://github.com/openshift/cluster-bootstrap/pull/55

They tested well together using clusterbot

deads2k:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-kube-apiserver-operator/1049/pull-ci-openshift-cluster-kube-apiserver-operator-master-e2e-aws/1362840467980423168 failed on kube-apiserver graceful termination, but we bootstrapped fine and all SCC failures related to missing restricted SCC were gone

/override ci/prow/e2e-aws

deads2k:
/override ci/prow/e2e-upgrade

This change doesn't impact apiserver availability and fixes our SCC issues

deads2k:
/override ci/prow/e2e-aws-operator

these are local tests. the bootstrapping is fine, these take a long time and have been flaky.  Let's get that SCC fixed.

PR #1050, opened 2021-02-17
config api-audiences along with SA issuer

stlaz:
I noticed that the API audiences were only being configured in the bootstrap phase to match the SA issuer, but during normal runtime, they would be rolled back to the default value.

Make the audiences always match the SA issuer.

/assign @marun

### DIFF
--- a/pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
+++ b/pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
@@ -89,13 +89,17 @@ func observedConfig(
 				"service-account-issuer": []interface{}{
 					newIssuer,
 				},
+				"api-audiences": []interface{}{
+					newIssuer,
+				},
 			},
 		}, errs
 	}
 
-	// if the issuer is not set, rely on the config-overrides.yaml to set it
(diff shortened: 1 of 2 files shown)


marun:
/hold

afaik bootstrap does need to set the audience to the custom issuer (if provided). Post-bootstrap similarly needs to synchronize the audience to the issuer (vs setting a static value), and investigation is suggested to ensure that is occurring.

marun:
/retest
/lgtm

@stlaz Maybe update the PR title and description? It doesn't seem to match the code.

deads2k:
based on my read, the default here was reasonable: 

```
      --api-audiences strings                             Identifiers of the API. The service account token authenticator will validate that tokens used against the API are bound to at least one of these audiences. If the --service-account-issuer flag is configured and this flag is not, this field defaults to a single element list containing the issuer URL.

```

but this looks more explicit

marun:
@stlaz The fact that the operator job has _never_ passed against this PR suggests it's not a flake.

/hold

marun:
/hold cancel

/retest

aws-operator is super flakey

PR #1052, opened 2021-02-18
make KAS-o upgradable even when webhook authenticators are set

stlaz:
/assign @deads2k 

/hold
until https://github.com/openshift/cluster-authentication-operator/pull/304 is ready


stlaz:
/retest
Graceful termination appears to be broken which is why the e2e-aws operator test breaks now (there's a fetch on clusteroperator during KAS rollout).

Let's hope it's going to magically get right during one of the reruns?

PR #1056, opened 2021-03-04
WIP: try exec probes for probes

rphillips:


rphillips:
hmm. this gets us a different error: 'kube-apiserver terminates within graceful termination period' fails

rphillips:
500 from the healthz endpoint `url: (22) The requested URL returned error: 500`

PR #1058, opened 2021-03-10
Switch to new image registry

marun:
s/registry.svc.ci.openshift.org/registry.ci.openshift.org/

/cc @sttts


marun:
/hold

Checking with ART to see if the change would be best be automated.

PR #1060, opened 2021-03-11
bug 1937916: add a flowschema to ensure that probes never get 429s

deads2k:
This addresses a CI failure that caused kube-apiservers to ripple fail

`Liveness probe failed: HTTP probe failed with statuscode: 429`


tkashem:
(we are adding a flowschema in the apf bootstrap configuration that exempts probes, so holding it for now)
 
https://github.com/kubernetes/kubernetes/pull/100678

/hold

PR #1062, opened 2021-03-14
Revert "bootstrap: add --advertise-address=pod.status.hostIP"

osherdp:
This reverts commit b713e9f553379558d9db34837079f6b190bd7b1a.


sttts:
Why?

osherdp:
I'm testing the behavior before this PR
It seems to cause the problems we have on assisted, so I want to confirm that (it's not for pushing)
/hold

osherdp:
Anyway, we probably figured out the problem now
Closing

PR #1065, opened 2021-03-15
bug 1939226: make kube-apiserver readiness checks, check readyz

deads2k:
readyz has different semantics than healthz

/assign @sttts


ehashman (commented):
LGTM but will leave the / for @sttts

sttts:
Different question: why do we have a readiness probe at all?

sttts:
Ic, this is for monitoring service? Make sense.

/retest
/approve

PR #1066, opened 2021-03-15
bug 1939227: make liveness hit livez

deads2k:
livez didn't exist when we started.  Now it does.  Update.

We may want a startup probe distinct from livez now. 

/assign @sttts


ehashman (commented):
LGTM, will leave / for @sttts

soltysh:
/override ci/prow/e2e-aws-operator
this test it known to be broken and it's pointless to re-test it, this is being actively worked on

PR #1069, opened 2021-03-16
[WIP] (test) enable audit logging of all probes

tkashem:


tkashem:
(for test purpose only)

cc @sttts @deads2k 

/hold

bindata/v4.1.0/kube-apiserver/audit-policies-cm.yaml
```diff
@@ -9,9 +9,6 @@ data:
     kind: Policy
     metadata:
       name: Default
-    # Don't generate audit events for all requests in RequestReceived stage.
-    omitStages:
-    - "RequestReceived"
```
deads2k, line 14: we want this omit after the probes, right?

tkashem:
oops, I forgot to run `update bindata`

bindata/v4.1.0/kube-apiserver/audit-policies-cm.yaml
tkashem: I thought we take the union of the `omitStages`

tkashem: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/audit/policy/checker.go#L38-L57

PR #1071, opened 2021-03-18
Switch p&f to v1beta1

soltysh:
/assign @sttts @tkashem


soltysh:
/close
the timeout doesn't fix the problem with tests

PR #1073, opened 2021-03-22
bug 1951705: add master CPU utilization alerts

deads2k:
This will give us warning before an individual kube-apiserver fails under load and if the cluster cannot sustain the load if a single member fails.

/assign @mfojtik


deads2k:
/hold

want to try to find resourceapply version of this rule

Also want to fix up the warning.

manifests/0000_90-kube-apiserver-operator_06_alerts-cpu.yaml
```diff
@@ -0,0 +1,51 @@
...
+spec:
+  groups:
+    - name: master-cpu-utilization
+      rules:
+        - alert: HighOverallMasterCPU
+          summary: >
+            CPU utilization across all three masters is higher than two masters can sustain; a single master outage may
+            cause a cascading failure; increase available CPU.
+          message: >
+            Given three masters, the overall CPU utilization may only only be about 2/3 of all available capacity.
```
sttts, line 19: double only

manifests/0000_90-kube-apiserver-operator_06_alerts-cpu.yaml
```diff
@@ -0,0 +1,51 @@
...
+            fail when they take the load.
+            To fix this, increase the CPU and memory on your masters.
+          expr: |
+            sum(
+              100 - (avg by (instance) (rate(node_cpu_seconds_total{mode="idle"}[5m])) * 100)
+              AND on (instance) label_replace( kube_node_role{role="master"}, "instance", "$1", "node", "(.+)" )
+            )
+            /
+            count(kube_node_role{role="master"})
+            > 60
```
sttts, line 31: who can review this? @s-urbaniak @tkashem @sanchezl ?

manifests/0000_90-kube-apiserver-operator_06_alerts-cpu.yaml
```diff
@@ -0,0 +1,51 @@
...
+metadata:
+  name: cpu-utilization
+  namespace: openshift-kube-apiserver
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
+spec:
+  groups:
+    - name: master-cpu-utilization
```
sttts, line 12: isn't the word master replaced by something else?

manifests/0000_90-kube-apiserver-operator_06_alerts-cpu.yaml
```diff
@@ -0,0 +1,51 @@
...
+            To fix this, increase the CPU and memory on your masters.
+          expr: |
+            sum(
+              100 - (avg by (instance) (rate(node_cpu_seconds_total{mode="idle"}[5m])) * 100)
+              AND on (instance) label_replace( kube_node_role{role="master"}, "instance", "$1", "node", "(.+)" )
+            )
+            /
+            count(kube_node_role{role="master"})
+            > 60
+          for: 10m
```
tkashem, line 32: - during a kube-apiserver roll out, it is natural for one instance to get more cpu usage, the alert may fire as a `warning` and on a somewhat busy cluster and that is fine, I assume.
- (from the incident we have seen) the cpu spikes briefly and the kube-apiserver crashes, so `5m` average may smooth out these spikes, is it feasible to do `1m` average to take into account those brief spikes. may be someone with monitoring expertise can answer this question.

this looks good to me

manifests/0000_90-kube-apiserver-operator_06_alerts-cpu.yaml
```diff
@@ -0,0 +1,51 @@
...
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
+spec:
+  groups:
+    - name: master-cpu-utilization
+      rules:
+        - alert: HighOverallMasterCPU
+          summary: >
+            CPU utilization across all three masters is higher than two masters can sustain; a single master outage may
+            cause a cascading failure; increase available CPU.
+          message: >
```
dofinn, line 18: Would this message be more suited for a runbook associated with the alert?

deads2k:
updated with library-go.  hopefully someone is collecting these in must-gatehr

pkg/operator/starter.go
```diff
@@ -60,6 +61,10 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 	if err != nil {
 		return err
 	}
+	dynamicClient, err := dynamic.NewForConfig(controllerContext.ProtoKubeConfig)
```
tkashem, line 64: I thought that the CRD does not work with the proto client

tkashem (commented):
/lgtm

deads2k:
/hold

ha.  Fun.  this is firing during our CI runs, we'll need to suppress it.

bindata/v4.1.0/alerts/cpu-utilization.yaml
```diff
@@ -0,0 +1,51 @@
...
+spec:
+  groups:
+    - name: control-plane-cpu-utilization
+      rules:
+        - alert: HighOverallControlPlaneCPU
+          summary: >
+            CPU utilization across all three control plane nodes is higher than two control plane nodes can sustain; a single control plane node outage may
+            cause a cascading failure; increase available CPU.
+          message: >
+            Given three control plane nodes, the overall CPU utilization may only only be about 2/3 of all available capacity.
```
ihcsim, line 19: s/only only/only

bindata/v4.1.0/alerts/cpu-utilization.yaml
```diff
@@ -0,0 +1,51 @@
...
+  namespace: openshift-kube-apiserver
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
+spec:
+  groups:
+    - name: control-plane-cpu-utilization
+      rules:
+        - alert: HighOverallControlPlaneCPU
```
lilic, line 14: I understand where you are coming from but a couple of things:

With CPU being a compressible resource it's very hard to write an alert that will actually be correct, as we see in CI, this alert fires but clearly everything went well as CI passed otherwise no? 

If there is CPU pressure it will just slow things down not stop working entirely? If yes inn which case a latency alert should fire for each component on the control plane.

What I would suggest instead is to write alerts for each control plane component and alert on that so as suggested latency alerts, error alerts, etc. For example, you could write alerts like:
* scheduling latency
* deployment controller reconcile loops taking long
* apiserver latency

This is very much [cause vs symptoms based alerting.](https://prometheus.io/docs/practices/alerting/#what-to-alert-on) The control plane alerts (symptoms) suggested above will also be able to catch other causes, not just high CPU utilization.

This might be a good read https://www.robustperception.io/when-to-alert-with-prometheus

deads2k: We see cases of clusters that are running along just fine.  All latencies are reasonable.  All external signs show the cluster is perfectly healthy.  But they are time bombs.  They're using say 80% of the CPU on their system.  When one shuts down, we get a cascading failure.

We want to provide a warning before the issue becomes catastrophic, but no "this loop is taking too long" check will ever come back positive.  Such a warning can tell you that you're in the middle of a crisis, but provides no value in terms of predicting that you're going to have a crisis.

bindata/v4.1.0/alerts/cpu-utilization.yaml
```diff
@@ -0,0 +1,51 @@
...
+            fail when they take the load.
+            To fix this, increase the CPU and memory on your control plane nodes.
+          expr: |
+            sum(
+              100 - (avg by (instance) (rate(node_cpu_seconds_total{mode="idle"}[5m])) * 100)
+              AND on (instance) label_replace( kube_node_role{role="master"}, "instance", "$1", "node", "(.+)" )
+            )
+            /
+            count(kube_node_role{role="master"})
+            > 60
```
sttts, line 31: I don't understand the formular. @lilic does it make sense?

Why 60 and not 66.6... ?

do we have an idea of the number on existing customer clusters?

sttts: In other words, will 80% of customer clusters have the alert after we ship this?

deads2k: > In other words, will 80% of customer clusters have the alert after we ship this?

we don't know.  But it's a warning.

deads2k:
verbal discussion with lili indicated the choices were acceptable as a warning and the query was well-formed.

sttts:
CPU warning sanity-checked by @lilic.

Otherwise
/lgtm

sttts:
/hold

for https://github.com/openshift/cluster-kube-apiserver-operator/pull/1114 including this.

PR #1074, opened 2021-03-22
[WIP] removal of audit pkg and use lib-go

EmilyM1:
This will be to remove the static files in the /audit package and instead have this operator use library-go and the audit package there conforming to the other operators

Doubling as a Proof PR for https://github.com/openshift/library-go/pull/1025


go.mod
```diff
@@ -36,3 +34,5 @@ require (
 	k8s.io/utils v0.0.0-20201110183641-67b214c5f920
 	sigs.k8s.io/kube-storage-version-migrator v0.0.3
 )
+
+replace github.com/openshift/library-go v0.0.0-20210318152630-323ad8a8f7d8 => /Users/emoss/library-go
```
tkashem, line 38: maybe go mod silently ignored the `replace` directive, I don't see `audit.go` in the library-go changes. the `replace` maybe malformed.
I think it should be 
```
replace github.com/openshift/library-go => /Users/emoss/library-go
```

EmilyM1: The replace is ok I think. It is not passing make build.

EmilyM1: @tkashem and @sttts after correcting the build errors I see the audit package now

`go: found github.com/openshift/library-go/pkg/operator/apiserver/audit in github.com/openshift/library-go v0.0.0-20210318152630-323ad8a8f7d8`

stlaz: How can the replace be ok? How is the build system supposed to know what's in `/Users/emoss/library-go`?

stlaz (changes requested):

pkg/cmd/render/render.go
```diff
@@ -251,7 +253,7 @@ func (r *renderOpts) Run() error {
 }
 
 func getDefaultConfigWithAuditPolicy() ([]byte, error) {
-	defaultPolicy, err := audit.DefaultPolicy()
+	defaultPolicy, err := libgoassets.DefaultPolicy("kube-apiserver-audit-policies", "openshift-kube-apiserver")
```
sttts, line 256: why do we care about namespace and configmap name?

The default policy (still) depends on whether we have deleted old non-sha256 oauthaccesstokens (safe vs. non-safe policy). I think we ignored this here because kube-apiserver does not log oauthaccesstokens. But for a general `DefaultPolicy` functions we need it.

go.mod
stlaz: Please look at `go.mod` in https://github.com/openshift/apiserver-library-go/pull/46 to see how it's done. You need to push your changes somewhere so that they are publicly accessible.

EmilyM1: @stlaz this is only the proof PR. 
For the real PR the changes will come from library-go master. As this was a local change only to library-go which is not pushed, that is why the option was to use local dir.

pkg/cmd/render/render.go
EmilyM1: Should:
- the ns and name parameter be removed and
- a conditional to pick secure or unsecure be added?

EmilyM1: I have renamed the DefaultPolicy function so that folks can tell the difference:
https://github.com/openshift/library-go/pull/1025/files#diff-e32a8349971df956b0a9ccb46a34e8a54f965b5d04e9a16a8a5d2d0d7dc005b6R26

PR #1075, opened 2021-03-24
PROOF? e2e: ignore all errors while waiting for clusteroperator conditions

stlaz:
It seems that, when used on a KAS deployment rollout, client calls
might get `connection refused` errors. This could possibly be caused
by failed graceful termination. Ignore these errors to unblock CI.


stlaz:
/close
I broke it

PR #1081, opened 2021-03-30
avoid creating revisions for oauth token timeouts we don't use

deads2k:
// TODO in 4.9, this is removed because the webhook is always available
// We determine readiness by...
// 1. if the webhook is not set, it is not ready.
// 2. if webhook is set and does not point to secret/openshift-oauth-apiserver, the webhook authenticator configuration is ready
// 3. if webhook is set and points to secret/openshift-oauth-apiserver and the existing observedConfig shows a webhook configuration
//    the webhook authenticator configuration is ready.  This allows stickiness after being set once.
// 4. if webhook is set and points to secret/openshift-oauth-apiserver and the authentication operator is at the same level as the
//    kube-apiserver, the webhook authenticator configuration is ready.  This ensures, the oauth-apiserver is ready to
//    answer webhook requests
// 5. if webhook is set and points to secret/openshift-oauth-apiserver and the clusterversion indicates this is an initial
//    install, the webhook authenticator is ready.  This avoids a late stage revision during install because some people care.
// overall, it is better to return true than false if the choice is ambiguous

In reading the kube-apiserver code, it appears that as soon as we specific a webhook configuration, we no longer honor the oauth token inactivity timeout.  I'm interested in whether a gap exists.

1. install 4.7
2. start upgrade to 4.8
3. kube-apiserver finishes upgrade, but doesn't use webhook auth because auth operator has not set auth.spec.webhookconfig
4. auth operator sets the auth.spec.webhookconfig *before* the oauth-apiserver has been updated
5. kube-apiserver rolls out the configuration change for webhook configuration, but the webhook consistently fails on that master until the oauth-apiserver is updated.

### DIFF
--- a/pkg/operator/configobservation/auth/webhook_authenticator.go
+++ b/pkg/operator/configobservation/auth/webhook_authenticator.go
@@ -3,6 +3,8 @@ package auth
 import (
 	"fmt"
 
+	"github.com/openshift/library-go/pkg/operator/status"
+
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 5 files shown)


pkg/operator/configobservation/auth/webhook_authenticator.go
```diff
@@ -45,8 +123,21 @@ func ObserveWebhookTokenAuthenticator(genericListers configobserver.Listers, rec
...
+		// if we aren't yet ready to use the webhook configuration, clear the webhook configuration
+		// don't sync anything and remove whatever we synced
+		resourceSyncer.SyncSecret(
+			resourcesynccontroller.ResourceLocation{Namespace: operatorclient.TargetNamespace, Name: "webhook-authenticator"},
+			resourcesynccontroller.ResourceLocation{Namespace: "", Name: ""},
+		)
+		return map[string]interface{}{}, nil
+	}
+
+	existingWebhookConfigured := len(existingWebhookAuthenticator) > 0
```
stlaz, line 140: I'd rather see this handled on the CAO side where the authenticators wouldn't be set until oauth-apiserver is ready

pkg/operator/configobservation/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,141 @@
...
+const (
+	defaultAccessTokenMaxAgeSeconds   = float64(86400) // a day
+	fieldAccessTokenMaxAgeSeconds     = "accessTokenMaxAgeSeconds"
+	fieldAccessTokenInactivityTimeout = "accessTokenInactivityTimeout"
+)
+
+// Standa: why does nothing call this?
+// ObserveAccessTokenMaxAgeSeconds returns an unstructured fragment of KubeAPIServerConfig that changes the default value for access token max age,
+// if there is a valid value for it in OAuth cluster config.
+func ObserveAccessTokenMaxAgeSeconds(genericlisters configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, errs []error) {
```
stlaz, line 32: this is unused

stlaz (changes requested):
I'd rather for the CAO webhook config controller to not set the webhook config until oauth-apiserver is ready.

pkg/operator/configobservation/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,141 @@
...
+	OAuthLister() configlistersv1.OAuthLister
+}
+
+const (
+	defaultAccessTokenMaxAgeSeconds   = float64(86400) // a day
+	fieldAccessTokenMaxAgeSeconds     = "accessTokenMaxAgeSeconds"
+	fieldAccessTokenInactivityTimeout = "accessTokenInactivityTimeout"
+)
+
+// Standa: why does nothing call this?
```
stlaz, line 29: Did not notice the question originally: the expiry of the token is set only during its creation in the oauth-server and it cannot be updated later, the authenticator simply checks whether `time.Now()` is before `tokenCreation + expiry`

stlaz:
/close
this is now handled in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1109

PR #1082, opened 2021-03-31
audit: keep 3 days and avoid write conflict

sttts:
- don't limit rotated files to 10, but keep 3 days. This is deterministic.
- rotate old audit.log on (re)start of container.


deads2k:
/hold

This needs to have a deterministic max-size per host.  Anything under 50GB or so is fine with me, but we need that deterministic size.  I've seen clusters producing a gig of audit logs an hour pretty easily.

Three days or xGB, whichever comes sooner.  I'm open to most values of x, but I see the size limit as a blocker for merge.

PR #1084, opened 2021-03-31
[WIP] Remove audit pkg to use library-go

EmilyM1:
This is the removal of pkg/operator/audit internal to the operator so that it can read the audit pkg from lib-go.


stlaz:
how is this PR different from https://github.com/openshift/cluster-kube-apiserver-operator/pull/1074?

tkashem:
@EmilyM1 looks like you need to update dependency, `make update-vendor` i think.

PR #1094, opened 2021-04-08
Bug 1949589: Add APIRemovedInNextEUSReleaseInUse Alert

sanchezl:


manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -148,3 +148,14 @@ spec:
...
         labels:
           severity: info
+      - alert: APIRemovedInNextEUSReleaseInUse
+        annotations:
+          message: >-
+            Deprecated API that will be removed in the next EUS version is being used. Removing the workload that is using
+            the {{"{{$labels.group}}"}}.{{"{{$labels.version}}"}}/{{"{{$labels.resource}}"}} API might be necessary for
+            a successful upgrade to the next EUS cluster version. Refer to the audit logs to identify the workload.
+        expr: |
+          group(apiserver_requested_deprecated_apis{removed_release="1.23"}) by (group,version,resource) and (sum by(group,version,resource) (rate(apiserver_request_total[4h]))) > 0
```
deads2k, line 158: you actually need everything removed from 1.19 to 1.23, right?

sanchezl: Updated

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -148,3 +148,14 @@ spec:
         for: 1h
         labels:
           severity: info
+      - alert: APIRemovedInNextEUSReleaseInUse
+        annotations:
+          message: >-
+            Deprecated API that will be removed in the next EUS version is being used. Removing the workload that is using
+            the {{"{{$labels.group}}"}}.{{"{{$labels.version}}"}}/{{"{{$labels.resource}}"}} API might be necessary for
+            a successful upgrade to the next EUS cluster version. Refer to the audit logs to identify the workload.
```
sttts, line 156: this will become a reference to the CRD, right?

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -148,3 +148,14 @@ spec:
...
+        expr: |
+          group(apiserver_requested_deprecated_apis{removed_release=~"1\\.2[123]"}) by (group,version,resource) and (sum by(group,version,resource) (rate(apiserver_request_total[4h]))) > 0
```
sttts, line 158: so there is nothing about being on the EUS stream here, right? So if somebody uses an API in 4.8 that still exists in 4.9, but goes away in 4.10, this alert will fire. What was the plan to avoid that? Or is there a plan at all?

deads2k: It appears that informational alert is intended to trigger on that.

deads2k:
replaced by https://github.com/openshift/cluster-kube-apiserver-operator/pull/1114

PR #1098, opened 2021-04-09
Bug 1948431: Bump openshift/api to get the necessary CSIMigration feature gates

huffmanca:
The CSI Migration tech preview feature is included in `openshift/api`, and I'm bumping the version here so that the FeatureGate can be propagated correctly.

Goal of this PR is to apply `--feature-gate=CSIMigrationAWS=true,CSIMigrationOpenStack=true` to kube-apiserver when `TechPreviewNoUpgrade featureSet` is enabled, which was introduced in openshift/api#880 .


jsafrane:
/retitle Bug 1948431: Bump openshift/api to get the necessary CSIMigration feature gates

PR #1099, opened 2021-04-12
Bug 1956606: Revert "Merge pull request #1060" (Use the flow schema introduced in upstream)

tkashem:
(Revert "Merge pull request #1060)

Use the flow schema introduced in upstream that exempts probes.
https://github.com/openshift/kubernetes/pull/656


wking:
Current manifest is a no-op which isn't pushed into the cluster because it declares no cluster profiles.  I've created a bug so we can drop it.

/retitle Bug 1956606: Revert "Merge pull request #1060" (Use the flow schema introduced in upstream)

tkashem:
we will see if the upstream PR for p&f auto-update merges in time, otherwise we will need to use cvo manifest for exempting health probes.

/hold

tkashem:
upstream PR picked

/hold cancel

PR #1109, opened 2021-04-19
Bug 1951637: don't roll a new revision on tokentimeout changes

stlaz:
New revisions on access token timeout changes are no longer necessary
if webhook authenticators are configured since the token timeout
would configuration would no longer have any effect.

/assign @deads2k


stlaz:
/retitle Bug 1951637: don't roll a new revision on tokentimeout changes

PR #1112, opened 2021-04-20
[POC] show the spot to fix up our static pod

deads2k:
/hold

This is just to demonstrate the size of the change we're talking about that ensure our static pod operators continue to function as designed.


deads2k:
in theory, kubelet admission mutates before writing to the API, so this may not necessary.

PR #1113, opened 2021-04-20
WIP Enable graceful rollout for single replica control plane

marun:
**NOTE** This PR depends on https://github.com/openshift/library-go/pull/1054

Preliminary work to enable graceful rollout for single replica control plane.

TODO
 - [ ] Enable log rotation
 - [ ] Ensure internal traffic is directed to the api server. Currently hardcoded. Possible solutions: 
        -  Have installer pod configure the graceful monitor with the node ip and cluster ips for `default/kubernetes` and `openshift-kube-apiserver/apiserver`
        - Fix kube-apiserver endpoint controller to properly set the endpoint port
        - Use a reverse proxy e.g. [caddy](https://caddyserver.com/docs/quick-starts/reverse-proxy)
 - [ ] Switch to checking pod health via the installer pod so that the monitor pod's only responsibility is ensuring a static configuration
 - [ ] e2e testing

/cc @sttts

### DIFF
--- a/Dockerfile-origin-release
+++ b/Dockerfile-origin-release
@@ -1,4 +1,4 @@
-FROM registry.svc.ci.openshift.org/openshift/origin-release:v4.0 as origin-release
+FROM registry.ci.openshift.org/openshift/origin-release:v4.0 as origin-release
 
 FROM centos:7 as jq
 ARG IMAGES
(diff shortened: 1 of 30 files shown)


marun:
/retest

I do so love it when a PR that is effectively no-op for non-SNO jobs ends up failing non-SNO jobs.

marun:
Not a viable approach

PR #1114, opened 2021-04-22
Bug 1949589: Add APIRemovedInNextEUSReleaseInUse Alert

deads2k:
updated version of https://github.com/openshift/cluster-kube-apiserver-operator/pull/1094 to reflect new API and use static-resource-controller instead of CVO

builds on https://github.com/openshift/cluster-kube-apiserver-operator/pull/1073 to avoid conflicts


bindata/v4.1.0/alerts/kube-apiserver-requests.yaml
```diff
@@ -0,0 +1,21 @@
...
+metadata:
+  name: kube-apiserver-requests
+  namespace: openshift-kube-apiserver
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
+spec:
+  groups:
+    - name: apiserver-requests-in-flight
```
sttts, line 12: where does this come from?

bindata/v4.1.0/alerts/api-usage.yaml
```diff
@@ -0,0 +1,37 @@
...
+metadata:
+  name: api-usage
+  namespace: openshift-kube-apiserver
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
+spec:
+  groups:
+    - name: pre-release-lifecycle
```
sttts, line 12: this move will break the e2e test checking the version in the expression

sttts:
I disagree with "use static-resource-controller instead of CVO". We shouldn't do this. It creates complexity.

bindata/v4.1.0/alerts/api-usage.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: api-usage
+  namespace: openshift-kube-apiserver
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
```
sttts, line 9: those don't make sense anymore

deads2k:
> I disagree with "use static-resource-controller instead of CVO". We shouldn't do this. It creates complexity.

our operator
1. reacts on a watch
2. produces events indicating status
3. sets operator status for the failing manifest
4. does a more intelligent diff
5. uses a lister cache instead of a GET

The CVO faces a watch fanout problem because of the namespace breadth.  In addition, it can never know the operator status itself.

The CVO is essential for getting the operator available and monitorable. Things that are inputs to the operator must be created by the CVO. But getting the operand functional is the responsibility of the operator. These files are for the operand, not the operator.

deads2k:
/hold
Need to figure out why the metrics skip didn't work

deads2k:
merged in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1115

PR #1115, opened 2021-04-22
bug 1949306: stop created deprecateapirequests

deads2k:
cleans up the CRD we never used.


deads2k:
/hold

holding to allow https://github.com/openshift/cluster-kube-apiserver-operator/pull/1114 to merge first.

deads2k:
/hold

Need to figure out why the skip doesn't work

PR #1118, opened 2021-04-27
WIP - Add etcd object counts alert

ihcsim:
This PR adds some new etcd alerts to warn the users on high and increasing
object counts in etcd. The thresholds used in these alerts are derived from a
load test performed on a 4.8.0-fc.0 cluster on GCP. The test observed the
effect of creating ~120K imagestream objects on etcd peer network latency,
leader's health, disk performance, failed peer proposals, and network bytes
sent/received.

The YAML of the imagestream used for testing can be found [here](https://gist.github.com/ihcsim/a9e612a91b4757d163a1de8f3dc510f2). It was 
copied from the OOB `ruby` imagestream.

JIRA: [API-1220](https://issues.redhat.com/browse/API-1220).

Pending:

* Is it worth repeating the same tests on Azure, which is known to experience etcd disk performance latency more often?
* Alerts description usually provides links to incident runbook. Assuming we don't have one atm, are we ok with more verbose alert description?
* This PR depends on Go code changes in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1073

<details>
<summary>Here are the results of my test run:</summary>

The number of objects increased from 20k to 120k over a 12-hour period:
image

The rate of increase of the object counts grouped by resources, seems to be
dominated by `Event`s, follow by `ImageStream`s. I am not sure if there is a
way to increase this rate, as it seems a more aggressive increase rate will
trigger the kind of errors some users saw. FWIW, replacing `rate` with 
`increase` yields an identical trend:
image

No issues with leader election:
image

image

No major degradation in disk performance with lowest value being 
6.25ms and peaked at 9.5ms:
image

image

Other than a 20-minute blip in network round trip latency (from 0.03 to 
0.06 seconds), everything looks normal: 
image

No network packets loss:
image

To get a better sense of the traffic quality between the peers, I also 
captured some `proposal_*` metrics. No prolonged failed and pending 
proposals communication in this test run:
image

image
</details>

### DIFF
--- a/bindata/v4.1.0/alerts/etcd-object-counts.yaml
+++ b/bindata/v4.1.0/alerts/etcd-object-counts.yaml
@@ -3,39 +3,40 @@ kind: PrometheusRule
 metadata:
   name: etcd-object-counts
   namespace: openshift-kube-apiserver
-  annotations:
-    include.release.openshift.io/self-managed-high-availability: "true"
-    include.release.openshift.io/single-node-developer: "true"
-    exclude.release.openshift.io/internal-openshift-hosted: "true"
 spec:
   groups:
     - name: etcd-object-counts
       rules:
-        - alert: etcdHighObjectCounts
+        - alert: etcdHighCRDObjectCounts
           annotations:
-            summary: "etcd recorded more than 100000 objects. Latest count: {{ $value }}"
+            summary: etcd recorded {{ $value }} {{ $labels.resource }} objects.
             description: |
-              A high object count of more than 100000 objects may lead to stability issue in the API server and etcd storage stack. If this value continues to increase, it may degrade the responsiveness of the API server and etcd.
-              To fix this, use the following query to determine the top-5 highest resource kinds:
-                  topk(5, sum(etcd_object_counts) by (apiserver,namespace,resource))
-              Depending on the resource kinds, additional metrics can be used to help identify the actual resources. For example,
-                  count(kube_secret_info) by (namespace)
-              shows the number of secrets grouped by namespaces.
-              Additional resource metrics can be found at https://github.com/kubernetes/kube-state-metrics/tree/master/docs
+              etcd recorded {{ $value }} {{ $labels.resource }} objects. A high number of CRD objects is known to have caused stability issues in the API server and etcd storage stack. Information on diagnosis and mitigation steps can be found in the runbook.
+            runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/etcdHighObjectCounts.md
           expr: |
-            sum(etcd_object_counts) > 100000
+            cluster:usage:resources:sum{resource!~".*k8s.io|^[^.]+"} > 200
           for: 10m
           labels:
             severity: warning
-        - alert: etcdIncreasingObjectCounts
+        - alert: etcdHighK8sObjectCounts
           annotations:
-            summary: "Increasing number of {{ $labels.resource }} objects in etcd."
+            summary: etcd recorded {{ $value }} {{ $labels.resource }} objects.
             description: |
-              The number of {{ $labels.resource }} objects have increased by {{ $value }} in the last 10 minutes. If this is unexpected, there are some additional metrics at https://github.com/kubernetes/kube-state-metrics/tree/master/docs that may be used to identify the actual resources. For example,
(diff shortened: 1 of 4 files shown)


bindata/v4.1.0/alerts/etcd-object-counts.yaml
```diff
@@ -0,0 +1,41 @@
...
+          expr: |
+            sum(etcd_object_counts) > 100000
+          for: 10m
+          labels:
+            severity: warning
+        - alert: etcdIncreasingObjectCounts
+          annotations:
+            summary: "Increasing number of {{ $labels.resource }} objects in etcd."
+            description: |
+              The number of {{ $labels.resource }} objects have increased by {{ $value }} in the last 10 minutes. If this is unexpected, there are some additional metrics at https://github.com/kubernetes/kube-state-metrics/tree/master/docs that may be used to identify the actual resources. For example,
```
smarterclayton, line 34: That's not what this alert does.  This alert is triggering if they have increased by 100 per second for 10m (which is 60000 objects in 10m).

bindata/v4.1.0/alerts/etcd-object-counts.yaml
```diff
@@ -0,0 +1,41 @@
...
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
+spec:
+  groups:
+    - name: etcd-object-counts
+      rules:
+        - alert: etcdHighObjectCounts
+          annotations:
+            summary: "etcd recorded more than 100000 objects. Latest count: {{ $value }}"
+            description: |
+              A high object count of more than 100000 objects may lead to stability issue in the API server and etcd storage stack. If this value continues to increase, it may degrade the responsiveness of the API server and etcd.
```
smarterclayton, line 18: This is very vague and probably not actionable.  I don't think 100k is going to be our bar.  Also, there's a huge difference between 1 object types with 100k and 100 object types with 100k

smarterclayton: Can you describe better actions?  How would I "fix" this alert (which is a requirement within alert guidelines https://github.com/openshift/enhancements/pull/637)

ihcsim:
Per offline convo with @smarterclayton, we need to:

* Refine the alerts to distinguish total counts between CRD and non-CRD
     * Define more accurate thresholds based on telemetry data in data hub 
     * Non-CRD could be 100k, it could be 250k; but CRD will definitely be a lot lower
     * Note that CRDs like images and imagestreams have built-in pruning
 * Define an `info` alert providing users with tangible steps to fix the active alerts
     * E.g., "these are the tools and commands to identify the resources' names and namespaces, and how to delete them"
     * Perhaps link the `warning` alerts to this `info` alert to avoid overly verbose warnings
 * Define and test an alert based on the resource response size metrics
     * Rate of kb pulled from resources is roughly proportional to the cost of the resources
     * E.g., `apiserver_response_sizes_bucket`

ihcsim:
@smarterclayton I made some updates to the alerts:

* added two etcd object count alerts; one for CRD and one for k8s.io objects
* added an alert to report on increasing rates in response size by resources
* the threshold values used are for testing purposes only (i still can't access data hub)
* following the [alert consistency enhancements](https://github.com/openshift/enhancements/blob/2368c99c86f718e719f1ac2921cf0938186f88cf/enhancements/monitoring/alerting-consistency.md), I used the `runbook_url` annotation to the alert, instead of walls of text. I think we can flesh out the "how to fix" details in the runbook

Here's what it looks like in the console:

Screenshot from 2021-04-29 15-24-35

I am hoping that between these alerts and the existing `Response Bytes per Second by Resource and Verb` and `etcd Object Count Inspect` dashboards in the console, we can provide users with more visibility into object counts and sizes.

LMKWYT, especially any errors in the `expr`.

ihcsim:
@sttts @smarterclayton I know there is also ongoing conversation around getting more visibility into apiserver rss, per object/response sizes using the `apiserver_response_sizes_*` metrics. I am not sure if that should be tackled separately from this JIRA card, or if they are different aspects towards the same goal. At the moment, this PR doesn't address any of the apiserver memory requirements stuff we talked about with @tkashem.

ihcsim:
Closing, pending further discussion in arch call on threshold values, whether this is the right metrics to capture etc.

PR #1119, opened 2021-04-28
Bug 1954481: Bring back flowcontrol.apiserver.k8s.io/v1alpha1

soltysh:


soltysh:
This isn't sufficient b/c that API is not available to re-enablement.

soltysh:
/assign @sttts 
This will work in companion with https://github.com/openshift/kubernetes/pull/713

PR #1120, opened 2021-04-28
bug 1953798: stop doing live gets for optional cert syncer files that don't exist on disk yet

deads2k:
/hold

This is just a proof PR for a library-go change.


deads2k:
real bump

/hold cancel

PR #1124, opened 2021-05-03
Add a nominal request to the kube-apiserver init container

lack:
This pod needs a nominal request so that the workload partitioning
kubelet mutation can act upon it.

Signed-off-by: Jim Ramsay <jramsay@redhat.com>


soltysh (approved):
/lgtm
/approve

PR #1125, opened 2021-05-03
Bug 1956081: Update bootstrap manifest namespace for SNO compatibility

marun:
Previously, the bootstrap manifest defined a pod in the `kube-system` namespace. After bootstrap of an SNO cluster, the CKAO updated the same same file (/etc/kubernetes/manifests/kube-apiserver-pod.yaml) to define a pod in the `openshift-kube-apiserver` namespace. The kubelet did not appear to consistently tear down the containers of the old pod. If the previous apiserver or insecure readyz containers continued to run, the new pod failed to start due to 6443 and/or 6080 remaining bound by the old pod's containers.

/cc @sttts


eranco74 (commented):
/lgtm

PR #1126, opened 2021-05-04
Bug 1953798: Add apiserver burn rate SLO alerts

ihcsim:
This PR copies the upstream [`KubeAPIErrorBudgetBurn` alerts](https://github.com/openshift/cluster-monitoring-operator/blob/6f22c3b33e2af7a61a3d385f0688d270519b49e4/assets/control-plane/prometheus-rule.yaml#L383-L436) and its recording
 rules to CKAO. It updates the rules to:

1. Exclude long-running requests to subresources like`proxy`, `logs`, `exec` etc., 
from the error budget calculation. As described in 
https://bugzilla.redhat.com/show_bug.cgi?id=1953798#c1, these long-running 
`GET`, `LIST` requests are causing flakiness in the e2e parallel and upgrade 
jobs, due to slow resource reads.
1. Separate the request latency calculation by the request `scope` (i.e. `resource`,
`namespace`, `cluster`).

Subsequent PRs to fix other parts of the e2e test suites will follow, if the changes
in this PR don't fully eliminate the flakiness.

~As part of the work to complete [1953798](https://bugzilla.redhat.com/show_bug.cgi?id=1953798#c1), a separate PR will also be submitted to~
~CMO to remove (disable, mask etc.) the same alerts.~

The PR to remove the same alerts from CMO is at 
https://github.com/openshift/cluster-monitoring-operator/pull/1164.

Signed-off-by: Ivan Sim <isim@redhat.com>

### DIFF
--- a/bindata/v4.1.0/alerts/kube-apiserver-slos.yaml
+++ b/bindata/v4.1.0/alerts/kube-apiserver-slos.yaml
@@ -9,7 +9,7 @@ spec:
     rules:
     - alert: KubeAPIErrorBudgetBurn
       annotations:
-        description: The API server is burning too much error budget.
+        description: The API server is burning too much error budget. This alert fires when too many requests are failing with high latency. Use the 'API Performance' monitoring dashboards to narrow down the request states and latency. The 'etcd' monitoring dashboards also provides metrics to help determine etcd stability and performance.
         summary: The API server is burning too much error budget.
       expr: |
         sum(apiserver_request:burnrate1h) > (14.40 * 0.01000)
@@ -22,7 +22,7 @@ spec:
         short: 5m
     - alert: KubeAPIErrorBudgetBurn
       annotations:
-        description: The API server is burning too much error budget.
+        description: The API server is burning too much error budget. This alert fires when too many requests are failing with high latency. Use the 'API Performance' monitoring dashboards to narrow down the request states and latency. The 'etcd' monitoring dashboards also provides metrics to help determine etcd stability and performance.
         summary: The API server is burning too much error budget.
       expr: |
         sum(apiserver_request:burnrate6h) > (6.00 * 0.01000)
@@ -35,7 +35,7 @@ spec:
         short: 30m
     - alert: KubeAPIErrorBudgetBurn
       annotations:
-        description: The API server is burning too much error budget.
+        description: The API server is burning too much error budget. This alert fires when too many requests are failing with high latency. Use the 'API Performance' monitoring dashboards to narrow down the request states and latency. The 'etcd' monitoring dashboards also provides metrics to help determine etcd stability and performance.
         summary: The API server is burning too much error budget.
       expr: |
         sum(apiserver_request:burnrate1d) > (3.00 * 0.01000)
@@ -48,7 +48,7 @@ spec:
         short: 2h
     - alert: KubeAPIErrorBudgetBurn
       annotations:
-        description: The API server is burning too much error budget.
+        description: The API server is burning too much error budget. This alert fires when too many requests are failing with high latency. Use the 'API Performance' monitoring dashboards to narrow down the request states and latency. The 'etcd' monitoring dashboards also provides metrics to help determine etcd stability and performance.
         summary: The API server is burning too much error budget.
       expr: |
         sum(apiserver_request:burnrate3d) > (1.00 * 0.01000)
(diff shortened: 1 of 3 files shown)


bindata/v4.1.0/alerts/kube-apiserver-slos.yaml
```diff
@@ -0,0 +1,401 @@
...
+              +
+              sum(rate(apiserver_request_duration_seconds_bucket{job="apiserver",verb=~"LIST|GET",scope="cluster",subresource!~"proxy|log|exec",le="5"}[1d]))
+            )
+          )
+          +
+          # errors
+          sum(rate(apiserver_request_total{job="apiserver",verb=~"LIST|GET",subresource!~"proxy|log|exec",code=~"5.."}[1d]))
+        )
+        /
+        sum(rate(apiserver_request_total{job="apiserver",verb=~"LIST|GET",subresource!~"proxy|log|exec"}[1d]))
```
smarterclayton, line 87: hrm, error rate of those api calls has to be included i think

bindata/v4.1.0/alerts/kube-apiserver-slos.yaml
```diff
@@ -0,0 +1,401 @@
...
+              )
+              +
+              sum(rate(apiserver_request_duration_seconds_bucket{job="apiserver",verb=~"LIST|GET",scope="namespace",subresource!~"proxy|log|exec",le="0.5"}[1d]))
+              +
+              sum(rate(apiserver_request_duration_seconds_bucket{job="apiserver",verb=~"LIST|GET",scope="cluster",subresource!~"proxy|log|exec",le="5"}[1d]))
+            )
+          )
+          +
+          # errors
+          sum(rate(apiserver_request_total{job="apiserver",verb=~"LIST|GET",subresource!~"proxy|log|exec",code=~"5.."}[1d]))
```
smarterclayton, line 84: same comment here, we want error rate but not latency rate

bindata/v4.1.0/alerts/kube-apiserver-slos.yaml
```diff
@@ -0,0 +1,401 @@
...
+            sum(rate(apiserver_request_duration_seconds_count{job="apiserver",verb=~"LIST|GET",subresource!~"proxy|log|exec"}[1d]))
+            -
+            (
+              (
+                sum(rate(apiserver_request_duration_seconds_bucket{job="apiserver",verb=~"LIST|GET",scope=~"resource|",subresource!~"proxy|log|exec",le="0.1"}[1d]))
+                or
+                vector(0)
+              )
+              +
+              sum(rate(apiserver_request_duration_seconds_bucket{job="apiserver",verb=~"LIST|GET",scope="namespace",subresource!~"proxy|log|exec",le="0.5"}[1d]))
```
smarterclayton, line 77: there are no cluster or namespace scoped subresource so you don’t need this exclusion

bindata/v4.1.0/alerts/kube-apiserver-slos.yaml
```diff
@@ -0,0 +1,401 @@
...
+    rules:
+    - expr: |
+        (
+          (
+            # too slow
+            sum(rate(apiserver_request_duration_seconds_count{job="apiserver",verb=~"LIST|GET",subresource!~"proxy|log|exec"}[1d]))
+            -
+            (
+              (
+                sum(rate(apiserver_request_duration_seconds_bucket{job="apiserver",verb=~"LIST|GET",scope=~"resource|",subresource!~"proxy|log|exec",le="0.1"}[1d]))
```
smarterclayton, line 72: in order to calculate this correctly you’re going to need to restructure the recording rule a bit (see how i did it in the bug, where each source of burn is given a label).  In order to calculate the contribution from slow requests, this clause and the total (above) need to exclude the subresource, but not for error rate or the others.

the math is gnarly but we roughly want:

contribution_a + contribution_b + contribution_c

vs the current ((t - (a + b + c)) + d) / e

ihcsim:
@smarterclayton Per slack convo, I replaced the old recording rules with the ones defined in the BZ, and excluded the subresources from the resource-scoped calculation. The results of the new and old rules look comparable when tested on my dev cluster, which makes me think that we can replace the old rules entirely. The new burn-type groupings also make things easier to reason about.

The trend of the aggregation rates looks almost identical at a lower time range. E.g. with `[30m]`:

new-recording-rules-30m

old-recording-rules-30m

At a higher time range, even though the shape of the charts looks different (due to the groupings), the values still fall closely within the same range. E.g., with `[1h]`:

new-recording-rules-1d

old-recording-rules-1d

bindata/v4.1.0/alerts/kube-apiserver-slos.yaml
```diff
@@ -0,0 +1,429 @@
...
+            sum(rate(apiserver_request_duration_seconds_count{job="apiserver",verb=~"LIST|GET",scope=~"namespace"}[30m]))
+          - sum(rate(apiserver_request_duration_seconds_bucket{job="apiserver",verb=~"LIST|GET",scope="namespace",le="0.5"}[30m]))
+          , "type", "slow-namespace", "_none_", "")
+          or
+          # cluster-scoped latency
+          label_replace(
+            sum(rate(apiserver_request_duration_seconds_count{job="apiserver",verb=~"LIST|GET",scope="cluster"}[30m]))
+            - sum(rate(apiserver_request_duration_seconds_bucket{job="apiserver",verb=~"LIST|GET",scope="cluster",le="5"}[30m]))
+          , "type", "slow-cluster", "_none_", "")
+        ) / scalar(sum(rate(apiserver_request_total{job="apiserver",verb=~"LIST|GET"}[30m])))
```
smarterclayton, line 122: Actually, with the resource exclusion, this denominator needs to move to each OR, and then on resource scoped you need to exclude subresource from the denominator `(a or b or c)/d` -> `a/d or b/d or c/d` -> `a (exclude resource) / d (exclude resource) or b/d or c/d`

ihcsim: should be fixed now. i added the denominator to the error contribution too.

bindata/v4.1.0/alerts/kube-apiserver-slos.yaml
```diff
@@ -0,0 +1,450 @@
...
+      for: 3h
+      labels:
+        long: 3d
+        severity: warning
+        short: 6h
+  - name: kube-apiserver.rules
+    rules:
+    - expr: |
+        # error
+        label_replace(
```
smarterclayton, line 66: I'm wondering whether I recommended a bad factoring here - it's possible that the rules with / and label matching this doesn't compose like it would in normal math rules.

bindata/v4.1.0/alerts/kube-apiserver-slos.yaml
```diff
@@ -0,0 +1,450 @@
...
+            (sum(rate(apiserver_request_duration_seconds_bucket{job="apiserver",verb=~"LIST|GET",subresource!~"proxy|log|exec",scope=~"resource|",le="0.1"}[5m])) or vector(0))
+          ) / scalar(sum(rate(apiserver_request_total{job="apiserver",verb=~"LIST|GET",subresource!~"proxy|log|exec"}[5m])))
+        , "type", "slow-resource", "_none_", "")
+        or
+        # namespace-scoped latency
+        label_replace(
+          (
+            sum(rate(apiserver_request_duration_seconds_count{job="apiserver",verb=~"LIST|GET",scope=~"namespace"}[5m]))
+          - sum(rate(apiserver_request_duration_seconds_bucket{job="apiserver",verb=~"LIST|GET",scope="namespace",le="0.5"}[5m]))
+          ) / scalar(sum(rate(apiserver_request_total{job="apiserver",verb=~"LIST|GET"}[5m])))
```
smarterclayton, line 85: Technically this has to be the same as the divisor above (subresource!~"...") because we're looking for "the sum of fractions of the requests we care about" and because we eliminate those subresources we "don't care about proxy|exec|log in the denominator for any fraction".  So you need subresource!~ here and in cluster scoped latency, but NOT in the error rate (since the error rate is the fraction of any error).  So it's possible to get over 100%, but it was possible before in the previous rule, and we're just trying to be as accurate as possible.  We'd rather overestimate sources of error than underestimate them in general and that's ok.

bindata/v4.1.0/alerts/kube-apiserver-slos.yaml
```diff
@@ -0,0 +1,450 @@
...
+        /
+        sum(rate(apiserver_request_total{job="apiserver",verb=~"POST|PUT|PATCH|DELETE"}[5m]))
+      labels:
+        verb: write
+      record: apiserver_request:burnrate5m
+    - expr: |
+        (
+          (
+            # too slow
+            sum(rate(apiserver_request_duration_seconds_count{job="apiserver",verb=~"POST|PUT|PATCH|DELETE"}[6h]))
```
smarterclayton, line 402: Doesn't exclude imageimport?   I guess we can handle this separately later.

ihcsim: Yeah, I added these write rules as-is from upstream. Do we foresee image resources contribute to significant burns? Don't image imports get auto-pruned anyway?

smarterclayton:
The read metrics look ok.  We can defer improving the write ones to later since they look correct today.

smarterclayton:
Can you link the PR to remove the old alerts here?

ihcsim:
The PR to remove the same alerts from CMO is at openshift/cluster-monitoring-operator#1164.

smarterclayton:
/test all

This is looking good - while we've talked about removing the alert, I'm tempted to say the current alert is sufficient to merge and then we can remove prior to 4.8.

ihcsim:
@sttts Per slack convo, I separated the PR into two commits, to make future tracking easier. The first commit has the original alerts and rules, and the second has the new rules.

sttts:
Factoring looks good. With @smarterclayton's lgtm https://github.com/openshift/cluster-kube-apiserver-operator/pull/1126#issuecomment-848944718 for the actual alert expressions.

/lgtm
/approve

PR #1127, opened 2021-05-06
bug 1957645: avoid warm loop for writing prometheus rules

deads2k:
1. brings in the lib-go fixes https://github.com/openshift/library-go/pull/1070
2. fixes the formatting for one naughty rule
3. writes up a dockerfile I used for testing.


sttts:
verify-deps.

Please rename to Dockerfile-dev. Then everybody knows what it is. Simple does not say anything.

PR #1128, opened 2021-05-07
Bug 1958094: Add flock to kube-apiserver startup

sttts:
We have seen evidence that kube-apiserver processes are not started after each other without overlap. This will make it visible.

### DIFF
--- a/bindata/v4.1.0/kube-apiserver/pod.yaml
+++ b/bindata/v4.1.0/kube-apiserver/pod.yaml
@@ -43,6 +43,15 @@ spec:
     command: ["/bin/bash", "-ec"]
     args:
         - |
+          LOCK=/var/log/kube-apiserver/.lock
+          echo -n "Acquiring exclusive lock ${LOCK}"
+          exec {LOCK_FD}>${LOCK} && flock -n "${LOCK_FD}" || {
+            echo "$(date -Iseconds -u) kubelet did not terminate old kube-apiserver before new one" >> /var/log/kube-apiserver/lock.log
+            echo -n ": WARNING: kubelet did not terminate old kube-apiserver before new one."
+            # we didn't get an exclusive lock. We keep going with the risk to corrupt audit logs.
+          }
+          echo
+
           if [ -f /etc/kubernetes/static-pod-certs/configmaps/trusted-ca-bundle/ca-bundle.crt ]; then
             echo "Copying system trust bundle"
             cp -f /etc/kubernetes/static-pod-certs/configmaps/trusted-ca-bundle/ca-bundle.crt /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
--- a/pkg/operator/v410_00_assets/bindata.go
+++ b/pkg/operator/v410_00_assets/bindata.go
@@ -1665,6 +1665,15 @@ spec:
     command: ["/bin/bash", "-ec"]
     args:
         - |
+          LOCK=/var/log/kube-apiserver/.lock
+          echo -n "Acquiring exclusive lock ${LOCK}"
+          exec {LOCK_FD}>${LOCK} && flock -n "${LOCK_FD}" || {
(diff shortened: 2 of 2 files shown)


bindata/v4.1.0/kube-apiserver/pod.yaml
```diff
@@ -43,6 +43,15 @@ spec:
     command: ["/bin/bash", "-ec"]
     args:
         - |
+          echo -n "Acquiring exclusive lock /var/log/kube-apiserver/.lock"
+          exec {lock_fd}>/var/log/kube-apiserver/.lock || exit 1
```
p0lyn0mial, line 47: so this is for opening a file,
where does `{lock_fd}` come from?

bindata/v4.1.0/kube-apiserver/pod.yaml
```diff
@@ -43,6 +43,15 @@ spec:
...
+          flock -n "${lock_fd}" || {
```
p0lyn0mial, line 48: this is for acquiring the lock. I have checked by default exclusive is applied and `-n` here means locking is a non-blocking operation.

p0lyn0mial: where the lock is released?

sttts: automatically by leaving the shell.

smarterclayton:
I'm extremely concerned about merging this into prod environments - the risk of wedging a cluster due to this is serious (the bug is bad, this could make it worse if we have ANY errors around lock release).  Are we 100% positive there is NO way this could cause us to fail to start at least one apiserver (that would be my bar for having this in longer than it takes to find the issue pre-4.8 release)?

bindata/v4.1.0/kube-apiserver/pod.yaml
```diff
@@ -43,6 +43,15 @@ spec:
...
+            echo "$(date -Iseconds -u) kubelet did not terminate old kube-apiserver before new one" >> /var/log/kube-apiserver/lock.log
+            echo " WARNING: kubelet did not terminate old kube-apiserver before new one." >&2
+            flock -s -n "${lock_fd}"
```
sttts, line 51: @smarterclayton please see here. We made it not fail anymore, but just switch to shared lock mode. So the next container will try the exclusive lock again and will warn again. This way we get all the debugging value, but don't fail starting the pod.

bindata/v4.1.0/kube-apiserver/pod.yaml
```diff
@@ -43,6 +43,15 @@ spec:
...
     args:
         - |
+          echo -n "Acquiring exclusive lock /var/log/kube-apiserver/.lock"
+          exec {lock_fd}>/var/log/kube-apiserver/.lock || exit 1
+          flock -n "${lock_fd}" || {
+            echo "$(date -Iseconds -u) kubelet did not terminate old kube-apiserver before new one" >> /var/log/kube-apiserver/lock.log
+            echo " WARNING: kubelet did not terminate old kube-apiserver before new one." >&2
+            # we didn't get an exclusive lock. We keep going with the risk to corrupt audit logs.
+          }
+          echo
```
p0lyn0mial, line 53: ah, to get a new line

bindata/v4.1.0/kube-apiserver/pod.yaml
smarterclayton: Ok, thanks

smarterclayton:
Is there a must gather bug?  It failed 3 times in a row the same way?

sttts:
@smarterclayton no bug, just a test updated here: https://github.com/openshift/origin/pull/26138

bindata/v4.1.0/kube-apiserver/pod.yaml
```diff
@@ -43,6 +43,15 @@ spec:
     command: ["/bin/bash", "-ec"]
     args:
         - |
+          LOCK=/var/log/kube-apiserver/.lock
+          echo -n "Acquiring exclusive lock ${LOCK}"
```
sttts, line 47: not that exclusive actually.

PR #1129, opened 2021-05-12
Bug 1967090: webhookauth observer: set cache TTL to 30s

stlaz:
Set webhook authenticator's cache TTL to 30s to prevent very early
oauthaccesstoken timeouts.

---
as described in https://github.com/openshift/enhancements/pull/403

/assign @deads2k 
cc @s-urbaniak


stlaz:
BZ tbd

slaskawi (approved):
LGTM but I think it would be good to add a test somewhere.

PR #1079, opened 2021-03-29
Bug 1943804: increases termination timeouts for AWS

p0lyn0mial:
Due to a known AWS issue: https://bugzilla.redhat.com/show_bug.cgi?id=1943804
the time needed for an LB to notice and remove unhealthy instances must be extended.

`shutdown-delay-duration` flag has been changed from `70s` to `210s` and `terminationGracePeriodSeconds` field from `135s` to` 275s`.

For `terminationGracePeriodSeconds` the initial `210s` is reserved for the minimal termination period.
Additional `60s` for finishing all in-flight requests and an extra `5s` to make sure a potential SIGTERM will be sent after a server terminates itself.

PR #1131, opened 2021-05-13
Bug 1943804: split encryption tests

p0lyn0mial:
The current timeout for all encryption tests is 4 hours.
Previously the time needed to roll out a new KAS was 15 minutes, due to a known issue on AWS with the NLB it can take now 22 minutes.
It means that the tests need ~25 min to make sure that all servers are on the same revision and it is safe to start the test.

Since the rotation scenario needs to enable the encryption first, then wait and check if all resources were migrated, then change the encryption key and check the migration again.
It means that only these tests will need ~1 hour just for installing the servers.

Thus this PR moves the rotation tests to a separate CI job.
It will shorten the overall time needed to test encryption but also increase the pass rate of the tests as we have seen them frequently failing to due to a timeout.

xref: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1079

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,6 +56,14 @@ test-e2e-encryption: GO_TEST_FLAGS += -p 1
 test-e2e-encryption: test-unit
 .PHONY: test-e2e-encryption
 
+# these are extremely slow serial e2e encryption rotation tests that modify the cluster's global state
+test-e2e-encryption-rotation: GO_TEST_PACKAGES :=./test/e2e-encryption-rotation/...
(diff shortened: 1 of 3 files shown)


p0lyn0mial:
/hold

I need to create a new CI job first before we can merge it

p0lyn0mial:
`e2e-aws-operator-encryption` failed on `e2e-aws-operator-encryption-gather-extra container test` the encryption tests passed.

p0lyn0mial:
`e2e-aws-operator-encryption-rotation` green!

PR #1134, opened 2021-05-14
Set proper root image in .ci-operator.yaml

soltysh:
We should be always using `rhel-8-release-golang-1.16-openshift-4.8` not `golang-1.16` since the other is deprecated and is missing a lot of tooling needed for build process. 

/assign @mfojtik


soltysh:
Setting valid-bug manually, since this is changing build configuration, not fixing bug.

PR #1143, opened 2021-06-08
Bug 1921139: remove override for fast cert rotation for release

sttts:
This switches back to normal 30 days rotation period (from 6h during development).


mfojtik:
/override ci/prow/e2e-upgrade

Adding override as we don't want to ship with short cert rotation by accident. The failures in the upgrade are unrelated to this change.

PR #1145, opened 2021-06-09
 kube-apiserver failed to load SNI cert and key

p0lyn0mial:
As of today, the dynamic certificates i.e. `kube-apiserver-certs` are accessed by at least two processes, namely the installer pod and the cert-syncer.
Up until now, there was no coordination between these processes that might have lead to many unexpected errors, like https://bugzilla.redhat.com/show_bug.cgi?id=1963730

This PR writes all certificates in an atomic way by first creating a temporary file, writing the content to it, and then renaming it to the original file. 

`os.Rename` calls `syscall.Rename` which in turn uses the rename syscall (Linux) which provides atomicity (https://man7.org/linux/man-pages/man2/rename.2.html)

The previous attempt didn't work as it would break the DR scripts - https://github.com/openshift/library-go/pull/1098

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/openshift/api v0.0.0-20210521075222-e273a339932a
 	github.com/openshift/build-machinery-go v0.0.0-20210423112049-9415d7ebd33e
 	github.com/openshift/client-go v0.0.0-20210521082421-73d9475a9142
-	github.com/openshift/library-go v0.0.0-20210608152959-aa07cb1d7909
+	github.com/openshift/library-go v0.0.0-20210611094144-35c8a075e255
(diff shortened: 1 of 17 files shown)


p0lyn0mial:
/hold

`ci/prow/e2e-gcp-operator ` seems to be broken

p0lyn0mial:
> ci/prow/e2e-gcp-operator

keeps failing on `Multus: [openshift-kube-apiserver/installer-13-ci-op-h6m9mzk7-f6035-jxnvh-master-1]: error getting pod: Unauthorized`

p0lyn0mial:
failed to create pod network sandbox will be fixed by https://bugzilla.redhat.com/show_bug.cgi?id=1972490

PR #1158, opened 2021-06-22
Bug 1974716: SA token issuer observer: fix observing api-audiences

stlaz:
api-audiences field was getting pruned which would result in a failing
cluster in case the service account issuer changed.

/assign @deads2k


s-urbaniak:
@stlaz can we have a test for the whole `ObserveServiceAccountIssuer` func or pull `configobserver.Prune` into `observedConfig`?

PR #1162, opened 2021-06-23
disable apiextensions.k8s.io/v1beta1 and admissionregistration.k8s.io/v1beta1

atiratree:
to disable it even before we hit k8s 1.22

### DIFF
--- a/bindata/v4.1.0/kube-apiserver/pod.yaml
+++ b/bindata/v4.1.0/kube-apiserver/pod.yaml
@@ -68,7 +68,7 @@ spec:
             fi
           done
           echo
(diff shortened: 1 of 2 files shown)


atiratree:
fix for the origin tests: https://github.com/openshift/origin/pull/26269

soltysh (approved):
/lgtm
/approve
/hold
until we get that origin PR merged

atiratree:
I think after this one we should be good to go https://github.com/openshift/origin/pull/26280/

soltysh:
The upgrade failures seemed to be related with a problem that was fixed in this revert https://github.com/openshift/cluster-monitoring-operator/pull/1259

PR #1174, opened 2021-07-11
Start using embed module for assets

Fedosin:
This commit starts using "embed" module for static assets and removes the generated ones, created by go-bindata. This should help developers, as they don't have to regenerate bindata after each change in the assets.


bindata/assets.go
```diff
@@ -0,0 +1,24 @@
+package v410_00_assets
```
sttts, line 1: do we still need this package name?

Fedosin: it's mostly done for backward compatibility. If you think it is reasonable to change it to something else, I'm absolutely fine with it.

Fedosin:
/hold

^ for tests to pass

PR #1180, opened 2021-07-16
introduces KubeAPIReadinessChecker used by startup monitor to assess Kube API server readiness/health condition

p0lyn0mial:
this PR implements the following checks

- checks if we are not dealing with the old kas
- checks kube-apiserver /healthz/etcd endpoint
- checks kube-apiserver /healthz endpoint
- checks kube-apiserver /readyz endpoint
- checks if the kas pod is running at the expected revision
- checks that kubelet has reporting readiness for the new pod

xref: openshift/enhancements#833

### DIFF
--- a/pkg/operator/startupmonitorreadiness/readiness_checks.go
+++ b/pkg/operator/startupmonitorreadiness/readiness_checks.go
@@ -1,0 +1,324 @@
+package startupmonitorreadiness
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"strconv"
+	"time"
+
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/operatorclient"
+	"github.com/openshift/library-go/pkg/operator/staticpod/startupmonitor"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilnet "k8s.io/apimachinery/pkg/util/net"
+	"k8s.io/client-go/kubernetes"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/transport"
+)
+
+// KubeAPIReadinessChecker is a struct that holds necessary data
+// to perform a set of checks against a Kube API server to assess its health condition
+type KubeAPIReadinessChecker struct {
+	// configuration for authN/authZ against the server
+	// populated from kubeconfig and set by the startup monitor pod
+	restConfig *rest.Config
+
+	// client we use to perform HTTP checks
+	client *http.Client
+
+	// defined here for easier testing
+	baseRawURL string
+
+	kubeClient *kubernetes.Clientset
+}
+
+var _ startupmonitor.ReadinessChecker = &KubeAPIReadinessChecker{}
+var _ startupmonitor.WantsRestConfig = &KubeAPIReadinessChecker{}
+
+// New creates a new Kube API readiness checker
+func New() *KubeAPIReadinessChecker {
+	return &KubeAPIReadinessChecker{
+		baseRawURL: "https://localhost:6443",
+	}
+}
+
+// SetRestConfig called by startup monitor to provide a valid configuration for authN/authZ against Kube API server
+func (ch *KubeAPIReadinessChecker) SetRestConfig(config *rest.Config) {
+	ch.restConfig = config
+
+	// note that we will be talking to Kube API over localhost and in case of an error/timeout requests will be retired for 5 min.
+	// setting the global timeout to a short value seems to be fine
+	ch.restConfig.Timeout = 4 * time.Second
+
+	ch.restConfig.Burst = 15
+	ch.restConfig.QPS = 10
+}
+
+// IsReady performs a series of checks for assessing Kube API server readiness condition
+func (ch *KubeAPIReadinessChecker) IsReady(ctx context.Context, revision int) ( /*ready*/ bool /*reason*/, string /*message*/, string /*err*/, error) {
+	if ch.restConfig == nil {
+		return false, "", "", fmt.Errorf("missing restConfig, use SetRestConfig() metod to set one")
+
+	}
+	if ch.client == nil {
+		client, err := createHTTPClient(ch.restConfig)
+		if err != nil {
+			return false, "", "", fmt.Errorf("failed to create an HTTP client due to %v", err)
+		}
+		ch.client = client
+	}
+
+	if ch.kubeClient == nil {
+		kubeClient, err := kubernetes.NewForConfig(ch.restConfig)
+		if err != nil {
+			return false, "", "", fmt.Errorf("failed to create kubernetes clientset due to %v", err)
+		}
+		ch.kubeClient = kubeClient
+	}
+
+	// loop through a list of ordered checks for assessing Kube API readiness condition
+	for _, checkFn := range []func(context.Context) (bool, string, string){
+		//	TODO: watch /var/log/kube-apiserver/termination.log for the first start-up attempt (beware of the race of startup-monitor startup and kube-apiserver startup). Set Reason=NeverStartedUp when this times out.
+		//	TODO: watch /var/log/kube-apiserver/termination.log for more than one start-up attempt. Set Reason=CrashLooping if more than one is found and the monitor times out.
+
+		// checks if we are not dealing with the old kas
+		noOldRevisionPodExists(ch.kubeClient.CoreV1().Pods(operatorclient.TargetNamespace), revision),
+
+		// check kube-apiserver /healthz/etcd endpoint
+		goodHealthzEtcdEndpoint(ch.client, ch.baseRawURL),
+
+		// check kube-apiserver /healthz endpoint
+		goodHealthzEndpoint(ch.client, ch.baseRawURL),
+
+		// check kube-apiserver /readyz endpoint
+		goodReadyzEndpoint(ch.client, ch.baseRawURL, 3, 5*time.Second),
+
+		// check if the kas pod is running at the expected revision
+		newRevisionPodExists(ch.kubeClient.CoreV1().Pods(operatorclient.TargetNamespace), revision),
+
+		// check that kubelet has reporting readiness for the new pod
+		newPodRunning(ch.kubeClient.CoreV1().Pods(operatorclient.TargetNamespace), revision),
+	} {
+		select {
+		case <-ctx.Done():
+			return false, "", "", ctx.Err()
+		default:
+		}
+
+		if ready, reason, message := checkFn(ctx); !ready {
+			return ready, reason, message, nil
+		}
+	}
+
+	// at this point Kube API is ready!
+	return true, "", "", nil
+}
+
+// newPodRunning checks if kas pod is in PodRunning phase and has PodReady condition set to true
+func newPodRunning(podClient corev1client.PodInterface, monitorRevision int) func(context.Context) (bool, string, string) {
+	return func(ctx context.Context) (bool, string, string) {
+		apiServerPods, err := podClient.List(ctx, metav1.ListOptions{LabelSelector: "apiserver=true"})
+		if err != nil {
+			return false, "PodListError", fmt.Sprintf("unable to check the pod's status, falied to get Kube API server pod due to %v", err)
+		}
+		if len(apiServerPods.Items) == 0 {
+			return false, "PodNotRunning", "unable to check the pod's status, waiting for Kube API server pod to show up"
+		}
+		if len(apiServerPods.Items) != 1 {
+			return false, "PodListError", fmt.Sprintf("unable to check the pod's status, unexpected number of Kube API server pods %d, expected only one pod", len(apiServerPods.Items))
+		}
+
+		kasPod := apiServerPods.Items[0]
+		if kasPod.Status.Phase != corev1.PodRunning {
+			return false, "PodNodReady", fmt.Sprintf("waiting for Kube API server pod to be in PodRunning phase, the current phase is %v", kasPod.Status.Phase)
+		}
+
+		if kasPod.Status.Phase == corev1.PodRunning && !func(pod corev1.Pod) bool {
+			for _, condition := range pod.Status.Conditions {
+				if condition.Type == corev1.PodReady && condition.Status == corev1.ConditionTrue {
+					return true
+				}
+			}
+			return false
+		}(kasPod) {
+			return false, "PodNodReady", "waiting for Kube API server pod to have PodReady state set to true"
+		}
+
+		return checkRevision(&kasPod, monitorRevision)
+	}
+}
+
+// newRevisionPodExists check if the kas pod is running at the expected revision
+func newRevisionPodExists(podClient corev1client.PodInterface, monitorRevision int) func(context.Context) (bool, string, string) {
+	return func(ctx context.Context) (bool, string, string) {
+		return checkRevisionOnPod(ctx, podClient, monitorRevision, true)
+	}
+}
+
+// noOldRevisionPodExists checks if we are not dealing with the old kas
+// it is useful when you want to avoid false positive - failing readyz check when the previous instance is still running
+//
+// note that:
+// it won't fail when getting the pod from the api server fails as that might mean the new instance is not ready/healthy
+func noOldRevisionPodExists(podClient corev1client.PodInterface, monitorRevision int) func(context.Context) (bool, string, string) {
+	return func(ctx context.Context) (bool, string, string) {
+		return checkRevisionOnPod(ctx, podClient, monitorRevision, false)
+	}
+}
+
+// checkRevisionOnPod checks if the kas pod is running at the expected revision
+//
+// strictMode controls whether a certain errors like: failing to get the pod or absence of the pod should be fatal
+// it is useful when you want to avoid false positive - failing readyz check when the previous instance is still running
+func checkRevisionOnPod(ctx context.Context, podClient corev1client.PodInterface, monitorRevision int, strictMode bool) (bool, string, string) {
+	apiServerPods, err := podClient.List(ctx, metav1.ListOptions{LabelSelector: "apiserver=true"})
+	if err != nil {
+		return !strictMode, "PodListError", fmt.Sprintf("unable to check a revison, failed to get Kube API server pod due to %v", err)
+	}
+	if len(apiServerPods.Items) == 0 {
+		return !strictMode, "PodNotRunning", "unable to check a revision, waiting for Kube API server pod to show up"
+	}
+	if len(apiServerPods.Items) != 1 {
+		return !strictMode, "PodListError", fmt.Sprintf("unable to check a revision, unexpected number of Kube API server pods %d, expected only one pod", len(apiServerPods.Items))
+	}
+
+	return checkRevision(&apiServerPods.Items[0], monitorRevision)
+}
+
+func checkRevision(kasPod *corev1.Pod, monitorRevision int) (bool, string, string) {
+	revisionString, found := kasPod.Labels["revision"]
+	if !found {
+		return false, "InvalidPod", fmt.Sprintf("pod %s doesn't have revision label", kasPod.Name)
+	}
+	if len(revisionString) == 0 {
+		return false, "InvalidRevision", fmt.Sprintf("empty revision label on %s pod", kasPod.Name)
+	}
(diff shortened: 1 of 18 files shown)


pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -0,0 +1,259 @@
...
+
+		// checks overall Kube API health condition
+		func() (bool, string, string) { return doKubeHealthCheck(ctx, ch.client, ch.baseRawURL) },
+
+		// check overall Kube API readiness condition
+		func() (bool, string, string) {
+			return doKubeReadyCheck(ctx, ch.client, ch.baseRawURL, 3, 5*time.Second)
+		},
+
+		// doRevisionCheck in "strict" mode
```
sttts, line 97: what is strict? Better describe here

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -0,0 +1,259 @@
...
+			return ready, reason, message, nil
+		}
+	}
+
+	// at this point Kube API is ready!
+	return true, "", "", nil
+}
+
+// doRevisionCheck checks if the kas pod is running at the expected revision
+func doRevisionCheck(ctx context.Context, podClient corev1client.PodInterface, monitorRevision int, strictMode bool) (bool, string, string) {
```
sttts, line 122: be more descriptive with strictMode. What does it mean?

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -0,0 +1,259 @@
...
+	apiServerPods, err := podClient.List(ctx, metav1.ListOptions{LabelSelector: "apiserver=true"})
+	if err != nil {
+		return !strictMode, "NotReadyError", fmt.Sprintf("falied to get Kube API server pod due to %v", err)
+	}
+	if len(apiServerPods.Items) == 0 {
+		return !strictMode, "NotReady", "waiting for Kube API server pod to show up"
+	}
+	if len(apiServerPods.Items) != 1 {
+		return !strictMode, "NotReady", fmt.Sprintf("unexpected number of Kube API server pods %d, expected only one pod", len(apiServerPods.Items))
```
sttts, line 131: do we change the name?

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -0,0 +1,259 @@
...
+	if len(revisionString) == 0 {
+		return false, "NotReady", fmt.Sprintf("empty revision label on %s pod", kasPod.Name)
+	}
+	revision, err := strconv.Atoi(revisionString)
+	if err != nil || revision < 0 {
+		return false, "NotReady", fmt.Sprintf("invalid revision label on pod %s: %q", kasPod.Name, revisionString)
+	}
+
+	if revision != monitorRevision {
+		return false, "NotReady", fmt.Sprintf("the running Kube API is at unexpected revision %d, expected %d", revision, monitorRevision)
```
sttts, line 148: use better reasons for all of these. NotReady should be reserved for the time we see the pod in the API, but it is not ready yet.

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -0,0 +1,259 @@
...
+
+	// at this point Kube API is ready!
+	return true, "", "", nil
+}
+
+// doRevisionCheck checks if the kas pod is running at the expected revision
+func doRevisionCheck(ctx context.Context, podClient corev1client.PodInterface, monitorRevision int, strictMode bool) (bool, string, string) {
+	apiServerPods, err := podClient.List(ctx, metav1.ListOptions{LabelSelector: "apiserver=true"})
+	if err != nil {
+		return !strictMode, "NotReadyError", fmt.Sprintf("falied to get Kube API server pod due to %v", err)
```
sttts, line 125: PodListError

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+	if len(apiServerPods.Items) == 0 {
+		return !strictMode, "NotReady", "waiting for Kube API server pod to show up"
```
sttts, line 128: PodNotRunning

pkg/operator/startupmonitorreadiness/readiness_checks.go
sttts: MultiplePods

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+	if len(apiServerPods.Items) != 1 {
+		return !strictMode, "NotReady", fmt.Sprintf("unexpected number of Kube API server pods %d, expected only one pod", len(apiServerPods.Items))
+	}
+
+	kasPod := apiServerPods.Items[0]
+	revisionString, found := kasPod.Labels["revision"]
+	if !found {
+		return false, "NotReady", fmt.Sprintf("pod %s doesn't have revision label", kasPod.Name)
```
sttts, line 137: PodNotRunning

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+	if len(revisionString) == 0 {
+		return false, "NotReady", fmt.Sprintf("empty revision label on %s pod", kasPod.Name)
```
sttts, line 140: PodNotRunning

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+	revision, err := strconv.Atoi(revisionString)
+	if err != nil || revision < 0 {
```
sttts, line 143: InvalidPod

pkg/operator/startupmonitorreadiness/readiness_checks.go
sttts: WrongRevision

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -0,0 +1,259 @@
...
+	tlsConfig, err := transport.TLSConfigFor(transportConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	client := &http.Client{
+		Transport: utilnet.SetTransportDefaults(&http.Transport{
+			TLSClientConfig: tlsConfig,
+		}),
+		Timeout: responseTimeout,
```
aojea, line 235: http.Client.Timeout may timeout even if the connection was sucesfull

https://cs.opensource.google/go/go/+/refs/tags/go1.16.6:src/net/http/client.go;l=90-94

I fixed recently a bug in kops because of this, https://github.com/kubernetes/kops/pull/11884/files#diff-8a8f23d88ae96f6029ac0558e279ed8850568082c913852bfd45c2b54e26166fR81-R93 

We are operating in a 2 seconds timeframe that seems to short, but still can race.
What about having more granularity (I'm making up the values here)?
```
	httpClient := &http.Client{
		Transport: &http.Transport{
			DialContext: (&net.Dialer{
				Timeout:   5 * time.Second,
				KeepAlive: 5 * time.Second,
			}).DialContext,
			TLSHandshakeTimeout:   5 * time.Second,
			ResponseHeaderTimeout: 5 * time.Second,
			IdleConnTimeout:       5 * time.Second,
		},
```

p0lyn0mial: This http client will be talking to Kube API over localhost. In case of an error/timeout requests will be retired for 5 min. 

Do we still think 2 seconds is too short?

aojea: ah, ok, this is always localhost, maybe I'm being  too paranoid :smile: 
these requests doesn't go through the apf thing, right? there is no chance the server can take more than 2 seconds to reply? The different timeouts help to known which part timed out exactly, but as you say this may not apply here

p0lyn0mial: > these requests doesn't go through the apf thing, right?

it does, we will authN as "system:master" which always is privileged.

> there is no chance the server can take more than 2 seconds to reply?

there is, for example getting a pod might take much longer if etcd is slow for example. I can increase the timeout to 5 seconds. 

The purpose of these checks is to assess the readiness of the Kube API server and fallback to the previous version in case of any issues. If the API server is unable to serve requests because the connection to etcd is "slow" we should fail.

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -0,0 +1,259 @@
...
+// New creates a new Kube API readiness checker
+func New() *KubeAPIReadinessChecker {
+	return &KubeAPIReadinessChecker{
+		baseRawURL: "https://localhost:6443",
+	}
+}
+
+// SetRestConfig called by startup monitor to provide a valid configuration for authN/authZ against Kube API server
+func (ch *KubeAPIReadinessChecker) SetRestConfig(restConfig *rest.Config) {
+	ch.restConfig = restConfig
```
aojea, line 51: is this restConfig limited by the client-go parameters `QPS, burst, RateLimiter`?
if affirmative, should we override it?

pkg/operator/startupmonitorreadiness/readiness_checks.go
aojea: I saw the Pod.List() and other api calls take 2 seconds on busy CIs, can this be a concern here?

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+// IsReady performs a series of checks for assessing Kube API server readiness condition
+func (ch *KubeAPIReadinessChecker) IsReady(ctx context.Context, revision int) ( /*ready*/ bool /*reason*/, string /*message*/, string /*err*/, error) {
+	if ch.restConfig == nil {
+		return false, "", "", fmt.Errorf("missing restConfig, use SetRestConfig() metod to set one")
+
+	}
+	if ch.client == nil {
+		client, err := createHTTPClient(2*time.Second, ch.restConfig)
```
aojea, line 61: nit, should this be a constant? 
```
const httpClientTimeout = 2* time.Second
```

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -0,0 +1,259 @@
...
+		if err != nil {
+			return false, "", "", fmt.Errorf("failed to create an HTTP client due to %v", err)
+		}
+		ch.client = client
+	}
+
+	if ch.kubeClient == nil {
+		kubeClient, err := kubernetes.NewForConfig(ch.restConfig)
```
aojea, line 69: should we add a timeout here too?

pkg/operator/startupmonitorreadiness/readiness_checks.go
aojea: ahh, this uses the kubeclient not the http.client

pkg/operator/startupmonitorreadiness/readiness_checks.go
p0lyn0mial: yes we should, I slightly reworked the PR

pkg/operator/startupmonitorreadiness/readiness_checks.go
p0lyn0mial: the default values QPS=5, Burst=10 should be enough, we are going to send just request per second.

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+}
+
+// SetRestConfig called by startup monitor to provide a valid configuration for authN/authZ against Kube API server
+func (ch *KubeAPIReadinessChecker) SetRestConfig(config *rest.Config) {
+	ch.restConfig = config
+
+	// note that we will be talking to Kube API over localhost and in case of an error/timeout requests will be retired for 5 min.
+	// setting the global timeout to a short value seems to be fine
+	ch.restConfig.Timeout = 4 * time.Second
```
p0lyn0mial, line 56: @aojea increased the timeout to 4s and set QPS to 10 and Burst to 15

pkg/operator/startupmonitorreadiness/readiness_checks_test.go
```diff
@@ -0,0 +1,446 @@
...
+
+		{
+			name: "scenario 3: unexpected err from the server",
+			rspWriterFn: func(w http.ResponseWriter) {
+				panic("bum")
+			},
+			healthy: false,
+			reason:  "UnhealthyError",
+			// we don't check the entire rsp from the server
+			msg: "falied while performing the check due to",
```
aojea, line 276: nit , you were carrying over this typo `falied`

aojea:
LGTM

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -0,0 +1,303 @@
...
+		}
+		ch.kubeClient = kubeClient
+	}
+
+	// loop through a list of ordered checks for assessing Kube API readiness condition
+	for _, checkFn := range []func() (bool, string, string){
+		//	TODO: watch /var/log/kube-apiserver/termination.log for the first start-up attempt (beware of the race of startup-monitor startup and kube-apiserver startup). Set Reason=NeverStartedUp when this times out.
+		//	TODO: watch /var/log/kube-apiserver/termination.log for more than one start-up attempt. Set Reason=CrashLooping if more than one is found and the monitor times out.
+
+		// doRevisionCheck in "lazy" mode to avoid false positive - failing readyz check when the previous instance hasn't terminated
```
sttts, line 89: what is lazy mode?

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -0,0 +1,303 @@
...
+		func() (bool, string, string) {
+			return doRevisionCheck(ctx, ch.kubeClient.CoreV1().Pods(operatorclient.TargetNamespace), revision, false)
+		},
+
+		// checks etcd health condition
```
sttts, line 94: be precise: this is "check etcd health how kube-apiserver sees it"

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+
+	// loop through a list of ordered checks for assessing Kube API readiness condition
+	for _, checkFn := range []func() (bool, string, string){
+		//	TODO: watch /var/log/kube-apiserver/termination.log for the first start-up attempt (beware of the race of startup-monitor startup and kube-apiserver startup). Set Reason=NeverStartedUp when this times out.
+		//	TODO: watch /var/log/kube-apiserver/termination.log for more than one start-up attempt. Set Reason=CrashLooping if more than one is found and the monitor times out.
+
+		// doRevisionCheck in "lazy" mode to avoid false positive - failing readyz check when the previous instance hasn't terminated
+		func() (bool, string, string) {
+			return doRevisionCheck(ctx, ch.kubeClient.CoreV1().Pods(operatorclient.TargetNamespace), revision, false)
```
sttts, line 91: what is a revision check if the revision does not matter?! versus line 107

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -0,0 +1,303 @@
...
+		func() (bool, string, string) {
+			return doKubeReadyCheck(ctx, ch.client, ch.baseRawURL, 3, 5*time.Second)
+		},
+
+		// doRevisionCheck in "strict" mode - at this point the running instance must match the expected revision
+		func() (bool, string, string) {
+			return doRevisionCheck(ctx, ch.kubeClient.CoreV1().Pods(operatorclient.TargetNamespace), revision, true)
+		},
+
+		// checks if kas pod is in PodRunning phase and has PodReady condition set to true
```
sttts, line 110: check that kubelet has reporting readiness of the pod

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -0,0 +1,321 @@
...
+	if ch.kubeClient == nil {
+		kubeClient, err := kubernetes.NewForConfig(ch.restConfig)
+		if err != nil {
+			return false, "", "", fmt.Errorf("failed to create kubernetes clientset due to %v", err)
+		}
+		ch.kubeClient = kubeClient
+	}
+
+	// loop through a list of ordered checks for assessing Kube API readiness condition
+	for _, checkFn := range []func(context.Context) (bool, string, string){
```
sttts, line 85: 😍  – nice and short

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -0,0 +1,321 @@
...
+		//	TODO: watch /var/log/kube-apiserver/termination.log for the first start-up attempt (beware of the race of startup-monitor startup and kube-apiserver startup). Set Reason=NeverStartedUp when this times out.
+		//	TODO: watch /var/log/kube-apiserver/termination.log for more than one start-up attempt. Set Reason=CrashLooping if more than one is found and the monitor times out.
+
+		// checks if we are not dealing with the old kas
+		noOldRevisionPodExists(ch.kubeClient.CoreV1().Pods(operatorclient.TargetNamespace), revision),
+
+		// check kube-apiserver /healthz/etcd endpoint
+		goodETCDEndpoint(ch.client, ch.baseRawURL),
```
sttts, line 93: nit: `good(KubeApiserver)HealthzEtcdEndpoint` – this sounds like we check etcd directly.

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -0,0 +1,321 @@
...
+func goodETCDEndpoint(client *http.Client, rawURL string) func(context.Context) (bool, string, string) {
+	return func(ctx context.Context) (bool, string, string) {
+		return doHTTPCheckAndTransform(ctx, client, fmt.Sprintf("%s/healthz/etcd", rawURL), "EtcdUnhealthy", doHTTPCheck)
+	}
+}
+
+func doHTTPCheckAndTransform(ctx context.Context, client *http.Client, rawURL string, checkName string, httpCheckFn func(ctx context.Context, client *http.Client, rawURL string) (int, string, error)) (bool, string, string) {
+	statusCode, response, err := httpCheckFn(ctx, client, rawURL)
+	if err != nil {
+		errMsg := fmt.Sprintf("failed while performing the check due to %v", err)
```
sttts, line 243: is this helpful. does err have e.g. the path? Would also follow the normal format with colon: `.... the check: %v`

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -0,0 +1,321 @@
...
+		goodHealthzEndpoint(ch.client, ch.baseRawURL),
+
+		// check kube-apiserver /readyz endpoint
+		goodReadyzEndpoint(ch.client, ch.baseRawURL, 3, 5*time.Second),
+
+		// check if the kas pod is running at the expected revision
+		newRevisionPodExists(ch.kubeClient.CoreV1().Pods(operatorclient.TargetNamespace), revision),
+
+		// check that kubelet has reporting readiness of the pod
+		newPodHasStateRunning(ch.kubeClient.CoreV1().Pods(operatorclient.TargetNamespace)),
```
sttts, line 105: I would like to see the last two linked into one check. It's the final check and it should ensure consistency. Now there is a slight race because we do two client calls.

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -0,0 +1,321 @@
...
+	if len(revisionString) == 0 {
+		return false, "InvalidPod", fmt.Sprintf("empty revision label on %s pod", kasPod.Name)
+	}
+	revision, err := strconv.Atoi(revisionString)
+	if err != nil || revision < 0 {
+		return false, "InvalidPod", fmt.Sprintf("invalid revision label on pod %s: %q", kasPod.Name, revisionString)
+	}
+
+	if revision != monitorRevision {
+		return false, "InvalidPod", fmt.Sprintf("the running Kube API is at unexpected revision %d, expected %d", revision, monitorRevision)
```
sttts, line 204: UnexpectedRevision

PR #1181, opened 2021-07-19
Bug 1986003: Bump k8s.io to 1.22.0-rc.0

soltysh:


Makefile
```diff
@@ -81,8 +81,6 @@ clean:
 export TP_DEPLOYMENT_YAML ?=./manifests/0000_20_kube-apiserver-operator_06_deployment.yaml
 export TP_CMD_PATH ?=./cmd/cluster-kube-apiserver-operator
 
-$(call verify-golang-versions,Dockerfile.rhel7)
```
sttts, line 84: where is this duplicated?

soltysh: we already have that done inside build-machinery-go https://github.com/openshift/build-machinery-go/blob/master/make/golang.mk#L11

p0lyn0mial: line 37 - https://github.com/openshift/cluster-kube-apiserver-operator/blob/22f2a00c6293c3a979804651e4446a57ec331ed6/Makefile#L37

PR #1183, opened 2021-07-23
staticpod/installer: get rid of sleep in sync loop

sttts:
Vendor in https://github.com/openshift/library-go/pull/1145.


sttts:
/retest

infra

PR #1186, opened 2021-07-26
kube-apiserver: improve downtime by utilizing SO_REUSEPORT correctly

mfojtik:
The kube-apiserver runs with the SO_REUSEPORT option, which means the port it binds to can be reused by other processes (typically a new API server).

### DIFF
--- a/bindata/assets/kube-apiserver/pod.yaml
+++ b/bindata/assets/kube-apiserver/pod.yaml
@@ -22,13 +22,8 @@ spec:
       command: ['/usr/bin/timeout', '105', '/bin/bash', '-ec'] # a bit more than 60s for graceful termination + 35s for minimum-termination-duration, 5s extra cri-o's graceful termination period
       args:
       - |
-        echo -n "Fixing audit permissions."
+        echo -n "Fixing audit permissions ..."
         chmod 0700 /var/log/kube-apiserver && touch /var/log/kube-apiserver/audit.log && chmod 0600 /var/log/kube-apiserver/*
-        echo -n "Waiting for port :6443 and :6080 to be released."
-        while [ -n "$(ss -Htan '( sport = 6443 or sport = 6080 )')" ]; do
-          echo -n "."
-          sleep 1
-        done
       securityContext:
         privileged: true
       resources:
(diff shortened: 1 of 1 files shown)


squeed:
suggestion: try `ss -Htan state listening '( sport = 6443 or sport = 6080 )'` instead.

I wouldn't trust the return code from curl. Too fragile; it depends exactly on how the connection fails, and can change.

bindata/assets/kube-apiserver/pod.yaml
```diff
@@ -44,30 +39,18 @@ spec:
...
         - |
           LOCK=/var/log/kube-apiserver/.lock
-          echo -n "Acquiring exclusive lock ${LOCK}"
-          exec {LOCK_FD}>${LOCK} && flock -n "${LOCK_FD}" || {
+          echo -n "Acquiring exclusive lock ${LOCK} ..."
+          exec {LOCK_FD}>${LOCK} && flock --verbose -w 600 "${LOCK_FD}" || {
             echo "$(date -Iseconds -u) kubelet did not terminate old kube-apiserver before new one" >> /var/log/kube-apiserver/lock.log
             echo -n ": WARNING: kubelet did not terminate old kube-apiserver before new one."
-            # we didn't get an exclusive lock. We keep going with the risk to corrupt audit logs.
+            exit 1
```
mfojtik, line 46: @sttts this is here after 10minutes pass and kubelet has still not terminated the process

bindata/assets/kube-apiserver/pod.yaml
```diff
@@ -44,30 +39,34 @@ spec:
     args:
         - |
           LOCK=/var/log/kube-apiserver/.lock
-          echo -n "Acquiring exclusive lock ${LOCK}"
-          exec {LOCK_FD}>${LOCK} && flock -n "${LOCK_FD}" || {
+          echo -n "Acquiring exclusive lock ${LOCK} ..."
+
+          # Waiting for 30s max for old kube-apiserver process to exit.
+          # If 30s pass and we fail, wait additional 5s for the port to be released.
+          # Do not tweak these timings without considering the livenessProbe initialDelaySeconds
```
sttts, line 46:
```suggestion
# Waiting for 30s max for old kube-apiserver's watch-termination process to exit and remove the lock.
# Two cases:
# 1. if kubelet does not start the old and new in parallel (i.e. works as expected), the flock will always succeed without any time.
# 2. if kubelet does overlap old and new pods for up to 30s, the flock will wait and immediate return when the old finishes.
# Note: we can increase 30s for a bigger expected overlap. But a higher value means less noise about the broken kubelet behaviour, i.e. we hide a bug.
```

mfojtik: :+1:

mfojtik:
closing in favor of https://github.com/openshift/cluster-kube-apiserver-operator/pull/1191 because I believe that is better solution

PR #1188, opened 2021-07-27
Bug 1979303: clear encryption conditions when there is no work to be done

p0lyn0mial:
fixes https://bugzilla.redhat.com/show_bug.cgi?id=1979303 and https://bugzilla.redhat.com/show_bug.cgi?id=1982398


aravindhp:
Patch manager comment: PR messages says it fixes Bug 1979303 and Bug 1982398. Bug 1982398 is in MODIFIED and has a different set of PRs associated with it. So should I wait for Bug 1982398 to also be in POST in case all the PRs need to go in together?

aravindhp:
> Patch manager comment: PR messages says it fixes Bug 1979303 and Bug 1982398. Bug 1982398 is in MODIFIED and has a different set of PRs associated with it. So should I wait for Bug 1982398 to also be in POST in case all the PRs need to go in together?

Clarified over slack

PR #1192, opened 2021-07-29
pkg/operator/targetconfigcontroller: block rollout and report on missing external resource

hexfusion:
The current code reports the invalid resource but does not block a new revision from being created. In the case that configmap openshift-config-managed/csr-controller-ca is not managed by cluster-kube-controller-manager-operator block if SNO. This will ensure we don't rollout and cause disruption before critical resource generation.


hexfusion:
/retest

cc @soltysh

hexfusion:
> error: unable to read image registry.build02.ci.openshift.org/ci-op-g6h0r2i8/stable-initial@sha256:6c05d74eb1fa37a77ed9215d83933265564d661dd96fa765a3e6799340c6fae5: received unexpected HTTP status: 504 Gateway Time-out

/test e2e-upgrade

PR #1195, opened 2021-08-02
Bug 1985997: Enable startup-monitor for SNO

sttts:
Enable startup-monitor static pod for SNO topology as described in https://github.com/openshift/enhancements/pull/833.

This also leaves `unsupportedConfigOverrides.startupMonitor` in place for testing and development in HA cluster.

PR #1198, opened 2021-08-02
Bug 1985997: Enable static pod fallback logic for SNO, with disruptive e2e test

p0lyn0mial:
Replacing https://github.com/openshift/cluster-kube-apiserver-operator/pull/1195. Enabling https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/startup-monitor.md for single node.

cc @romfreiman @eranco74

### DIFF
--- a/test/e2e-sno-disruptive/sno_disruptive_test.go
+++ b/test/e2e-sno-disruptive/sno_disruptive_test.go
@@ -1,7 +1,240 @@
 package e2e_sno_disruptive
 
-import "testing"
+import (
+	"context"
+	"encoding/json"
+	"strconv"
+	"testing"
+	"time"
 
-func TestFallback(t *testing.T) {
-	t.Log("implement me")
+	"github.com/stretchr/testify/require"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/operator/staticpod/startupmonitor/annotations"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	commontesthelpers "github.com/openshift/library-go/test/library/encryption"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
+)
+
+func TestFallback(tt *testing.T) {
+	t := commontesthelpers.NewE(tt)
+	cs := getClients(t)
+
+	t.Log("Starting the fallback test")
+	clusterStateWaitPollTimeout, clusterMustBeReadyFor, waitForFallbackDegradedConditionTimeout := fallbackTimeoutsForCurrentPlatform(t, cs)
+
+	// before starting a new test make sure the current state of the cluster is good
+	ensureClusterInGoodState(t, cs, clusterStateWaitPollTimeout, clusterMustBeReadyFor)
+
+	// cause a disruption
+	cfg := getDefaultUnsupportedConfigForCurrentPlatform(t, cs)
+	cfg["apiServerArguments"] = map[string][]string{"non-existing-flag": {"true"}}
+	setUnsupportedConfig(t, cs, cfg)
+
+	// validate if the fallback condition is reported and the cluster is stable
+	waitForFallbackDegradedCondition(t, cs, waitForFallbackDegradedConditionTimeout)
+	nodeName, failedRevision := assertFallbackOnNodeStatus(t, cs)
+	assertKasPodAnnotatedOnNode(t, cs, failedRevision, nodeName)
+
+	// clean up
+	setUnsupportedConfig(t, cs, getDefaultUnsupportedConfigForCurrentPlatform(t, cs))
+	err := waitForClusterInGoodState(t, cs, clusterStateWaitPollTimeout, clusterMustBeReadyFor)
+	require.NoError(t, err)
+}
+
+// ensureClusterInGoodState makes sure the cluster is not progressing for mustBeReadyFor period
+// in addition in an HA env it applies getDefaultUnsupportedConfigForCurrentPlatform so that the feature is enabled before the tests starts
+func ensureClusterInGoodState(t testing.TB, cs clientSet, waitPollTimeout, mustBeReadyFor time.Duration) {
+	setUnsupportedConfig(t, cs, getDefaultUnsupportedConfigForCurrentPlatform(t, cs))
+	err := waitForClusterInGoodState(t, cs, waitPollTimeout, mustBeReadyFor)
+	require.NoError(t, err)
+}
+
+// waitForClusterInGoodState checks if the cluster is not progressing
+func waitForClusterInGoodState(t testing.TB, cs clientSet, waitPollTimeout, mustBeReadyFor time.Duration) error {
+	t.Helper()
+
+	startTs := time.Now()
+	t.Logf("Waiting %s for the cluster to be in a good condition, interval = 10s, timeout %v", mustBeReadyFor.String(), waitPollTimeout)
+
+	return wait.Poll(10*time.Second, waitPollTimeout, func() (bool, error) {
+		ckaso, err := cs.Operator.Get(context.TODO(), "cluster", metav1.GetOptions{})
+		if err != nil {
+			t.Log(err)
+			return false, nil /*retry*/
+		}
+
+		for _, ns := range ckaso.Status.NodeStatuses {
+			if ckaso.Status.LatestAvailableRevision != ns.CurrentRevision || ns.TargetRevision > 0 {
+				t.Logf("Node %s is progressing, latestAvailableRevision: %v, currentRevision: %v, targetRevision: %v", ns.NodeName, ckaso.Status.LatestAvailableRevision, ns.CurrentRevision, ns.TargetRevision)
+				return false, nil /*retry*/
+			}
+		}
+
+		if time.Since(startTs) > mustBeReadyFor {
+			t.Logf("The cluster has been in good condition for %s", mustBeReadyFor.String())
+			return true, nil /*done*/
+		}
+		return false, nil /*wait a bit more*/
+	})
+}
+
+// setUnsupportedConfig simply sets UnsupportedConfigOverrides config to the provided cfg
+func setUnsupportedConfig(t testing.TB, cs clientSet, cfg map[string]interface{}) {
+	t.Helper()
+
+	t.Logf("Setting UnsupportedConfigOverrides to %v", cfg)
+	raw, err := json.Marshal(cfg)
+	require.NoError(t, err)
+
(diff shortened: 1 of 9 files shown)


test/e2e-sno-disruptive/sno_disruptive_test.go
```diff
@@ -1,7 +1,261 @@
...
+	// TODO: check if pod is ready ?
+
+	// clean up
+	setUnsupportedConfig(t, cs, getDefaultUnsupportedConfig(t, cs))
+	if err := clusterInGoodState(t, cs, 10*time.Second, 6*time.Minute, 1*time.Minute); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func ensureClusterInGoodState(t testing.TB, cs clientSet, waitPollInterval, waitPollTimeout, mustBeReadyFor time.Duration) {
```
sttts, line 51: does it need waitPollInterval ?

test/e2e-sno-disruptive/sno_disruptive_test.go
```diff
@@ -1,7 +1,261 @@
...
+}
+
+func clusterInGoodState(t testing.TB, cs clientSet, waitPollInterval, waitPollTimeout, mustBeReadyFor time.Duration) error {
+	t.Helper()
+
+	startTs := time.Now()
+	t.Logf("Waiting %s for the cluster to be in a good condition, interval = %v, timeout %v", mustBeReadyFor.String(), waitPollInterval, waitPollTimeout)
+
+	return wait.Poll(waitPollInterval, waitPollTimeout, func() (bool, error) {
+		monitorPod, err := cs.Kube.CoreV1().Pods("openshift-kube-apiserver").Get(context.TODO(), "kube-apiserver-startup-monitor", metav1.GetOptions{})
```
sttts, line 67: these are static pod with another name (node appended)

test/e2e-sno-disruptive/sno_disruptive_test.go
```diff
@@ -1,7 +1,261 @@
...
+			t.Log("expected to find the kube-apiserver static pod but haven't found any")
+			return false, nil /*retry*/
+		}
+		for _, kasPod := range kasPods.Items {
+			if fallbackFor, ok := kasPod.Annotations[annotations.FallbackForRevision]; ok {
+				t.Logf("%v pod has %v annotation set to: %v", kasPod.Name, annotations.FallbackForRevision, fallbackFor)
+				return false, nil /*retry*/
+			}
+		}
+		// TODO: on an HA cluster we could also check if pods are on the same revision
```
sttts, line 91: I would consider a cluster no in-progress as good (i.e. nodeStatus has no transitioning pod). Why do we need all of this?

test/e2e-sno-disruptive/sno_disruptive_test.go
```diff
@@ -1,7 +1,261 @@
...
+		return false, nil /*wait a bit more*/
+	})
+}
+
+func setUnsupportedConfig(t testing.TB, cs clientSet, cfg map[string][]string) {
+	t.Helper()
+
+	t.Logf("Setting UnsupportedConfigOverrides to %v", cfg)
+	err := retry.OnError(retry.DefaultRetry, func(error) bool { return true }, func() error {
+		raw, err := json.Marshal(cfg)
```
sttts, line 116: nit: once outside of the loop is enough

test/e2e-sno-disruptive/sno_disruptive_test.go
```diff
@@ -1,7 +1,261 @@
...
+		return err
+	})
+	require.NoError(t, err)
+}
+
+func waitForFallbackDegradedCondition(t testing.TB, cs clientSet) {
+	t.Helper()
+
+	t.Log("Waiting for StaticPodFallbackRevisionDegraded condition, interval = 20s, timeout = 6min")
+	err := wait.Poll(20*time.Second, 6*time.Minute, func() (bool, error) {
```
sttts, line 140: why 6min?

test/e2e-sno-disruptive/sno_disruptive_test.go
```diff
@@ -1,7 +1,261 @@
...
+		ckaso, err := cs.Operator.Get(context.TODO(), "cluster", metav1.GetOptions{})
+		if err != nil {
+			t.Logf("unable to get kube-apiserver-operator resource: %v", err)
+			return false, nil /*retry*/
+		}
+
+		if v1helpers.IsOperatorConditionFalse(ckaso.Status.Conditions, "StaticPodFallbackRevisionDegraded") {
```
sttts, line 147: why not check for True and return true? Seems to be more direct in a loop like this.

test/e2e-sno-disruptive/sno_disruptive_test.go
```diff
@@ -1,7 +1,261 @@
...
+			t.Logf("StaticPodFallbackRevisionDegraded condition hasn't been set yet")
+			return false, nil /*retry*/
+		}
+
+		return true, nil /*done*/
+	})
+	require.NoError(t, err)
+}
+
+func assertNodeStatus(t testing.TB, cs clientSet) (string, int32) {
```
sttts, line 157: assert what?

test/e2e-sno-disruptive/sno_disruptive_test.go
```diff
@@ -1,7 +1,261 @@
...
+			nodeName = ns.NodeName
+			failedRevision = ns.LastFailedRevision
+		}
+	}
+
+	t.Logf("The fallback has been reported on node %v, failed revision is %v", nodeName, failedRevision)
+	return nodeName, failedRevision
+}
+
+func assertKasPodAnnotatedOnNode(t testing.TB, cs clientSet, expectedFailedRevision int32, expectedFallbackReason, expectedFallbackMessage, nodeName string) {
```
sttts, line 181: I don't think we have to check the message. It makes maintenance hard. We have unit tests for that.

If you really want to specific, just check for a keyword in the string (strings.Contains).

test/e2e-sno-disruptive/sno_disruptive_test.go
```diff
@@ -1,7 +1,261 @@
...
+	for _, potentialKasPods := range kasPods {
+		if potentialKasPods.Spec.NodeName == currentNodeName {
+			filteredKasPods = append(filteredKasPods, potentialKasPods)
+		}
+	}
+
+	return filteredKasPods
+}
+
+func getDefaultUnsupportedConfig(t testing.TB, cs clientSet) map[string][]string {
```
sttts, line 249: not sure about this func. Name is unspecific, and it does abstract something that's better explicit.

p0lyn0mial: `getPlatformSpecificConfig` would be better?

test/e2e-sno-disruptive/sno_disruptive_test.go
p0lyn0mial: what's sure as well, there might be cases that would break etcd, I can remove.

test/e2e-sno-disruptive/sno_disruptive_test.go
p0lyn0mial: just playing safe, checking if `nodeStatus` contain only `currentRevision` and `nodeName` would be enough?

test/e2e-sno-disruptive/sno_disruptive_test.go
p0lyn0mial: do you want to use `wait.UntilWithContext` ?

sttts: no, would just make this constant. 10s or 30s.

test/e2e-sno-disruptive/sno_disruptive_test.go
sttts: I think it is enough to start with. We also don't want to start from some "dirty" state.

test/e2e-sno-disruptive/sno_disruptive_test.go
sttts: I think the infra check is nothing one should hide in a func

test/e2e-sno-disruptive/sno_disruptive_test.go
```diff
@@ -1,7 +1,243 @@
...
+	if err := clusterInGoodState(t, cs, waitPollTimeout, mustBeReadyFor); err != nil {
+		t.Logf("Trying to bring the cluster back to normal state by removing all unsupportedConfigOverrides entries, err: %v", err)
+		setUnsupportedConfig(t, cs, getDefaultUnsupportedConfigForCurrentPlatform(t, cs))
+		err := clusterInGoodState(t, cs, waitPollTimeout, mustBeReadyFor)
+		require.NoError(t, err)
+	}
+}
+
+// clusterInGoodState checks if the cluster is stable: is not progressing and doesn't have a failed revision for mustBeReadyFor period
+func clusterInGoodState(t testing.TB, cs clientSet, waitPollTimeout, mustBeReadyFor time.Duration) error {
```
sttts, line 65: waitForClusterInGoodState

test/e2e-sno-disruptive/sno_disruptive_test.go
```diff
@@ -1,7 +1,243 @@
...
+	// before starting a new test make sure the current state of the cluster is good
+	ensureClusterInGoodState(t, cs, clusterStateWaitPollTimeout, clusterMustBeReadyFor)
+
+	// cause a disruption
+	cfg := getDefaultUnsupportedConfigForCurrentPlatform(t, cs)
+	cfg["apiServerArguments"] = map[string][]string{"non-existing-flag": {"true"}}
+	setUnsupportedConfig(t, cs, cfg)
+
+	// validate if the fallback condition is reported and the cluster is stable
+	waitForFallbackDegradedCondition(t, cs, waitForFallbackDegradedConditionTimeout)
```
sttts, line 40: can't we derive `waitForFallbackDegradedConditionTimeout` from some per-platform number? E.g. `oneNodeRolloutTimeout`

p0lyn0mial: added `TODO`

test/e2e-sno-disruptive/sno_disruptive_test.go
p0lyn0mial: done.

p0lyn0mial:
pulled https://github.com/openshift/library-go/pull/1177
please retag

sttts:
flakes:
```
: [sig-arch][Early] Managed cluster should start all core operators [Skipped:Disconnected] [Suite:openshift/conformance/parallel] expand_less | 0s
-- | --
fail [github.com/onsi/ginkgo@v4.7.0-origin.0+incompatible/internal/leafnodes/runner.go:113]: Aug  4 16:40:38.918: Some cluster operators are not ready: authentication (Degraded=True OAuthServerDeployment_UnavailablePod: OAuthServerDeploymentDegraded: 1 of 3 requested instances are unavailable for oauth-openshift.openshift-authentication ())
```

PR #1199, opened 2021-08-02
Bug 1998552: Enforce OpenShift's defined kubelet version skew policies

sanchezl:
Add `KubeletVersionSkewController` in support of enhancement user story: [APIServer - Enforce OpenShift's defined kubelet version skew policies](https://github.com/openshift/enhancements/blob/master/enhancements/update/eus-upgrades-mvp.md)

### Condition

```go
// KubeletVersionSkewController sets Upgradeable=False if the kubelet
// version on a node prevents upgrading to a supported OCP version.
//
// For odd OCP minor versions, kubelet versions 0 or 1 minor version
// behind the API server version are supported.
//
// For even OCP minor versions, kubelet versions 0, 1, or 2 minor
// versions behind the API server version are supported.
const KubletVersionSkewLimitUpgradeable = "KubletVersionSkewLimitUpgradeable"
```

### Reasons

Reason | Upgradeable? | Description
-- | -- | --
KubeletVersionUnknown | Unknown | Error determining kubelet version.
KubeletMinorVersionsSynced | True | Kubelet and API server version minor versions are synced.
KubeletMinorVersionSupportedNextUpgrade | True | Kubelet version on a node will be supported in the next OCP minor version upgrade.
KubeletMinorVersionUnsupportedNextUpgrade | False | Kubelet version on a node will not be supported in the next OCP minor version upgrade.
KubeletMinorVersionUnsupported | False | Unsupported kubelet minor version on a node is too far behind the expected API server version.
KubeletMinorVersionAhead | Unknown | Unsupported kubelet minor version on a node is newer than expected API server version.

### Kublet minor version skew limits

* If OCP minor version is odd, kubelet versions older than 1 minor version of the API server version set `Upgradeable=False`.
* If OCP minor version is even, kubelet versions must be in sync with API server version in order to set `Upgradeable=True`.

### Supported kublet minor versions

A re-statement of the of the kubelet minor version skew limits above as supported kubelet minor versions:

* If OCP minor version is odd, kubelet versions 0 or 1 minor version behind the API server version are supported.
* If OCP minor version is even, kubelet versions 0, 1, or 2 minor versions behind the API server version are supported.
 
### Upstream kubelet compatibility

For reference, here is the upstream specification for kubelet/api server compatibility:

* https://kubernetes.io/releases/version-skew-policy/#kubelet

### DIFF
--- a/pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
+++ b/pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
@@ -1,0 +1,255 @@
+package kubeletversionskewcontroller
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"sort"
+	"strings"
+
+	"github.com/blang/semver"
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/management"
+	"github.com/openshift/library-go/pkg/operator/status"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/runtime"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+)
+
+const (
+	KubeletMinorVersionUpgradeableConditionType = "KubeletMinorVersionUpgradeable"
+
+	KubeletVersionUnknownReason                     = "KubeletVersionUnknown"
+	KubeletMinorVersionSyncedReason                 = "KubeletMinorVersionsSynced"
+	KubeletMinorVersionSupportedNextUpgradeReason   = "KubeletMinorVersionSupportedNextUpgrade"
+	KubeletMinorVersionUnsupportedNextUpgradeReason = "KubeletMinorVersionUnsupportedNextUpgrade"
+	KubeletMinorVersionUnsupportedReason            = "KubeletMinorVersionUnsupported"
+	KubeletMinorVersionAheadReason                  = "KubeletMinorVersionAhead"
+)
+
+// KubeletVersionSkewController sets Upgradeable=False if the kubelet
+// version on a node prevents upgrading to a supported OpenShift version.
+//
+// For odd OpenShift minor versions, kubelet versions 0 or 1 minor
+// versions behind the API server version are supported.
+//
+// For even OpenShift minor versions, kubelet versions 0, 1, or 2
+// minor versions behind the API server version are supported.
+type KubeletVersionSkewController interface {
+	factory.Controller
+}
+
+func NewKubeletVersionSkewController(
+	operatorClient v1helpers.OperatorClient,
+	kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces,
+	recorder events.Recorder,
+) *kubeletVersionSkewController {
+	openShiftVersion := semver.MustParse(status.VersionForOperatorFromEnv())
+	nextOpenShiftVersion := semver.Version{Major: openShiftVersion.Major, Minor: openShiftVersion.Minor + 1}
+	c := &kubeletVersionSkewController{
+		operatorClient:              operatorClient,
+		nodeLister:                  kubeInformersForNamespaces.InformersFor("").Core().V1().Nodes().Lister(),
+		apiServerVersion:            semver.MustParse(status.VersionForOperandFromEnv()),
+		minSupportedSkew:            minSupportedKubeletSkewForOpenShiftVersion(openShiftVersion),
+		minSupportedSkewNextVersion: minSupportedKubeletSkewForOpenShiftVersion(nextOpenShiftVersion),
+	}
+	c.Controller = factory.New().
+		WithSync(c.sync).
+		WithInformers(kubeInformersForNamespaces.InformersFor("").Core().V1().Nodes().Informer()).
+		ToController("KubeletVersionSkewController", recorder.WithComponentSuffix("kubelet-version-skew-controller"))
+	return c
+}
+
+func minSupportedKubeletSkewForOpenShiftVersion(v semver.Version) int {
+	switch v.Minor % 2 {
+	case 0: // even OpenShift versions
+		return -2
+	case 1: // odd OpenShift versions
+		return -1
+	default:
+		panic("should not happen")
+	}
+}
+
+type kubeletVersionSkewController struct {
+	factory.Controller
+	operatorClient              v1helpers.OperatorClient
+	nodeLister                  corev1listers.NodeLister
+	apiServerVersion            semver.Version
+	minSupportedSkew            int
+	minSupportedSkewNextVersion int
+}
+
+func (c *kubeletVersionSkewController) sync(_ context.Context, _ factory.SyncContext) error {
+	operatorSpec, _, _, err := c.operatorClient.GetOperatorState()
+	if err != nil {
+		return err
+	}
+	if !management.IsOperatorManaged(operatorSpec.ManagementState) {
+		return nil
+	}
+
+	nodes, err := c.nodeLister.List(labels.Everything())
+	if err != nil {
+		return err
+	}
+	sort.Sort(byName(nodes))
+
+	var errors nodeKubeletInfos
+	var skewedUnsupported nodeKubeletInfos
+	var skewedLimit nodeKubeletInfos
+	var skewedButOK nodeKubeletInfos
+	var synced nodeKubeletInfos
+	var unsupported nodeKubeletInfos
+
+	// for each node, check kubelet version
+	for _, node := range nodes {
+		kubeletVersion, err := nodeKubeletVersion(node)
+		if err != nil {
+			runtime.HandleError(fmt.Errorf("unable to determine kubelet version on node %s: %w", node.Name, err))
+			errors = append(errors, nodeKubeletInfo{node: node.Name, err: err})
+			continue
+		}
+		skew := int(kubeletVersion.Minor - c.apiServerVersion.Minor)
+		// Assume that an OpenShift minor version upgrade also bumps to the next kube minor version. Revisit
+		// this in the future if an OpenShift minor version upgrade ever skips or repeats a kube minor version.
+		skewNextVersion := skew - 1
+		switch {
+		case skew == 0:
+			// synced
+			synced = append(synced, nodeKubeletInfo{node: node.Name, version: &kubeletVersion})
+		case skew < c.minSupportedSkew:
+			// already in an unsupported state
+			skewedUnsupported = append(skewedUnsupported, nodeKubeletInfo{node: node.Name, version: &kubeletVersion})
+		case skewNextVersion < c.minSupportedSkewNextVersion:
+			// upgrading to next minor version of API server would result in an unsupported config
+			skewedLimit = append(skewedLimit, nodeKubeletInfo{node: node.Name, version: &kubeletVersion})
+		case skew < 0:
+			// behind, but upgrading to next minor version of API server is supported
+			skewedButOK = append(skewedButOK, nodeKubeletInfo{node: node.Name, version: &kubeletVersion})
+		default:
+			// kubelet version newer than api server version. possibly in the middle of a rollback.
+			unsupported = append(unsupported, nodeKubeletInfo{node: node.Name, version: &kubeletVersion})
+		}
+	}
+
+	condition := operatorv1.OperatorCondition{Type: KubeletMinorVersionUpgradeableConditionType}
+	// use the most "severe" reason to set the condition status
+	switch {
+	case len(skewedUnsupported) > 0:
+		condition.Reason = KubeletMinorVersionUnsupportedReason
+		condition.Status = operatorv1.ConditionFalse
+		switch len(skewedUnsupported) {
+		case 1:
+			condition.Message = fmt.Sprintf("Unsupported kubelet minor version (%v) on node %s is too far behind the target API server version (%v).", skewedUnsupported.version(), skewedUnsupported.nodes(), c.apiServerVersion)
+		case 2, 3:
+			condition.Message = fmt.Sprintf("Unsupported kubelet minor versions on nodes %s are too far behind the target API server version (%v).", skewedUnsupported.nodes(), c.apiServerVersion)
+		default:
+			condition.Message = fmt.Sprintf("Unsupported kubelet minor versions on %d nodes are too far behind the target API server version (%v).", len(skewedUnsupported), c.apiServerVersion)
+		}
+	case len(unsupported) > 0:
+		condition.Reason = KubeletMinorVersionAheadReason
+		condition.Status = operatorv1.ConditionUnknown
+		switch len(unsupported) {
+		case 1:
+			condition.Message = fmt.Sprintf("Unsupported kubelet minor version (%v) on node %s is ahead of the target API server version (%v).", unsupported.version(), unsupported.nodes(), c.apiServerVersion)
+		case 2, 3:
+			condition.Message = fmt.Sprintf("Unsupported kubelet minor versions on nodes %s are ahead of the target API server version (%v).", unsupported.nodes(), c.apiServerVersion)
+		default:
+			condition.Message = fmt.Sprintf("Unsupported kubelet minor versions on %d nodes are ahead of the target API server version (%v).", len(unsupported), c.apiServerVersion)
+		}
+	case len(errors) > 0:
+		condition.Reason = KubeletVersionUnknownReason
+		condition.Status = operatorv1.ConditionUnknown
+		switch len(errors) {
+		case 1:
+			condition.Message = fmt.Sprintf("Unable to determine the kubelet version on node %s: %v", errors.nodes(), errors.error())
+		case 2, 3:
+			condition.Message = fmt.Sprintf("Unable to determine the kubelet version on nodes %s.", errors.nodes())
+		default:
+			condition.Message = fmt.Sprintf("Unable to determine the kubelet version on %d nodes.", len(errors))
+		}
+	case len(skewedLimit) > 0:
+		condition.Reason = KubeletMinorVersionUnsupportedNextUpgradeReason
+		condition.Status = operatorv1.ConditionFalse
+		switch len(skewedLimit) {
+		case 1:
+			condition.Message = fmt.Sprintf("Kubelet minor version (%v) on node %s will not be supported in the next OpenShift minor version upgrade.", skewedLimit.version(), skewedLimit.nodes())
+		case 2, 3:
+			condition.Message = fmt.Sprintf("Kubelet minor versions on nodes %s will not be supported in the next OpenShift minor version upgrade.", skewedLimit.nodes())
+		default:
+			condition.Message = fmt.Sprintf("Kubelet minor versions on %d nodes will not be supported in the next OpenShift minor version upgrade.", len(skewedLimit))
+		}
+	case len(skewedButOK) > 0:
+		condition.Reason = KubeletMinorVersionSupportedNextUpgradeReason
+		condition.Status = operatorv1.ConditionTrue
+		switch len(skewedButOK) {
+		case 1:
+			condition.Message = fmt.Sprintf("Kubelet minor version (%v) on node %s is behind the expected API server version; nevertheless, it will continue to be supported in the next OpenShift minor version upgrade.", skewedButOK.version(), skewedButOK.nodes())
+		case 2, 3:
+			condition.Message = fmt.Sprintf("Kubelet minor versions on nodes %s are behind the expected API server version; nevertheless, they will continue to be supported in the next OpenShift minor version upgrade.", skewedButOK.nodes())
+		default:
+			condition.Message = fmt.Sprintf("Kubelet minor versions on %d nodes are behind the expected API server version; nevertheless, they will continue to be supported in the next OpenShift minor version upgrade.", len(skewedButOK))
+		}
+	default:
+		condition.Reason = KubeletMinorVersionSyncedReason
+		condition.Status = operatorv1.ConditionTrue
+		condition.Message = "Kubelet and API server minor versions are synced."
+	}
+
+	_, _, err = v1helpers.UpdateStatus(c.operatorClient, v1helpers.UpdateConditionFn(condition))
+	return err
+}
+
+type nodeKubeletInfo struct {
+	node    string
+	version *semver.Version
+	err     error
+}
+
+type nodeKubeletInfos []nodeKubeletInfo
+
+func (n nodeKubeletInfos) nodes() string {
+	var s []string
+	for _, i := range n {
+		s = append(s, i.node)
+	}
+	switch len(s) {
+	case 0, 1:
+	case 2:
+		return strings.Join(s, " and ")
+	default:
+		s[len(s)-1] = "and " + s[len(s)-1]
+	}
+	return strings.Join(s, ", ")
+}
+
+func (n nodeKubeletInfos) error() error {
+	if len(n) > 0 {
+		return n[0].err
+	}
+	return nil
+}
+
+func (n nodeKubeletInfos) version() *semver.Version {
+	if len(n) > 0 {
+		return n[0].version
+	}
+	return nil
+}
+
+func nodeKubeletVersion(node *corev1.Node) (semver.Version, error) {
+	return semver.Parse(strings.TrimPrefix(node.Status.NodeInfo.KubeletVersion, "v"))
+}
+
+var byNodeRegexp = regexp.MustCompile(`node [^ ]*`)
+
+type byName []*corev1.Node
+
+func (n byName) Len() int           { return len(n) }
+func (n byName) Swap(i, j int)      { n[i], n[j] = n[j], n[i] }
+func (n byName) Less(i, j int) bool { return strings.Compare(n[i].Name, n[j].Name) < 0 }
--- a/pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller_test.go
+++ b/pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller_test.go
@@ -1,0 +1,180 @@
+package kubeletversionskewcontroller
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/blang/semver"
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/diff"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
+)
+
+func Test_kubeletVersionSkewController_Sync(t *testing.T) {
+
+	evenOpenShiftVersion := "4.8.0"
+	oddOpenShiftVersion := "4.9.0"
+	apiServerVersion := "1.21.1"
+	skewedKubeletVersions := func(s ...int) []string {
+		var v []string
+		for i, skew := range s {
+			v = append(v, fmt.Sprintf("1.%d.%d", 21+skew, i))
+		}
+		return v
+	}
+
+	testCases := []struct {
+		name             string
+		ocpVersion       string
+		kubeletVersions  []string
+		expectedStatus   operatorv1.ConditionStatus
+		expectedReason   string
+		expectedMsgLines string
+	}{
+		{
+			name:             "Synced/Even",
+			ocpVersion:       evenOpenShiftVersion,
+			kubeletVersions:  skewedKubeletVersions(0, 0, 0),
+			expectedStatus:   operatorv1.ConditionTrue,
+			expectedReason:   KubeletMinorVersionSyncedReason,
+			expectedMsgLines: "Kubelet and API server minor versions are synced.",
+		},
+		{
+			name:             "Synced/Odd",
+			ocpVersion:       oddOpenShiftVersion,
+			kubeletVersions:  skewedKubeletVersions(0, 0, 0),
+			expectedStatus:   operatorv1.ConditionTrue,
+			expectedReason:   KubeletMinorVersionSyncedReason,
+			expectedMsgLines: "Kubelet and API server minor versions are synced.",
+		},
+		{
+			name:             "ErrorParsingKubeletVersion",
+			ocpVersion:       oddOpenShiftVersion,
+			kubeletVersions:  []string{"Invalid", "1.21.2", "1.20.3"},
+			expectedStatus:   operatorv1.ConditionUnknown,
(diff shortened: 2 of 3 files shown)


pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
```diff
@@ -0,0 +1,138 @@
...
+
+	KubeletVersionUnknownReason           = "KubeletVersionUnknown"
+	KubeletVersionSyncedReason            = "KubeletVersionSynced"
+	KubeletVersionSkewLessThanLimitReason = "KubeletVersionSkewLessThanLimit"
+	KubeletVersionSkewAtLimitReason       = "KubeletVersionSkewAtLimit"
+	KubeletVersionUnsupportedReason       = "KubeletVersionUnsupported"
+)
+
+// KubeletVersionSkewController sets Upgradeable=False if the kubelet
+// version on a node falls too far behind the api server version.
```
sttts, line 31: what is "too far behind" ?

sanchezl: More than 1 version (N-1) behind.

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
```diff
@@ -0,0 +1,138 @@
...
+	kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces,
+	recorder events.Recorder,
+) *kubeletVersionSkewController {
+	c := &kubeletVersionSkewController{
+		operatorClient:   operatorClient,
+		nodeLister:       kubeInformersForNamespaces.InformersFor("").Core().V1().Nodes().Lister(),
+		apiServerVersion: semver.MustParse(status.VersionForOperandFromEnv()),
+	}
+	// check if major version updated and no one updated this code
+	if c.apiServerVersion.Minor < 2 {
```
sttts, line 47: what does 2 mean here?

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
```diff
@@ -0,0 +1,138 @@
...
+		panic(fmt.Errorf("unable to determine kubelet version skew limit"))
```
tkashem, line 50: i am not clear on what we do here - can you explain a bit?

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
```diff
@@ -0,0 +1,138 @@
...
+	var errorMsgs []string
+	var skewedToTheLimitMsgs []string
+	var skewedButOKMsgs []string
+	var syncedMsgs []string
+	var unsupportedMsgs []string
+
+	// for each node, check kubelet version
+	for _, node := range nodes {
+		kubeletVersion, err := nodeKubeletVersion(node)
+		skew := int(kubeletVersion.Minor - c.apiServerVersion.Minor)
```
tkashem, line 89: what if `kubeletVersion` is nil due to an `err`? If we always have a `kubeletVersion` then maybe we want `nodeKubeletVersion(node *corev1.Node) semver.Version`?

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
```diff
@@ -0,0 +1,138 @@
...
+		switch {
+		case err != nil:
+			errorMsgs = append(errorMsgs, fmt.Sprintf("unable to determine kubelet version on node %s: %s", node.Name, err))
```
tkashem, line 92: nit: `Unable`

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
```diff
@@ -0,0 +1,138 @@
...
+		condition.Status = operatorv1.ConditionTrue
+	default:
+		condition.Reason = KubeletVersionSyncedReason
+		condition.Status = operatorv1.ConditionTrue
+		condition.Message = "Kubelet and API server versions are synced."
+	}
+
+	_, _, err = v1helpers.UpdateStatus(c.operatorClient, v1helpers.UpdateConditionFn(condition))
+	return err
+
```
tkashem, line 131: nit: extra new line

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
```diff
@@ -0,0 +1,138 @@
...
+func NewKubeletVersionSkewController(
+	operatorClient v1helpers.OperatorClient,
+	kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces,
+	recorder events.Recorder,
+) *kubeletVersionSkewController {
+	c := &kubeletVersionSkewController{
+		operatorClient:   operatorClient,
+		nodeLister:       kubeInformersForNamespaces.InformersFor("").Core().V1().Nodes().Lister(),
+		apiServerVersion: semver.MustParse(status.VersionForOperandFromEnv()),
```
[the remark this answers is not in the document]
sanchezl: Comes from the operator's (in this case kas-o) `OPERAND_IMAGE_VERSION`  environment variable which is in turn updated by the CVO(?).

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
sanchezl: Kubelet version on a node is two minor versions older than the API server. Updated doc with clarification.

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
sanchezl: Added comment to clarify.

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
```diff
@@ -0,0 +1,140 @@
...
+
+	// for each node, check kubelet version
+	for _, node := range nodes {
+		kubeletVersion, err := nodeKubeletVersion(node)
+		skew := int(kubeletVersion.Minor - c.apiServerVersion.Minor)
+		switch {
+		case err != nil:
+			errorMsgs = append(errorMsgs, fmt.Sprintf("unable to determine kubelet version on node %s: %s", node.Name, err))
+		case skew <= -2: // behind too far
+			skewedToTheLimitMsgs = append(skewedToTheLimitMsgs, fmt.Sprintf("Kubelet version (%s) on node %s is behind the expected API server version (%s) by at least the maximum allowable skew.", kubeletVersion, node.Name, c.apiServerVersion))
```
tkashem, line 94: kubelet.minor = `1` and apiserver.minor = `4` -> skew = 1 - 4 = -3.  `-3 <= -2 = true` so is `-2 <= -2` . Are we putting skew of -3 and -2 in the same category as `skewedToTheLimitMsgs`. I thought a skew of `2` is acceptable.

I thought it would be 
```
switch {
case skew > 0:
  // kubelet version is newer, rollack
case skew == 0:
  // synced
case skew >= -2:
  // skew within limit of 2
default:
  // too far
}
```

May be add some comments to clarify this?

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
```diff
@@ -0,0 +1,140 @@
...
+		case skew == 0: // synced
+			syncedMsgs = append(syncedMsgs, fmt.Sprintf("Kublet version on node %s is in synce with the expected API server version.", node.Name))
+		default: // kubelet version newer than api server version. possibly a rollback.
+			unsupportedMsgs = append(unsupportedMsgs, fmt.Sprintf("Kubelet version (%s) on node %s is ahead of the expected API server version (%s).", kubeletVersion, node.Name, c.apiServerVersion))
+		}
+	}
+
+	// update status
+
+	// combine all msgs
```
tkashem, line 106: nit: do we need the above comments - it seems obvious

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
```diff
@@ -0,0 +1,140 @@
...
+	_, _, err = v1helpers.UpdateStatus(c.operatorClient, v1helpers.UpdateConditionFn(condition))
+	return err
+
+}
+
+func nodeKubeletVersion(node *corev1.Node) (semver.Version, error) {
+	return semver.Parse(strings.TrimPrefix(node.Status.NodeInfo.KubeletVersion, "v"))
+}
+
+func (c *kubeletVersionSkewController) calculateSkew(kubeletVersion semver.Version) uint64 {
```
tkashem, line 138: `calculateSkew` is not being used, it seems.

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
sanchezl: > Are we putting skew of -3 and -2 in the same category as `skewedToTheLimitMsgs`. I thought a skew of `2` is acceptable.

Skew of `-2` is supported, but this is about preventing you from upgrading to the the **next** minor version of API server, in which case you would end up in an unsupported state, thus the `-2` is included in the same case as the `-3`.

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
sanchezl: Updated.

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
sanchezl: Updated.

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
sanchezl: Added clarifications to the comments.

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
sanchezl: Deleted.

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
sanchezl: Deleted.

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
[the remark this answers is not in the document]
wking: actually via `oc`.  Because the operator's deployment manifest has some [`0.0.1-snapshot` placeholders][1], and [those are replaced with the version string by `oc` at release-image-build time][2] ([code][3]).

[1]: https://github.com/openshift/cluster-openshift-apiserver-operator/blob/4da70a9e816206146c43bd4de0d6f7655a52d9a0/manifests/0000_30_openshift-apiserver-operator_07_deployment.yaml#L54-L57
[2]: https://github.com/openshift/enhancements/blob/f97876821d3bb506d28fee565271d3bebbbc682c/dev-guide/cluster-version-operator/dev/clusteroperator.md#what-should-be-the-contents-of-clusteroperator-custom-resource-in-manifests
[3]: https://github.com/openshift/oc/blob/f94afb52dc8a3185b3b9eacaf92ec34d80f8708d/pkg/cli/admin/release/image_mapper.go#L324-L334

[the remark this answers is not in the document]
s-urbaniak: @wking out of curiosity: How and where do those `0.0.1-snapshot` mappings translate to environment variables? I think the `OPERATOR_IMAGE_VERSION` env var is derived from the constant name `operator` in https://github.com/openshift/cluster-openshift-apiserver-operator/blob/4da70a9e816206146c43bd4de0d6f7655a52d9a0/manifests/0000_30_openshift-apiserver-operator_08_clusteroperator.yaml#L12-L13

But how is `OPERAND_IMAGE_VERSION` being derived if theoretically multiple operand versions can be specified https://github.com/openshift/cluster-openshift-apiserver-operator/blob/4da70a9e816206146c43bd4de0d6f7655a52d9a0/manifests/0000_30_openshift-apiserver-operator_08_clusteroperator.yaml#L14-L15 ?

I didn't find the code base for setting those env vars, do you have a pointer?

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
```diff
@@ -0,0 +1,135 @@
...
+	var syncedMsgs []string
+	var unsupportedMsgs []string
+
+	// for each node, check kubelet version
+	for _, node := range nodes {
+		kubeletVersion, err := nodeKubeletVersion(node)
+		var skew int
+		if err == nil {
+			skew = int(kubeletVersion.Minor - c.apiServerVersion.Minor)
+		}
```
tkashem, line 92: does this seem a little bit more concise?
```
kubeletVersion, err := nodeKubeletVersion(node)
if err != nil {
   errorMsgs = append(errorMsgs, fmt.Sprintf("Unable to determine kubelet version on node %s: %s", node.Name, err))
   continue
}

skew := int(kubeletVersion.Minor - c.apiServerVersion.Minor)
switch {
...
}
```

tkashem:
LGTM

(will let others to give their feedback)

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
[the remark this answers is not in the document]
sanchezl: @s-urbaniak It's just a simple string replacement. In your example, it is used twice since the versions there should always be in sync and match the overall OCP release version. Another common replacement is for `1.0.0-snapshot-kubernetes` which corresponds to the upstream `kubernetes` "component" version. I think the release version and the various component versions (of which I only know of `kubernetes`) are extracted from the release's ImageStream. Hopefully @wking can clarify.

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
sanchezl: Sure does :-) 
Done.

sanchezl:
Reviewers, please see the description for the change to a skew policy based on even/odd OCP versions. (FYI @sdodson)

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
sanchezl: Updated GoDoc with new even/odd skew policy details.

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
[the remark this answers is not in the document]
wking: It's not my code, so I don't have a complete handle on this, but the godocs for [the code I linked earlier][1] seem reasonably clear.  They say `0.0.1-snapshot-[component]` though, so it would be `0.0.1-snapshot-kubernetes`, etc., not `1.0.0-snapshot-kubernetes`.  [Looks][2] [like][3] the component versions are pulled out of a `io.openshift.build.versions` annotation in the ImageStream feeding the release image (and maybe from the image config for stuff that's not sucked in off an ImageStream?).  Spot-checking 4.8.5:

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-upgrade/1425332773236772864/artifacts/release/artifacts/release-images-latest | jq -r '.spec.tags[] | .annotations["io.openshift.build.versions"] as $v | select($v != null) | .name + " " + $v'
hyperkube kubernetes=1.21.1
machine-os-content machine-os=48.84.202108062347-0
```

[1]: https://github.com/openshift/oc/blob/f94afb52dc8a3185b3b9eacaf92ec34d80f8708d/pkg/cli/admin/release/image_mapper.go#L324-L334
[2]: https://github.com/openshift/oc/blob/f94afb52dc8a3185b3b9eacaf92ec34d80f8708d/pkg/cli/admin/release/image_mapper.go#L174-L188
[3]: https://github.com/openshift/oc/blob/f94afb52dc8a3185b3b9eacaf92ec34d80f8708d/pkg/cli/admin/release/annotations.go#L19-L24

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
```diff
@@ -0,0 +1,163 @@
...
+	for _, node := range nodes {
+		kubeletVersion, err := nodeKubeletVersion(node)
+		if err != nil {
+			errorMsgs = append(errorMsgs, fmt.Sprintf("Unable to determine kubelet version on node %s: %s", node.Name, err))
+			continue
+		}
+		skew := int(kubeletVersion.Minor - c.apiServerVersion.Minor)
+		switch {
+		case skew == 0: // synced
+			syncedMsgs = append(syncedMsgs, fmt.Sprintf("Kubelet version on node %s is in synced with the expected API server version.", node.Name))
```
wking, line 111: probably point out here (and possibly also in the `KubeletVersionSynced` reason) that "synced" means major/minor match, but does not imply a patch match.  Because we don't want to give folks with 1.21.1 kubelets around the false impression that they are completely caught up with a cluster that would prefer 1.21.99 kubelets.  The APIs will be synchronized, but the implementations will differ in bug exposure.  Maybe `KubeletAPIVersionSynced` or some such might be enough to get the focus onto the API and off of the implementations?  Or something else?

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
wking: Currently this case includes both "still supported, but `Upgradeable=False` because bumping would leave you unsupported" and "you are unsupported right now".  I'm trying to wrap my head around what I think I want instead...

wking: I think we need to directly address the expected Kube-API version of the next OpenShift minor, unless we can just assume that each OpenShift minor bump will come with a matching Kube-API bump.  Even if we do assume that, writing it down in a line of Go would make things more clear:
 
 ```go
expectedSkewVsNextOpenShiftMinor := skew - 1 // assume that every OpenShift minor bump includes a Kube API minor bump
```

Then:

* `skew == 0`: synced.  Fill `syncedMsgs`.
* `skew < c.minUpgradeableSkew`: so far back it's currently unsupported.  Fill `unsupportedMsgs`.
* `expectedSkewVsNextOpenShiftMinor <= c.minUpgradeableSkewNextMinor`: upgrading to next minor version of API server would result in an unsupported config.  Fill `skewedToTheLimitMsgs`.
* `skew < 0`: behind, but upgrading to next minor version of API server is supported.  Fill `skewedButOKMsgs`.
* `default`: kubelet version newer than API server version.  Possibly in the middle of a rollback.  Fill `unsupportedMsgs`.

Working some examples, fiddling with the even/odd `minUpgradeableSkew` settings vs. your current `minUpgradeableKubeletSkewForOCPVersion` implementation:

* OCP 4.9, `minUpgradeableSkew = -1; minUpgradeableSkewNextMinor = -2`.  API 1.22 (expected next API 1.23).  Kubelet 1.22, `skew = 0; expectedSkewVsNextOpenShiftMinor = -1`.  `syncedMsgs`.
* OCP 4.9, `minUpgradeableSkew = -1; minUpgradeableSkewNextMinor = -2`.  API 1.22 (expected next API 1.23).  Kubelet 1.21, `skew = -1; expectedSkewVsNextOpenShiftMinor = -2`.  `skewedButOKMsgs`.
* OCP 4.9, `minUpgradeableSkew = -1; minUpgradeableSkewNextMinor = -2`.  API 1.22 (expected next API 1.23).  Kubelet 1.20, `skew = -2; expectedSkewVsNextOpenShiftMinor = -3`.  `unsupportedMsgs`.
* OCP 4.8, `minUpgradeableSkew = -2; minUpgradeableSkewNextMinor = -1`.  API 1.21 (expected next API 1.22).  Kubelet 1.21, `skew = 0; expectedSkewVsNextOpenShiftMinor = -1`.  `syncedMsgs`.
* OCP 4.8, `minUpgradeableSkew = -2; minUpgradeableSkewNextMinor = -1`.  API 1.21 (expected next API 1.22).  Kubelet 1.20, `skew = -1; expectedSkewVsNextOpenShiftMinor = -2`.  `skewedToTheLimitMsgs`.
* OCP 4.8, `minUpgradeableSkew = -2; minUpgradeableSkewNextMinor = -1`.  API 1.21 (expected next API 1.22).  Kubelet 1.19, `skew = -2; expectedSkewVsNextOpenShiftMinor = -3`.  `skewedToTheLimitMsgs`.
* OCP 4.8, `minUpgradeableSkew = -2; minUpgradeableSkewNextMinor = -1`.  API 1.21 (expected next API 1.22).  Kubelet 1.18, `skew = -3; expectedSkewVsNextOpenShiftMinor = -4`.  `unsupported`.

File: pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go [hunk 13]
```diff
@@ -0,0 +1,163 @@
...
+			unsupportedMsgs = append(unsupportedMsgs, fmt.Sprintf("Kubelet version (%s) on node %s is ahead of the expected API server version (%s).", kubeletVersion, node.Name, c.apiServerVersion))
+		}
+	}
+
+	msgs := append(errorMsgs, append(skewedToTheLimitMsgs, append(skewedButOKMsgs, append(syncedMsgs, unsupportedMsgs...)...)...)...)
+	// sort messages to avoid flapping. sort by node name for aesthetics.
+	sort.Sort(byNode(msgs))
+	msg := strings.Join(msgs, "\n")
+
+	condition := operatorv1.OperatorCondition{Type: KubeletVersionSkewLimitUpgradeableConditionType, Message: msg}
```
wking, line 126: For clusters with hundreds of nodes, this is going to be hard to read if we include all the messages.  Can we drop `msg` and swap in a join of the most-severe message slice in the `switch` below?  E.g.:

```go
case len(errorMsgs) > 0:
	condition.Reason = KubeletVersionUnknownReason
	condition.Status = operatorv1.ConditionFalse
	sort.Sort(byNode(errorMsgs)) // avoid flapping.  Sort by node name for aesthetics
	condition.Message = strings.Join(errorMsgs, "\n")
```

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
```diff
@@ -0,0 +1,163 @@
...
+	// use the most "severe" reason to set the condition status
+	switch {
+	case len(errorMsgs) > 0:
+		condition.Reason = KubeletVersionUnknownReason
+		condition.Status = operatorv1.ConditionUnknown
+	case len(unsupportedMsgs) > 0:
+		condition.Reason = KubeletVersionUnsupportedReason
+		condition.Status = operatorv1.ConditionUnknown
```
wking, line 134: If `unsupportedMsgs` [gets updated][1] to include "Kubelets already too far behind to be supported", we'd want `False` here.  If we use separate slices for "ahead" and "too far behind", which is probably a good idea, we'd want `True` for "Kubelets ahead" (yes, please give me a newer API-server to close this gap) and `False` for "Kubelets already way behind" (gap is already too big; don't make it worse!).  It's not unknown.

[1]: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1199#discussion_r690791985

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
wking: actually, probably point this out in the reason and your `Kubelet and API server versions are synced` string.  I don't think we need `syncedMsgs` at all, because the `KubeletVersionSyncedReason` case is the `default` below.  This assumes you buy [my message-denoising argument][1].

[1]: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1199#discussion_r690794774

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
sanchezl: > Currently this case includes both "still supported, but `Upgradeable=False` because bumping would leave you unsupported" and "you are unsupported right now". 

It's not clear which version of the switch you are referring to as it has changed a bit, so pasting the current one here w/ comments.

```go
skew := int(kubeletVersion.Minor - c.apiServerVersion.Minor)
switch {
case skew == 0:
	// Minor versions synced.
	syncedMsgs = append(syncedMsgs, ...)
case skew < c.minUpgradeableSkew:
	// Upgrading to next minor version of OCP would result in an unsupported OCP state.
	// Note that we might already be in an unsupported OCP state (kubelet too far
	// behind), but the end result is the same, we don't want you to upgrade to the
	// next minor version of OCP.
	skewedToTheLimitMsgs = append(skewedToTheLimitMsgs, ...)
case skew < 0:
	// Behind, but upgrading to next minor version of API server is supported.
	// This can only occur when OCP minor version is odd.
	skewedButOKMsgs = append(skewedButOKMsgs, ...)
default:
	// Kubelet version newer than api server version for the current OCP version.
	// Possibly in the middle of a rollback. Let's just stay out of the way.
	unsupportedMsgs = append(unsupportedMsgs, ...)
}
```

I don't see the need to split the `skew < c.minUpgradeableSkew` case into _"unsupported already"_ and _"unsupported if updated"_ for the scope of this particular controller as the result is still the same: `Upgradeable=False`.

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
sanchezl: > probably point out here (and possibly also in the `KubeletVersionSynced` reason) that "synced" means major/minor match, but does not imply a patch match.
 
I agree, I'll re-work the reasons/messages to make that clear.

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
wking: > ...for the scope of this particular controller as the result is still the same: `Upgradeable=False`.

In both cases, `Upgradeable` should be `False`.  But in the will-become-unsupported case, the messaging should be "not a big deal, but you'll want to fix this before you can bump minors", while the currently-unsupported case should be "this is very bad right now.  [Nasal demons)][1]".  And ideally there would be an alert going off... but we can come back to look at the alert if we actually see this in condition Telemetry (another reason to have different reasons for the "nodes too far in front" and "nodes too far behind" cases, and possibly also a reason for "both at once, since they are both equally serious").

[1]: https://en.wikipedia.org/wiki/Undefined_behavior

deads2k: For the immediate future, we can assume every openshift comes with a new kube and that we never skip a kube.  If that assumption breaks in the future, we can backport a specific change to directly handle it.

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
```diff
@@ -0,0 +1,258 @@
...
+	corev1listers "k8s.io/client-go/listers/core/v1"
+)
+
+const (
+	KubeletMinorVersionUpgradeableConditionType = "KubeletMinorVersionUpgradeable"
+
+	KubeletVersionUnknownReason                     = "KubeletVersionUnknown"
+	KubeletMinorVersionSyncedReason                 = "KubeletMinorVersionsSynced"
+	KubeletMinorVersionSupportedNextUpgradeReason   = "KubeletMinorVersionSupportedNextUpgrade"
+	KubeletMinorVersionUnsupportedNextUpgradeReason = "KubeletMinorVersionUnsupportedNextUpgradeReason"
```
wking, line 29: nit: probably drop `Reason` in the value here (but keep it in the variable name)

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
```diff
@@ -0,0 +1,258 @@
...
+
+	// for each node, check kubelet version
+	for _, node := range nodes {
+		kubeletVersion, err := nodeKubeletVersion(node)
+		if err != nil {
+			runtime.HandleError(fmt.Errorf("unable to determine kubelet version on node %s: %w", node.Name, err))
+			errors = append(errors, nodeKubeletInfo{node: node.Name, err: err})
+			continue
+		}
+		minSupportedSkew := c.minUpgradeableSkew - 1
```
wking, line 113: I don't think this is right.  Can we think about this in terms of what skew is supported, instead of what skew allows `Upgradeable=False`  or `Upgradeable=True`?  Then we can derive `Upgradeable` from support.  Constraints from Scott:

* If OCP minor version is odd, kubelet versions older than 1 minor version of the API server version set `Upgradeable=False`.
* If OCP minor version is even, kubelet versions must be in sync with API server version in order to set `Upgradeable=True`.

So 4.9 (odd) is `Upgradeable=True` with skew 0 and skew 1, which means 4.10 (even) supports skews 0 through 2.

And 4.8 (even) is `Upgradeable=True` with skew 0, which means 4.9 (odd) supports skew 0 and skew 1.

So for odd, `minUpgradeableSkew` is 1, but `minSupportedSkew` is also 1, right?
And for even, `minUpgradeableSkew` is 0 and `minSupportedSkew` is 2.
i.e. even is generous, but knows that odd will be stingy.  Odd is stingy, but knows even will be generous.

sanchezl: I have re-worked the logic to work with a `minSupportedSkew` and `minSupportedSkewNextVersion` instead of a `minUpgradableSkew`.

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
sanchezl: Done.

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
sanchezl: See https://github.com/openshift/cluster-kube-apiserver-operator/pull/1199#discussion_r693237388 for a follow-up.

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
sanchezl: Done.

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
```diff
@@ -0,0 +1,255 @@
...
+}
+
+func minSupportedKubeletSkewForOpenShiftVersion(v semver.Version) int {
+	switch v.Minor % 2 {
+	case 0: // even OpenShift versions
+		return -2
+	case 1: // odd OpenShift versions
+		return -1
+	default:
+		panic("should not happen")
```
sttts, line 74: I would claim it cannot happen :)

PR #1201, opened 2021-08-04
Bug 1985073: use 1m resolution for control plane cpu alerts

tkashem:


tkashem:
image
image

`5m`, `2m` vs `1m` - looks like we should go with `1m` resolution.

bindata/assets/alerts/cpu-utilization.yaml
```diff
@@ -42,7 +42,7 @@ spec:
               kube-apiservers are also under-provisioned.
               To fix this, increase the CPU and memory on your control plane nodes.
           expr: |
-            100 - (avg by (instance) (rate(node_cpu_seconds_total{mode="idle"}[5m])) * 100) > 90 AND on (instance) label_replace( kube_node_role{role="master"}, "instance", "$1", "node", "(.+)" )
+            100 - (avg by (instance) (rate(node_cpu_seconds_total{mode="idle"}[1m])) * 100) > 90 AND on (instance) label_replace( kube_node_role{role="master"}, "instance", "$1", "node", "(.+)" )
```
wking, line 45: While we're refactoring this, I think it's easier to read if we rephrase from `100 - (avg idle) > 90` to `avg idle < 10`.

Also, this still doesn't account for holes in the `node_cpu_seconds_total` metric.  My understanding of the `rate` call is that if the covered minute has any `node_cpu_seconds_total` data, but `node_cpu_seconds_total` is ticking up at 20% for 10s, while `node_cpu_seconds_total` is missing for the other 50s, it will look like 0.2 * 0.1 = 0.02 = 2% idle, so the `expr` would match despite 20% being > 10%

tkashem: >  it will look like 0.2 * 0.1 = 0.02 = 2% idle

`rate` does extrapolation based on the slope of the first and the last sample under the window.
`rate` also avoids extrapolating too far, extrapolation extends to half the sample interval when the first or the last sample is too far away from the window
so prometheus does not handle the missing data/gap as above. since `rate` already extrapolates for us i don't see it's necessary for the alert to take into account any gap in its calculation.

(slack thread where we discussed it - https://coreos.slack.com/archives/C01CQA76KMX/p1626752724386400)

PR #1205, opened 2021-08-16
WIP: Testing DO NOT MERGE: Enable Tracing with defaults

sallyom:
Testing

Signed-off-by: Sally O'Malley <somalley@redhat.com>

### DIFF
--- a/bindata/assets/kube-apiserver/pod.yaml
+++ b/bindata/assets/kube-apiserver/pod.yaml
@@ -84,7 +84,7 @@ spec:
             cp -f /etc/kubernetes/static-pod-certs/configmaps/trusted-ca-bundle/ca-bundle.crt /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
           fi
 
-          exec watch-termination --termination-touch-file=/var/log/kube-apiserver/.terminating --termination-log-file=/var/log/kube-apiserver/termination.log --graceful-termination-duration={{.GracefulTerminationDuration}}s --kubeconfig=/etc/kubernetes/static-pod-resources/configmaps/kube-apiserver-cert-syncer-kubeconfig/kubeconfig -- hyperkube kube-apiserver --openshift-config=/etc/kubernetes/static-pod-resources/configmaps/config/config.yaml --advertise-address=${HOST_IP} {{.Verbosity}} --permit-address-sharing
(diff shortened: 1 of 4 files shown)


sallyom:
@tkashem I'm missing something, maybe you can tell me b4 I find it.

I see in **kube-apiserver bootstrap logs**:
`1 feature_gate.go:246] feature gates: &{map[APIPriorityAndFairness:true APIServerTracing:true]}` 
and 
`flags.go:59] FLAG: --tracing-config-file="/etc/kubernetes/static-pod-resources/configmaps/kube-apiserver-tracing-config/tracing-config.yaml"`
so that's good, but then
`Error: failed to read tracing config: unable to read tracing configuration from "/etc/kubernetes/static-pod-resources/configmaps/kube-apiserver-tracing-config/tracing-config.yaml": open /etc/kubernetes/static-pod-resources/configmaps/kube-apiserver-tracing-config/tracing-config.yaml: no such file or directory`  
how do I get that configmap file mounted? where are the others? I see other files, like 
`--tls-private-key-file="/etc/kubernetes/secrets/kube-apiserver-service-network-server.key", --tls-cert-file="/etc/kubernetes/secrets/kube-apiserver-service-network-server.crt"` but can't see in the code or yaml where these files are mounted, just that they are listed in the Assets like the tracing-config-file here: 
https://github.com/openshift/cluster-kube-apiserver-operator/blob/f147b1eaf7e491d604f812dcead312a4800dc370/pkg/operator/starter.go#L128-#L155  
If I ack for those, that's the only place I see them in the kaso code

tkashem:
I think you need to add it to the static resource controller

it might help if you follow Stefan's PR to add `audit-policies-cm.yaml`   https://github.com/openshift/cluster-kube-apiserver-operator/pull/1155/files#diff-0d623dfd885adb20f991bda4c2453aebd732ca6dbb4d1d4be6e79805c3b48de6L156

I think you just need to add your configmap asset to `staticResourceController := staticresourcecontroller.NewStaticResourceController(`. 
The path to your asset would be `assets/kube-apiserver/tracing-config.yaml`. Give it a try and let me know if it works.

sallyom:
> I think you need to add it to the static resource controller

> I think you just need to add your configmap asset to `staticResourceController := staticresourcecontroller.NewStaticResourceController(`.
> The path to your asset would be `assets/kube-apiserver/tracing-config.yaml`. Give it a try and let me know if it works.

That's what I understood, too, but that isn't working - I had that configmap asset added like so: https://github.com/openshift/cluster-kube-apiserver-operator/blob/f147b1eaf7e491d604f812dcead312a4800dc370/pkg/operator/starter.go#L151, but it's not found :shrug: I'll take a closer look at other PRs, thanks

PR #1208, opened 2021-08-18
Bug 1994857: targetcontroller: check external dependencies (csr-controller-ca configmap) only on SNO

soltysh:
/assign @sttts 

@hexfusion  based on comments from Stefan we might want to change the deps check, he says that relying on managed fields is not reliable enough. What was the annotation you initially wanted to check?


sttts:
Even this is not the solution. We must not use managedField for some definitive answer. It is best effort from the apiserver and information only.

soltysh:
/hold
alternative being discussed in #1209

PR #1211, opened 2021-08-21
Bug 1994643: remove startup-send-retry-after-until-ready option

tkashem:


bindata/bootkube/config/bootstrap-config-overrides.yaml
```diff
@@ -83,6 +83,8 @@ apiServerArguments:
     - /etc/kubernetes/secrets/kube-apiserver-service-network-server.crt
   tls-private-key-file:
     - /etc/kubernetes/secrets/kube-apiserver-service-network-server.key
+  startup-send-retry-after-until-ready:
+    - "false"
```
sttts, line 87: why?

PR #1216, opened 2021-08-27
deadman: add command to reset clusteroperators conditions to unknown

mfojtik:
Background: Under some circumstances, when kubelet is not able to run operator pods, but it is able to run static pods, the `oc get clusteroperators` will lie about the current cluster state. It reports the "last known" state for cluster operators, but no operators are actually running to reconcile the status.
For example, this happens today on SNO when you shut down the cluster for some extended period of time and the CSR kubelet use expires and new CSR's require manual approval.
To mitigate, we need a process that will reset the cluster operator status to "unknown". The process needs to have a working connection to the API server, so a side-car to kube-apiserver is a natural choice.

The process here will:

* Store timestamp to file on disk every 60 seconds
* When the process comes up (after shutdown, restart, redeploy, etc..) and the difference between the current timestamp and last heartbeat timestamp is more than 60 minutes it triggers a callback function (once).
* The callback function will reset all cluster operators conditions to "Unknown" unless they made progress in the last 30 minutes (once).
* When operators come back, they should reconcile and update the conditions

### DIFF
--- a/bindata/assets/kube-apiserver/pod.yaml
+++ b/bindata/assets/kube-apiserver/pod.yaml
@@ -177,6 +177,30 @@ spec:
     volumeMounts:
     - mountPath: /etc/kubernetes/static-pod-resources
       name: resource-dir
+  - name: kube-apiserver-deadman
+    env:
+      - name: POD_NAMESPACE
+        valueFrom:
+          fieldRef:
+            fieldPath: metadata.namespace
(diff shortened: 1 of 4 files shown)


deads2k:
replaced by https://github.com/openshift/cluster-kube-apiserver-operator/pull/1423

PR #1218, opened 2021-08-31
bug 1998031: Deploy PDB to prevent more than one replica going unavailable

ingvagabund:
In case master nodes are drained to quickly in a row there's no
waiting until instances go back to available. So it's possible
for two or more replicas to go unavailable which may lead
to all instances going unavailable.

Based on https://github.com/openshift/library-go/pull/1056/commits/8efd1883d406cc389eb25e2a257c4451dfbd668c

### DIFF
--- a/bindata/assets/kube-apiserver/pdb.yaml
+++ b/bindata/assets/kube-apiserver/pdb.yaml
@@ -1,0 +1,11 @@
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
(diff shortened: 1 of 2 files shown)


ingvagabund:
The approach might not work here. Alternative in https://github.com/openshift/library-go/pull/1095.

ingvagabund:
Wrong repo, https://github.com/openshift/cluster-openshift-apiserver-operator is the right one.
/close

PR #1236, opened 2021-10-01
Bump library-go for support GCP and vSphere out-of-tree cloud providers

lobziik:
Updated library-go dependency with changes from openshift/library-go#1221. 
Needed for proceeding with GCP and vSphere out of tree providers related work.


soltysh (approved):
/lgtm
/approve
/retest

PR #1239, opened 2021-10-06
Bug 2014614: Exempt metrics scrapes from APF.

benluddy:
Currently, metrics scraping requests can be rejected when all
workload-high seats are occupied, which results in missed samples.


tkashem:
/lgtm

Thanks

PR #1245, opened 2021-10-19
set kube-apiserver degraded=true if a webhook service is missing or down

sanchezl:
Add the following conditions to `kubeapiservers`:

* `MutatingAdmissionWebhookConfigurationDegraded`
* `ValidatingAdmissionWebhookConfigurationDegraded`
* `CRDConversionWebhookConfigurationDegraded`

The conditions are set to `True` if the webhook's service does not exist or the service's endpoints are not ready. Reason is set to the following:

* `WebhookServiceNotFound` - A webhook's service was not found.
* `WebhookServiceConnectionError` - Connection to service failed.
* `WebhookServiceNotReady` - A combination of the above two reasons.


pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
```diff
@@ -0,0 +1,104 @@
...
+	msgs := append(serviceMsgs, tlsMsgs...)
+	sort.Strings(msgs)
+	condition.Message = strings.Join(msgs, "\n")
+
+	return v1helpers.UpdateConditionFn(condition)
+}
+
+// assertService checks that the referenced service resource exists.
+func (c *webhookSupportabilityController) assertService(reference *serviceReference) error {
+	_, err := c.serviceLister.Services(reference.Namespace).Get(reference.Name)
```
sttts, line 76: How long is the inertia of these checks regarding the operator degraded condition? Wondering whether we run into false positives e.g. when CVO or OLM install a new CRD. An inertia of 1 or 2 min would be good IMO.

sanchezl: The status controller [defaults to a 2 minute inertia on degraded status](https://github.com/openshift/library-go/blob/master/pkg/operator/status/status_controller.go#L75).

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
```diff
@@ -0,0 +1,104 @@
...
+	host := reference.Name + "." + reference.Namespace + ".svc"
+	port := "443"
+	if reference.Port != nil {
+		port = fmt.Sprintf("%d", *reference.Port)
+	}
+	rootCAs := x509.NewCertPool()
+	if len(caBundle) > 0 {
+		rootCAs.AppendCertsFromPEM(caBundle)
+	}
+	client := &http.Client{
```
p0lyn0mial, line 94: I think we should set a `client.Timeout`

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
```diff
@@ -0,0 +1,104 @@
...
+	_, err := c.serviceLister.Services(reference.Namespace).Get(reference.Name)
+	if err != nil {
+		return fmt.Errorf("unable to find find service %s.%s: %v", reference.Name, reference.Namespace, err)
+	}
+	return nil
+}
+
+// assertConnect performs a dns lookup of service, opens a tcp connection, and performs a tls handshake.
+func (c *webhookSupportabilityController) assertConnect(ctx context.Context, reference *serviceReference, caBundle []byte) error {
+	host := reference.Name + "." + reference.Namespace + ".svc"
```
p0lyn0mial, line 85: since we are using `serviceReference` here we should take into account the `Path`

```
// `path` is an optional URL path which will be sent in any request to
// this service.
// +optional
Path *string `json:"path,omitempty" protobuf:"bytes,3,opt,name=path"`
```

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
p0lyn0mial: I think we could just establish a TLS connection not an HTTP

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
p0lyn0mial: otherwise for an HTTP request we will get 404

pkg/operator/webhooksupportabilitycontroller/degraded_webhook_admission.go
```diff
@@ -0,0 +1,69 @@
...
+
+func (c *webhookSupportabilityController) updateMutatingAdmissionWebhookConfigurationDegraded(ctx context.Context) v1helpers.UpdateStatusFunc {
+	condition := operatorv1.OperatorCondition{
+		Type:   MutatingAdmissionWebhookConfigurationDegradedType,
+		Status: operatorv1.ConditionUnknown,
+	}
+	webhookConfigurations, err := c.mutatingWebhookLister.List(labels.Everything())
+	if err != nil {
+		condition.Message = err.Error()
+		return v1helpers.UpdateConditionFn(condition)
```
p0lyn0mial, line 19: this won't put the operator into a degraded state, is that okay?

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
```diff
@@ -0,0 +1,104 @@
...
+
+// updateWebhookConfigurationDegraded updates the condition specified after
+// checking that the services associated with the specified webhooks exist
+// and have at least one ready endpoint.
+func (c *webhookSupportabilityController) updateWebhookConfigurationDegraded(ctx context.Context, condition operatorv1.OperatorCondition, webhookInfos []webhookInfo) v1helpers.UpdateStatusFunc {
+	var serviceMsgs []string
+	var tlsMsgs []string
+	for _, webhook := range webhookInfos {
+		if webhook.Service != nil {
+			err := c.assertService(webhook.Service)
```
p0lyn0mial, line 39: is this necessary? since we will hit the service network anyway.

pkg/operator/webhooksupportabilitycontroller/webhook_supportability_controller.go
```diff
@@ -0,0 +1,73 @@
...
+	apiextensionslistersv1 "k8s.io/apiextensions-apiserver/pkg/client/listers/apiextensions/v1"
+	"k8s.io/client-go/kubernetes"
+	admissionregistrationlistersv1 "k8s.io/client-go/listers/admissionregistration/v1"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+)
+
+type webhookSupportabilityController struct {
+	factory.Controller
+	operatorClient          v1helpers.StaticPodOperatorClient
+	kubeClient              *kubernetes.Clientset
```
p0lyn0mial, line 20: is this client used?

pkg/operator/webhooksupportabilitycontroller/webhook_supportability_controller.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	apiextensionsinformers "k8s.io/apiextensions-apiserver/pkg/client/informers/externalversions"
+	apiextensionslistersv1 "k8s.io/apiextensions-apiserver/pkg/client/listers/apiextensions/v1"
+	"k8s.io/client-go/kubernetes"
+	admissionregistrationlistersv1 "k8s.io/client-go/listers/admissionregistration/v1"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+	discoveryv1listers "k8s.io/client-go/listers/discovery/v1"
+)
+
+type webhookSupportabilityController struct {
```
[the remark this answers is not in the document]
sanchezl: Done.

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
```diff
@@ -0,0 +1,99 @@
...
+	requirement, err := labels.NewRequirement(discoveryv1.LabelServiceName, selection.Equals, []string{reference.Name})
+	if err != nil {
+		return err
+	}
+	endpointSlices, err := c.endpointSliceLister.EndpointSlices(reference.Namespace).List(labels.NewSelector().Add(*requirement))
+	var ok bool
+search:
+	for _, endpointSlice := range endpointSlices {
+		for _, endpoint := range endpointSlice.Endpoints {
+			if endpoint.Conditions.Ready != nil && *endpoint.Conditions.Ready {
```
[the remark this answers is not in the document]
sanchezl: No longer checking endpointslices.

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
sanchezl: Done. Added a 1s timeout and switched to only establishing a tls handshake.

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
sanchezl: No longer making http request.

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
sanchezl: This gives us a chance to give them a higher level "service does not exist" message vs. just saying "can't connect".

pkg/operator/webhooksupportabilitycontroller/webhook_supportability_controller.go
sanchezl: No. Removed.

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
```diff
@@ -0,0 +1,120 @@
...
+	rootCAs := x509.NewCertPool()
+	if len(caBundle) > 0 {
+		rootCAs.AppendCertsFromPEM(caBundle)
+	}
+	var err error
+	// retry up to 3 times on error
+	for i := 0; i < 3; i++ {
+		select {
+		case <-ctx.Done():
+			return nil
```
p0lyn0mial, line 100: should we return `ctx.Err()` ?

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
```diff
@@ -0,0 +1,120 @@
...
+				RootCAs:    rootCAs,
+			},
+		}
+		var conn net.Conn
+		conn, err = dialer.DialContext(ctx, "tcp", net.JoinHostPort(host, port))
+		if err != nil {
+			runtime.HandleError(err)
+			continue
+		}
+		runtime.HandleError(conn.Close())
```
p0lyn0mial, line 116: shouldn't we assign a potential error to the global `err` (line 119)?

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
```diff
@@ -0,0 +1,120 @@
...
+			Config: &tls.Config{
+				ServerName: host,
+				RootCAs:    rootCAs,
+			},
+		}
+		var conn net.Conn
+		conn, err = dialer.DialContext(ctx, "tcp", net.JoinHostPort(host, port))
+		if err != nil {
+			runtime.HandleError(err)
+			continue
```
p0lyn0mial, line 114: ok, in that case `conn` will be `nil`

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
```diff
@@ -0,0 +1,120 @@
...
+			NetDialer: &net.Dialer{Timeout: 1 * time.Second},
+			Config: &tls.Config{
+				ServerName: host,
+				RootCAs:    rootCAs,
+			},
+		}
+		var conn net.Conn
+		conn, err = dialer.DialContext(ctx, "tcp", net.JoinHostPort(host, port))
+		if err != nil {
+			runtime.HandleError(err)
```
p0lyn0mial, line 113: does it mean the error won't be visible in the log file?

p0lyn0mial: ok, it will be visible in `condition.Message`

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
sanchezl: No. Whatever caused the context to be cancelled should not be affecting the degraded condition.

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
sanchezl: I log it here, because only the last error (up to 3 iterations) will end up in `condition.Message`.

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
sanchezl: The thought here is that whatever is causing `conn.Close()` to return an error should not affect the degraded status. So I log it instead.

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
p0lyn0mial: in that case, I think we could update the `library-go` and pass a context to `v1helpers.UpdateStatus` function - it won't update the condition if the `ctx` expired

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
p0lyn0mial: ok

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
p0lyn0mial: ok

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	var err error
+	// retry up to 3 times on error
+	for i := 0; i < 3; i++ {
+		select {
+		case <-ctx.Done():
+			return nil
+		default:
+		}
+		dialer := &tls.Dialer{
```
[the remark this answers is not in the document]
sanchezl: > shouldn't the dialer be reused?

Not sure. It has some fields like `LocalAddr` and `Deadline` where I might run into trouble if I reuse. 

> how frequent will the assertConnect method be called?

`assertConnect()` if trggered on changes to `*WebHookConfiguration`, `Service`, and `CRD` resources.

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
sanchezl: I check for a cancelled context before the call to `UpdateStatus`. I'll bump `library-go` in a subsequent PR.

pkg/operator/webhooksupportabilitycontroller/degraded_webhook_admission.go
sanchezl: Yes. This is typical behavior.

p0lyn0mial:
/hold

it doesn't compile `webhook_supportability_controller.go:72:37: not enough arguments in call to v1helpers.UpdateStatus`

sanchezl:
/hold cancel

rebased

p0lyn0mial:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-kube-apiserver-operator/1245/pull-ci-openshift-cluster-kube-apiserver-operator-master-e2e-aws-serial/1460690453379682304 failed on `Some cluster operators are not ready: kube-apiserver (Degraded=True MutatingAdmissionWebhookConfiguration_WebhookServiceConnectionError: MutatingAdmissionWebhookConfigurationDegraded: pod-identity-webhook.amazonaws.com: remote error: tls: internal error)`

p0lyn0mial:
/hold

the following error seems to be getting common, check the link below. I fear once this one merges it might block other PRs.

`Operator degraded (MutatingAdmissionWebhookConfiguration_WebhookServiceConnectionError): MutatingAdmissionWebhookConfigurationDegraded: pod-identity-webhook.amazonaws.com: remote error: tls: internal error`

https://search.ci.openshift.org/?search=MutatingAdmissionWebhookConfigurationDegraded%3A+pod-identity-webhook.amazonaws.com&maxAge=48h&context=1&type=bug%2Bjunit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job


it looks like we need to sort out how to fix it before mergin this pr.

sanchezl:
/hold cancel 

PR openshift/cloud-credential-operator#421, addressing  [this comment](https://github.com/openshift/cluster-kube-apiserver-operator/pull/1245#issuecomment-973266981), has merged.

PR #1249, opened 2021-11-02
Bug 2033784: set kube-apiserver degraded=true if webhook matches a virtual resource

sanchezl:
Add new condition: `VirtualResourceAdmissionDegraded`
Add new reason: `AdmissionWebhookMatches`

### DIFF
--- a/pkg/operator/webhooksupportabilitycontroller/conditions.go
+++ b/pkg/operator/webhooksupportabilitycontroller/conditions.go
@@ -12,6 +12,10 @@ const (
 	// CRDConversionWebhookConfigurationDegradedType is true when there
 	// is a problem with a custom resource definition conversion webhook service.
 	CRDConversionWebhookConfigurationDegradedType = "CRDConversionWebhookConfigurationDegraded"
+
+	// VirtualResourceAdmissionDegradedType is true when a dynamic admission webhook matches
+	// a virtual resource.
+	VirtualResourceAdmissionDegradedType = "VirtualResourceAdmissionDegraded"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


pkg/operator/webhooksupportabilitycontroller/conditions.go
```diff
@@ -25,4 +29,8 @@ const (
 	// WebhookServiceNotReadyReason indicates that webhook services are having a variety of
 	// problems.
 	WebhookServiceNotReadyReason = "WebhookServiceNotReady"
+
+	// DynamicAdmissionWebhookMatchesReason indicates that a dynamic admission webhook matches
+	// a virtual resource.
+	DynamicAdmissionWebhookMatchesReason = "DynamicAdmissionWebhookMatches"
```
sttts, line 35: not sure I would understand this in a real cluster. would drop the "dynamic". VirtualResourceAdmissionWebHook

sanchezl: Done.

pkg/operator/webhooksupportabilitycontroller/degraded_webhook_on_virtual_resource.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+	}
+
+	if len(msgs) > 0 {
+		sort.Strings(msgs)
+		condition.Message = strings.Join(msgs, "\n")
+		condition.Reason = AdmissionWebhookMatchesVirtualResourceReason
+		condition.Status = operatorv1.ConditionTrue
+	} else {
+		condition.Status = operatorv1.ConditionFalse
```
benluddy, line 75: Does `condition.Message` need to be cleared here?

PR #1257, opened 2021-11-24
remove unnecessary service annotations

atiratree:


bindata/assets/kube-apiserver/svc.yaml
```diff
@@ -3,9 +3,6 @@ kind: Service
 metadata:
   namespace: openshift-kube-apiserver
   name: apiserver
-  annotations:
```
soltysh, line 6: This is fine, but you're missing labels which are being matched here: https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/manifests/0000_90_kube-apiserver-operator_03_servicemonitor.yaml#L30-L32

soltysh (changes requested):

bindata/assets/kube-apiserver/svc.yaml
soltysh: That's actually wrong. We're scraping the metrics from default namespace kubernetes service. So just ignore my comment here.

soltysh (approved):
/lgtm
/approve

PR #1262, opened 2021-11-30
Bug 2008462: config: disable pod security admission

s-urbaniak:
This disable pod security admission for 4.10. We plan to activate it again in 4.11.

### DIFF
--- a/bindata/assets/config/defaultconfig.yaml
+++ b/bindata/assets/config/defaultconfig.yaml
@@ -42,6 +42,8 @@ apiServerArguments:
     - /etc/kubernetes/static-pod-resources/configmaps/kube-apiserver-audit-policies/policy.yaml
   client-ca-file:
     - /etc/kubernetes/static-pod-certs/configmaps/client-ca/ca-bundle.crt
+  disable-admission-plugins:
+    - PodSecurity
   enable-admission-plugins:
     - CertificateApproval
     - CertificateSigning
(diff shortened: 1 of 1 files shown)


s-urbaniak:
@sttts @stlaz ptal

bindata/assets/config/defaultconfig.yaml
```diff
@@ -42,6 +42,8 @@ apiServerArguments:
     - /etc/kubernetes/static-pod-resources/configmaps/kube-apiserver-audit-policies/policy.yaml
   client-ca-file:
     - /etc/kubernetes/static-pod-certs/configmaps/client-ca/ca-bundle.crt
+  disable-admission-plugins:
```
stlaz, line 45: What about bootstrap?

s-urbaniak: afaiu this also applies to bootstrap: https://github.com/openshift/cluster-kube-apiserver-operator/blob/cb650638006a9e6dc87c24a6faab846359a0adbb/pkg/cmd/render/render.go#L275-L276

stlaz: I see, in that case we may want to use this PR as an opportunity to remove this line:
https://github.com/openshift/cluster-kube-apiserver-operator/blob/d3025b6784e883e323f37428962e4caf4c66e3fb/bindata/bootkube/config/bootstrap-config-overrides.yaml#L60

s-urbaniak: true, fixed :+1:

enxebre:
@s-urbaniak any plans to re-enable this soon? re https://github.com/openshift/hypershift/pull/1243

PR #1263, opened 2021-12-02
test/e2e/TestNamedCertificates: sign using root CA

s-urbaniak:
Currently, server certs are signed using the server private key.
They should be signed by the CA's private key.

This fixes it.

/cc @openshift/openshift-team-auth-maintainers @soltysh @sttts


soltysh (approved):
/lgtm
/approve

PR #1267, opened 2021-12-03
reduce alert severity to warning where appropriate

tkashem:
relates to https://issues.redhat.com/browse/API-1361

### DIFF
--- a/bindata/assets/alerts/cpu-utilization.yaml
+++ b/bindata/assets/alerts/cpu-utilization.yaml
@@ -12,6 +12,7 @@ spec:
             summary: >-
               CPU utilization across all three control plane nodes is higher than two control plane nodes can sustain; a single control plane node outage may
               cause a cascading failure; increase available CPU.
+            runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-kube-apiserver-operator/ExtremelyHighIndividualControlPlaneCPU.md
             description: >-
               Given three control plane nodes, the overall CPU utilization may only be about 2/3 of all available capacity.
               This is because if a single control plane node fails, the remaining two must handle the load of the cluster in order to be HA.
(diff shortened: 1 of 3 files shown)


bindata/assets/alerts/kube-apiserver-slos-basic.yaml
```diff
@@ -19,7 +19,7 @@ spec:
       labels:
         long: 1h
         namespace: openshift-kube-apiserver
-        severity: critical
+        severity: warning
```
dgrisonnet, line 22: I am on the fence regarding both of the severity changes of the KubeAPIErrorBudgetBurn alerts. Although these alerts are hardly actionable, they are usually symptoms of things going very wrong in the cluster. Most of the time, these availability issues will propagate all over the cluster since a lot of applications depend on the apiserver. As such, I still think that customers would want to be informed right away when such a thing happens, and we should give them the keys to solve this kind of issue, even though it might need further investigation from their side.

It is also worth mentioning that some OSD folks wrote this runbook to help diagnose and mitigate the issues reported by these alerts: https://github.com/openshift/runbooks/blob/master/alerts/cluster-kube-apiserver-operator/KubeAPIErrorBudgetBurn.md. If we think that it is lacking we could maybe improve it? wdyt @tkashem?

bindata/assets/alerts/cpu-utilization.yaml
```diff
@@ -12,6 +12,7 @@ spec:
             summary: >-
               CPU utilization across all three control plane nodes is higher than two control plane nodes can sustain; a single control plane node outage may
               cause a cascading failure; increase available CPU.
+            runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-kube-apiserver-operator/ExtremelyHighIndividualControlPlaneCPU.md
```
dgrisonnet, line 15: I think it is fine to share the runbook between `ExtremelyHighIndividualControlPlaneCPU` and `HighOverallControlPlaneCPU`, but we should mention `HighOverallControlPlaneCPU` in the runbook which is not the case today.

tkashem: okay, i can add a few lines in the runbook

PR #1272, opened 2021-12-15
prometheus rules for client-go

aojea:
Kubernetes is a distributed system, so network is a fundamental part
of the system.

At the heart of the system we can find the apiservers and the etcd
cluster, both have a good instrumentation that allow users and
developers to diagnose and troubleshoot networking problems.

However, there are other components like the kube-scheduler,
kube-controller-manager and kubelets that depend on its connectivity
against the apiserver to work.

These communication can be achieved via a direct network connection,
and intermeadiate load balancer or via the Kubernetes Services
implementation (CNI, ...)

Recording the latency and the error rate of these components against
the apiserver will allow users and developers to understand and
troubleshoot the networking problem caused by external factors.

Also, before adding SLOs we need a better understanding on how these
SLIs behave, since they depend a lot of the environment, per example,
cloud providers load balancers vs BareMetal keepalived+haproxy or
customer load balancer, ...

Add 4 rules to store the latency against the internal apiserver, localhost and loopback connections

### DIFF
--- a/bindata/assets/alerts/client-go-requests.yaml
+++ b/bindata/assets/alerts/client-go-requests.yaml
@@ -16,13 +16,13 @@ spec:
                 label_replace(rest_client_request_duration_seconds_bucket{verb="GET",url=~"https?://api-int.*"},
                   "host","$1","url","https?://([^/\\s]+).*")[5m:30s]
                 )) by (le,host,service,namespace,node)
-        - record: service:rest_client_request_duration_seconds:sum:rate
+        - record: service:rest_client_request_duration_seconds:rate5m
           expr: |
               sum(rate(
                 label_replace(rest_client_request_duration_seconds_bucket{verb="GET",url!~"https?://(api-int|\\[::1\\]|127\\.0\\.0\\.1|localhost).*"},
                   "host","$1","url","https?://([^/\\s]+).*")[5m:30s]
                 )) by (le,host,service,namespace)
-        - record: pod:rest_client_request_duration_seconds:sum:rate
+        - record: pod:rest_client_request_duration_seconds:rate5m
           expr: |
               sum(rate(
                 label_replace(rest_client_request_duration_seconds_bucket{verb="GET",url=~"https?://(\\[::1\\]|127\\.0\\.0\\.1|localhost).*"},
@@ -33,17 +33,17 @@ spec:
           labels:
             quantile: "0.99"
           expr: |
-            histogram_quantile(0.99, sum by (le) (internal_load_balancer:rest_client_request_duration_seconds:sum:rate))
+            histogram_quantile(0.99, sum by (le) (internal_load_balancer:rest_client_request_duration_seconds:rate5m))
         - record: service:rest_client_request_duration_seconds:histogram_quantile
           labels:
             quantile: "0.99"
           expr: |
-            histogram_quantile(0.99, sum by (le) (service:rest_client_request_duration_seconds:sum:rate))
+            histogram_quantile(0.99, sum by (le) (service:rest_client_request_duration_seconds:rate5m))
         - record: pod:rest_client_request_duration_seconds:histogram_quantile
           labels:
             quantile: "0.99"
           expr: |
-            histogram_quantile(0.99, sum by (le) (pod:rest_client_request_duration_seconds:sum:rate))
+            histogram_quantile(0.99, sum by (le) (pod:rest_client_request_duration_seconds:rate5m))
         # error by host and job, the rest_client metrics use the code "<error>" to aggregate
         # any non-http error and avoid cardinality problems.
         # xref: https://github.com/kubernetes/kubernetes/blob/66931c9b8f11a3f223ba1890e4f390cff74ce1a6/staging/src/k8s.io/client-go/rest/request.go#L785-L799
@@ -63,12 +63,12 @@ spec:
             / ignoring (host,service,namespace) group_left
             sum(rate(rest_client_requests_total{host=~"(\\[::1\\]|127\\.0\\.0\\.1|localhost).*"}[5m]))
         # errors by destination
-        - record: internal_load_balancer:rest_client_requests_errors:sum
+        - record: internal_load_balancer:rest_client_requests_errors:rate5m:sum
           expr: |
             sum(internal_load_balancer:rest_client_requests_errors:ratio_rate5m)
-        - record: service:rest_client_requests:errors:sum
+        - record: service:rest_client_requests:errors:rate5m:sum
           expr: |
             sum(service:rest_client_requests:errors:ratio_rate5m)
-        - record: pod:rest_client_requests:errors:sum
+        - record: pod:rest_client_requests:errors:rate5m:sum
           expr: |
             sum(pod:rest_client_requests:errors:ratio_rate5m)
--- a/bindata/assets/alerts/client-go-requests.yaml
+++ b/bindata/assets/alerts/client-go-requests.yaml
@@ -1,0 +1,74 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: client-go
+  namespace: openshift-kube-apiserver
+spec:
+  groups:
+    - name: control-plane-client-go
+      rules:
+        # record request duration rate by:
+        # - job (kubelet, apiserver. controller-manager, ...)
+        # - destination host: internal-lb, services, localhost
+        - record: internal_load_balancer:rest_client_request_duration_seconds:rate5m
+          expr: |
+              sum(rate(
+                label_replace(rest_client_request_duration_seconds_bucket{verb="GET",url=~"https?://api-int.*"},
+                  "host","$1","url","https?://([^/\\s]+).*")[5m:30s]
+                )) by (le,host,service,namespace,node)
+        - record: service:rest_client_request_duration_seconds:sum:rate
+          expr: |
+              sum(rate(
+                label_replace(rest_client_request_duration_seconds_bucket{verb="GET",url!~"https?://(api-int|\\[::1\\]|127\\.0\\.0\\.1|localhost).*"},
+                  "host","$1","url","https?://([^/\\s]+).*")[5m:30s]
+                )) by (le,host,service,namespace)
+        - record: pod:rest_client_request_duration_seconds:sum:rate
+          expr: |
+              sum(rate(
+                label_replace(rest_client_request_duration_seconds_bucket{verb="GET",url=~"https?://(\\[::1\\]|127\\.0\\.0\\.1|localhost).*"},
+                  "host","$1","url","https?://([^/\\s]+).*")[5m:30s]
+                )) by (le,host,service,namespace)
+        # latency by destination: internal-lb, services, localhost
+        - record: internal_load_balancer:rest_client_request_duration_seconds:histogram_quantile
+          labels:
+            quantile: "0.99"
+          expr: |
+            histogram_quantile(0.99, sum by (le) (internal_load_balancer:rest_client_request_duration_seconds:sum:rate))
+        - record: service:rest_client_request_duration_seconds:histogram_quantile
+          labels:
+            quantile: "0.99"
+          expr: |
+            histogram_quantile(0.99, sum by (le) (service:rest_client_request_duration_seconds:sum:rate))
+        - record: pod:rest_client_request_duration_seconds:histogram_quantile
+          labels:
+            quantile: "0.99"
+          expr: |
+            histogram_quantile(0.99, sum by (le) (pod:rest_client_request_duration_seconds:sum:rate))
+        # error by host and job, the rest_client metrics use the code "<error>" to aggregate
+        # any non-http error and avoid cardinality problems.
+        # xref: https://github.com/kubernetes/kubernetes/blob/66931c9b8f11a3f223ba1890e4f390cff74ce1a6/staging/src/k8s.io/client-go/rest/request.go#L785-L799
+        - record: internal_load_balancer:rest_client_requests_errors:ratio_rate5m
+          expr: |
+            sum(rate(rest_client_requests_total{code="<error>",host=~"api-int.*"}[5m])) by (host,service,namespace)
+            / ignoring (host,service,namespace) group_left
+            sum(rate(rest_client_requests_total{host=~"api-int.*"}[5m]))
+        - record: service:rest_client_requests:errors:ratio_rate5m
+          expr: |
+            sum(rate(rest_client_requests_total{code="<error>",host!~"(api-int|\\[::1\\]|127\\.0\\.0\\.1|localhost).*"}[5m])) by (host,service,namespace)
+            / ignoring (host,service,namespace) group_left
+            sum(rate(rest_client_requests_total{host!~"(api-int|\\[::1\\]|127\\.0\\.0\\.1|localhost).*"}[5m]))
+        - record: pod:rest_client_requests:errors:ratio_rate5m
+          expr: |
+            sum(rate(rest_client_requests_total{code="<error>",host=~"(\\[::1\\]|127\\.0\\.0\\.1|localhost).*"}[5m])) by (host,service,namespace)
+            / ignoring (host,service,namespace) group_left
+            sum(rate(rest_client_requests_total{host=~"(\\[::1\\]|127\\.0\\.0\\.1|localhost).*"}[5m]))
+        # errors by destination
+        - record: internal_load_balancer:rest_client_requests_errors:sum
+          expr: |
+            sum(internal_load_balancer:rest_client_requests_errors:ratio_rate5m)
+        - record: service:rest_client_requests:errors:sum
+          expr: |
+            sum(service:rest_client_requests:errors:ratio_rate5m)
+        - record: pod:rest_client_requests:errors:sum
+          expr: |
+            sum(pod:rest_client_requests:errors:ratio_rate5m)
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -149,6 +149,7 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		"assets/kube-apiserver/storage-version-migration-prioritylevelconfiguration.yaml",
 		"assets/alerts/api-usage.yaml",
 		"assets/alerts/audit-errors.yaml",
+		"assets/alerts/client-go-requests.yaml",
 		"assets/alerts/cpu-utilization.yaml",
 		"assets/alerts/kube-apiserver-requests.yaml",
 		"assets/alerts/kube-apiserver-slos-basic.yaml",


aojea:
/assign @dgrisonnet @tkashem 
@smarterclayton FYI  I'd love to have some telemetry from users on these metrics

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,23 @@
...
+    - name: control-plane-client-go
+      rules:
+        - record: client_go_request_duration:method_get:histogram_quantile99
+          expr: |
+            histogram_quantile(0.99,
+              sum(rate(
+                label_replace(rest_client_request_duration_seconds_bucket{verb="GET"},
+                  "url_host","$1$2","url","https?://([^:^/]*)(:[0-9]+)?.*?")[5m:30s]
+                )
+              ) by (url_host,job)
```
dgrisonnet, line 17:
```suggestion
              ) by (le,url_host,job)
```

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,23 @@
...
+spec:
+  groups:
+    - name: control-plane-client-go
+      rules:
+        - record: client_go_request_duration:method_get:histogram_quantile99
+          expr: |
+            histogram_quantile(0.99,
+              sum(rate(
+                label_replace(rest_client_request_duration_seconds_bucket{verb="GET"},
+                  "url_host","$1$2","url","https?://([^:^/]*)(:[0-9]+)?.*?")[5m:30s]
```
dgrisonnet, line 15: I'd prefer `host` as a label instead of `url_host`, `url` feels unnecessary since the metric is about requests

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: client-go
+  namespace: openshift-kube-apiserver
+spec:
+  groups:
+    - name: control-plane-client-go
+      rules:
+        - record: client_go_request_duration:method_get:histogram_quantile99
```
dgrisonnet, line 10: naming suggestion:
```
cluster:rest_client_request_duration_seconds:histogram_quantile
```
essentially the format is: `<scope>:<metric>:<aggregation>`

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,23 @@
...
+          expr: |
+            histogram_quantile(0.99,
+              sum(rate(
+                label_replace(rest_client_request_duration_seconds_bucket{verb="GET"},
+                  "url_host","$1$2","url","https?://([^:^/]*)(:[0-9]+)?.*?")[5m:30s]
+                )
+              ) by (url_host,job)
+            )
+        - record: client_go_request_total:network_errors:rate5m
```
dgrisonnet, line 19:
```suggestion
        - record: cluster:rest_client_requests:rate
```

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,23 @@
...
+          expr: |
+            sum(rate(rest_client_requests_total{code="<error>"}[5m])) by (job,host)
```
dgrisonnet, line 21: Since we want insights on the proxy, I don't think we care about 4xx errors
```suggestion
            sum(rate(rest_client_requests_total{code=~"5.."}[5m])) by (job,host)
```

bindata/assets/alerts/client-go-requests.yaml
dgrisonnet, line 10: we could make the recording rule generic here by adding a quantile label as in: https://github.com/openshift/cluster-monitoring-operator/blob/master/assets/cluster-monitoring-operator/prometheus-rule.yaml#L420-L423. This would allow having multiple `cluster:rest_client_request_duration_seconds:histogram_quantile` and select which one we want to use by specifying the quantile label. In our case
```
cluster:rest_client_request_duration_seconds:histogram_quantile{quantile="0.99"}
```

bindata/assets/alerts/client-go-requests.yaml
aojea: this is about error rate, better `cluster:rest_client_requests_errors:rate`

bindata/assets/alerts/client-go-requests.yaml
aojea: 5.. means the request reached the apiserver, we have those metrics in the audit log and the apiserver metrics.
we want the ones that are not being able to contact the apiserver, these are the request that the client send and get lost

aojea: see
```go
// updateURLMetrics is a convenience function for pushing metrics.
// It also handles corner cases for incomplete/invalid request data.
func updateURLMetrics(ctx context.Context, req *Request, resp *http.Response, err error) {
	url := "none"
	if req.c.base != nil {
		url = req.c.base.Host
	}

	// Errors can be arbitrary strings. Unbound label cardinality is not suitable for a metric
	// system so we just report them as `<error>`.
	if err != nil {
		metrics.RequestResult.Increment(ctx, "<error>", req.verb, url)
	} else {
		//Metrics for failure codes
		metrics.RequestResult.Increment(ctx, strconv.Itoa(resp.StatusCode), req.verb, url)
	}
}
```
in https://github.com/kubernetes/kubernetes/blob/66931c9b8f11a3f223ba1890e4f390cff74ce1a6/staging/src/k8s.io/client-go/rest/request.go#L785-L799

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,28 @@
...
+        # - job (kubelet, apiserver. controller-manager, ...)
+        # - destination host: internal-lb, kubernetes.default service, direct connection ...
+        - record: cluster:rest_client_request_duration_seconds:histogram_quantile
+          labels:
+            quantile: "0.99"
+          expr: |
+            histogram_quantile(0.99,
+              sum(rate(
+                label_replace(rest_client_request_duration_seconds_bucket{verb="GET"},
+                  "url_host","$1","url","https?://([^/\\s]+).*")[5m:30s]
```
aojea, line 20: this regex matches anything until the next `/` or whitespace, it works fine

bindata/assets/alerts/client-go-requests.yaml
aojea: added

bindata/assets/alerts/client-go-requests.yaml
aojea: sorry, it was a question, I clicked too fast

bindata/assets/alerts/client-go-requests.yaml
aojea: I think that I choose that because I was afraid `host` already existed as label

bindata/assets/alerts/client-go-requests.yaml
aojea: yeah :upside_down_face:

bindata/assets/alerts/client-go-requests.yaml
aojea: updated to use host

bindata/assets/alerts/client-go-requests.yaml
dgrisonnet: wow `<error>` is so confusing, it really feels like a placeholder. Thanks for providing the details, I completely thought it was a placeholder at first

aojea: heh, indeed it is, let me add a comment

bindata/assets/alerts/client-go-requests.yaml
aojea: added the one that you suggested

bindata/assets/alerts/client-go-requests.yaml
aojea: done

aojea:
/hold
let's think this more

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,64 @@
...
+            )
+        - record: cluster:rest_client_request_duration_seconds:other:histogram_quantile
+          labels:
+            quantile: "0.99"
+          expr: |
+            histogram_quantile(0.99,
+              sum(rate(
+                label_replace(rest_client_request_duration_seconds_bucket{verb="GET",url!~"https?://(api-int|\\[::1\\]|127\\.0\\.0\\.1|localhost).*"},
+                  "host","$1","url","https?://([^/\\s]+).*")[5m:30s]
+                ) TODO JUST THE ONE THAT DOES NOT MATCH CLUSTERIP* on (cluster_ip) group_left kube_service_info{service="kubernetes"}
```
aojea, line 54: @dgrisonnet these are the missing ones, I can't figure out how to query this

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,53 @@
...
+            quantile: "0.99"
+          expr: |
+            histogram_quantile(0.99,
+              sum(rate(
+                label_replace(rest_client_request_duration_seconds_bucket{verb="GET",url!~"https?://(api-int|\\[::1\\]|127\\.0\\.0\\.1|localhost).*"},
+                  "host","$1","url","https?://([^/\\s]+).*")[5m:30s]
+                ) * on (cluster_ip) group_left group(kube_service_info{cluster_ip!~"None",namespace=~"(openshift|kube|default).*"}) by (service,namespace,cluster_ip)
+              ) by (le,host,job,service,namespace)
+            )
+        - record: cluster:rest_client_request_duration_seconds:localhost:histogram_quantile
```
aojea, line 35: image

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,53 @@
...
+          labels:
+            quantile: "0.99"
+          expr: |
+            histogram_quantile(0.99,
+              sum(rate(
+                label_replace(rest_client_request_duration_seconds_bucket{verb="GET",url=~"https?://api-int.*"},
+                  "host","$1","url","https?://([^/\\s]+).*")[5m:30s]
+                )
+              ) by (le,host,service,namespace,node)
+            )
```
aojea, line 23: image

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,53 @@
...
+          labels:
+            quantile: "0.99"
+          expr: |
+            histogram_quantile(0.99,
+              sum(rate(
+                label_replace(rest_client_request_duration_seconds_bucket{verb="GET",url!~"https?://(api-int|\\[::1\\]|127\\.0\\.0\\.1|localhost).*"},
+                  "host","$1","url","https?://([^/\\s]+).*")[5m:30s]
+                ) * on (cluster_ip) group_left group(kube_service_info{cluster_ip!~"None",namespace=~"(openshift|kube|default).*"}) by (service,namespace,cluster_ip)
+              ) by (le,host,service,namespace)
+            )
```
aojea, line 34: image

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,53 @@
...
+          labels:
+            quantile: "0.99"
+          expr: |
+            histogram_quantile(0.99,
+              sum(rate(
+                label_replace(rest_client_request_duration_seconds_bucket{verb="GET",url=~"https?://(\\[::1\\]|127\\.0\\.0\\.1|localhost).*"},
+                  "host","$1","url","https?://([^/\\s]+).*")[5m:30s]
+                )
+              ) by (le,host,service,namespace)
+            )
```
aojea, line 45: image

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,53 @@
...
+        # error by host and job, the rest_client metrics use the code "<error>" to aggregate
+        # any non-http error and avoid cardinality problems.
+        # xref: https://github.com/kubernetes/kubernetes/blob/66931c9b8f11a3f223ba1890e4f390cff74ce1a6/staging/src/k8s.io/client-go/rest/request.go#L785-L799
+        - record: cluster:rest_client_requests:errors:rate
+          expr: |
+            sum(rate(rest_client_requests_total{code="<error>"}[5m])) by (job,host,service,namespace)
+            / ignoring (job, host,service,namespace) group_left
+            sum(rate(rest_client_requests_total{}[5m]))
```
aojea, line 53: image

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,53 @@
...
+              ) by (le,host,service,namespace)
+            )
+        # error by host and job, the rest_client metrics use the code "<error>" to aggregate
+        # any non-http error and avoid cardinality problems.
+        # xref: https://github.com/kubernetes/kubernetes/blob/66931c9b8f11a3f223ba1890e4f390cff74ce1a6/staging/src/k8s.io/client-go/rest/request.go#L785-L799
+        - record: cluster:rest_client_requests:errors:rate
+          expr: |
+            sum(rate(rest_client_requests_total{code="<error>"}[5m])) by (host,service,namespace)
+            / ignoring (host,service,namespace) group_left
+            sum(rate(rest_client_requests_total{}[5m]))
```
aojea, line 53: image

bindata/assets/alerts/client-go-requests.yaml
aojea: this is a great view of the internal load balancer , we can also differentiate by kubelet and node

bindata/assets/alerts/client-go-requests.yaml
aojea: this will yield all the services

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,53 @@
...
+  name: client-go
+  namespace: openshift-kube-apiserver
+spec:
+  groups:
+    - name: control-plane-client-go
+      rules:
+        # latency by:
+        # - job (kubelet, apiserver. controller-manager, ...)
+        # - destination host: internal-lb, kubernetes.default service, direct connection ...
+        - record: cluster:rest_client_request_duration_seconds:internal_load_balancer:histogram_quantile
```
dgrisonnet, line 13: The following might be better suited since we want the scope of the recording rule before the metric name:

- cluster:internal_load_balancer:rest_client_request_duration_seconds:histogram_quantile
- internal_load_balancer:rest_client_request_duration_seconds:histogram_quantile

bindata/assets/alerts/client-go-requests.yaml
dgrisonnet: Yes, this recording rule is great to see the latency of the internal load balancer. The only thing I would change for telemetry purposes is: `by (le,host,service,namespace,node)`. As you mentioned it is great to differentiate kubelet and other services, but we can't afford this granularity in telemetry as each additional serie is expensive to store. My suggestion would be to keep this granularity for the dashboards and alerts, but reduce it to `by (le)` for telemetry so that we can reduce the number of metrics we push for the internal load balancer latency to 1.

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,53 @@
...
+            quantile: "0.99"
+          expr: |
+            histogram_quantile(0.99,
+              sum(rate(
+                label_replace(rest_client_request_duration_seconds_bucket{verb="GET",url=~"https?://api-int.*"},
+                  "host","$1","url","https?://([^/\\s]+).*")[5m:30s]
+                )
+              ) by (le,host,service,namespace,node)
+            )
+        - record: cluster:rest_client_request_duration_seconds:services:histogram_quantile
```
dgrisonnet, line 24:
```suggestion
        - record: service:rest_client_request_duration_seconds:histogram_quantile
```

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,53 @@
...
+          labels:
+            quantile: "0.99"
+          expr: |
+            histogram_quantile(0.99,
+              sum(rate(
+                label_replace(rest_client_request_duration_seconds_bucket{verb="GET",url!~"https?://(api-int|\\[::1\\]|127\\.0\\.0\\.1|localhost).*"},
+                  "host","$1","url","https?://([^/\\s]+).*")[5m:30s]
+                ) * on (cluster_ip) group_left group(kube_service_info{cluster_ip!~"None",namespace=~"(openshift|kube|default).*"}) by (service,namespace,cluster_ip)
```
dgrisonnet, line 32: This part seems superflux:
```suggestion
                ) * on (cluster_ip) group_left kube_service_info{cluster_ip!~"None",namespace=~"(openshift|kube|default).*"}
```

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,53 @@
...
+              ) by (le,host,service,namespace)
```
dgrisonnet, line 33: Same comment as above, this is great for dashboards and alerts but for telemetry we would prefer `by (le)`

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,53 @@
...
+            )
+        - record: cluster:rest_client_request_duration_seconds:localhost:histogram_quantile
```
dgrisonnet, line 35: Maybe `pod` is better suited for the scope wdyt?
```suggestion
        - record: pod:rest_client_request_duration_seconds:histogram_quantile
```

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,53 @@
...
+          labels:
+            quantile: "0.99"
+          expr: |
+            histogram_quantile(0.99,
+              sum(rate(
+                label_replace(rest_client_request_duration_seconds_bucket{verb="GET",url=~"https?://(\\[::1\\]|127\\.0\\.0\\.1|localhost).*"},
+                  "host","$1","url","https?://([^/\\s]+).*")[5m:30s]
+                )
+              ) by (le,host,service,namespace)
```
dgrisonnet, line 44: Same comment as above, this is great for dashboards and alerts but for telemetry we would prefer `by (le)`

bindata/assets/alerts/client-go-requests.yaml
dgrisonnet: I would expect to have 3 different rules for the error rate since we have dedicated rules for latency. Ultimately we want to know the availability of each proxy separately, so we should also differentiate their error rate in telemetry. So I would expect 3 recording rules, one for the internal load balancer, one for the services, and another one for direct connections. Each of them yielding only 1 serie (telemetry requirements), so that we can easily differentiate them in telemetry later and maybe even reuse them to create dedicated alerts with different SLOs.

bindata/assets/alerts/client-go-requests.yaml
aojea: I'm going with cluster:internal_load_balancer, since internal_load_balancer is an specific configuration of openshift clusters

bindata/assets/alerts/client-go-requests.yaml
aojea: yep

bindata/assets/alerts/client-go-requests.yaml
aojea: we need to aggregate or it can fail with errors like
```
found duplicate series for the match group {cluster_ip="None"} on the right hand-side of the operation: [{__name__="kube_service_info", cluster_ip="None", container="kube-rbac-proxy-main", endpoint="https-main", job="kube-state-metrics", namespace="openshift-cluster-machine-approver", prometheus="openshift-monitoring/k8s", service="machine-approver"}, {__name__="kube_service_info", cluster_ip="None", container="kube-rbac-proxy-main", endpoint="https-main", job="kube-state-metrics", namespace="kube-system", prometheus="openshift-monitoring/k8s", service="kubelet"}];many-to-many matching not allowed: matching labels must be unique on one side
```

bindata/assets/alerts/client-go-requests.yaml
aojea: is pod really the scope?
I think is better scoped by the destination than by the origin, is more about control_plane or something like that,

bindata/assets/alerts/client-go-requests.yaml
aojea: ack

bindata/assets/alerts/client-go-requests.yaml
aojea: it seems that services with external_ips are duplicated too, we have to aggregate by service/namespace that is unique, but I don't know if that is the correct statement though, I just assumed group was doing that and it actually works

aojea: there are some services that are referred by url
I think that we better use here the rest of connections that doesn't use the load balancer or the localhost address

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,64 @@
...
+          expr: |
+            histogram_quantile(0.99, sum by (le) (pod:rest_client_request_duration_seconds:sum:rate))
+        # error by host and job, the rest_client metrics use the code "<error>" to aggregate
+        # any non-http error and avoid cardinality problems.
+        # xref: https://github.com/kubernetes/kubernetes/blob/66931c9b8f11a3f223ba1890e4f390cff74ce1a6/staging/src/k8s.io/client-go/rest/request.go#L785-L799
+        - record: internal_load_balancer:rest_client_requests:errors:rate
+          expr: |
+            sum(rate(rest_client_requests_total{code="<error>",host=~"api-int.*"}[5m])) by (host,service,namespace)
+            / ignoring (host,service,namespace) group_left
+            sum(rate(rest_client_requests_total{host=~"api-int.*"}[5m]))
```
dgrisonnet, line 54: We will have the same problem as with the latency for telemetry. We have to create separate recording rules that will only yield one serie such as:
```
sum(internal_load_balancer:rest_client_requests:errors:rate)
```
At this point it is a bit hard to name them, but maybe something like:
```
internal_load_balancer:rest_client_requests:errors:sum
```

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,64 @@
...
+            histogram_quantile(0.99, sum by (le) (service:rest_client_request_duration_seconds:sum:rate))
+        - record: pod:rest_client_request_duration_seconds:histogram_quantile
+          labels:
+            quantile: "0.99"
+          expr: |
+            histogram_quantile(0.99, sum by (le) (pod:rest_client_request_duration_seconds:sum:rate))
+        # error by host and job, the rest_client metrics use the code "<error>" to aggregate
+        # any non-http error and avoid cardinality problems.
+        # xref: https://github.com/kubernetes/kubernetes/blob/66931c9b8f11a3f223ba1890e4f390cff74ce1a6/staging/src/k8s.io/client-go/rest/request.go#L785-L799
+        - record: internal_load_balancer:rest_client_requests:errors:rate
```
dgrisonnet, line 50:
```suggestion
        - record: internal_load_balancer:rest_client_requests_errors:ratio_rate5m
```

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,64 @@
...
+          expr: |
+            sum(rate(rest_client_requests_total{code="<error>",host=~"api-int.*"}[5m])) by (host,service,namespace)
+            / ignoring (host,service,namespace) group_left
+            sum(rate(rest_client_requests_total{host=~"api-int.*"}[5m]))
+        - record: service:rest_client_requests:errors:rate
+          expr: |
+            sum(rate(rest_client_requests_total{code="<error>",host!~"(api-int|\\[::1\\]|127\\.0\\.0\\.1|localhost).*"}[5m])) by (host,service,namespace)
```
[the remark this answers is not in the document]
aojea: services can use the IP in numeric format too

bindata/assets/alerts/client-go-requests.yaml
aojea: done

bindata/assets/alerts/client-go-requests.yaml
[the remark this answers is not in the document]
aojea: is ok this way, we are not going to achieve perfection at first shot ;)

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,74 @@
...
+        # record request duration rate by:
+        # - job (kubelet, apiserver. controller-manager, ...)
+        # - destination host: internal-lb, services, localhost
+        - record: internal_load_balancer:rest_client_request_duration_seconds:rate5m
+          expr: |
+              sum(rate(
+                label_replace(rest_client_request_duration_seconds_bucket{verb="GET",url=~"https?://api-int.*"},
+                  "host","$1","url","https?://([^/\\s]+).*")[5m:30s]
+                )) by (le,host,service,namespace,node)
+        - record: service:rest_client_request_duration_seconds:sum:rate
```
dgrisonnet, line 19:
```suggestion
        - record: service:rest_client_request_duration_seconds:rate5m
```

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,74 @@
...
+          expr: |
+              sum(rate(
+                label_replace(rest_client_request_duration_seconds_bucket{verb="GET",url!~"https?://(api-int|\\[::1\\]|127\\.0\\.0\\.1|localhost).*"},
+                  "host","$1","url","https?://([^/\\s]+).*")[5m:30s]
+                )) by (le,host,service,namespace)
+        - record: pod:rest_client_request_duration_seconds:sum:rate
```
dgrisonnet, line 25:
```suggestion
        - record: pod:rest_client_request_duration_seconds:rate5m
```

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,74 @@
...
+              sum(rate(
+                label_replace(rest_client_request_duration_seconds_bucket{verb="GET",url=~"https?://(\\[::1\\]|127\\.0\\.0\\.1|localhost).*"},
+                  "host","$1","url","https?://([^/\\s]+).*")[5m:30s]
+                )) by (le,host,service,namespace)
+        # latency by destination: internal-lb, services, localhost
+        - record: internal_load_balancer:rest_client_request_duration_seconds:histogram_quantile
+          labels:
+            quantile: "0.99"
+          expr: |
+            histogram_quantile(0.99, sum by (le) (internal_load_balancer:rest_client_request_duration_seconds:sum:rate))
```
dgrisonnet, line 36:
```suggestion
            histogram_quantile(0.99, sum by (le) (internal_load_balancer:rest_client_request_duration_seconds:rate5m))
```

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,74 @@
...
+        - record: service:rest_client_request_duration_seconds:histogram_quantile
+          labels:
+            quantile: "0.99"
+          expr: |
+            histogram_quantile(0.99, sum by (le) (service:rest_client_request_duration_seconds:sum:rate))
```
dgrisonnet, line 41:
```suggestion
            histogram_quantile(0.99, sum by (le) (service:rest_client_request_duration_seconds:rate5m))
```

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,74 @@
...
+        - record: pod:rest_client_request_duration_seconds:histogram_quantile
+          labels:
+            quantile: "0.99"
+          expr: |
+            histogram_quantile(0.99, sum by (le) (pod:rest_client_request_duration_seconds:sum:rate))
```
dgrisonnet, line 46:
```suggestion
            histogram_quantile(0.99, sum by (le) (pod:rest_client_request_duration_seconds:rate5m))
```

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,74 @@
...
+            sum(rate(rest_client_requests_total{code="<error>",host!~"(api-int|\\[::1\\]|127\\.0\\.0\\.1|localhost).*"}[5m])) by (host,service,namespace)
+            / ignoring (host,service,namespace) group_left
+            sum(rate(rest_client_requests_total{host!~"(api-int|\\[::1\\]|127\\.0\\.0\\.1|localhost).*"}[5m]))
+        - record: pod:rest_client_requests:errors:ratio_rate5m
+          expr: |
+            sum(rate(rest_client_requests_total{code="<error>",host=~"(\\[::1\\]|127\\.0\\.0\\.1|localhost).*"}[5m])) by (host,service,namespace)
+            / ignoring (host,service,namespace) group_left
+            sum(rate(rest_client_requests_total{host=~"(\\[::1\\]|127\\.0\\.0\\.1|localhost).*"}[5m]))
+        # errors by destination
+        - record: internal_load_balancer:rest_client_requests_errors:sum
```
dgrisonnet, line 66: seeing the recording rule name again, the following might be better suited to not confuse with an actual sum:
```suggestion
        - record: internal_load_balancer:rest_client_requests_errors:rate5m:sum
```

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,74 @@
...
+          expr: |
+            sum(internal_load_balancer:rest_client_requests_errors:ratio_rate5m)
+        - record: service:rest_client_requests:errors:sum
```
dgrisonnet, line 69:
```suggestion
        - record: service:rest_client_requests:errors:rate5m:sum
```

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,74 @@
...
+          expr: |
+            sum(service:rest_client_requests:errors:ratio_rate5m)
+        - record: pod:rest_client_requests:errors:sum
```
dgrisonnet, line 72:
```suggestion
        - record: pod:rest_client_requests:errors:rate5m:sum
```

dgrisonnet (commented):
Final comments! :tada:

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,74 @@
...
+        - record: internal_load_balancer:rest_client_requests_errors:ratio_rate5m
+          expr: |
+            sum(rate(rest_client_requests_total{code="<error>",host=~"api-int.*"}[5m])) by (host,service,namespace)
+            / ignoring (host,service,namespace) group_left
+            sum(rate(rest_client_requests_total{host=~"api-int.*"}[5m]))
+        - record: service:rest_client_requests:errors:ratio_rate5m
+          expr: |
+            sum(rate(rest_client_requests_total{code="<error>",host!~"(api-int|\\[::1\\]|127\\.0\\.0\\.1|localhost).*"}[5m])) by (host,service,namespace)
+            / ignoring (host,service,namespace) group_left
+            sum(rate(rest_client_requests_total{host!~"(api-int|\\[::1\\]|127\\.0\\.0\\.1|localhost).*"}[5m]))
```
dgrisonnet, line 59: This is the same as:
```suggestion
            sum(rate(rest_client_requests_total{code="<error>",host!~"(api-int|\\[::1\\]|127\\.0\\.0\\.1|localhost).*"}[5m]))
            /
            sum(rate(rest_client_requests_total{host!~"(api-int|\\[::1\\]|127\\.0\\.0\\.1|localhost).*"}[5m]))
```

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,74 @@
...
+        - record: pod:rest_client_requests:errors:ratio_rate5m
+          expr: |
+            sum(rate(rest_client_requests_total{code="<error>",host=~"(\\[::1\\]|127\\.0\\.0\\.1|localhost).*"}[5m])) by (host,service,namespace)
+            / ignoring (host,service,namespace) group_left
+            sum(rate(rest_client_requests_total{host=~"(\\[::1\\]|127\\.0\\.0\\.1|localhost).*"}[5m]))
```
dgrisonnet, line 64: This is the same as:
```suggestion
            sum(rate(rest_client_requests_total{code="<error>",host=~"(\\[::1\\]|127\\.0\\.0\\.1|localhost).*"}[5m]))
            /
            sum(rate(rest_client_requests_total{host=~"(\\[::1\\]|127\\.0\\.0\\.1|localhost).*"}[5m]))
```

bindata/assets/alerts/client-go-requests.yaml
```diff
@@ -0,0 +1,74 @@
...
+          expr: |
+            histogram_quantile(0.99, sum by (le) (pod:rest_client_request_duration_seconds:rate5m))
+        # error by host and job, the rest_client metrics use the code "<error>" to aggregate
+        # any non-http error and avoid cardinality problems.
+        # xref: https://github.com/kubernetes/kubernetes/blob/66931c9b8f11a3f223ba1890e4f390cff74ce1a6/staging/src/k8s.io/client-go/rest/request.go#L785-L799
+        - record: internal_load_balancer:rest_client_requests_errors:ratio_rate5m
+          expr: |
+            sum(rate(rest_client_requests_total{code="<error>",host=~"api-int.*"}[5m])) by (host,service,namespace)
+            / ignoring (host,service,namespace) group_left
+            sum(rate(rest_client_requests_total{host=~"api-int.*"}[5m]))
```
dgrisonnet, line 54: This is the same as:
```suggestion
            sum(rate(rest_client_requests_total{code="<error>",host=~"api-int.*"}[5m]))
            /
            sum(rate(rest_client_requests_total{host=~"api-int.*"}[5m]))
```

bindata/assets/alerts/client-go-requests.yaml
aojea: hmm, since your change to remove the url path right?
or am I missing something?

aojea: hmm, I think that is not the same, it aggregates based on those 3 fields, I can't remember now the reason but I'm sure that I had to do it this way because of a reason (duplicates??)

aojea:
we no longer need this

PR #1275, opened 2021-12-17
bug 2005901: Guard controller pdb

ingvagabund:
Create a new guard controller which:
- create one guarding pod per each static operand (the guarding pod assume the operand has /readyz endpoint up and running)
- create a PDB with minAvailable when the topology is not SNO
- adjust PDB's minAvailable based on the number of master nodes available
- serves as a controller which recreates guarding pods there were deleted/terminated


ingvagabund:
/hold
until https://github.com/openshift/library-go/pull/1268#discussion_r770769105 gets resolved through https://github.com/openshift/library-go/pull/1272

soltysh (approved):
/lgtm

PR #1276, opened 2021-12-17
Bug 2034484: library-go bump

jerpeter1:
pull in to grab fix described here openshift/library-go#1272


sanchezl (commented):
I don't see that the fix you mentioned has been pulled in. (needs another bump?)

qJkee:
@sanchezl 
fix is in `eca2c46` commit. As you can see, the same hash in `go.mod`

PR #1284, opened 2021-12-21
Bug 2034559: alerts: update latency thresholds

dgrisonnet:
Update latency thresholds used to compute the kube-apiserver SLO to
align with results from sig-scalability.

https://github.com/kubernetes/community/blob/master/sig-scalability/slos/api_call_latency.md#definition

### DIFF
--- a/bindata/assets/alerts/kube-apiserver-slos-basic.yaml
+++ b/bindata/assets/alerts/kube-apiserver-slos-basic.yaml
@@ -49,7 +49,7 @@ spec:
           (
             sum(rate(apiserver_request_duration_seconds_count{job="apiserver",verb=~"LIST|GET",subresource!~"proxy|log|exec",scope="resource"}[5m]))
           -
-            (sum(rate(apiserver_request_duration_seconds_bucket{job="apiserver",verb=~"LIST|GET",subresource!~"proxy|log|exec",scope="resource",le="0.1"}[5m])) or vector(0))
+            (sum(rate(apiserver_request_duration_seconds_bucket{job="apiserver",verb=~"LIST|GET",subresource!~"proxy|log|exec",scope="resource",le="1"}[5m])) or vector(0))
           ) / scalar(sum(rate(apiserver_request_total{job="apiserver",verb=~"LIST|GET",subresource!~"proxy|log|exec"}[5m])))
         , "type", "slow-resource", "_none_", "")
         or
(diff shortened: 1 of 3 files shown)


soltysh:
/hold
there's a clear issue with k8s-e2e-gcp since it didn't pass ever, so I'd rather we don't continue re-testing

dgrisonnet:
Holding until https://github.com/openshift/origin/pull/26739 is merged to not regress.

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -92,7 +92,7 @@ spec:
       sourceLabels:
       - __name__
     - action: drop
-      regex: apiserver_request_duration_seconds_bucket;(0.15|0.25|0.3|0.35|0.4|0.45|0.6|0.7|0.8|0.9|1.25|1.5|1.75|2.5|3|3.5|4.5|6|7|8|9|15|25|30|50)
+      regex: apiserver_request_duration_seconds_bucket;(0.4|0.6|0.8|1.25|1.5|3|6|8|15)
```
dgrisonnet, line 95: I removed some of the buckets that were dropped but aren't in the 1.23 list of buckets (https://github.com/kubernetes/kubernetes/pull/106306/commits/a8bd20a5af1174ce9f9a185b6d49f025a1045b03) and I also remove the 30s bucket since it is now used in the alert.

dgrisonnet:
PTAL @tkashem

bindata/assets/alerts/kube-apiserver-slos-basic.yaml
```diff
@@ -49,23 +49,23 @@ spec:
           (
             sum(rate(apiserver_request_duration_seconds_count{job="apiserver",verb=~"LIST|GET",subresource!~"proxy|log|exec",scope="resource"}[5m]))
           -
-            (sum(rate(apiserver_request_duration_seconds_bucket{job="apiserver",verb=~"LIST|GET",subresource!~"proxy|log|exec",scope="resource",le="0.1"}[5m])) or vector(0))
+            (sum(rate(apiserver_request_duration_seconds_bucket{job="apiserver",verb=~"LIST|GET",subresource!~"proxy|log|exec",scope="resource",le="1"}[5m])) or vector(0))
```
tkashem, line 52: if scope="resource", then verb should always be `GET`, and if scope is `namespace` verb should always be `LIST`, no? 

when we sync the alert with upstream then we can fix these. for now, let's merge it as is

dgrisonnet: that's correct and yeah that might be worth fixing upstream although they are not really differentiating between verb but rather "read" and "write" requests. So they are just taking "LIST|GET" as the verbs that would refer to read requests no matter the scope

PR #1307, opened 2022-02-03
Bug 2050407: revert dev cert rotation

tkashem:


tkashem:
/hold

we can cancel the hold before we start tracking 4.11

PR #1309, opened 2022-02-09
Bug 2052513: disable webhook supportability during upgrade

sanchezl:
Give customers running mis-configured webhooks before 4.10 a chance to upgrade to 4.10 before they are forced to fix their webhooks.

### DIFF
--- a/pkg/operator/webhooksupportabilitycontroller/webhook_supportability_controller.go
+++ b/pkg/operator/webhooksupportabilitycontroller/webhook_supportability_controller.go
@@ -3,6 +3,9 @@ package webhooksupportabilitycontroller
 import (
 	"context"
 
+	configv1 "github.com/openshift/api/config/v1"
+	configinformers "github.com/openshift/client-go/config/informers/externalversions"
+	configv1listers "github.com/openshift/client-go/config/listers/config/v1"
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/management"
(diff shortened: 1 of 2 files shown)


pkg/operator/webhooksupportabilitycontroller/webhook_supportability_controller.go
```diff
@@ -59,6 +66,18 @@ func (c *webhookSupportabilityController) sync(ctx context.Context, controllerCo
...
 
+	// do nothing while an upgrade is in progress
+	clusterVersion, err := c.clusterVersionLister.Get("version")
+	if err != nil {
+		return err
+	}
+	desired := clusterVersion.Status.Desired.Version
+	history := clusterVersion.Status.History
+	// upgrade is in progress if there is no history, or the latest history entry matches the desired version and is not completed
+	if len(history) == 0 || history[0].Version != desired || history[0].State != configv1.CompletedUpdate {
```
wking, line 77: The `desired` condition is probably overkill.  It's fine if you stay `Degraded` in any window where the CVO has accepted the retarget but not yet written a new `history` entry.  But it doesn't hurt to have the `desired` check either, if you want to keep it.

pkg/operator/webhooksupportabilitycontroller/webhook_supportability_controller.go
```diff
@@ -59,6 +66,18 @@ func (c *webhookSupportabilityController) sync(ctx context.Context, controllerCo
...
+		return nil
```
wking, line 78: If we return `nil` here, does something go through and actually clear any of these `Degraded` cases from the status?  Because we could also have situations like:

1. Cluster is 4.11.0.
2. Broke webhook installed.
3. Controller goes degraded.
4. Admin requests an update to 4.11.1.
5. Controller notices it's updating, and clears these Degraded webhook health-checks.
6. Update to 4.11.1 completes.
7. Controller notices it's no longer updating, and resets the Degraded webhook health-checks.

tkashem:
/lgtm
/approve
/hold

@wking you had a comment, if you are okay with this version, please feel free to cancel the hold.

tkashem:
/retitle Bug 2052339: disable webhook supportability during upgrade

pkg/operator/webhooksupportabilitycontroller/webhook_supportability_controller.go
wking: I think blocking 4.11 -> 4.11 and similar may be too strict, but we can always relax that later if I convince people.

/hold cancel

PR #1311, opened 2022-02-09
Bug 2052744: kube-apiserver Degraded=True if webhook references kubernetes.default.svc

sanchezl:
We ran into a case where the webhook was implemented in the apiserver, so the webhook configuration referenced the `kubernetes.default.svc` service.
* The webhook configuration did not specify a CABundle to use[1], so the webhook check failed since it attempted to connect without the needed CA certs. The 
* Nevertheless. the webhook should still work due to the [special case in the webhook impl](https://github.com/openshift/kubernetes/blob/9dd59017c4c218135b12949a2bd3af07e892199a/staging/src/k8s.io/apiserver/pkg/util/webhook/authentication.go#L200) that loads the CA certs provided by the service account when accessing the `kubernetes.default.svc` service. 

[1]: It asked for one to be injected, but the apiserver certs are not managed by the service-ca operator that injects CABundles.

### DIFF
--- a/pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
+++ b/pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
@@ -13,6 +13,7 @@ import (
 	operatorv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 	"k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
 )
 
@@ -101,8 +102,15 @@ func (c *webhookSupportabilityController) assertConnect(ctx context.Context, ref
 		port = fmt.Sprintf("%d", *reference.Port)
 	}
 	rootCAs := x509.NewCertPool()
-	if len(caBundle) > 0 {
+	switch {
+	case net.JoinHostPort(host, port) == "kubernetes.default.svc:443":
+		// ignore any caBundle specified and get one from the service account
+		rootCAs = serviceAccountCABundle()
+	case len(caBundle) > 0:
 		rootCAs.AppendCertsFromPEM(caBundle)
+	default:
+		// if no caBundle was specified, use the one from the service account
+		rootCAs = serviceAccountCABundle()
 	}
 	// the last error that occurred in the loop below
 	var err error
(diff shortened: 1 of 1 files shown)


aojea:
can there be some attack angle here?

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
```diff
@@ -101,8 +101,15 @@ func (c *webhookSupportabilityController) assertConnect(ctx context.Context, ref
 		port = fmt.Sprintf("%d", *reference.Port)
 	}
 	rootCAs := x509.NewCertPool()
-	if len(caBundle) > 0 {
+	switch {
+	case net.JoinHostPort(host, port) == "kubernetes.default.svc:443":
+		// ignore any caBundle specified and get one from the service account
```
aojea, line 106: that leaves caBundle field an ignored field for kubernetes.default.svc, right?

sanchezl: Correct. This matched the behavior in the apiserver when calling the webhook.

pkg/operator/webhooksupportabilitycontroller/webhook_supportability_controller.go
```diff
@@ -43,6 +46,7 @@ func NewWebhookSupportabilityController(
 		serviceLister:           kubeInformersForAllNamespaces.Core().V1().Services().Lister(),
 		crdLister:               apiExtensionsInformers.Apiextensions().V1().CustomResourceDefinitions().Lister(),
 		clusterVersionLister:    configInformers.Config().V1().ClusterVersions().Lister(),
+		serviceAccountCAs:       defaultCABundle(),
```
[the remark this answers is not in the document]
sanchezl: I removed the caching from the code.

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
```diff
@@ -101,8 +102,15 @@ func (c *webhookSupportabilityController) assertConnect(ctx context.Context, ref
...
+		rootCAs = serviceAccountCABundle()
+	case len(caBundle) > 0:
 		rootCAs.AppendCertsFromPEM(caBundle)
+	default:
+		// if no caBundle was specified, use the one from the service account
+		rootCAs = serviceAccountCABundle()
```
aojea, line 113: talked with @deads2k in slack

> if the host != "kubernetes.default.svc", then serviceaccount/ca.crt should not be used by default.

it seems we should not default it here , then?

PR #1314, opened 2022-02-10
[release-4.10] Bug 2052339: degraded webhook conditions to errors

tkashem:
Rename the following conditions so that they do not get rolled up into the overall Degraded status:

-MutatingAdmissionWebhookConfigurationDegraded
-ValidatingAdmissionWebhookConfigurationDegraded
-CRDConversionWebhookConfigurationDegraded
-VirtualResourceAdmissionDegraded

We will gather insights from connected customers in the future to determine if we can re-add these conditions to the overall Degraded status without impacting customer clusters.

cherry pick from https://github.com/openshift/cluster-kube-apiserver-operator/pull/1313


sanchezl (approved):
/lgtm

PR #1317, opened 2022-02-11
[WIP] Bug 2053268: track static pod lifecycle

tkashem:


tkashem:
cc @deads2k

pkg/operator/missingstaticpodcontroller/missingstaticpodcontroller.go
```diff
@@ -0,0 +1,209 @@
...
+			// we expect every installer pod to have a installerPodRevision in its name, unexpected error here
+			return fmt.Errorf("failed to get installerPodRevision for installer pod %q - %w", latestInstallerPodOnNode.Name, err)
+		}
+
+		finishedAt, ok := installerPodFinishedAt(latestInstallerPodOnNode)
+		if !ok {
+			// either it's in the process of running, or it ran into an error
+			continue
+		}
+		threshold := terminationGracePeriodSeconds(latestInstallerPodOnNode) + 10*time.Second
```
p0lyn0mial, line 75: I think this should take into account the grace period of an operand, not an installer pod, right? 
for example, `kas` needs more time to terminate than the other components, so we need to know how long it can take plus some buffor.

pkg/operator/missingstaticpodcontroller/missingstaticpodcontroller.go
```diff
@@ -0,0 +1,209 @@
...
+func getInstallerPodsByNode(pods []*corev1.Pod) map[string][]*corev1.Pod {
+	byNodes := map[string][]*corev1.Pod{}
+	for i := range pods {
+		pod := pods[i]
+		if !strings.HasPrefix(pod.Name, "installer-") {
+			continue
+		}
+
+		nodeName := pod.Spec.NodeName
+		if len(nodeName) == 0 {
```
p0lyn0mial, line 142: technically this should never happen, not sure if we would like to return an error just in case

pkg/operator/missingstaticpodcontroller/missingstaticpodcontroller.go
```diff
@@ -0,0 +1,209 @@
...
+		WithInformers(
+			operatorClient.Informer(),
+			kubeInformersForTargetNamespace.Core().V1().Pods().Informer(),
+		).
+		WithSync(c.sync).
+		WithSyncDegradedOnError(operatorClient).
+		ToController("MissingStaticPodController", eventRecorder)
+}
+
+func (c MissingStaticPodController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
```
p0lyn0mial, line 50: make it a pointer receiver

pkg/operator/missingstaticpodcontroller/missingstaticpodcontroller.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	return mostRecentInstallerPodByNode
+}
+
+func getInstallerPodsByNode(pods []*corev1.Pod) map[string][]*corev1.Pod {
+	byNodes := map[string][]*corev1.Pod{}
+	for i := range pods {
+		pod := pods[i]
+		if !strings.HasPrefix(pod.Name, "installer-") {
```
dgrisonnet, line 137: do we really need this check since all the pods passed to this function will have the `app: installer` label?

pkg/operator/missingstaticpodcontroller/missingstaticpodcontroller.go
```diff
@@ -0,0 +1,209 @@
...
+		operatorClient: operatorClient,
+		podLister:      kubeInformersForTargetNamespace.Core().V1().Pods().Lister(),
+	}
+	return factory.New().
+		WithInformers(
+			operatorClient.Informer(),
+			kubeInformersForTargetNamespace.Core().V1().Pods().Informer(),
+		).
+		WithSync(c.sync).
+		WithSyncDegradedOnError(operatorClient).
```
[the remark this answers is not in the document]
dgrisonnet: I think it is ok to report a degraded condition whenever an error occurs since they are reported whenever we can't verify the status of a static pod.

As for reporting status degraded for missing static pods, we could accumulate all the missing pods into one error that would be reported inside of the degraded status, wdyt?

pkg/operator/missingstaticpodcontroller/missingstaticpodcontroller.go
```diff
@@ -0,0 +1,209 @@
...
+	value := pod.Spec.TerminationGracePeriodSeconds
+	if value == nil {
+		return 0
+	}
+
+	return time.Duration(*value * int64(time.Second))
+}
+
+func getMostRecentInstallerPodByNode(pods []*corev1.Pod) map[string]*corev1.Pod {
+	mostRecentInstallerPodByNode := map[string]*corev1.Pod{}
```
dgrisonnet, line 120: this will not work if there are multiple installer pods inside of the same namespace. Are we certain that this will never be the case?

pkg/operator/missingstaticpodcontroller/missingstaticpodcontroller.go
p0lyn0mial: it is true today, what if in the future we will introduce something new?
I think we are only interested in `installer-revision` pods. I would leave it as it is.

pkg/operator/missingstaticpodcontroller/missingstaticpodcontroller.go
p0lyn0mial: we discussed it offline, Damien wanted to know if an installer-pod can install more than one pod.

pkg/operator/missingstaticpodcontroller/missingstaticpodcontroller.go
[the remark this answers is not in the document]
p0lyn0mial: > As for reporting status degraded for missing static pods, we could accumulate all the missing pods into one error that would be reported inside of the degraded status, wdyt?

yes, we should, otherwise, we might be updating the status too frequently.

pkg/operator/missingstaticpodcontroller/missingstaticpodcontroller.go
```diff
@@ -0,0 +1,209 @@
...
+			staticPodRevisionOnThisNode < installerPodRevision {
+			// if we are here:
+			//  a: the latest installer pod successfully completed at finishedAt
+			//  b: it has been more than 'terminationGracePeriodSeconds' + 10s since the
+			//     installer pod has completed at finishedAt
+			//  c. the static pod is not at correct installerPodRevision yet
+			// then all of the above conditions are true, so we should go degraded
+			//
+			// TODO: set a degraded condition
+			syncCtx.Recorder().Eventf("MissingStaticPod", "static pod lifecycle failure - installer pod: %q with revision: %d completed at: %s, static pod revision: %d",
```
p0lyn0mial, line 87: it can produce many events, one per every `sync` loop, we should find a way to avoid it.

pkg/operator/missingstaticpodcontroller/missingstaticpodcontroller.go
```diff
@@ -0,0 +1,209 @@
...
+
+	installerPods, err := c.podLister.List(labels.SelectorFromSet(labels.Set{"app": "installer"}))
+	if err != nil {
+		return err
+	}
+
+	// get the most recent installer pod for each node
+	latestInstallerPodsByNode := getMostRecentInstallerPodByNode(installerPods)
+	for node, latestInstallerPodOnNode := range latestInstallerPodsByNode {
+		installerPodRevision, err := getInstallerPodRevision(latestInstallerPodOnNode)
```
p0lyn0mial, line 64: it should work during an upgrade as well, right?

since upgrades can be installed on a `healthy` cluster all previous revisions would be `good`.

pkg/operator/missingstaticpodcontroller/missingstaticpodcontroller.go
p0lyn0mial: on an `sno` cluster, in case of a fallback, it can produce events forever.

PR #1328, opened 2022-03-10
Config Observer and Latency Controller for nodes.config.openshift.io WorkerLatencyProfile

swghosh:
- Config observer for [Worker Latency Profiles](https://github.com/openshift/enhancements/blob/master/enhancements/worker-latency-profile/worker-latency-profile.md) intended to update the following kube-api-server arguments on the basis of latency profile values found in [config.openshift.io/v1/node](https://github.com/openshift/api/blob/master/config/v1/types_node.go) resource.
   1. default-not-ready-toleration-seconds
   2. default-unreachable-toleration-seconds
- LatencyController to sync rollout status in config.openshift.io/v1/node/cluster object.


rphillips:
Looks like this PR needs a `go mod tidy` run on it.

pkg/operator/configobservation/node/observe_latency_profile.go
```diff
@@ -0,0 +1,127 @@
...
+	case node.Spec.WorkerLatencyProfile == configV1.DefaultUpdateDefaultReaction:
+		observedDefaultNotReadyTolerationSeconds = strconv.Itoa(configV1.DefaultNotReadyTolerationSeconds)
+	case node.Spec.WorkerLatencyProfile == configV1.MediumUpdateAverageReaction:
+		observedDefaultNotReadyTolerationSeconds = strconv.Itoa(configV1.MediumNotReadyTolerationSeconds)
+	case node.Spec.WorkerLatencyProfile == configV1.LowUpdateSlowReaction:
+		observedDefaultNotReadyTolerationSeconds = strconv.Itoa(configV1.LowNotReadyTolerationSeconds)
+	// in case of empty worker latency profile, do not update config
+	case node.Spec.WorkerLatencyProfile == "":
+		return existingConfig, errs
+	}
```
p0lyn0mial, line 46: what if we add a new profile in the future?

pkg/operator/configobservation/node/observe_latency_profile.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+)
+
+var notReadyTolerationSecondsPath = []string{"apiServerArguments", "default-not-ready-toleration-seconds"}
+var unreachableTolerationSecondsPath = []string{"apiServerArguments", "default-unreachable-toleration-seconds"}
+
+// ObserveNotReadyTolerationSeconds observes the value that should be set for default-not-ready-toleration-seconds
+// api server argument on the basis of provided worker latency profile from config node object.
+func ObserveNotReadyTolerationSeconds(genericListers configobserver.Listers, _ events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, errs []error) {
```
p0lyn0mial, line 21: `ObserveNotReadyTolerationSeconds` misses unit tests

pkg/operator/configobservation/node/observe_latency_profile.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+		return observedConfig, errs
+	}
+	// nothing has changed return the original configuration
+	return existingConfig, errs
+}
+
+// ObserveUnreachableTolerationSeconds observes the value that should be set for default-unreachable-toleration-seconds
+// api server argument on the basis of provided worker latency profile from config node object.
+func ObserveUnreachableTolerationSeconds(genericListers configobserver.Listers, _ events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, errs []error) {
```
p0lyn0mial, line 76: please add unit tests

pkg/operator/configobservation/node/observe_latency_profile.go
```diff
@@ -0,0 +1,127 @@
...
+	nodeLister := genericListers.(configobservation.Listers).NodeLister
+	node, err := nodeLister.Get("cluster")
+	if err != nil && !apierrors.IsNotFound(err) {
+		// we got an error so without the node object we are not able to determine worker latency profile
+		return existingConfig, append(errs, err)
+	}
+
+	// read the observed value
+	var observedDefaultUnreachableTolerationSeconds string
+	switch {
```
p0lyn0mial, line 91: consider handling unsupported/future values

pkg/operator/configobservation/node/observe_latency_profile.go
```diff
@@ -0,0 +1,127 @@
...
+	case node.Spec.WorkerLatencyProfile == "":
+		return existingConfig, errs
+	}
+
+	// read the current value
+	var currentDefaultUnreachableTolerationSeconds string
+	currentDefaultUnreachableTolerationSecondsSlice, _, err := unstructured.NestedStringSlice(
+		existingConfig, unreachableTolerationSecondsPath...)
+	if err != nil {
+		errs = append(errs, fmt.Errorf("unable to extract not ready toleration seconds from the existing config: %v", err))
```
p0lyn0mial, line 108: update the message to match `unreachableTolerationSecondsPath`

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,216 @@
...
+	controlplanev1 "github.com/openshift/api/kubecontrolplane/v1"
+	configinformers "github.com/openshift/client-go/config/informers/externalversions"
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/operatorclient"
+)
+
+const (
+	operatorNamespace = operatorclient.OperatorNamespace
+	targetNamespace   = operatorclient.TargetNamespace
+
+	ApiServerConfigMapName = "config"
```
p0lyn0mial, line 33: it looks like all those fields could be private.

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,216 @@
...
+	eventRecorder events.Recorder,
+) factory.Controller {
+
+	ret := &LatencyProfileController{
+		operatorClient:  operatorClient,
+		configMapClient: v1helpers.CachedConfigMapGetter(kubeClient.CoreV1(), kubeInformersForNamespaces),
+		nodeLister:      configInformer.Config().V1().Nodes().Lister(),
+		nodeClient:      configClient.ConfigV1().Nodes(),
+	}
+	configClient.ConfigV1().Nodes()
```
p0lyn0mial, line 63: remove

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,216 @@
...
+	configClient configv1client.Clientset,
+	configInformer configinformers.SharedInformerFactory,
+	kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces,
+	kubeClient kubernetes.Interface,
+	eventRecorder events.Recorder,
+) factory.Controller {
+
+	ret := &LatencyProfileController{
+		operatorClient:  operatorClient,
+		configMapClient: v1helpers.CachedConfigMapGetter(kubeClient.CoreV1(), kubeInformersForNamespaces),
```
p0lyn0mial, line 59: a `lister` won't do?

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,216 @@
...
+
+	ApiServerConfigMapName = "config"
+	ApiServerConfigMapKey  = "config.yaml"
+
+	DefaultNotReadyTolerationSecondsArgument    = "default-not-ready-toleration-seconds"
+	DefaultUnreachableTolerationSecondsArgument = "default-unreachable-toleration-seconds"
+)
+
+// LatencyProfileController
+type LatencyProfileController struct {
```
p0lyn0mial, line 41: it would be nice if you could document the purpose of this controller

p0lyn0mial, line 41: please add unit tests

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,216 @@
...
+	configInformer configinformers.SharedInformerFactory,
+	kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces,
+	kubeClient kubernetes.Interface,
+	eventRecorder events.Recorder,
+) factory.Controller {
+
+	ret := &LatencyProfileController{
+		operatorClient:  operatorClient,
+		configMapClient: v1helpers.CachedConfigMapGetter(kubeClient.CoreV1(), kubeInformersForNamespaces),
+		nodeLister:      configInformer.Config().V1().Nodes().Lister(),
```
p0lyn0mial, line 60: could we change the name to `configNodeLister` as it is confusing with a regular `node lister`?

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,216 @@
...
+		nodeClient:      configClient.ConfigV1().Nodes(),
+	}
+	configClient.ConfigV1().Nodes()
+
+	return factory.New().WithInformers(
+		kubeInformersForNamespaces.InformersFor(targetNamespace).Core().V1().ConfigMaps().Informer(),
```
p0lyn0mial, line 66: why do we need a reactor for this informer?

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,216 @@
...
+		operatorClient.Informer(),
+		configInformer.Config().V1().Nodes().Informer(),
+	).ResyncEvery(time.Second).WithSync(ret.sync).ToController("LatencyProfileController", eventRecorder)
```
p0lyn0mial, line 69: we could `WithSyncDegradedOnError`

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+func (c *LatencyProfileController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	syncMethod := c.ensureProfileSyncedToAllNodes
+	err := syncMethod(ctx, syncCtx)
+	utilruntime.HandleError(err)
```
p0lyn0mial, line 75: this seems unnecessery.

pkg/operator/latencyprofilecontroller/controller.go
p0lyn0mial, line 69: why not to call `c.ensureProfileSyncedToAllNodes` directly?

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,216 @@
...
+	return err
+}
+
+// ensureProfileSyncedToAllNodes
+func (c *LatencyProfileController) ensureProfileSyncedToAllNodes(ctx context.Context, syncCtx factory.SyncContext) error {
```
p0lyn0mial, line 80: could you please explain the purpose of this method?

p0lyn0mial: Having the observers is not enough?

pkg/operator/configobservation/node/observe_latency_profile.go
```diff
@@ -0,0 +1,127 @@
...
+		// keep going, we are only interested in the observed value which will overwrite the current configuration anyway
+	}
+	if len(currentDefaultNotReadyTolerationSecondsSlice) > 0 {
+		currentDefaultNotReadyTolerationSeconds = currentDefaultNotReadyTolerationSecondsSlice[0]
+	}
+
+	// see if the current and the observed value differ
+	observedConfig := map[string]interface{}{}
+	if currentDefaultNotReadyTolerationSeconds != observedDefaultNotReadyTolerationSeconds {
```
p0lyn0mial, line 62: I run into the same issue recently. It seems there is no need to compare with the current value. For simplicity, you could always write the observed value.

pkg/operator/configobservation/node/observe_latency_profile.go
p0lyn0mial, line 21: should those observes be also set on an `SNO` (single node) cluster?

pkg/operator/configobservation/node/observe_latency_profile.go
```diff
@@ -0,0 +1,127 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	configV1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+)
+
+var notReadyTolerationSecondsPath = []string{"apiServerArguments", "default-not-ready-toleration-seconds"}
+var unreachableTolerationSecondsPath = []string{"apiServerArguments", "default-unreachable-toleration-seconds"}
```
p0lyn0mial, line 17: What happens on a downgrade? I think nothing will overwrite previously set values. Could you please double-check it?

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,216 @@
...
+			LastTransitionTime: metav1.Now(),
+			Message:            "Worker latency profile not set",
+		}
+		c.updateConfigNodeStatus(configNodeObj, &newCondition)
+	}
+
+	// For each revision, check that the configmap for that revision have
+	// correct argument values or not
+	revisionsHaveSynced := true
+	for _, revision := range uniqueRevisions {
```
p0lyn0mial, line 132: shouldn't we check only a single revision? 

I think we should wait until all API servers are on the same revision and then simply check the latest.

pkg/operator/latencyprofilecontroller/controller.go
swghosh: Removed (it was erroneous).

pkg/operator/latencyprofilecontroller/controller.go
swghosh: Added a summary.
```

// LatencyProfileController periodically lists the config/v1/node object
// and fetches the worker latency profile applied on the cluster which is used to
// updates the status of the config node object. The status updates reflect the
// state of kube-api-server(s) running on control plane node(s) and their
// observed config for default-not-ready-toleration-seconds and
// default-unreachable-toleration-seconds match the applied arguments.
```

pkg/operator/latencyprofilecontroller/controller.go
swghosh: Yes, made them as unexported variables.

pkg/operator/configobservation/node/observe_latency_profile.go
swghosh: Added `TestObserveNotReadyTolerationSeconds`.

pkg/operator/configobservation/node/observe_latency_profile.go
swghosh: Added `TestObserveUnreachableTolerationSeconds`.

pkg/operator/latencyprofilecontroller/controller.go
swghosh: We'd want to add the latency profile controller apart from just the config observers in favour of the following desired outcomes:
- having a controller would ensure us to update the config node CR status condition with appropriate status of K-apiserver
- we have certain edge cases where we would like the user refrain the user from transitioning latency profiles to/from extremeties, having a controller would allow rejecting such use cases back in the config observer
- we want to ensure that all control plane nodes in the cluster have reached the desired state i.e. applied the required K-apiserver flags before another update is trigerred via config observers, this helps in cluster to stay functional/usable

More: https://coreos.slack.com/archives/C0392N52Q5P/p1649867165534869

pkg/operator/latencyprofilecontroller/controller.go
swghosh: Renamed to `LatencyController.configNodeLister`.

pkg/operator/latencyprofilecontroller/controller.go
swghosh: Used `WithSyncDegradedOnError` now.

swghosh: Done, renamed `c.ensureProfileSyncedToAllNodes` to `c.sync` and used it within `.WithSync`; so we don't have an extraneous function now. 

IMHO, `ensureProfileSyncedToAllNodes` was incorrectly named as it was more like a function to: update config node object status on the basis of whether all node revisions have the specific argument rolled out or not.

pkg/operator/latencyprofilecontroller/controller.go
swghosh: Modified it; also: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1328#discussion_r854073835.

pkg/operator/latencyprofilecontroller/controller.go
swghosh: While implementing this:
- took majority of my inspiration from: 

https://github.com/openshift/cluster-kube-apiserver-operator/blob/10569df18e9577663802c0df9861bce540b16e2b/pkg/operator/boundsatokensignercontroller/controller.go#L238

https://github.com/openshift/cluster-kube-apiserver-operator/blob/10569df18e9577663802c0df9861bce540b16e2b/pkg/operator/boundsatokensignercontroller/controller.go#L257-L292

- this should be able to take care of the fact that more/less irrespective of different revision rollout(s) by various config observers the worker latency profile status stays up to date via checks on the given node revision if the required apiServerArgument is present or not
- also: (https://coreos.slack.com/archives/CB48XQ4KZ/p1648141150207779?thread_ts=1648139872.258959&cid=CB48XQ4KZ)
> > In case the KubeAPIServer resource on the cluster is changed twice within a short interval of time (assuming the changes were done by config observers), by design: will the operator's revision control pods transition the kube-apiserver static pods from revision x directly to revision x+2 or attempt the revisions consecutively i.e. revision x+1, revision x+2?
>
> It depends on timing and your luck, but I think it is possible to go straight to x+2.
At worst (assuming revision creation interval < time to apply revision to a single node), only one node will get x+1 and then x+2 , subsequent nodes will go directly to x+2. 

- so, in case of n frequent rollout(s) by different config observers nodes could attain revision x+n directly instead of x+1, x+2, ... x+n eventually; in case we wait for all nodes to attain target revision we the controller would keep waiting / take very long / never attain profile rollout Complete status.

pkg/operator/latencyprofilecontroller/controller.go
swghosh: The configNodeLister is used to get config/v1/node/cluster object and it's respective spec.workerLatencyProfile, however the other configV1Client is used to update status of that object.

The updated objects are as follows:
```golang
func NewLatencyProfileController(
	operatorClient v1helpers.StaticPodOperatorClient,
	configClient configv1.ConfigV1Interface,
	nodeInformer configv1informers.NodeInformer,
	kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces,
	kubeClient kubernetes.Interface,
	eventRecorder events.Recorder,
) factory.Controller {

	ret := &LatencyProfileController{
		operatorClient:   operatorClient,
		configClient:     configClient,
		configMapClient:  v1helpers.CachedConfigMapGetter(kubeClient.CoreV1(), kubeInformersForNamespaces),
		configNodeLister: nodeInformer.Lister(),
	}
```

pkg/operator/configobservation/node/observe_latency_profile.go
swghosh: For single node cluster(s), the behavior applied by these 2 config observer funcs should be quite idempotent.
Even in case, the cluster topology is SNO applying any worker latency profile shouldn't harm the cluster stability in any way. The two different k-apiserver arguments used here: `default-not-ready-toleration-seconds` and `default-unreachable-toleration-seconds` are essentially latency settings acting only when interaction between multiple control plane nodes come into play. Since this is our very first iteration of worker latency profiles, we're following a very straight forward approach and applied updated args in SNO clusters shouldn't cause any issue.

pkg/operator/configobservation/node/observe_latency_profile.go
swghosh: Added a default case for now. In the future if more profile(s) have to be added their respective arguments have to be included in this observer and controller anyway.

pkg/operator/configobservation/node/observe_latency_profile.go
swghosh: Similar to: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1328#discussion_r854181024.

pkg/operator/configobservation/node/observe_latency_profile.go
swghosh: In case of a downgrade this config observer, latency controller as well as config node CRD object will be absent; the absence of config observer would ensure that the argument gets unset? unset is default value.

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,188 @@
...
+		if _, ok := revisionMap[revision]; !ok {
+			revisionMap[revision] = struct{}{}
+			uniqueRevisions = append(uniqueRevisions, revision)
+		}
+	}
+
+	// Collect the current latency profile
+	desiredApiServerArgumentVals := map[string]string{}
+	configNodeObj, err := c.configNodeLister.Get("cluster")
+	if err != nil {
```
p0lyn0mial, line 100: can we expect to always get this resource? In the observers, you allow it to be missing.

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,188 @@
...
+	case apiconfigv1.LowUpdateSlowReaction:
+		desiredApiServerArgumentVals[DefaultNotReadyTolerationSecondsArgument] = strconv.Itoa(apiconfigv1.LowNotReadyTolerationSeconds)
+		desiredApiServerArgumentVals[DefaultUnreachableTolerationSecondsArgument] = strconv.Itoa(apiconfigv1.LowUnreachableTolerationSeconds)
+	}
+
+	condition := metav1.Condition{
+		Type:   apiconfigv1.KubeControllerManagerDegraded,
+		Status: metav1.ConditionUnknown,
+	}
+	if configNodeObj.Spec.WorkerLatencyProfile == "" {
```
p0lyn0mial, line 120: this is a case we should handle right after reading the `c.configNodeLister.Get("cluster")`

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,188 @@
...
+	uniqueRevisions := []int32{}
+	for _, nodeStatus := range operatorStatus.NodeStatuses {
+		revision := nodeStatus.CurrentRevision
+		if _, ok := revisionMap[revision]; !ok {
+			revisionMap[revision] = struct{}{}
+			uniqueRevisions = append(uniqueRevisions, revision)
+		}
+	}
+
+	// Collect the current latency profile
```
p0lyn0mial, line 97: I'd move it before collecting the revisions (line 86). Then I'd read the profile and check if there is work to do. Does it make sense?

pkg/operator/latencyprofilecontroller/status_utils.go
```diff
@@ -0,0 +1,74 @@
...
+	for i := range conditions {
+		if conditions[i].Type == conditionType {
+			return &conditions[i]
+		}
+	}
+
+	return nil
+}
+
+func (c *LatencyProfileController) updateConfigNodeStatus(newCondition metav1.Condition) (updated bool, err error) {
```
p0lyn0mial, line 49: I think you could use `operatorv1helpers.UpdateStatus`

pkg/operator/configobservation/node/observe_latency_profile.go
```diff
@@ -0,0 +1,105 @@
...
+		observedDefaultNotReadyTolerationSeconds = strconv.Itoa(configV1.DefaultNotReadyTolerationSeconds)
+	case configV1.MediumUpdateAverageReaction:
+		observedDefaultNotReadyTolerationSeconds = strconv.Itoa(configV1.MediumNotReadyTolerationSeconds)
+	case configV1.LowUpdateSlowReaction:
+		observedDefaultNotReadyTolerationSeconds = strconv.Itoa(configV1.LowNotReadyTolerationSeconds)
+	// in case of empty worker latency profile, do not update config
+	case "":
+		return existingConfig, errs
+	default:
+		return existingConfig, append(errs, fmt.Errorf("unknown worker latency profile found: %v", node.Spec.WorkerLatencyProfile))
```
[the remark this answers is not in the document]
swghosh: This would probably be a too extreme case, cause we already added k8s enum validation on the node CRD here:
- https://github.com/openshift/api/blob/441fe135b2fce3b4aac41a1488bea0b79cc6c800/config/v1/0000_10_config-operator_01_node.crd.yaml#L46-L52
- https://github.com/openshift/api/blob/master/config/v1/types_node.go

So essentially k8s API should reject any unknown values though only an empty spec results in `configNodeObj.Spec.workerLatencyProfile=""` and any other updates by operator/user rejects the update (both behaviours were tested manually).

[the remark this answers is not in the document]
p0lyn0mial: During an upgrade, the API server will be installed first, then some other controller could change the current profile based on some data even before this operator would be upgraded to a newer version.

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,188 @@
...
+)
+
+// LatencyProfileController periodically lists the config/v1/node object
+// and fetches the worker latency profile applied on the cluster which is used to
+// updates the status of the config node object. The status updates reflect the
+// state of kube-api-server(s) running on control plane node(s) and their
+// observed config for default-not-ready-toleration-seconds and
+// default-unreachable-toleration-seconds match the applied arguments.
+
+type LatencyProfileController struct {
```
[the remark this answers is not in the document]
swghosh: We can keep it unexported, shouldn't be a problem.

pkg/operator/configobservation/node/observe_latency_profile.go
```diff
@@ -0,0 +1,105 @@
...
+		[]string{observedDefaultNotReadyTolerationSeconds},
+		notReadyTolerationSecondsPath...); err != nil {
+		return existingConfig, append(errs, err)
+	}
+	return observedConfig, errs
+}
+
+// ObserveUnreachableTolerationSeconds observes the value that should be set for default-unreachable-toleration-seconds
+// api server argument on the basis of provided worker latency profile from config node object.
+func ObserveUnreachableTolerationSeconds(genericListers configobserver.Listers, _ events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, errs []error) {
```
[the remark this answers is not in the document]
swghosh: Do-able! Lemme update and put in a single function.

pkg/operator/configobservation/node/observe_latency_profile_test.go
```diff
@@ -0,0 +1,146 @@
...
+			})
+		}
+	}
+}
+
+func TestObserveNotReadyTolerationSeconds(t *testing.T) {
+	scenarios := []workerLatencyProfileTestScenario{
+		// scenario 1: empty worker latency profile
+		{
+			name:                  "default value is not applied when worker latency profile is unset",
```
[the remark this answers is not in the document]
swghosh: Yeah, but in practice found that [this empty spec config node CR](https://github.com/openshift/cluster-config-operator/blob/846caeae1b01bc48062f6cdab530196a26470b5e/empty-resources/0000_10_config-operator_02_node.cr.yaml#L10) that automatically comes up with the cluster / comes up with cluster upgrade would yield `configNodeObj.Spec.workerLatencyProfile=""` and hence testing this scenario here as well as `case ""` in config observer section.

pkg/operator/latencyprofilecontroller/status_utils.go
swghosh: Initially, was planning to use the same: operatorv1helpers.UpdateStatus but seems like we cannot, because that expects an OperatorClient and it's status object need to be operatorv1.Condition whereas in our case we're configv1.Node uses metav1.Condition in its status conditions. (hard luck till we have strong support for go generics I guess 😀)

Essentially, I copied over majority of the logic from there itself to attain similar status updation behaviour.

pkg/operator/latencyprofilecontroller/status_utils.go
```diff
@@ -0,0 +1,74 @@
...
+	"k8s.io/client-go/util/retry"
+)
+
+const (
+	reasonWorkerLatencyProfileUpdated = "ProfileUpdated"
+	reasonWorkerLatencyProfileEmpty   = "ProfileEmpty"
+)
+
+// setWLPStatusCondition is used to set condition in config node object status.workerLatencyProfileStatus
+func setWLPStatusCondition(conditions *[]metav1.Condition, newCondition metav1.Condition) {
```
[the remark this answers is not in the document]
swghosh: Same knock: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1328#discussion_r856530481.

pkg/operator/latencyprofilecontroller/controller.go
swghosh: Updated error accordingly, similar to what's done in config observer.

pkg/operator/latencyprofilecontroller/controller.go
swghosh: Updated!

pkg/operator/latencyprofilecontroller/controller.go
swghosh: Yeah, did that too.

pkg/operator/latencyprofilecontroller/status_utils.go
```diff
@@ -0,0 +1,74 @@
...
+		setWLPStatusCondition(&newStatus.WorkerLatencyProfileStatus.Conditions, newCondition)
+
+		if equality.Semantic.DeepEqual(oldStatus, newStatus) {
+			updated = false
+			// We return the newStatus which is a deep copy of oldStatus but with set condition applied.
+			return nil
+		}
+
+		configNodeObj.Status = *newStatus
+		_, err = c.configClient.Nodes().UpdateStatus(context.TODO(), configNodeObj, metav1.UpdateOptions{})
```
p0lyn0mial, line 68: if you want to set the status for a different operator then please move this controller to this operator. This operator should only manage its own status.

pkg/operator/configobservation/node/observe_latency_profile.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+)
+
+var notReadyTolerationSecondsPath = []string{"apiServerArguments", "default-not-ready-toleration-seconds"}
+var unreachableTolerationSecondsPath = []string{"apiServerArguments", "default-unreachable-toleration-seconds"}
+
+// ObserveNotReadyTolerationSeconds observes the value that should be set for default-not-ready-toleration-seconds
+// api server argument on the basis of provided worker latency profile from config node object.
+func ObserveNotReadyTolerationSeconds(genericListers configobserver.Listers, _ events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, errs []error) {
```
p0lyn0mial, line 21: I think you will need similar logic for `kube-controller-manager` and `kube-scheduler` operators. I think we could write a generic observer that could be placed in `library-go`.

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,198 @@
...
+// default-unreachable-toleration-seconds match the applied arguments.
+
+type latencyProfileController struct {
+	operatorClient   v1helpers.StaticPodOperatorClient
+	configClient     configv1.ConfigV1Interface
+	configMapClient  corev1client.ConfigMapsGetter
+	configNodeLister listerv1.NodeLister
+}
+
+func NewLatencyProfileController(
```
p0lyn0mial, line 50: if you decide to set the status for this operator (https://github.com/openshift/cluster-kube-apiserver-operator/pull/1328#discussion_r857344419) then we will need a similar controller for `kube-controller-manager` and `kube-scheduler`. We could write a generic controller and reuse it.

pkg/operator/configobservation/node/observe_latency_profile.go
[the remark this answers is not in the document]
p0lyn0mial: as discussed over the chat, it can break backward compatibility. Either document the risk and steps to ensure compatibility in the API or tolerate unknown values here and report an error from the operator.

pkg/operator/latencyprofilecontroller/status_utils.go
swghosh: Actually, we don't have a separate operator for config/v1/node, and if you check the [worker latency profile EP](https://github.com/openshift/enhancements/blob/master/enhancements/worker-latency-profile/worker-latency-profile.md#workerlatencyprofile) the idea is that the respective operand's (operand being Kubelet, Kube Controller Manager, Kube API Server) operator should update the config node status to propagate information about the latency profile rollout.

For now, we have [kubelet's latency profile status being updated by MCO](https://github.com/openshift/machine-config-operator/pull/3015) on the config node object.

p0lyn0mial: since the resource isn't managed by an operator what is the point of updating it? The status won't be visible by an end-user.

mfojtik: i do agree with @p0lyn0mial that updating node status from KAS operator looks like an anti-pattern to what we do in operators. only kubelet should update the node status, having multiple actors updating this status might only lead to problems and conflicts.

mfojtik: nit: please wire the context.TODO() properly :-)

swghosh: Using function signature as `func (c *latencyProfileController) updateConfigNodeStatus(ctx context.Context, newCondition metav1.Condition) (updated bool, err error) {` now to avoid using `context.TODO()`.

swghosh: @p0lyn0mial 
The status is visible to end user via the following command which prints the relevant status from the config node object.
```
$ oc get node.config -o yaml
  status:
    workerLatencyProfileStatus:
      conditions:
      - lastTransitionTime: "2022-04-25T09:28:02Z"
        message: worker latency profile not set on cluster
        reason: ProfileEmpty
        status: "False"
        type: KubeControllerManagerDegraded
      - lastTransitionTime: "2022-04-25T10:03:51Z"
        message: Machine Config Pool Updated
        reason: MachineConfigPoolUpdated
        status: "True"
        type: KubeletComplete
      - lastTransitionTime: "2022-04-25T10:03:50Z"
        message: kube-api-server(s) are updating the latency profile
        reason: ProfileUpdated
        status: "True"
        type: KubeAPIServerProgressing
```

@mfojtik 
For this feature to work we essentially need the args for Kubelet, KubeAPIServer and KubeControllerManager to be updated.
- In case of Kubelet, MCO updates the arg(s)
- In case of APIServer, this operator i.e. KubeAPIServer Operator should update the args(s)
- In case of ControllerManager, KubeControllerManager Operator should update the args(s)

This is documented in detail in the [Worker Latency Profile EP](https://github.com/openshift/enhancements/blob/master/enhancements/worker-latency-profile/worker-latency-profile.md) after discussing with @soltysh, @sttts and @deads2k.

Cc: @harche @rphillips

harche: Another example of how multiple operators let user know about the progress of the worker latency profile roll out. 

```
spec:
  workerLatencyProfile: Default
status:
  workerLatencyProfileStatus:
    conditions:
    - lastTransitionTime: "2022-04-26T16:09:39Z"
      message: ""
      reason: ProfileUpdated
      status: "False"
      type: KubeControllerManagerDegraded
    - lastTransitionTime: "2022-04-26T18:30:24Z"
      message: Machine Config Pool Updated
      reason: MachineConfigPoolUpdated
      status: "True"
      type: KubeletComplete
    - lastTransitionTime: "2022-04-26T17:26:38Z"
      message: ""
      reason: ProfileUpdated
      status: "False"
      type: KubeAPIServerProgressing
    - lastTransitionTime: "2022-04-26T16:57:06Z"
      message: all kube-api-server(s) have updated latency profile
      reason: ProfileUpdated
      status: "True"
      type: KubeAPIServerComplete
    - lastTransitionTime: "2022-04-26T18:24:37Z"
      message: ""
      reason: ProfileUpdated
      status: "False"
      type: KubeControllerManagerProgressing
    - lastTransitionTime: "2022-04-26T18:24:37Z"
      message: all kube-controller-manager(s) have updated latency profile
      reason: ProfileUpdated
      status: "True"
      type: KubeControllerManagerComplete
```

soltysh (commented):
All my comments from https://github.com/openshift/cluster-kube-controller-manager-operator/pull/611 apply here as well, :+1: to what @p0lyn0mial writes about moving the shared code to library-go.

swghosh:
/hold till openshift/library-go#1364 is merged.
We'll be using config observer(s) and controller constructed/invoked from implementation in library-go to add the functionality in this operator.

p0lyn0mial:
once the `library-go` PR merges, please split into two commits, one for the bump and the second for actual changes.

pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -138,6 +143,14 @@ func NewConfigObserver(
 			network.ObserveServicesSubnet,
 			network.ObserveExternalIPPolicy,
 			network.ObserveServicesNodePortRange,
+			nodeobserver.NewLatencyProfileObserver(
+				node.LatencyConfigs,
+				nodeobserver.NewSuppressConfigUpdateUntilSameProfileFunc(
+					operatorClient.(v1helpers.StaticPodOperatorClient),
```
p0lyn0mial, line 149: update `NewConfigObserver` to accept `operatorClient v1helpers.StaticPodOperatorClient`. I think it is a simple change.

swghosh: Changes done!

swghosh:
Invoked the implementation of latency profile config observer and latency profile controller introduced in openshift/library-go#1364.

@p0lyn0mial you may PTAL.

pkg/operator/latencyprofilecontroller/status_utils.go
swghosh: We're updating status in the operator now 🎉.

pkg/operator/configobservation/node/observe_latency_profile.go
swghosh: We're in library-go. (openshift/library-go#1364)

pkg/operator/latencyprofilecontroller/controller.go
swghosh: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1328#discussion_r871977546

PR #1331, opened 2022-03-23
Bug 2067384: OCP 4.10 should be firing APIRemovedInNextEUSReleaseInUse for APIs removed in 1.25

sanchezl:
* `APIRemovedInNextEUSReleaseInUse` should alert on use of APIs removed in OpenShift 4.11 & 4.12 (kube api server versions 1.24 and 1.25).
* `APIRemovedInNextReleaseInUse` should alert on use of APIs removed in OpenShift v4.11 (kube api server version 1.24).
* Added `TestAPIRemovedInNextEUSReleaseInUse` e2e test.

Once OpenShift 4.11 has been rebased onto kube apiserver 1.24, this will need to be updated again.


tkashem:
> Once OpenShift 4.11 has been rebased onto kube apiserver 1.24, this will need to be updated again.

do we already have a BZ for 4.11 so we can keep track?

PR #1334, opened 2022-03-31
Move node CPU alerts to MCO

dgrisonnet:
Move the node CPU alerts to the machine config operator which is their
rightful owner.

Add the release.openshift.io/delete: "true" annotation to the manifest
in order to have it cleaned up in 4.11. After 4.11 is released, we will
be able to remove the manifest for 4.12.

/cc @tkashem @wking

### DIFF
--- a/bindata/assets/alerts/cpu-utilization.yaml
+++ b/bindata/assets/alerts/cpu-utilization.yaml
@@ -3,6 +3,8 @@ kind: PrometheusRule
 metadata:
   name: cpu-utilization
   namespace: openshift-kube-apiserver
+  annotations:
+    release.openshift.io/delete: "true"
 spec:
   groups:
     - name: control-plane-cpu-utilization


bindata/assets/alerts/cpu-utilization.yaml
```diff
@@ -1,71 +0,0 @@
-apiVersion: monitoring.coreos.com/v1
-kind: PrometheusRule
-metadata:
-  name: cpu-utilization
-  namespace: openshift-kube-apiserver
```
wking, line 5: To avoid leaking this PrometheusRule resource in clusters updated from the old situation to the new situation, you should keep this manifest around (at least for 4.11, you can drop it in 4.12) with [the `release.openshift.io/delete: "true"` annotation][1].

[1]: https://github.com/openshift/enhancements/blob/master/enhancements/update/object-removal-manifest-annotation.md

dgrisonnet: Oh wow, I wasn't aware of that annotation, thank you for sharing. I used to write some logic in operators to clean up the manifests.

PR #1337, opened 2022-04-06
e2e: TestNamedCertificates: fix on SNO

sanchezl:
On a single node "cluster", the `TestNamedCertificates` tests would sometimes run before the test configuration was applied. Added a check to ensure at least one revision has been rolled out after the test configuration is created.

### DIFF
--- a/test/e2e/user_certs_test.go
+++ b/test/e2e/user_certs_test.go
@@ -24,9 +24,6 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 	configclient "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
-	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/operatorclient"
-	testlibraryapi "github.com/openshift/library-go/test/library/apiserver"
-
 	test "github.com/openshift/cluster-kube-apiserver-operator/test/library"
 )
 
@@ -50,9 +47,14 @@ func TestNamedCertificates(t *testing.T) {
 	configClient, err := configclient.NewForConfig(kubeConfig)
 	require.NoError(t, err)
 
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
 	// before starting our test make sure that all apis are on the same revision
 	// a previous test might have triggered a new revision and failed
-	testlibraryapi.WaitForAPIServerToStabilizeOnTheSameRevision(t, kubeClient.Pods(operatorclient.TargetNamespace))
+	revision, err := test.WaitForAPIServerPodsToStabilizeOnRevision(ctx, kubeClient, test.AnyRevision())
+	require.NoError(t, err)
+	t.Logf("initial revision: %d", revision)
 
 	// create secrets for named serving certificates
 	for _, info := range testCertInfoById {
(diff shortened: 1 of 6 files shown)


test/e2e/user_certs_test.go
```diff
@@ -205,7 +210,15 @@ func TestNamedCertificates(t *testing.T) {
 	}
 
 	// wait until a new version has been rolled out with the new configuration
-	testlibraryapi.WaitForAPIServerToStabilizeOnTheSameRevision(t, kubeClient.Pods(operatorclient.TargetNamespace))
+	for {
```
deads2k, line 213: how does this exit if the test times out?

test/e2e/user_certs_test.go
```diff
@@ -223,6 +236,22 @@ func TestNamedCertificates(t *testing.T) {
 	}
 }
 
+func apiserverPodRevision(kubeClient *clientcorev1.CoreV1Client) (int, error) {
+	var revision int
+	err := retry.RetryOnConnectionErrors(context.Background(), func(ctx context.Context) (bool, error) {
```
deads2k, line 241: I think you want https://github.com/kubernetes/client-go/blob/master/tools/watch/until.go#L127

test/e2e/user_certs_test.go
```diff
@@ -205,7 +210,15 @@ func TestNamedCertificates(t *testing.T) {
 	}
 
 	// wait until a new version has been rolled out with the new configuration
-	testlibraryapi.WaitForAPIServerToStabilizeOnTheSameRevision(t, kubeClient.Pods(operatorclient.TargetNamespace))
+	for {
+		testlibraryapi.WaitForAPIServerToStabilizeOnTheSameRevision(t, kubeClient.Pods(operatorclient.TargetNamespace))
```
p0lyn0mial, line 214: do we know why having this method is not enough?

sanchezl: When we call it the second time, sometimes the rollout of the new configuration has not yet "started" and we are still "synced" on previous revision.

test/e2e/user_certs_test.go
sanchezl: I took a stab at it. PTAL.

test/e2e/user_certs_test.go
sanchezl: go test causes a panic in the test executable.

test/e2e/user_certs_test.go
p0lyn0mial: the intention of `WaitForAPIServerToStabilizeOnTheSameRevision` method was to wait some time, even if nothing has changed before returning. I think this is the use case we have here.

sanchezl: > the intention of `WaitForAPIServerToStabilizeOnTheSameRevision` method was to wait some time, even if nothing has changed before returning. I think this is the use case we have here.

We were still missing the rollout of the configuration changes sometimes if the rollout was slow enough. What I really want is for my config changes to be rolled out, which means at least the next revision. Simply checking for the next revision isn't perfect either, but seems good enough given that all the e2e-operator tests are run serially (i.e. all in same package, no calls to `t.Parallel()`).

As an added bonus, when things are running smoothly, the `e2e-operator` tests spends less time waiting. I don't care past the next 'stable' revision as the tests are running serially.

p0lyn0mial: > We were still missing the rollout of the configuration changes sometimes if the rollout was slow enough

I have just checked and the `library.WaitForPodsToStabilizeOnTheSameRevision` waits 6 minutes to ensure the APIs are stable and on the same revision.

p0lyn0mial: I have also written a unit test to ensure it can handle a single pod and it seems to be working.

p0lyn0mial: are you sure the issue is with the `WaitForPodsToStabilizeOnTheSameRevision` method?

p0lyn0mial: ptal, https://github.com/openshift/library-go/pull/1350

PR #1340, opened 2022-04-19
scc bindata: fix hostmount-anyuid clusterrole to refer to the hostnmont-anyuid SCC

stlaz:
/assign @deads2k


deads2k:
Figure out what to do with the old one.

/approve
/hold

You can release the hold once you and the apiserver team have figured out what to do with the old clusterrole.  One possible suggestion: a controller that removes it if there is no SCC at the name (just because we don't create it, doesn't mean it doesn't exist)

PR #1347, opened 2022-04-26
[rebase v1.24]: remove insecure-port from kas args

tkashem:
`insecure-port` server option has been removed in 1.24 - https://github.com/kubernetes/kubernetes/blob/v1.23.7-rc.0/staging/src/k8s.io/apiserver/pkg/server/options/deprecated_insecure_serving.go

This is causing cluster bootstrap to fail in rebase PR with the following error:
```
E0426 04:06:12.761688       1 run.go:74] "command failed" err="unknown flag: --insecure-port"
```


soltysh (approved):
/lgtm

PR #1349, opened 2022-04-29
add HighAdmissionWebhookLatency alert

sanchezl:
Add `critical` alert `HighAdmissionWebhookLatency` that fires when the average duration of an admission webhook is > 1s.

### DIFF
--- a/bindata/assets/alerts/kube-apiserver-webhook.yaml
+++ b/bindata/assets/alerts/kube-apiserver-webhook.yaml
@@ -1,0 +1,27 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: webhooks
+  namespace: openshift-kube-apiserver
+spec:
+  groups:
+    - name: webhooks
+      rules:
+        - alert: HighAdmissionWebhookLatency
+          annotations:
+            summary: Webhook Latency is high.
+            description: >-
+              The maximum latency of the {{ $labels.name }} (type="{{ $labels.type }}") webhook 
+              has been over 1s for the past five minutes. Review the logs of the webhook workload to 
+              determine the cause of the high latency.
+            runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-kube-apiserver-operator/HighAdmissionWebhookLatency.md
+          expr: |
+            histogram_quantile(0.99, 
+              sum by (le,name)(
+                rate(apiserver_admission_webhook_admission_duration_seconds_bucket[5m])
+              )
+            ) > 1
+          for: 10m
+          labels:
+            namespace: openshift-kube-apiserver
+            severity: critical
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -153,6 +153,7 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		"assets/alerts/cpu-utilization.yaml",
 		"assets/alerts/kube-apiserver-requests.yaml",
 		"assets/alerts/kube-apiserver-slos-basic.yaml",
+		"assets/alerts/kube-apiserver-webhook.yaml",
 	}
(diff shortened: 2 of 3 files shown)


bindata/assets/alerts/kube-apiserver-webhook.yaml
```diff
@@ -0,0 +1,20 @@
...
+  namespace: openshift-kube-apiserver
+spec:
+  groups:
+    - name: webhooks
+      rules:
+        - alert: HighAdmissionWebhookLatency
+          annotations:
+            summary: Webhook Latency is high.
+            description: >-
+              The average latency due to the {{ $labels.name }} (type="{{ $labels.type }}") webhook is over 1s.
```
dgrisonnet, line 14: This description does not match what the expression is computing. The following expression returns true if any of the requests during the last 5 minutes took more than 1s. 
```
(sum(rate(apiserver_admission_webhook_admission_duration_seconds_sum[5m]))by(name)-sum(rate(apiserver_admission_webhook_admission_duration_seconds_bucket{le="1"}[5m]))by(name))>0
```

If you want to alert based on the average you should use the following expression:
```
sum(rate(apiserver_admission_webhook_admission_duration_seconds_sum[5m]))by(name)-sum(rate(apiserver_admission_webhook_admission_duration_seconds_count[5m]))by(name)>1000
```

But I don't think alerting on the average is good enough, maybe on the 95th / 99th percentile might be better, wdyt?

bindata/assets/alerts/kube-apiserver-webhook.yaml
```diff
@@ -0,0 +1,20 @@
...
+          expr: |
+            (sum(rate(apiserver_admission_webhook_admission_duration_seconds_sum[5m]))by(name)-sum(rate(apiserver_admission_webhook_admission_duration_seconds_bucket{le="1"}[5m]))by(name))>0
+          for: 10m
+          labels:
+            namespace: openshift-kube-apiserver
+            severity: critical
```
dgrisonnet, line 20: Because this alert is critical, we have to introduce a runbook to it.

bindata/assets/alerts/kube-apiserver-webhook.yaml
sanchezl: I have reworked to check for 99th percentile.

simonpasquier (commented):
pinging @openshift/sre-alert-sme since this adds a critical alert to OCP.

mmazur:
From a fleet SRE perspective the primary question always is – how flappy will this be? Across hundreds/thousands of Rosa clusters how often will this fire for 5–15 minutes and then autoresolve before an SRE will time to figure out what's wrong.

Is there not a way to run such queries against historical data to figure such things out beforehand?

bindata/assets/alerts/kube-apiserver-webhook.yaml
```diff
@@ -0,0 +1,27 @@
...
+    - name: webhooks
+      rules:
+        - alert: HighAdmissionWebhookLatency
+          annotations:
+            summary: Webhook Latency is high.
+            description: >-
+              The maximum latency of the {{ $labels.name }} (type="{{ $labels.type }}") webhook 
+              has been over 1s for the past five minutes. Review the logs of the webhook workload to 
+              determine the cause of the high latency.
+            runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-kube-apiserver-operator/HighAdmissionWebhookLatency.md
```
dofinn, line 17: runbook doesnt exist

bindata/assets/alerts/kube-apiserver-webhook.yaml
dofinn: the runbook mentions that this condition is slowing down API server requests - should we confirm this statement within the `expr` and also detect that PUTs/POSTs are indeed increasing in duration?

PR #1354, opened 2022-05-05
Update library-go to 80f9619c2

pacevedom:
Relates to https://github.com/openshift/library-go/pull/1347, where we increase missing static pods timeout in case we are in single node.


p0lyn0mial:
Could you split it into 2 commits? One for the bump and the second for the actual change. Thanks.

pacevedom:
> Could you split it into 2 commits? One for the bump and the second for the actual change. Thanks.

Sure. Should I do the same for all 4 operators?

p0lyn0mial:
/lgtm
/approve

Thanks!

p0lyn0mial:
> Sure. Should I do the same for all 4 operators?

Yes, please

PR #1358, opened 2022-05-26
Bug 2086519: exempt build controller SA from PodSecurity admission

stlaz:
/assign @deads2k


bindata/assets/config/defaultconfig.yaml
```diff
@@ -20,6 +20,9 @@ admission:
           audit-version: "latest"
           warn: "restricted"
           warn-version: "latest"
+        exemptions:
+          usernames:
+          - system:serviceaccount:openshift-infra:build-controller
```
deads2k, line 25: add comments explaining why this is safe.

This does match my expectations.

PR #1359, opened 2022-06-03
Bug 2074031: Support tuning GOGC within a limited range.

benluddy:
From the upstream release notes:

> Kubernetes 1.24 bumped version of golang it is compiled with to go1.18, which introduced significant changes to its garbage collection algorithm. As a result, we observed an increase in memory usage for kube-apiserver in larger an heavily loaded clusters up to ~25% (with the benefit of API call latencies drop by up to 10x on 99th percentiles). If the memory increase is not acceptable for you you can mitigate by setting GOGC env variable (for our tests using GOGC=63 brings memory usage back to original value, although the exact value may depend on usage patterns on your cluster).

Since kube-apiserver is managed by an operator on OpenShift, they'll need a configuration knob to tune this value if desired.

### DIFF
--- a/bindata/assets/kube-apiserver/pod.yaml
+++ b/bindata/assets/kube-apiserver/pod.yaml
@@ -127,6 +127,8 @@ spec:
         valueFrom:
           fieldRef:
             fieldPath: status.hostIP
+      - name: GOGC
+        value: "{{ .GOGC }}"
     securityContext:
       privileged: true
   - name: kube-apiserver-cert-syncer
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -412,25 +412,21 @@ func proxyMapToEnvVars(proxyConfig map[string]string) []corev1.EnvVar {
 	return envVars
 }
 
-func gracefulTerminationDurationFromConfig(operatorSpec *operatorv1.StaticPodOperatorSpec) (int, error) {
+func gracefulTerminationDurationFromConfig(config map[string]interface{}) (int, error) {
+	// 135s is our default value
+	//   the initial 70s is reserved fo the minimal termination period
+	//   additional 60s for finishing all in-flight requests
+	//   an extra 5s to make sure the potential SIGTERM will be sent after the server terminates itself
+	const defaultDuration = 135
+
 	var gracefulTerminationDurationPath = []string{"gracefulTerminationDuration"}
 
-	mergedConfigs, err := resourcemerge.MergeProcessConfig(map[string]resourcemerge.MergeFunc{}, operatorSpec.ObservedConfig.Raw, operatorSpec.UnsupportedConfigOverrides.Raw)
-	if err != nil {
-		return 0, err
-	}
-
-	// read the watch termination from the observed configuration
-	observedConfig := map[string]interface{}{}
-	if err := json.NewDecoder(bytes.NewBuffer(mergedConfigs)).Decode(&observedConfig); err != nil {
-		return 0, err
-	}
-	observedGracefulTerminationDurationStr, _, err := unstructured.NestedString(observedConfig, gracefulTerminationDurationPath...)
+	observedGracefulTerminationDurationStr, _, err := unstructured.NestedString(config, gracefulTerminationDurationPath...)
 	if err != nil {
 		return 0, fmt.Errorf("unable to extract gracefulTerminationDuration from the observed config: %v, path = %v", err, gracefulTerminationDurationPath)
 	}
 	if len(observedGracefulTerminationDurationStr) == 0 {
-		return 0, nil
+		return defaultDuration, nil
 	}
 	observedGracefulTerminationDuration, err := strconv.Atoi(observedGracefulTerminationDurationStr)
 	if err != nil {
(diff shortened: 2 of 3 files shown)


vendor/github.com/openshift/api/operator/v1/types_kubeapiserver.go
```diff
@@ -29,6 +29,10 @@ type KubeAPIServer struct {
 
 type KubeAPIServerSpec struct {
 	StaticPodOperatorSpec `json:",inline"`
+
+	// todo: haggle over the api
+	// +optional
+	RuntimeParameters *RuntimeParameters `json:"runtimeParameters"`
```
deads2k, line 35: Instead of this, you could make the value settable only via unsupportedconfigoverrides. It would give you a way to handle the issue in the field if it arises, without growing new, risky, and hard to understand configuration options that will not be needed by most users.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -412,20 +412,21 @@ func proxyMapToEnvVars(proxyConfig map[string]string) []corev1.EnvVar {
 	return envVars
 }
 
-func gracefulTerminationDurationFromConfig(operatorSpec *operatorv1.StaticPodOperatorSpec) (int, error) {
+func gracefulTerminationDurationFromConfig(config map[string]interface{}) (duration int, err error) {
+	defer func() {
```
deads2k, line 416: This looks wrong.  If you want a default value, how about returning it in the case you want it used.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -412,20 +412,21 @@ func proxyMapToEnvVars(proxyConfig map[string]string) []corev1.EnvVar {
 	return envVars
 }
 
-func gracefulTerminationDurationFromConfig(operatorSpec *operatorv1.StaticPodOperatorSpec) (int, error) {
+func gracefulTerminationDurationFromConfig(config map[string]interface{}) (duration int, err error) {
```
deads2k, line 415: no named return values.  Too many accidents in kube, so we disallow it in the vast majority of cases.

bindata/assets/kube-apiserver/pod.yaml
```diff
@@ -127,6 +127,8 @@ spec:
         valueFrom:
           fieldRef:
             fieldPath: status.hostIP
+      - name: GOGC
+        value: {{ printf "%q" .GOGC }}
```
aojea, line 131: naive question, why can not this be templated directly and we have to use a printf?

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -440,12 +441,54 @@ func gracefulTerminationDurationFromConfig(operatorSpec *operatorv1.StaticPodOpe
 	return observedGracefulTerminationDuration, nil
 }
 
+func gogcFromConfig(config map[string]interface{}) (string, error) {
```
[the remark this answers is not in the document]
benluddy: The new config field consumed by this function is logically an int, but the runtime accepts both integers and the string "off" for the GOGC environment variable itself. The function signature is making the pedantic distinction (i.e. its return value holds a GOGC value), even though Go 1.18 only supports one non-int value that this function should never return. The precedent is there, so the idea was to avoid requiring the caller of gogcFromConfig to know what constitutes a valid GOGC value.

Do you have a strong preference against? I don't mind changing it, but I wrote it both ways and didn't feel there was a downside. It's future-proofed for the world in which we want gogcFromConfig to sometime return `100,flyingCars="yes, please"` without changing `manageTemplate`.

bindata/assets/kube-apiserver/pod.yaml
benluddy: It's related reasoning to https://github.com/openshift/cluster-kube-apiserver-operator/pull/1359#discussion_r896925534. This mostly avoids encoding too much information about GOGC into the template, since there's precedent for acceptable non-integer GOGC values. `"{{ .GOGC }}"` would produce invalid YAML if at some point a quotation mark can be present in the value of GOGC.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -440,12 +436,54 @@ func gracefulTerminationDurationFromConfig(operatorSpec *operatorv1.StaticPodOpe
...
+
+	gcPercentage, ok, err := unstructured.NestedString(config, gcPercentagePath...)
+	if err != nil {
+		return "", fmt.Errorf("unable to extract %q from the observed config: %v", strings.Join(gcPercentagePath, "."), err)
+	}
+	if !ok {
+		return "100", nil
+	}
+
+	gogc, err := strconv.Atoi(gcPercentage)
```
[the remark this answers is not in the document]
benluddy: That's right. The config field is expected to be a percentage (`garbageCollectionTargetPercentage`) and if a user tries to set it to anything else then it should produce an error. So in that way it's not just a pass-through to GOGC.

sanchezl (approved):

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -440,12 +436,54 @@ func gracefulTerminationDurationFromConfig(operatorSpec *operatorv1.StaticPodOpe
...
 
+func gogcFromConfig(config map[string]interface{}) (string, error) {
+	var gcPercentagePath = []string{"garbageCollectionTargetPercentage"}
+
+	gcPercentage, ok, err := unstructured.NestedString(config, gcPercentagePath...)
+	if err != nil {
+		return "", fmt.Errorf("unable to extract %q from the observed config: %v", strings.Join(gcPercentagePath, "."), err)
+	}
+	if !ok {
+		return "100", nil
```
p0lyn0mial, line 447: so, this is our default value

p0lyn0mial:
/approve

deffering lgtm to others

PR #1360, opened 2022-06-07
Bug 2100347: Bump(openshift/library-go) latency profile observer, controller

swghosh:
Bumps library-go with new changes in Latency Profile Observer and Controller.
- It fixes the problem described in PR comment: https://github.com/openshift/library-go/pull/1372/#discussion_r902344399 , due to which while transitioning from any latency profile -> "" (empty profile) old config values are retained instead of unsetting the latency config values
    - bumped latency controller will unset latency config arg values in observedConfig when re-activating empty profile

Signed-off-by: Swarup Ghosh <swghosh@redhat.com>


p0lyn0mial:
@swghosh please split into three commits: one for pinning the new library-go, one for actual bump and one for actual changes in the code. It will be easier to review

swghosh:
> please split into three commits: one for pinning the new library-go, one for actual bump and one for actual changes in the code. It will be easier to review

@p0lyn0mial done using 3 separate commits now.

p0lyn0mial:
thanks!

/approve

PR #1361, opened 2022-06-22
Bug 2100155: Add new alert on Pod Security violations

stlaz:
Cluster administrators should be aware when a Pod Security violation
occurs as that might cause workloads to break in the next .y OpenShift
version.

/cc @ibihim @s-urbaniak


soltysh (approved):
/approve

PR #1372, opened 2022-08-23
PodSecurity is on since 1.23, and locked to on in 1.25

soltysh:
/assign @tkashem


soltysh:
/hold 
for https://github.com/openshift/api/pull/1260

PR #1375, opened 2022-09-03
Bug 2073220: routes/status resources can leak sensitive data

aojea:
bump library-go to carry the fix


tkashem:
/retitle OCPBUGS-3293: routes/status resources can leak sensitive data

PR #1376, opened 2022-09-07
Bug 2094362: Duplicate prometheus rules for API SLOs after upgrade

sanchezl:
- use conditional static resource functionality
- delete prometheusrules/kube-apiserver-slos


iNecas:
Any updates on the PR? It seems this issue is causing quite a lot of noise to be generated (both for our customers as well as on fleet level when analyzing the overall health of connected clusters) and fixing it would help us better focus on more severe issues.

PR #1380, opened 2022-09-30
Bug 2100429: Allow ephemeral volumes in all SCCs

jsafrane:
"`ephemeral`" volumes, called as [Generic Ephemeral Volumes](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#generic-ephemeral-volumes) in upstream docs, are volumes represented by a [PersistentVolumeClaim template in a Pod](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#generic-ephemeral-volumes).

`ephemeral` is already part of upstream `restricted` policy: https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted

Using these volumes is safe for any user, as users with the capability to use "`ephemeral`" volumes can only fool Kubernetes to create PVCs, in the same fashion as users that can create StatefulSets can fool Kubernetes to create PVCs.

In both cases, the created PVC is subject of namespace's PVC quota. From Kubernetes/OCP point of view, it's just a regular PVC. Users won't get ability to create PVs (except for dynamic provisioning) nor ability to refer to an existing volume.

There were multiple discussions on Slack about this:
* https://coreos.slack.com/archives/CB48XQ4KZ/p1655465586780419
* @deads2k's semi-approval: https://coreos.slack.com/archives/CE4L0F143/p1663272051841919

**Note: this is a different feature than "CSI In-line Ephemeral Volumes", which *could* be unsafe and we're working on a way how to fix it in https://github.com/openshift/enhancements/pull/1240**

### DIFF
--- a/bindata/bootkube/scc-manifests/0000_20_kube-apiserver-operator_00_scc-anyuid.yaml
+++ b/bindata/bootkube/scc-manifests/0000_20_kube-apiserver-operator_00_scc-anyuid.yaml
@@ -37,6 +37,7 @@ volumes:
 - configMap
 - downwardAPI
 - emptyDir
+- ephemeral
 - persistentVolumeClaim
 - projected
 - secret
--- a/bindata/bootkube/scc-manifests/0000_20_kube-apiserver-operator_00_scc-hostaccess.yaml
+++ b/bindata/bootkube/scc-manifests/0000_20_kube-apiserver-operator_00_scc-hostaccess.yaml
@@ -41,6 +41,7 @@ volumes:
 - configMap
 - downwardAPI
 - emptyDir
+- ephemeral
 - hostPath
 - persistentVolumeClaim
 - projected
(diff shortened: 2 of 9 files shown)


jsafrane:
/assign @stlaz
For auth / SCC review

jsafrane:
/retest-required
Pathologic events in openshift-marketplace + disruption/image-registry connection/reused was unreachable

jsafrane:
> disruption/service-load-balancer-with-pdb connection/new was unreachable during disruption:  for at least 19s of 1h11m33s (maxAllowed=18s):

/retest

jsafrane:
/retest-required
`failed to create cluster: failed to apply Terraform`

dobsonj:
/retest-required
```
event happened 25 times, something is wrong: ns/openshift-config-operator pod/openshift-config-operator-58f7c8c7bb-qz788 node/ip-10-0-186-171.ec2.internal - reason/ProbeError Liveness probe error: Get "https://10.129.0.28:8443/healthz": net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

jsafrane:
/cherry-pick release-4.12
I wonder if it works with BZs

PR #1381, opened 2022-10-03
OCPBUGS-2198: AUTH-309: Wire support for trusted service account issuers

mfojtik:
This replaces cumbersome https://github.com/openshift/cluster-kube-apiserver-operator/pull/1379 with more clean logic.

The controller properly syncs the value from the authentication config into Kubernetes apiserver operator status and handles ALL cases.
The config observer then only uses the Kubernetes apiserver operator status as the source of truth and sets the flags according to the status fields.

TODO: replace the openshift/api bump with real bump once https://github.com/openshift/api/pull/1313 merge


pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
```diff
@@ -123,6 +127,34 @@ func observedConfig(
...
+// issuersChanged compares the command line flags used for KAS and the operator status service account issuers.
+// if these two sets are different, we have a change and we need to update the KAS.
+func issuersChanged(kasIssuers []string, activeOperatorIssuer string, trustedOperatorIssuers []string) bool {
+	sort.Strings(kasIssuers)
+	operatorAllIssuers := append([]string{activeOperatorIssuer}, trustedOperatorIssuers...)
+	sort.Strings(operatorAllIssuers)
+	return !sets.NewString(kasIssuers...).Equal(sets.NewString(operatorAllIssuers...))
+}
+
+func getTrustedServiceAccountIssuers(issuers []operatorv1.ServiceAccountIssuerStatus) []string {
```
s-urbaniak, line 139: nit: this is both present here and in the other commit in pkg/operator/serviceaccountissuercontroller/serviceaccountissuer_controller.go. It is a trivial func, however common logic. Suggestion is to centralize.

pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
```diff
@@ -123,6 +127,34 @@ func observedConfig(
...
+	result := []string{}
+	for i := range issuers {
+		if issuers[i].ExpirationTime != nil {
+			result = append(result, issuers[i].Name)
+		}
+	}
+	return result
+}
+
+func getActiveServiceAccountIssuer(issuers []operatorv1.ServiceAccountIssuerStatus) string {
```
s-urbaniak, line 149: same as `getTrustedServiceAccountIssuers`

pkg/operator/serviceaccountissuercontroller/serviceaccountissuer_controller.go
```diff
@@ -0,0 +1,206 @@
...
+				currentIssuer, activeIssuer, c.nowFn().Add(defaultTrustedServiceAccountIssuerExpirationDuration),
+			)
+		}
+		return err
+	}
+	return nil
+}
+
+// getActiveServiceAccountIssuer gets the active (currently used to generate new bound tokens) issuer.
+// This is the issue without expiration time set.
```
s-urbaniak, line 136: `This is the issuer`

pkg/operator/serviceaccountissuercontroller/serviceaccountissuer_controller.go
```diff
@@ -0,0 +1,206 @@
...
+		// at this point, we must have the active issuer (the one without expiration time).
+		// if we don't it means somebody changed the status deliberately.
+		// in this case, we correct it by setting the configured value as active.
+		// NOTE: this is an error/edge case
+		return c.makeActiveIssuerTrusted(ctx, currentIssuer, currentIssuer, operator)
+	}
+
+	// the issuer configured in auth config and the active issuer we have in operator status matches.
+	// this is no-op configuration wise, but we prune the list from expired issuers.
+	if currentIssuer == activeIssuer {
```
s-urbaniak, line 106: stylistic nit for better readability:
```
issuerChanged := currentIssuer != activeIssuer

if issuerChanged {
  c.makeActiveIssuerTrusted(...)
} else {
  c.pruneExpiredServiceAccountIssuers(...)
}
```

pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
```diff
@@ -123,6 +127,34 @@ func observedConfig(
 	}, errs
 }
 
+// issuersChanged compares the command line flags used for KAS and the operator status service account issuers.
+// if these two sets are different, we have a change and we need to update the KAS.
+func issuersChanged(kasIssuers []string, activeOperatorIssuer string, trustedOperatorIssuers []string) bool {
```
s-urbaniak, line 132: nit, this can be simplified for understandability as `issuersChanged(old []string, current ...string)`

and the caller-side: 
```
issuersChanged(
			existingIssuers,
			getActiveServiceAccountIssuer(operator.Status.ServiceAccountIssuers),
			getTrustedServiceAccountIssuers(operator.Status.ServiceAccountIssuers)...,
		)
```

pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
```diff
@@ -68,34 +68,38 @@ func observedConfig(
...
+	if err := checkIssuer(newActiveIssuer); err != nil {
 		return existingConfig, append(errs, err)
 	}
 
-	if len(newIssuer) != 0 {
-		issuerChanged = existingIssuer != newIssuer
+	if len(newActiveIssuer) > 0 {
+		issuerChanged = issuersChanged(
+			existingIssuers,
+			getActiveServiceAccountIssuer(operator.Status.ServiceAccountIssuers),
```
s-urbaniak, line 90: nit: this can be replaced with `newActiveIssuer ` and avoids double function call

pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
```diff
@@ -68,34 +68,38 @@ func observedConfig(
...
+			getTrustedServiceAccountIssuers(operator.Status.ServiceAccountIssuers),
+		)
+
+		value := []interface{}{newActiveIssuer}
+		for _, i := range getTrustedServiceAccountIssuers(operator.Status.ServiceAccountIssuers) {
```
s-urbaniak, line 95: nit: introduce `currentTrustedServiceAccountIssuers := getTrustedServiceAccountIssuers(operator.Status.ServiceAccountIssuers)` above line 88, use it for `issuersChanged` and here.

s-urbaniak:
general comment: `operator.Status.ServiceAccountIssuers` should have a maximum upper bound. Currently, it is easy to overwhelm the operator with configuring millions of service account issuers in a loop.

pkg/operator/serviceaccountissuercontroller/serviceaccountissuer_controller.go
mfojtik: I don't like if-elsing that much... in my opinion, it makes the code read worse, and in addition, the comment blocks are nested.

pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
mfojtik: @s-urbaniak, on the other hand, if I import the controller, I will make two controllers depend on each-other, which might, in the future, result in import loops. Since it is a simple trivial func, I believe it is fine to copy because the dependency hell is not worth it :-)

pkg/operator/serviceaccountissuercontroller/serviceaccountissuer_controller.go
s-urbaniak: you have a point with else-structures. really just a nit, ignore at will

pkg/operator/serviceaccountissuercontroller/serviceaccountissuer_controller.go
```diff
@@ -0,0 +1,213 @@
...
+			continue
+		}
+		// handle the case when new issuer is already in the trusted list
+		// this will remove it from the list
+		if server.Status.ServiceAccountIssuers[i].Name == newIssuer {
+			continue
+		}
+		updated = append(updated, server.Status.ServiceAccountIssuers[i])
+	}
+	fmt.Printf("updated=%d\n", len(updated))
```
s-urbaniak, line 173: i guess this needs to be removed

pkg/operator/serviceaccountissuercontroller/serviceaccountissuer_controller.go
```diff
@@ -0,0 +1,206 @@
...
+
+// getActiveServiceAccountIssuer gets the active (currently used to generate new bound tokens) issuer.
+// This is the issue without expiration time set.
+func getActiveServiceAccountIssuer(issuers []operatorv1.ServiceAccountIssuerStatus) string {
+	for i := range issuers {
+		if issuers[i].ExpirationTime == nil {
+			return issuers[i].Name
+		}
+	}
+	return ""
```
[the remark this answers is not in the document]
mfojtik: to correct that edge case one would need to read the KAS args and compare the active against the two... i am not sure if that is worth the complexity

mfojtik:
/override ci/prow/e2e-upgrade

the upgrade tests already passed on this PR and this is blocking 3 other backports

PR #1384, opened 2022-10-07
certrotation: rotate faster in test infrastructure

mfojtik:
This is kind of dirty, but it gets the job done. With this change, we no longer have to open Jira issues to remind us to revert short rotations before we ship. This way is also not doable in openshift.next when we ship much faster.

### DIFF
--- a/pkg/operator/certrotationcontroller/certrotationcontroller.go
+++ b/pkg/operator/certrotationcontroller/certrotationcontroller.go
@@ -2,7 +2,9 @@ package certrotationcontroller
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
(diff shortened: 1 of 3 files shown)


pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -123,9 +130,18 @@ func newCertRotationController(
...
-		// This must be reverted before we ship
-		rotationDay = rotationDay / 60
+		// For the internal CI testing, make the cert rotation much faster.
+		// This is RH infrastructure specific, dirty, but it gets the job done without need to revert/apply this on every dev cycle start
+		if !cache.WaitForCacheSync(ctx.Done(), configInformer.Config().V1().Infrastructures().Informer().HasSynced) {
+			utilruntime.HandleError(fmt.Errorf("caches did not sync"))
+			return nil, errors.New("failed to wait for infrastructure informer")
+		}
+		if infra, err := ret.infrastructureLister.Get("cluster"); err == nil {
+			if strings.HasPrefix(infra.Status.InfrastructureName, "ci-op") && strings.Contains(infra.Status.APIServerURL, "dev.rhcloud.com") {
```
wangke19, line 140: @mfojtik Please also add one more condition 'devcluster.openshift.com' of the cluster APIServer URL for QE testing. We have been using this rotationDay setting in our DR Cert testing.

PR #1389, opened 2022-10-11
tolerate newlines in alert expr

sanchezl:
Add `s` flag (let `.` match `\n`) to regexes so they work if alert expression has newlines.


soltysh (approved):
/approve

PR #1390, opened 2022-10-13
Fix typo in operator status serviceAccountIssuer name field

mfojtik:


mfojtik:
self-lgtm because this is only typo fix

PR #1393, opened 2022-10-17
serviceaccountissuer: fix case when default value is being used and not trusted after change

mfojtik:
As documented in https://github.com/openshift/cluster-kube-apiserver-operator/blob/005a95607cf9f8db490e962b549811d8bc0c5eaf/bindata/assets/config/config-overrides.yaml#L8 we set the default value for service account issuer if one is not specified in authentication config. This was missed in the original controller design, and we have to account for this default value. When the user specifies a custom issuer in auth config, we must preserve the default value previously used and keep it as a trusted value.

### DIFF
--- a/pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
+++ b/pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
@@ -26,6 +26,10 @@ var (
 	jwksURIPath              = []string{"apiServerArguments", "service-account-jwks-uri"}
 )
 
+// defaultServiceAccountIssuerValue is a value used when no service account issuer is configured.
+// This is in sync with bootstrap and post-bootstrap config overrides.
+const defaultServiceAccountIssuerValue = "https://kubernetes.default.svc"
+
 // ObserveServiceAccountIssuer changes apiServerArguments.service-account-issuer from
 // the default value if Authentication.Spec.ServiceAccountIssuer specifies a valid
 // non-empty value.
@@ -49,7 +53,7 @@ func observedConfig(existingConfig map[string]interface{},
 
 	errs := []error{}
 	var issuerChanged bool
-	var existingActiveIssuer, newActiveIssuer string
+	var existingConfigIssuer, observedActiveIssuer string
 	// when the issuer will change, indicate that by setting `issuerChanged` to true
 	// to emit the informative event
 	defer func() {
@@ -57,18 +61,28 @@ func observedConfig(existingConfig map[string]interface{},
 			recorder.Eventf(
 				"ObserveServiceAccountIssuer",
 				"ServiceAccount issuer changed from %v to %v",
-				existingActiveIssuer, newActiveIssuer,
+				existingConfigIssuer, observedActiveIssuer,
 			)
 		}
 	}()
 
-	existingIssuers, _, err := unstructured.NestedStringSlice(existingConfig, serviceAccountIssuerPath...)
+	existingConfigIssuers, _, err := unstructured.NestedStringSlice(existingConfig, serviceAccountIssuerPath...)
 	if err != nil {
 		errs = append(errs, fmt.Errorf("unable to extract service account issuer from unstructured: %v", err))
 	}
 
-	if len(existingIssuers) > 0 {
-		existingActiveIssuer = existingIssuers[0]
+	for i := range existingConfigIssuers {
+		if len(existingConfigIssuers[i]) > 0 {
+			existingConfigIssuer = existingConfigIssuers[i]
+			break
+		}
+	}
+
+	// If the issuer is not set, it is safe to assume it is being defaulted by the config override.
+	// However, if there is no issuer set in KAS, we need to default it here and also change the "service-account-jwks-uri".
+	if len(existingConfigIssuer) == 0 {
+		existingConfigIssuer = defaultServiceAccountIssuerValue
+		existingConfigIssuers = []string{existingConfigIssuer}
 	}
 
 	operator, err := getOperator("cluster")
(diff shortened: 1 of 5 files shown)


pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
```diff
@@ -71,6 +71,12 @@ func observedConfig(existingConfig map[string]interface{},
 		existingActiveIssuer = existingIssuers[0]
 	}
 
+	// if the issuer is not set, it is safe to assume it is being defaulted by the config override and we don't have
+	// to trigger revision rollout just to sync operator status with config.
+	if len(existingActiveIssuer) == 0 {
```
mfojtik, line 76: This is needed because the config this observer sync on is a lie :-) There is never an empty existingActiveIssuer; if it is, it has defaulted.
This change will default it here, so this observer won't trigger a new revision, thinking there is now a "new" value for the issuer.

mfojtik:
https://issues.redhat.com/browse/OCPBUGS-2436 this is why all the failures....

pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
```diff
@@ -71,6 +71,13 @@ func observedConfig(existingConfig map[string]interface{},
...
+		existingActiveIssuer = "https://kubernetes.default.svc"
```
stlaz, line 77: this should be a constant with a self-explanatory name

pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
```diff
@@ -71,6 +71,13 @@ func observedConfig(existingConfig map[string]interface{},
...
+		existingIssuers = []string{existingActiveIssuer}
```
stlaz, line 78: A corner case: `len(existingIssuers) > 0`, `existingIssuer[0] = ""; existingIssuer[1] = "https://nice.issuer"`.

In that case you're rewriting `existingIssuers`, which is a field that you read from the observed config, with `{ <default_issuer> }`.

Perhaps looking for a first non-empty issuer in `existingIssuers` and then doing this would be safer. It shouldn't happen that an issuer in the observed slice would be empty, but you can never be too sure :eyes:

pkg/operator/serviceaccountissuercontroller/serviceaccountissuer_controller.go
```diff
@@ -60,31 +73,27 @@ func (c *ServiceAccountIssuerController) sync(ctx context.Context, controllerCon
...
 		}
 		return statusUpdateErr
 	}
 
-	// in this case, user deleted service account issuer value, which means we have to remove all values from the status
-	// if we don't do this and instead just remove first value, the new tokens will be issued with "previous" service
-	// account issuer which is not desired
-	if len(currentIssuer) == 0 && len(operator.Status.ServiceAccountIssuers) != 0 {
+	// this is a case when issuer is not set in auth config and the operator status already has the default issuer set.
+	if isDefaultServiceAccountIssuer(currentIssuer, operator.Status.ServiceAccountIssuers) {
```
stlaz, line 87: why don't we instead set `currentIssuer` to `https://kubernetes.default.svc` if `len(currentIssuer) == 0`?

pkg/operator/serviceaccountissuercontroller/serviceaccountissuer_controller.go
```diff
@@ -60,31 +73,27 @@ func (c *ServiceAccountIssuerController) sync(ctx context.Context, controllerCon
...
+		return nil
+	}
+
+	// in this case the issuer was previously set in auth config, and later it was removed.
+	// this case must result into having default issuer set.
+	// this allows to instantly remove compromised issuers if needed.
+	if len(currentIssuer) == 0 {
 		operatorCopy := operator.DeepCopy()
-		operatorCopy.Status.ServiceAccountIssuers = []operatorv1.ServiceAccountIssuerStatus{}
+		operatorCopy.Status.ServiceAccountIssuers = defaultServiceAccountIssuerValue
```
stlaz, line 96: We cannot assume this, no? The previous issuer must still be trusted or your cluster gets bricked when you decide that you don't need to have your own issuer.

mfojtik: @stlaz i believe this is the corner case of "my cluster was hacked and I need to change the issuer right now". in that case, people might not care about bricking their cluster for ~1h if they can invalidate all tokens immediately

pkg/operator/serviceaccountissuercontroller/serviceaccountissuer_controller.go
mfojtik: because that will need a write to authConfig and I don't want to write default into auth config spec from a controller :-)

stlaz: I meant (line 67)
```go
currentIssuer := authConfig.Spec.ServiceAccountIssuer
if len(currentIssuer) == 0 {
    currentIssuer = "https://kubernetes.default.svc"
}

```

pkg/operator/serviceaccountissuercontroller/serviceaccountissuer_controller.go
stlaz: We do need some kind of a revocation mechanism, true. This should be documented in the API, though.

mfojtik: @stlaz i plan to update the product documentation with this (as was agreed in API review with Joel)

pkg/operator/serviceaccountissuercontroller/serviceaccountissuer_controller.go
mfojtik: @stlaz it will complicate the the revocation case and when I do this,random unit test start to fail (like pruning :-)

pkg/operator/serviceaccountissuercontroller/serviceaccountissuer_controller.go
```diff
@@ -60,31 +73,27 @@ func (c *ServiceAccountIssuerController) sync(ctx context.Context, controllerCon
...
-	if len(currentIssuer) == 0 && len(operator.Status.ServiceAccountIssuers) != 0 {
+	// this is a case when issuer is not set in auth config and the operator status already has the default issuer set.
+	if isDefaultServiceAccountIssuer(currentIssuer, operator.Status.ServiceAccountIssuers) {
+		return nil
+	}
+
+	// in this case the issuer was previously set in auth config, and later it was removed.
+	// this case must result into having default issuer set.
+	// this allows to instantly remove compromised issuers if needed.
+	if len(currentIssuer) == 0 {
```
stlaz, line 94: the below code only differs from the `if len(currentIssuer) == 0 && len(operator.Status.ServiceAccountIssuers) == 0 {` case by a different logging message.

I wonder if grouping them into a single branch could make this easier to read.

pkg/operator/serviceaccountissuercontroller/serviceaccountissuer_controller.go
stlaz: This must be a part of the API documentation as well. It's an action taken per a specific API field value, if it's not documented in the API, it is a bug.

mfojtik: @stlaz separate, existing bug :-) feel free to jira it.

pkg/operator/serviceaccountissuercontroller/serviceaccountissuer_controller.go
mfojtik: `len(operator.Status.ServiceAccountIssuers) == 0` should never happen as that means there is no issuer set :-)

stlaz: `len(operator.Status.ServiceAccountIssuers) == 0` is the initial state

mfojtik: gotcha, I combined those two cases, unit tests seems to pass :-)

pkg/operator/serviceaccountissuercontroller/serviceaccountissuer_controller_test.go
```diff
@@ -83,22 +83,30 @@ func TestController(t *testing.T) {
...
 				ServiceAccountIssuer: "newIssuer",
 			}},
 			operator: operatorv1.KubeAPIServer{
 				ObjectMeta: metav1.ObjectMeta{
 					Name: "cluster",
 				},
+				Status: operatorv1.KubeAPIServerStatus{
+					ServiceAccountIssuers: defaultServiceAccountIssuerValue,
+				},
 			},
```
stlaz, line 101: Add a test for when you configure the issuer specifically to `defaultServiceAccountIssuerValue[0].Name` and previous issuers are set.

The result should be that previous issuer is still trusted and the default issuer is the current one - compared to when you would set the issuer to an empty string.

pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
```diff
@@ -79,52 +93,54 @@ func observedConfig(existingConfig map[string]interface{},
 		return existingConfig, append(errs, err)
 	}
 
-	newActiveIssuer = getActiveServiceAccountIssuer(operator.Status.ServiceAccountIssuers)
-	if err := checkIssuer(newActiveIssuer); err != nil {
+	// desiredActiveIssuer is the desired service account issuer set in KAS operator status
+	// This apiServerArgumentValue is being synced using serviceaccountissuer controller.
+	desiredActiveIssuer = getActiveServiceAccountIssuer(operator.Status.ServiceAccountIssuers)
```
stlaz, line 98: nit: we would usually call this an `observedActiveIssuer`

pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
```diff
@@ -79,52 +93,54 @@ func observedConfig(existingConfig map[string]interface{},
...
+	// if desired active issuer is not set (the serviceaccountissuer for some reason has not defaulted it)
+	// then make sure, we default it here, because we have to set the jwks-uri correctly.
+	if existingConfigIssuer == defaultServiceAccountIssuerValue && len(desiredActiveIssuer) == 0 {
```
stlaz, line 101: `if existingConfigIssuer != defaultServiceAccountIssuerValue` but `desiredActiveIssuer == ""` we don't set `desiredActiveIssuer=defaultServiceAccountIssuerValue`?

pkg/operator/serviceaccountissuercontroller/serviceaccountissuer_controller_test.go
mfojtik: test added

pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
mfojtik: i struggle how that can happen other than somebody manually tweaking the KAS config :-)

mfojtik:
/retest

networking test flake

PR #1409, opened 2022-11-22
OCPBUGS-3985: update for featureset rendering

deads2k:
This was added to library-go and used in cluster-config-operator, this brings it in for future use here.

/assign @stlaz


deads2k:
now with fake bump to update library-go handling code for consumption *after* substitution.

PR #1410, opened 2022-11-29
OCPBUGS-3041: guard controller: set an explicit hostname to avoid name collisions

ingvagabund:
To avoid:
```
Oct 28 23:58:55.693391 ci-op-3hj6pnwf-4f6ab-lv57z-master-1 kubenswrapper[1657]: E1028 23:58:55.693346    1657 kubelet_pods.go:413] "Hostname for pod was too long, truncated it" podName="openshift-kube-scheduler-guard-ci-op-3hj6pnwf-4f6ab-lv57z-master-1" hostnameMaxLen=63 truncatedHostname="openshift-kube-scheduler-guard-ci-op-3hj6pnwf-4f6ab-lv57z-maste"
Oct 28 23:59:03.735726 ci-op-3hj6pnwf-4f6ab-lv57z-master-0 kubenswrapper[1670]: E1028 23:59:03.735671    1670 kubelet_pods.go:413] "Hostname for pod was too long, truncated it" podName="openshift-kube-scheduler-guard-ci-op-3hj6pnwf-4f6ab-lv57z-master-0" hostnameMaxLen=63 truncatedHostname="openshift-kube-scheduler-guard-ci-op-3hj6pnwf-4f6ab-lv57z-maste"
Oct 28 23:59:11.168082 ci-op-3hj6pnwf-4f6ab-lv57z-master-2 kubenswrapper[1667]: E1028 23:59:11.168041    1667 kubelet_pods.go:413] "Hostname for pod was too long, truncated it" podName="openshift-kube-scheduler-guard-ci-op-3hj6pnwf-4f6ab-lv57z-master-2" hostnameMaxLen=63 truncatedHostname="openshift-kube-scheduler-guard-ci-op-3hj6pnwf-4f6ab-lv57z-maste"
```

with identical `truncatedHostname="openshift-kube-scheduler-guard-ci-op-3hj6pnwf-4f6ab-lv57z-maste"` hostname. The new hostname is set to:
- `guard-ci-op-3hj6pnwf-4f6ab-lv57z-master-0`
- `guard-ci-op-3hj6pnwf-4f6ab-lv57z-master-1`
- `guard-ci-op-3hj6pnwf-4f6ab-lv57z-master-2`

Testing https://github.com/openshift/library-go/pull/1427

PR #1416, opened 2022-12-13
OCPBUGS-4478: guard controller: set an explicit hostname to avoid name collisions

ingvagabund:
Backporting https://github.com/openshift/cluster-kube-apiserver-operator/pull/1410


soltysh (approved):
/lgtm

PR #1417, opened 2022-12-20
Drop flags removed in k8s 1.26

soltysh:
These flags were removed in https://github.com/kubernetes/kubernetes/pull/112120, based on [the defaults](https://github.com/kubernetes/enhancements/blob/master/keps/sig-instrumentation/2845-deprecate-klog-specific-flags-in-k8s-components/README.md#logging-defaults)  we should be good with just dropping these flags.


sanchezl (approved):
/lgtm

PR #1419, opened 2022-12-21
make the bootstrap kube-apiserver honor cluster-wide featuregates

deads2k:
This is a precursor to a future update that will enable the new discovery API once we update to 1.26.  We want the *only* form of discovery ever available to be aggregated so that we can flush out problems in our new install clusters instead of having clients coast on caches.

/assign @tkashem
/cc @soltysh

### DIFF
--- a/bindata/bootkube/config/bootstrap-config-overrides.yaml
+++ b/bindata/bootkube/config/bootstrap-config-overrides.yaml
@@ -49,13 +49,8 @@ apiServerArguments:
     - /etc/kubernetes/secrets/etcd-client.key
   etcd-servers: {{range .EtcdServerURLs}}
     - {{.}}{{end}}
-  feature-gates:
-    - "APIPriorityAndFairness=true"
-    - "RotateKubeletServerCertificate=true"
-    - "DownwardAPIHugePages=true"
-{{- if .ServiceCIDR | len | eq 2}}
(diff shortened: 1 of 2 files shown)


bindata/bootkube/config/bootstrap-config-overrides.yaml
```diff
@@ -49,13 +49,8 @@ apiServerArguments:
     - /etc/kubernetes/secrets/etcd-client.key
   etcd-servers: {{range .EtcdServerURLs}}
     - {{.}}{{end}}
-  feature-gates:
-    - "APIPriorityAndFairness=true"
-    - "RotateKubeletServerCertificate=true"
-    - "DownwardAPIHugePages=true"
-{{- if .ServiceCIDR | len | eq 2}}
-    - "IPv6DualStack=true"
```
stlaz, line 57: Won't this feature gate be missing with the new code?

deads2k: IPv6 was forced to always true in 1.24 and removed in 1.25: https://github.com/kubernetes/kubernetes/blob/release-1.24/pkg/features/kube_features.go#L945

sanchezl (approved):

deads2k:
merge after https://github.com/openshift/cluster-kube-apiserver-operator/pull/1427

PR #1423, opened 2023-01-04
create two stage deadman switch for clusteroperator status

deads2k:
updating https://github.com/openshift/cluster-kube-apiserver-operator/pull/1216

This implements https://github.com/openshift/enhancements/pull/1025.  It creates a two stage check

1. every few minutes it modifies clusteroperator.status.condition[*].message to have a prefix about checking staleness.  Every operator is expected to continually update their status to reflect the actual state of the cluster so this should be written back.
2. every few minutes it checks to see if the special clusteroperator.status.condition[*].message prefix remains.  If the prefix remains for more than a minute, it will update the condition status, message, reason, and lasttransitiontime to indicate that the condition status is unknown and something is incorrect with the operator.  The Degraded condition is set to "True" because the operator is degraded even though we don't yet know why.

I hit another customer case over the break where something went badly wrong and it appears that operators were not setting their status conditions properly.


deads2k:
we'll see a few stages of failure:

1. if an operator is never resync-ing, we'll see test failure on the degraded condition.
2. manual inspection will show operators that are degraded and we don't expect that
3. a test can be written to see if we ever have an operator with the reason being `OperatorFailedStalenessCheck`
4. we can then extend the timeouts to something like 5 minute staleness challenge and another 5 minutes for the staleness failure.  That will account for image pull times in the field I think.

pkg/cmd/deadman/deadman.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func (o *Options) Run(ctx context.Context) error {
+	configClient, err := configeversionedclient.NewForConfig(o.controllerContext.KubeConfig)
+	if err != nil {
+		return fmt.Errorf("failed to create config client: %w", err)
+	}
+	configInformers := configv1informers.NewSharedInformerFactory(configClient, 1*time.Minute)
+
+	conditionChallenger := NewOperatorConditionChallenger(configClient, configInformers, 1*time.Minute)
```
mfojtik, line 76: Should the 1-minute duration between checks be configurable?

pkg/cmd/deadman/operator_condition_challenger.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	if isCheckingForStaleness(clusterOperator) {
+		// staleness check is in progress already, just return
+		return nil
+	}
+
+	mostRecentOperatorUpdate := getMostRecentConditionTransitionTime(clusterOperator)
+	if mostRecentOperatorUpdate == nil {
+		// this means we have no conditions, so there's no work to do.
```
mfojtik, line 101: Is it? Do we want to cover a case where the operator never updates its status, so the condition is stale/empty? Or a chance when third-party actors erase conditions and the operator never sets then back?

pkg/cmd/deadman/operator_condition_challenger.go
```diff
@@ -0,0 +1,179 @@
...
+func (c *OperatorConditionChallenger) syncHandler(ctx context.Context, key string) error {
+	clusterOperator, err := c.clusterOperatorLister.Get(key)
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	if isCheckingForStaleness(clusterOperator) {
```
mfojtik, line 94: Do we need to handle the case when the operator is unmanaged?

pkg/cmd/deadman/operator_condition_challenger.go
```diff
@@ -0,0 +1,179 @@
...
+	if now.Before(nextLivenessCheckTime) {
+		return nil
+	}
+
+	// if we're past the deadline since our last check, so we will perform a staleness check by resetting the message.
+	// only modify the message to avoid changing machine-readable reasons or status in the common case when operators
+	// are properly managing their status.
+	clusterOperatorToWriteAsStale := clusterOperator.DeepCopy()
+	for i, condition := range clusterOperator.Status.Conditions {
+		clusterOperatorToWriteAsStale.Status.Conditions[i].Message = "Checking for stale status, the active operator will reset this message: " + condition.Message
```
mfojtik, line 128: I think if we don't want to add a special field into the Condition type that indicates the staleness, to be more explicit.

pkg/cmd/deadman/operator_condition_challenger.go
```diff
@@ -0,0 +1,179 @@
...
+	newestTime := time.Time{}
+	for _, condition := range clusterOperator.Status.Conditions {
+		if condition.LastTransitionTime.After(newestTime) {
+			newestTime = condition.LastTransitionTime.Time
+		}
+	}
+	return &newestTime
+}
+
+func (c *OperatorConditionChallenger) syncHandler(ctx context.Context, key string) error {
```
mfojtik, line 85: Why not use library-go https://github.com/openshift/library-go/tree/master/pkg/controller/factory ?

pkg/cmd/deadman/operator_condition_challenger.go
```diff
@@ -0,0 +1,179 @@
...
+		return err
+	}
+
+	// mark the next time we should check for staleness
+	c.operatorToTimeToCheckForStaleness[clusterOperator.Name] = now.Add(c.durationAllowedBetweenStalenessChecks)
+
+	return nil
+}
+
+// Run starts the controller and blocks until stopCh is closed.
```
mfojtik, line 142: lies :-) stopCh is a thing of past ;-)

pkg/cmd/deadman/operator_staleness_check.go
```diff
@@ -0,0 +1,177 @@
...
+		},
+		1*time.Minute,
+	)
+
+	return c
+}
+
+func isCheckingForStaleness(clusterOperator *configv1.ClusterOperator) bool {
+	for _, condition := range clusterOperator.Status.Conditions {
+		if strings.HasPrefix(condition.Message, "Checking for stale status") {
```
mfojtik, line 73: We can avoid this with explicit field, however I think this is fine for initial implementation :-)

pkg/cmd/deadman/operator_staleness_check.go
```diff
@@ -0,0 +1,177 @@
...
+		return nil
+	}
+
+	// if we're past the deadline and some messsages still indicate we're checking for staleness, every condition should
+	// be marked as unknown and indicated as stale.
+	clusterOperatorToWriteAsStale := clusterOperator.DeepCopy()
+	for i, condition := range clusterOperator.Status.Conditions {
+		switch condition.Type {
+		case configv1.OperatorDegraded:
+			// if the operator is not setting status, it is Degraded.
```
mfojtik, line 110: One can argue that the "degraded" state indicates the "operand" is malfunctioning, not the operator. But I think it is fine, otherwise we are going to invent meta-operators here :-)

pkg/cmd/deadman/operator_staleness_check.go
```diff
@@ -0,0 +1,177 @@
...
+		default:
+			// The other conditions are Unknown.
+			clusterOperatorToWriteAsStale.Status.Conditions[i].Status = configv1.ConditionUnknown
+		}
+
+		if condition.Status != clusterOperatorToWriteAsStale.Status.Conditions[i].Status {
+			clusterOperatorToWriteAsStale.Status.Conditions[i].LastTransitionTime = metav1.Time{Time: now}
+		}
+		clusterOperatorToWriteAsStale.Status.Conditions[i].Reason = "OperatorFailedStalenessCheck"
+		clusterOperatorToWriteAsStale.Status.Conditions[i].Message = "Operator is not setting status.  Last status was: " + condition.Message
```
mfojtik, line 128: "Operator is not responding" ?
"Operator is not updating status for longer than N minutes" ?

pkg/cmd/deadman/operator_staleness_check.go
```diff
@@ -0,0 +1,177 @@
...
+
+		default:
+			// The other conditions are Unknown.
+			clusterOperatorToWriteAsStale.Status.Conditions[i].Status = configv1.ConditionUnknown
+		}
+
+		if condition.Status != clusterOperatorToWriteAsStale.Status.Conditions[i].Status {
+			clusterOperatorToWriteAsStale.Status.Conditions[i].LastTransitionTime = metav1.Time{Time: now}
+		}
+		clusterOperatorToWriteAsStale.Status.Conditions[i].Reason = "OperatorFailedStalenessCheck"
```
mfojtik, line 127: This is perhaps event worthy? It will help us write synthetic tests :-)

pkg/cmd/deadman/deadman.go
deads2k: > Should the 1-minute duration between checks be configurable?

I'm open to this.

pkg/cmd/deadman/operator_condition_challenger.go
deads2k: > Is it? Do we want to cover a case where the operator never updates its status, so the condition is stale/empty? Or a chance when third-party actors erase conditions and the operator never sets then back?

If we have no conditions, it's not clear what this controller should set.  I think this logic would only fail during failed installations.

pkg/cmd/deadman/operator_condition_challenger.go
deads2k: > Do we need to handle the case when the operator is unmanaged?

Meaning you're concerned that an unmanaged operator wouldn't set status?  Seems like it would need to continually set an "I'm not sure" status, so it would be a bug either way.

pkg/cmd/deadman/operator_condition_challenger.go
deads2k: > I think if we don't want to add a special field into the Condition type that indicates the staleness, to be more explicit.

Conditions are standardized in metav1.  Our type pre-dates metav1, but we should remain consistent and not add any fields.

pkg/cmd/deadman/operator_condition_challenger.go
deads2k: > Why not use library-go https://github.com/openshift/library-go/tree/master/pkg/controller/factory ?

Doesn't allow customization of resync interval and once you add enough, "let me tweak this little thing", you've produced a new language.

pkg/cmd/deadman/operator_staleness_check.go
deads2k: > Do we want to include the last reason here as well?

Sure, I can add that.

pkg/cmd/deadman/operator_staleness_check.go
deads2k: > This is perhaps event worthy? It will help us write synthetic tests :-)

Yeah, it probably is.  I'll add it.

pkg/cmd/deadman/operator_condition_challenger.go
mfojtik: I'm not sure, but if you set Managed=False, we go upgradeable=False... In the past, people used managed=false to disable operators completely (for whatever reason). I'm asking if this should be reflected in the status/message (iow. "checking for stale status" vs. "operator unmanaged, status is stale, waiting for it to come back")

pkg/cmd/deadman/operator_condition_challenger.go
mfojtik: 👍

pkg/cmd/deadman/operator_condition_challenger.go
mfojtik: It does allow that. I only asked because all controllers in this operator use that factory, but writing these two from scratch is fine (just wanted to know the reason ;-)

pkg/cmd/deadman/operator_condition_challenger.go
deads2k: the data shouldn't be stale because the operator should still be setting *something* indicating that it isn't doing work.

deads2k:
comments addressed.  Also add logging and event emission to be sure its actually running.

deads2k:
now with unit tests. It appears to function fine in the e2e

bindata/assets/kube-apiserver/pod.yaml
```diff
@@ -179,6 +179,29 @@ spec:
...
+      - name: POD_NAMESPACE
+        valueFrom:
+          fieldRef:
+            fieldPath: metadata.namespace
+    image: {{.OperatorImage}}
+    imagePullPolicy: IfNotPresent
+    terminationMessagePolicy: FallbackToLogsOnError
+    command: ["cluster-kube-apiserver-operator", "deadman"]
+    args:
+      - --kubeconfig=/etc/kubernetes/static-pod-resources/configmaps/kube-apiserver-cert-syncer-kubeconfig/kubeconfig
```
mfojtik, line 193: @deads2k, maybe a TODO for the future to give this its cert? :-)

bindata/assets/kube-apiserver/pod.yaml
```diff
@@ -179,6 +179,29 @@ spec:
...
+      - --namespace=$(POD_NAMESPACE)
+      - -v=2
+    resources:
+      requests:
+        memory: 50Mi
+        cpu: 5m
+    volumeMounts:
+      - mountPath: /etc/kubernetes/static-pod-resources
+        name: resource-dir
+      - mountPath: /var/log/kube-apiserver
```
mfojtik, line 203: do we mount/write here?

pkg/cmd/deadman/deadman.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+type Options struct {
+	controllerContext *controllercmd.ControllerContext
+
+	challengeInterval             time.Duration
+	controllerResponseGracePeriod time.Duration
+}
+
+func NewDeadmanCommand(ctx context.Context) *cobra.Command {
```
mfojtik, line 26: nit: can you please add some godoc here explaining why we added this and some example cases when then helps? (for future generations)

mfojtik:
A couple of nits, but otherwise, this LGTM to me.

/cc @tkashem 
(to have non-manager reviewing this :-)

/lgtm
/hold

(unhold if you don't want to wait for enhancement)

tkashem:
> (unhold if you don't want to wait for enhancement)

we have some questions around the current approach as described in the enhancement, let's resolve those first.

/hold

pkg/cmd/deadman/operator_condition_challenger.go
```diff
@@ -0,0 +1,185 @@
...
+	// only modify the message to avoid changing machine-readable reasons or status in the common case when operators
+	// are properly managing their status.
+	clusterOperatorToWriteAsStale := clusterOperator.DeepCopy()
+	for i, condition := range clusterOperator.Status.Conditions {
+		clusterOperatorToWriteAsStale.Status.Conditions[i].Message = "Checking for stale status, the active operator will reset this message: " + condition.Message
+	}
+
+	klog.V(2).Infof("Challenging clusteroperator/%v for status write", clusterOperator.Name)
+	// don't use apply here because we want to conflict if someone else has written status.
+	if _, err := c.clusterOperatorClient.Update(ctx, clusterOperatorToWriteAsStale, metav1.UpdateOptions{}); err != nil {
```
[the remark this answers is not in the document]
2uasimojo: ...except when I tried it on my cluster (4.12.3), I had to `oc edit` with `--subresource=status` to make it work.

deads2k:
operators lacking active reconciliation
1. monitoring - slow
2. machine-api - missing
3. openshift-samples - missing at least on GCP
4. cluster-autoscaler - missing on gcp
5. cloud-credential - slow?
6.

pkg/cmd/deadman/operator_condition_challenger.go
```diff
@@ -0,0 +1,191 @@
...
+	if now.Before(nextLivenessCheckTime) {
+		return nil
+	}
+
+	// if we're past the deadline since our last check, so we will perform a staleness check by resetting the message.
+	// only modify the message to avoid changing machine-readable reasons or status in the common case when operators
+	// are properly managing their status.
+	clusterOperatorToWriteAsStale := clusterOperator.DeepCopy()
+	for i, condition := range clusterOperator.Status.Conditions {
+		clusterOperatorToWriteAsStale.Status.Conditions[i].Message = "Checking for stale status, the active operator will reset this message: " + condition.Message
```
sanchezl, line 139: I think this reads as if "this message" refers to the text after the colon only, not the entire message itself.

sanchezl (commented):
I would propose the following text to make it clearer for someone running into these messages for the first time:

**Challenge:**
Requesting that the operator update this condition.  _\<original message here\>_

**Stale:**
Condition has not been updated by the operator in at least 1m0s. Last reason was: _blah, blah, blah..._

pkg/cmd/deadman/deadman.go
```diff
@@ -0,0 +1,93 @@
...
+	o := &Options{
+		challengeInterval:             1 * time.Minute,
+		controllerResponseGracePeriod: 1 * time.Minute,
```
mfojtik, line 29: @deads2k can you try to make this 11 minutes? it look like some operator have much longer default resync intervals that they should (samples-operator has 10 minutes resync)....

alternative is to make them reconcile immediately after change to their clusteroperator, but in some case that would require understanding how different teams write their operators :-)

pkg/cmd/deadman/operator_condition_challenger.go
```diff
@@ -0,0 +1,195 @@
...
+)
+
+const (
+	challengePrefix = "Requesting that the operator update this condition. "
+)
+
+type OperatorConditionChallenger struct {
+	// operatorToTimeToCheckForStaleness is a map from operator name to the time at which the condition messages will be
+	// reset to a "checking for activity" message.
+	operatorToTimeToCheckForStaleness map[string]time.Time
```
sanchezl, line 31: You will need to synchronize access to this map (and the similar one in OperatorStalenessChecker) if you intend to have more than one worker.

sanchezl:
@deads2k this needs an update to vendor and/or rebase

deads2k:
> @deads2k this needs an update to vendor and/or rebase

did both.

PR #1427, opened 2023-01-05
remove use of deprecated klog flags

sanchezl:
Previously deprecated klog flags will be removed completely in kube 1.26.
Logs of bootstrap pods were previously using the --log-file flag, will be available in the cri-o container logs instead.

### DIFF
--- a/bindata/bootkube/bootstrap-manifests/kube-apiserver-pod.yaml
+++ b/bindata/bootkube/bootstrap-manifests/kube-apiserver-pod.yaml
@@ -45,7 +45,6 @@ spec:
       hyperkube kube-apiserver
       --openshift-config=/etc/kubernetes/config/{{ .ConfigFileName }}
       --v=2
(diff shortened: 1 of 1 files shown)


soltysh (approved):
/lgtm

deads2k:
/lgtm cancel
/hold

This change preserves the file, but now produces the potential for creating an enormous file that will never be rotated and never be replaced.  This shouldn't merge in the current fashion.

Also, it didn't build on the revert, so it now conflicts.  The revert will *always* merge, fixes built on top will not have conflict failures like this.

deads2k:
bootkube isn't used in upgrade

/override ci/prow/e2e-aws-ovn-upgrade

omertuc:
https://issues.redhat.com/browse/OCPQE-13723

I think a library-go bump to github.com/openshift/library-go/commit/86dfb3653d218ba7ebc3fdff8e02cd6c011e2625 is needed here

PR #1430, opened 2023-01-17
increase audit log size to contain an entire upgrade+e2e run

deads2k:
We need this for debugging.


dgrisonnet (approved):
Sounds reasonable to me.

/lgtm

PR #1431, opened 2023-01-18
API-1520: Update SLO alerts based on upstream improvements

dgrisonnet:
Update the SLO alerts to include some improvements from upstream and revert some changes that were made downstream which broke the alerts.

Improvements:
- use new `apiserver_request_slo_duration_seconds` metric which excludes webhook latency and errors 
- exclude `attach` and `portforward` subresources from latency calculation since they are long-running requests
- revert changes from https://github.com/openshift/cluster-kube-apiserver-operator/commit/622c08f101555be4584cb897f68f772777b32ada since this is not a typo, it is meant to catch metrics for which requests don't have a scope and add them to the resource-scoped SLO. Not sure if this is still relevant but I'd rather not diverge from upstream for something so minor that doesn't have any impact
- revert expression related changes from https://github.com/openshift/cluster-kube-apiserver-operator/commit/db7cab78817d8c780d6691eea2df5ea408f0772e as the split that was made broke the way SLO is being computed. Instead of taking into account both errors and latency, it created one SLO for errors and one for latency.

### DIFF
--- a/bindata/assets/alerts/kube-apiserver-slos-basic.yaml
+++ b/bindata/assets/alerts/kube-apiserver-slos-basic.yaml
@@ -40,138 +40,110 @@ spec:
   - name: kube-apiserver.rules
     rules:
     - expr: |
-        # error
-        label_replace(
-          sum(rate(apiserver_request_total{job="apiserver",verb=~"LIST|GET",code=~"5.."}[5m]))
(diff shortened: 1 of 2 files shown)


dgrisonnet:
Test failures are unrelated.

/retest-required

PR #1451, opened 2023-03-02
OCPBUGS-8235: Guard pod set readiness probe endpoint explicitly

ingvagabund:
Backporting https://github.com/openshift/cluster-kube-apiserver-operator/pull/1448


soltysh (approved):
/approve
/label backport-risk-assessed

PR #1452, opened 2023-03-02
DO NOT MERGE: Check current encryption CI state

dgrisonnet:
/hold


dgrisonnet:
This is not needed anymore.

PR #1453, opened 2023-03-06
OCPBUGS-8477: e2e: disable issuer revocation test

mfojtik:
The service account issuer mechanism has a special revocation procedure. When the issuer is set to an empty string, the service account token issuer is **immediately** revoked, and existing tokens with the previous issuer are invalid.
This is to support a case when the cluster is compromised. Unfortunately, when we try to do this in an e2e test, the immediate revocation causes some components to use tokens with invalid issuers and therefore, the tokens fail authentication. This manifests as an "Unauthorized" error.

### DIFF
--- a/test/e2e/serviceaccountissuer_test.go
+++ b/test/e2e/serviceaccountissuer_test.go
@@ -28,6 +28,7 @@ func TestServiceAccountIssuer(t *testing.T) {
 	require.NoError(t, err)
 
 	t.Run("serviceaccountissuer set in authentication config results in apiserver config", func(t *testing.T) {
+		t.Skip()
 		setServiceAccountIssuer(t, authConfigClient, "https://first.foo.bar")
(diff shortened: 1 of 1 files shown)


mfojtik:
note: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1453/files#diff-597e6d73559f42ea48030e5e97565dd0745c09fd68363b97d33aac6f18384987R39 should make sure the original issuer is still trusted and the cluster should keep working normally.

sanchezl (commented):
`t.Skip()` ?

dgrisonnet:
For the record I opened https://issues.redhat.com/browse/API-1550 to re-enable the test in the future and https://issues.redhat.com/browse/API-1549 for the backport.

mfojtik:
https://github.com/openshift/cluster-kube-apiserver-operator/pull/1454 backport

dgrisonnet:
/retitle OCPBUGS-8477: e2e: disable issuer revocation test

stlaz:
/close
not it

PR #1455, opened 2023-03-06
OCPBUGS-8478: Disable TestBoundTokenSignerController

dgrisonnet:
We need to disable this test for now because it puts the CI cluster in an unstable state.

https://issues.redhat.com/browse/OCPBUGS-8529 is a blocker bug against OpenShift 4.13 that tracks the effort of fixing this behavior.


dgrisonnet:
`e2e-gcp-operator` failure is unrelated to what we have seen before:
```
 ERRO[2023-03-07T14:24:30Z] 
  * could not run steps: step e2e-gcp-operator failed: status 502 Bad Gateway, statusCode 502 releasing us-central1--gcp-quota-slice-02 
```

PR #1459, opened 2023-03-08
[wip] Read featuregates from cluster state 

deads2k:
For https://github.com/openshift/enhancements/pull/1373

Part 4 of a multi-step plan to make individual featuregates observable via the API. This was previously avoided because when featuregates get promoted from TechPreview to Default, during an upgrade, an old operator may try to enable a TechPreview variant of a feature.

This can be addressed by keying the featuregates by version. Let's see how ugly that will be in practice now that we have a decent build-out of library-go.

Unit tests are still owed. Proof is still owed. cluster-config-operator controller is still owed.

/hold

This is proving the concept, not to be merged until it shows as working end-to-end.

Other steps include

1. need to render out a version of FeatureGates that includes status for the featuregates to use during bootstrapping and plumb that through the installer.
2. need to read a FeatureGates manifest from disk instead of the single arg
3. enhancements write up describing how it works end to end.

fyi @JoelSpeed


JoelSpeed (commented):
No initial feedback on the first pass, will review again when this is ready for an e2e proof

PR #1463, opened 2023-03-15
[4.13] OCPBUGS-8710: turn PSa to logging mode by default again

stlaz:
/cc @deads2k @ibihim


stlaz:
/hold
contains fake bump

PR #1475, opened 2023-03-28
DO NOT MERGE: Re-enable bound token signer tests

dgrisonnet:
/hold


dgrisonnet:
/remove-lifecycle rotten
/reopen

I need to look into that again.

PR #1481, opened 2023-04-20
OCPBUGS-10831: pod security: use v1 api

stlaz:
/assign @openshift/openshift-team-auth


liouk:
> Are there any changes from v1beta to v1?

According to [this blog post](https://kubernetes.io/blog/2022/08/25/pod-security-admission-stable/#what-s-new-since-beta), the changes since beta are improved violation messages, improved namespace warnings, and some changes in the pod security standards pod OS field which aren't in use here as far as I can tell.

PR #1482, opened 2023-04-20
OCPBUGS-8404: pkg/operator/configobserver: check that the serving certificate refer…

vrutkovs:
…ence exists before applying

Ensure that .spec.servingCerts.namedCertificates[*].servingCertificate.Name exists before applying. This prevents a typo or a missing/inaccessible certificate from being referenced in kube-apiserver pod and prevents crashlooping and recovery operation.

Tested this manually and now when invalid secretref is set, no rollouts occur and the operator sets Degraded message "ConfigObservationDegraded: secret "foobar" not found", but Degraded=False


dgrisonnet (approved):
/lgtm
/approve

PR #1484, opened 2023-04-25
OCPBUGS-3986: dashboard: use recording rules for most metrics

vrutkovs:
Add more recording rules to reduce the load on Thanos querier and Prometheus. Adding more recording rules allows Prometheus to store metrics aggregation up front and request less information every time a dashboard is reloaded.

This removes "auto" interval as it can't be cached via recording rules.

Verified by deploying this as an additional dashboard and comparing the result - graphs are almost identical.

### DIFF
--- a/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
+++ b/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
@@ -138,3 +138,121 @@ spec:
     - record: cluster:apiserver_current_inflight_requests:sum:max_over_time:2m
       expr: |
         max_over_time(sum(apiserver_current_inflight_requests{apiserver=~"openshift-apiserver|kube-apiserver"}) by (apiserver,request_kind)[2m:])
+  - name: api-performance
+    rules:
+    - record: resource_verb:apiserver_request_duration_seconds_bucket:rate:1m
+      expr: sum(rate(apiserver_request_duration_seconds_bucket{apiserver=~"openshift-apiserver|kube-apiserver",subresource!~"proxy|attach|log|exec|portforward",verb!~"WATCH|WATCHLIST|PROXY"}[1m])) by (apiserver, resource, verb, le)
+    - record: resource_verb:apiserver_request_duration_seconds_bucket:rate:5m
+      expr: sum(rate(apiserver_request_duration_seconds_bucket{apiserver=~"openshift-apiserver|kube-apiserver",subresource!~"proxy|attach|log|exec|portforward",verb!~"WATCH|WATCHLIST|PROXY"}[5m])) by (apiserver, resource, verb, le)
+    - record: list:apiserver_request_duration_seconds_bucket:rate1m
+      expr: sum(rate(apiserver_request_duration_seconds_bucket{apiserver=~"openshift-apiserver|kube-apiserver",verb=~"LIST|GET"}[1m])) by (apiserver, le)
+    - record: list:apiserver_request_duration_seconds_bucket:rate5m
+      expr: sum(rate(apiserver_request_duration_seconds_bucket{apiserver=~"openshift-apiserver|kube-apiserver",verb=~"LIST|GET"}[5m])) by (apiserver, le)
+    - record: write:apiserver_request_duration_seconds_bucket:rate1m
+      expr: sum(rate(apiserver_request_duration_seconds_bucket{apiserver=~"openshift-apiserver|kube-apiserver",verb=~"POST|PUT|PATCH|UPDATE|DELETE"}[1m])) by (apiserver, le)
+    - record: write:apiserver_request_duration_seconds_bucket:rate5m
+      expr: sum(rate(apiserver_request_duration_seconds_bucket{apiserver=~"openshift-apiserver|kube-apiserver",verb=~"POST|PUT|PATCH|UPDATE|DELETE"}[5m])) by (apiserver, le)
+    - record: verb:apiserver_request_duration_seconds_bucket:rate1m
+      expr: sum(rate(apiserver_request_duration_seconds_bucket{apiserver=~"openshift-apiserver|kube-apiserver",subresource!~"proxy|attach|log|exec|portforward",verb!~"WATCH|WATCHLIST|PROXY"}[1m])) by (apiserver, verb, le)
+    - record: verb:apiserver_request_duration_seconds_bucket:rate5m
+      expr: sum(rate(apiserver_request_duration_seconds_bucket{apiserver=~"openshift-apiserver|kube-apiserver",subresource!~"proxy|attach|log|exec|portforward",verb!~"WATCH|WATCHLIST|PROXY"}[5m])) by (apiserver, verb, le)
+    - record: operation:etcd_request_duration_seconds_bucket:rate1m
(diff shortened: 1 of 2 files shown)


dinhxuanvu (commented):
/lgtm

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -138,3 +138,121 @@ spec:
     - record: cluster:apiserver_current_inflight_requests:sum:max_over_time:2m
       expr: |
         max_over_time(sum(apiserver_current_inflight_requests{apiserver=~"openshift-apiserver|kube-apiserver"}) by (apiserver,request_kind)[2m:])
+  - name: api-performance
+    rules:
+    - record: verb_resource:apiserver_request_duration_seconds_bucket:rate:1m
+      expr: sum(rate(apiserver_request_duration_seconds_bucket{apiserver=~"openshift-apiserver|kube-apiserver",subresource!="log",verb!~"WATCH|WATCHLIST|PROXY"}[1m])) by (apiserver, resource, verb, le)
```
dgrisonnet, line 144: notes for the future:
```
apiserver=~"openshift-apiserver|kube-apiserver"
```
we should also most likely include the oauth-apiserver to this dashboard.

```
subresource!="log"
```
this needs to be updated to `subresource!~"proxy|attach|log|exec|portforward"` based on what we recently changed in the SLO alerts

we might also want to use `apiserver_request_sli_duration_seconds_bucket` that excludes pnf & webhooks latency  instead of the e2e latency metric

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -138,3 +138,121 @@ spec:
...
+    - record: verb_resource:apiserver_request_duration_seconds_bucket:rate:5m
```
dgrisonnet, line 145: these are the only recording rules using the `verb_resource` prefix instead of `resource_verb`. It would be great to make that consistent

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
vrutkovs: >we should also most likely include the oauth-apiserver to this dashboard.

Neither of the mentioned metrics has oauth-apiserver in `apiserver` label

vrutkovs: >>subresource!="log"

>this needs to be updated to subresource!~"proxy|attach|log|exec|portforward" based on what we recently changed in the SLO alerts

good idea, done

vrutkovs: >we might also want to use apiserver_request_sli_duration_seconds_bucket that excludes pnf & webhooks latency instead of the e2e latency metric

I think its best being done in the new PR - to make this PR just a pure recording rules transition - also not sure how far can we backport this

manifests/0000_90_kube-apiserver-operator_05_api_performance_dashboard.yaml
```diff
@@ -81,7 +81,7 @@ data:
           "steppedLine": false,
           "targets": [
             {
-              "expr": "histogram_quantile(0.99, sum(rate(apiserver_request_duration_seconds_bucket{apiserver=\"$apiserver\",subresource!=\"log\",verb!~\"WATCH|WATCHLIST|PROXY\"}[$period])) by(verb,le))",
+              "expr": "histogram_quantile(0.99, sum(verb_resource:apiserver_request_duration_seconds_bucket:rate:$period{apiserver=\"$apiserver\"}) by (verb, le))",
```
dgrisonnet, line 84: we should reflect the name change here

manifests/0000_90_kube-apiserver-operator_05_api_performance_dashboard.yaml
```diff
@@ -297,7 +297,7 @@ data:
           "steppedLine": false,
           "targets": [
             {
-              "expr": "topk(20, histogram_quantile(0.99, sum(rate(apiserver_request_duration_seconds_bucket{apiserver=\"$apiserver\",subresource!=\"log\",verb!~\"WATCH|WATCHLIST|PROXY\"}[$period])) by(resource,verb,le)))",
+              "expr": "topk(20, histogram_quantile(0.99, verb_resource:apiserver_request_duration_seconds_bucket:rate:$period{apiserver=\"$apiserver\"}))",
```
dgrisonnet, line 300: we should reflect the name change here

manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
dgrisonnet: > Neither of the mentioned metrics has oauth-apiserver in apiserver label

maybe it has another value, but there should be metrics for the oauth-apiserver deployed by https://github.com/openshift/cluster-authentication-operator/ and we most likely also want to incude them in the dashboards in the future

dgrisonnet: > I think its best being done in the new PR - to make this PR just a pure recording rules transition - also not sure how far can we backport this

yeah I totally agree. My intention with the original comment was to take notes of things I noticed while reviewing to fix later

dgrisonnet: Looking at an OCP cluster, we indeed don't have any metrics for the oauth-apiserver, that's pretty odd, maybe we are just not exposing any metrics there

dgrisonnet: So we are not collection metrics for the openshift-apiserver at all despite the fact that it has a ServiceMonitor, but it is not picked up by Prometheus. I opened https://github.com/openshift/cluster-authentication-operator/pull/616 in an attempt to fix that.

manifests/0000_90_kube-apiserver-operator_05_api_performance_dashboard.yaml
vrutkovs: done

vrutkovs:
`[sig-network][Feature:EgressIP][apigroup:operator.openshift.io] [external-targets][apigroup:user.openshift.io][apigroup:security.openshift.io] pods should have the assigned EgressIPs and EgressIPs can be deleted and recreated [Skipped:azure][apigroup:route.openshift.io] [Serial] [Suite:openshift/conformance/serial]` is [permafailing](https://sippy.dptools.openshift.org/sippy-ng/tests/4.14/analysis?test=%5Bsig-network%5D%5BFeature%3AEgressIP%5D%5Bapigroup%3Aoperator.openshift.io%5D+%5Bexternal-targets%5D%5Bapigroup%3Auser.openshift.io%5D%5Bapigroup%3Asecurity.openshift.io%5D+pods+should+have+the+assigned+EgressIPs+and+EgressIPs+can+be+deleted+and+recreated+%5BSkipped%3Aazure%5D%5Bapigroup%3Aroute.openshift.io%5D+%5BSerial%5D+%5BSuite%3Aopenshift%2Fconformance%2Fserial%5D), doesn't seem related to this PR

PR #1487, opened 2023-04-28
Restricted featuregate reader

deads2k:
Update library-go to get a more restrictive featuregate reader to avoid naughty usage.


deads2k:
external ccm bug

deads2k:
/hold

in case https://github.com/openshift/cluster-kube-apiserver-operator/pull/1488 tests good first.

PR #1488, opened 2023-05-01
Read feature manifest

deads2k:
read input manifests to find FeatureGates


deads2k:
lgtm'd via slack. updating to test consumption.

PR #1491, opened 2023-05-08
OCPBUGS-13547: Remove featureset flag and use only the manifest

deads2k:
The manifest now exists in every install and can be used by bootstrapping.  The installer passes the arg now, so we can update the operator to only honor the manifest and eliminate interpretation.


deads2k:
last cleanup for command related to external CCM feature unrevert

dinhxuanvu (approved):
/lgtm

deads2k:
cleaning up remnants for external cloud provider

/retest

PR #1492, opened 2023-05-10
OCPBUGS-13303: pkg/operator/startupmonitor: skip openshift-apiserver readiness check…

vrutkovs:
… for kube-api's readyz

For SNO installations cluster-kube-apiserver-operator should consider kube-apiserver healthy even without openshift-apiserver available. This would allow operator to rollout sooner, speeding up the installation process.

The tradeoff is that users will not be able to login during install until apiserver/oauth servers are ready

### DIFF
--- a/pkg/operator/startupmonitorreadiness/readiness_checks.go
+++ b/pkg/operator/startupmonitorreadiness/readiness_checks.go
@@ -244,24 +244,28 @@ func checkRevision(kasPod *corev1.Pod, monitorRevision int) (bool, string, strin
 }
 
 // goodReadyzEndpoint performs HTTP checks against readyz?verbose=true endpoint
+// some checks may be excluded - see getExcludedChecks func
 //
 //	returns true, "", "", when we got HTTP 200 "successThreshold" times
 //	returns false, "NotReady", EntireResponseBody (if any) on HTTP != 200
 //	returns false, "NotReadyError", EntireResponseBody (if any) in case of any error or timeout
 func goodReadyzEndpoint(client *http.Client, rawURL string, successThreshold int, interval time.Duration) func(ctx context.Context) (bool, string, string) {
+	readyzURL := appendExcludedChecksToAddress(fmt.Sprintf("%s/readyz?verbose=true", rawURL))
 	return func(ctx context.Context) (bool, string, string) {
-		return doHTTPCheckAndTransform(ctx, client, fmt.Sprintf("%s/readyz?verbose=true", rawURL), "NotReady", doHTTPCheckMultipleTimes(successThreshold, interval))
+		return doHTTPCheckAndTransform(ctx, client, readyzURL, "NotReady", doHTTPCheckMultipleTimes(successThreshold, interval))
 	}
 }
 
 // goodHealthzEndpoint performs an HTTP check against healthz?verbose=true endpoint
+// some checks may be excluded - see getExcludedChecks func
 //
 //	returns true, "", "", on HTTP 200
 //	returns false, "Unhealthy", EntireResponseBody (if any) on HTTP != 200
 //	returns false, "UnhealthyError", EntireResponseBody (if any) in case of any error or timeout
 func goodHealthzEndpoint(client *http.Client, rawURL string) func(context.Context) (bool, string, string) {
+	healthzURL := appendExcludedChecksToAddress(fmt.Sprintf("%s/healthz?verbose=true", rawURL))
 	return func(ctx context.Context) (bool, string, string) {
-		return doHTTPCheckAndTransform(ctx, client, fmt.Sprintf("%s/healthz?verbose=true", rawURL), "Unhealthy", doHTTPCheck)
+		return doHTTPCheckAndTransform(ctx, client, healthzURL, "Unhealthy", doHTTPCheck)
 	}
 }
 
@@ -376,4 +380,27 @@ func filterByNodeName(kasPods []corev1.Pod, currentNodeName string) []corev1.Pod
 	}
 
 	return filteredKasPods
+}
+
+func getExcludedChecks() []string {
+	// Speed up SNO install by declaring kube-apiserver healthy before openshift-apiserver is ready
+	// so that CVO could proceed with other components sooner
+	return []string{
+		// Defined at https://github.com/openshift/kubernetes/blob/53f3066f47d3a594d6b93843916e8711ee673715/openshift-kube-apiserver/openshiftkubeapiserver/sdn_readyz_wait.go#L22
+		"api-openshift-apiserver-available",
+		// Defined at https://github.com/openshift/kubernetes/blob/53f3066f47d3a594d6b93843916e8711ee673715/openshift-kube-apiserver/openshiftkubeapiserver/sdn_readyz_wait.go#L26
+		"api-openshift-oauth-apiserver-available",
+		// Defined at https://github.com/openshift/kubernetes/blob/53f3066f47d3a594d6b93843916e8711ee673715/openshift-kube-apiserver/openshiftkubeapiserver/patch.go#L119
+		"poststarthook/openshift.io-openshift-apiserver-reachable",
+		// Defined at https://github.com/openshift/kubernetes/blob/53f3066f47d3a594d6b93843916e8711ee673715/openshift-kube-apiserver/openshiftkubeapiserver/patch.go#L123
+		"poststarthook/openshift.io-oauth-apiserver-reachable",
+	}
+}
+
+func appendExcludedChecksToAddress(rawURL string) string {
+	newURL := rawURL
+	for _, excludedCheck := range getExcludedChecks() {
+		newURL = fmt.Sprintf("%s&exclude=%s", newURL, excludedCheck)
+	}
+	return newURL
 }
--- a/pkg/operator/startupmonitorreadiness/readiness_checks_test.go
+++ b/pkg/operator/startupmonitorreadiness/readiness_checks_test.go
@@ -273,7 +273,6 @@ func TestGoodReadyzEndpoint(t *testing.T) {
 			name:    "scenario 1: happy path, HTTP 200, empty reason and msg",
 			healthy: true,
 		},
-
 		{
 			name: "scenario 2: HTTP 500, unhealthy reason and msg",
 			rspWriterFn: func(w http.ResponseWriter) {
@@ -330,6 +329,16 @@ func TestGoodReadyzEndpoint(t *testing.T) {
 			msg:       "waiting for kube-apiserver static pod to listen on port 6443",
 			customURL: "https://localhost:1234",
 		},
+		{
+			name: "scenario 6: unexpected err from the server",
+			rspWriterFn: func(w http.ResponseWriter) {
+				panic("bum")
+			},
+			healthy: false,
+			reason:  "NetworkError",
+			// we don't check the entire rsp from the server
+			msg: fmt.Sprintf("%s\": EOF", appendExcludedChecksToAddress("/readyz?verbose=true")),
+		},
 	}
 
 	for _, scenario := range scenarios {
@@ -345,7 +354,7 @@ func TestGoodReadyzEndpoint(t *testing.T) {
 					w.Write([]byte(fmt.Sprintf("a req received at unexpected path: %v", r.URL.Path)))
 					return
 				}
-				if r.URL.RawQuery != "verbose=true" {
+				if r.URL.RawQuery != appendExcludedChecksToAddress("verbose=true") {
 					w.WriteHeader(http.StatusBadRequest)
 					w.Write([]byte(fmt.Sprintf("unexpected query params received: %v", r.URL.RawQuery)))
 					return
@@ -405,7 +414,7 @@ func TestGoodHealthzEndpoint(t *testing.T) {
 			healthy: false,
 			reason:  "NetworkError",
 			// we don't check the entire rsp from the server
-			msg: "/healthz?verbose=true\": EOF",
+			msg: fmt.Sprintf("%s\": EOF", appendExcludedChecksToAddress("/healthz?verbose=true")),
 		},
 		{
 			name: "scenario 4: no rsp from the server",
@@ -438,7 +447,7 @@ func TestGoodHealthzEndpoint(t *testing.T) {
 					w.Write([]byte(fmt.Sprintf("a req received at unexpected path: %v", r.URL.Path)))
 					return
 				}
-				if r.URL.RawQuery != "verbose=true" {
+				if r.URL.RawQuery != appendExcludedChecksToAddress("verbose=true") {
 					w.WriteHeader(http.StatusBadRequest)
 					w.Write([]byte(fmt.Sprintf("unexpected query params received: %v", r.URL.RawQuery)))
 					return
@@ -458,7 +467,9 @@ func TestGoodHealthzEndpoint(t *testing.T) {
 			}
 
 			// act and validate
-			doCheckAndValidate(t, func() (bool, string, string) { return goodHealthzEndpoint(client, url)(context.TODO()) }, scenario.healthy, scenario.reason, scenario.msg)
+			doCheckAndValidate(t, func() (bool, string, string) {
+				return goodHealthzEndpoint(client, url)(context.TODO())
+			}, scenario.healthy, scenario.reason, scenario.msg)
 		})
 	}
 }


pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -95,6 +99,20 @@ func (ch *KubeAPIReadinessChecker) IsReady(ctx context.Context, revision int) (
 		return false, "", "", fmt.Errorf("a node name is required, use the SetNodeName method")
 	}
 
+	if ch.isSNO == nil {
```
p0lyn0mial, line 102: Isn't the `IsReady` method called only on the SNO platform ?

Worth checking, if it is we could simply remove the code that tries to determine the platform type.

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -248,9 +266,13 @@ func checkRevision(kasPod *corev1.Pod, monitorRevision int) (bool, string, strin
 //	returns true, "", "", when we got HTTP 200 "successThreshold" times
 //	returns false, "NotReady", EntireResponseBody (if any) on HTTP != 200
 //	returns false, "NotReadyError", EntireResponseBody (if any) in case of any error or timeout
-func goodReadyzEndpoint(client *http.Client, rawURL string, successThreshold int, interval time.Duration) func(ctx context.Context) (bool, string, string) {
+func goodReadyzEndpoint(client *http.Client, rawURL string, successThreshold int, interval time.Duration, isSNO *bool) func(ctx context.Context) (bool, string, string) {
+	readyzURL := fmt.Sprintf("%s/readyz?verbose=true", rawURL)
+	if *isSNO {
+		readyzURL = fmt.Sprintf("%s&exclude=api-openshift-apiserver-available", readyzURL)
```
p0lyn0mial, line 272: Is the behaviour of the `exclude` query param documented somewhere ?

p0lyn0mial: it looks like it is - https://kubernetes.io/docs/reference/using-api/health-checks/

p0lyn0mial: where is `api-openshift-apiserver-available` defined ? 
what will we do if the name ever changes ? do we care ?

p0lyn0mial: > where is api-openshift-apiserver-available defined ?

okay, it looks like I have found it here:
https://github.com/openshift/kubernetes/blob/53f3066f47d3a594d6b93843916e8711ee673715/openshift-kube-apiserver/openshiftkubeapiserver/sdn_readyz_wait.go#L21

p0lyn0mial: adding a test for checking the name of the check might be very expensive.

maybe at least we should add a comment (sdn_readyz_wait.go) saying that the name should be stable otherwise this check will fail again?

p0lyn0mial: it looks like the readyz EP prints a warning if the excluded check doesn't match - i hope we do log the output from the EP somewhere.

pkg/operator/startupmonitorreadiness/readiness_checks.go
vrutkovs: `IsReady` being called for all platforms, its part of library-go interface: https://github.com/openshift/library-go/blob/e749b54aff208942299351536cf2bf2e09df2dfb/pkg/operator/staticpod/startupmonitor/cmd.go#L24

(there is a different startup probe for static pods though - https://github.com/vrutkovs/cluster-kube-apiserver-operator/blob/6556d4c60747aad03920ffdffc38bc5ea3132220/pkg/operator/starter.go#L261-L282)

pkg/operator/startupmonitorreadiness/readiness_checks.go
vrutkovs: >maybe at least we should add a comment (sdn_readyz_wait.go) saying that the name should be stable otherwise this check will fail again?

iiuc its assumed that check names are persistent. Not sure if its part of the enhancement to establish that - I can dig it. In any case, the check won't fail if we specify an exclude which doesn't match anything

vrutkovs: >it looks like the readyz EP prints a warning if the excluded check doesn't match - i hope we do log the output from the EP somewhere.

kube-apiserver prints the list of checks which were run and passed/failed - along with the excluded ones (`[ ] openshift-apiserver - excluded`)

pkg/operator/startupmonitorreadiness/readiness_checks.go
p0lyn0mial: I think that `KubeAPIReadinessChecker` is used by the `startup-monitor` and it is run only on the SNO platform.

Check out https://github.com/openshift/cluster-kube-apiserver-operator/blob/ffb092644a35398a8b848522a96f60ac5a7297ef/cmd/cluster-kube-apiserver-operator/main.go#L60


Where is the code that runs this check on all platform?

pkg/operator/startupmonitorreadiness/readiness_checks.go
p0lyn0mial: > iiuc its assumed that check names are persistent. Not sure if its part of the enhancement to establish that 

I don't think there was an enhancement for the openshift readiness check, it was added ad-hoc.

I'm fine with either adding a test that verifies that the check we are interested in shows up on the list or with adding a comment with some instructions what to do in case the name of the check changes.

pkg/operator/startupmonitorreadiness/readiness_checks.go
vrutkovs: Oh, so in `main.go` we create startupmonitorreadiness, but start it for SNO only in starter.go? Okay, its safe to remove infra check in this PR then

p0lyn0mial: Yes. I think that the startup monitor runs only on the SNO platform and as of today it is sole consumer of `KubeAPIReadinessChecker`.

pkg/operator/startupmonitorreadiness/readiness_checks.go
vrutkovs: Looks like we don't have a unit/e2e test checking kube-apiserver returned status, so I added instructions on how to update the exclude list in the commit message

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -377,3 +379,17 @@ func filterByNodeName(kasPods []corev1.Pod, currentNodeName string) []corev1.Pod
...
+func getExcludedChecks() []string {
+	return []string{
+		"api-openshift-apiserver-available",
+	}
+}
+
+func appendExcludedChecksToAddress(rawURL string) string {
+	newURL := rawURL
+	for _, excludedCheck := range getExcludedChecks() {
+		newURL = fmt.Sprintf("%s&exclude=%s", newURL, excludedCheck)
```
p0lyn0mial, line 392: this won't work when `getExcludedChecks` returns more than one item, am I right?

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -377,3 +379,17 @@ func filterByNodeName(kasPods []corev1.Pod, currentNodeName string) []corev1.Pod
 
 	return filteredKasPods
 }
+
+func getExcludedChecks() []string {
+	return []string{
+		"api-openshift-apiserver-available",
```
p0lyn0mial, line 385: could you add a comment explaining why we exclude this check?

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -260,8 +261,9 @@ func goodReadyzEndpoint(client *http.Client, rawURL string, successThreshold int
 //	returns false, "Unhealthy", EntireResponseBody (if any) on HTTP != 200
 //	returns false, "UnhealthyError", EntireResponseBody (if any) in case of any error or timeout
 func goodHealthzEndpoint(client *http.Client, rawURL string) func(context.Context) (bool, string, string) {
+	healthzURL := appendExcludedChecksToAddress(fmt.Sprintf("%s/healthz?verbose=true", rawURL))
```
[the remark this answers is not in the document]
p0lyn0mial: the same applies to the `goodReadyzEndpoint`

pkg/operator/startupmonitorreadiness/readiness_checks_test.go
```diff
@@ -330,6 +329,16 @@ func TestGoodReadyzEndpoint(t *testing.T) {
 			msg:       "waiting for kube-apiserver static pod to listen on port 6443",
 			customURL: "https://localhost:1234",
 		},
+		{
+			name: "scenario 6: unexpected err from the server",
+			rspWriterFn: func(w http.ResponseWriter) {
+				panic("bum")
```
p0lyn0mial, line 335: mhm, won't this fail the test ?
I haven't found any recover routine. 
Why not to:
```
	rspWriterFn: func(w http.ResponseWriter) {
				w.WriteHeader(http.StatusInternalServerError)
				w.Write([]byte("kube is on fire"))
			},
``` 
?

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -377,3 +379,17 @@ func filterByNodeName(kasPods []corev1.Pod, currentNodeName string) []corev1.Pod
 
 	return filteredKasPods
 }
+
+func getExcludedChecks() []string {
```
p0lyn0mial, line 383: should we also exclude the poststarthook for the openshift readiness?

https://github.com/openshift/kubernetes/blob/53f3066f47d3a594d6b93843916e8711ee673715/openshift-kube-apiserver/openshiftkubeapiserver/patch.go

```
	genericConfig.AddPostStartHookOrDie("openshift.io-openshift-apiserver-reachable", func(context genericapiserver.PostStartHookContext) error {
		go openshiftAPIServiceReachabilityCheck.checkForConnection(context)
		return nil
	})
```

pkg/operator/startupmonitorreadiness/readiness_checks.go
vrutkovs: It would, as we keep appending stuff to `newURL`, not `rawURL`. 

Tested this in unit-tests:
```
unexpected message "unexpected query params received: verbose=true&exclude=api-openshift-apiserver-available&exclude=more-stuff", expected "/healthz?verbose=true&exclude=api-openshift-apiserver-available\": EOF"
```

pkg/operator/startupmonitorreadiness/readiness_checks.go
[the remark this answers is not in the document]
vrutkovs: I think we do, yeah - no reason to restart kube-apiserver during its lifetime if openshift-apiserver is not yet up, especially if we allowed to start it in that state

pkg/operator/startupmonitorreadiness/readiness_checks_test.go
vrutkovs: iiuc `setupServerClient` will panic, but the test continues and `msg` is still being checked:
```
=== RUN   TestGoodHealthzEndpoint/scenario_3:_unexpected_err_from_the_server
2023/05/11 16:17:05 http: panic serving 127.0.0.1:40822: bum
goroutine 216 [running]:
net/http.(*conn).serve.func1()
        /usr/lib/sdk/golang/src/net/http/server.go:1854 +0xbf
panic({0x1dec240, 0x2507d80})
        /usr/lib/sdk/golang/src/runtime/panic.go:890 +0x263
github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/startupmonitorreadiness.TestGoodHealthzEndpoint.func2({0xc0000a2aa8?, 0x4ac693?})
        /var/home/vrutkovs/src/github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/startupmonitorreadiness/readiness_checks_test.go:412 +0x27
github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/startupmonitorreadiness.TestGoodHealthzEndpoint.func4.2({0x252d8e0, 0xc00077c540}, 0xc000722a00)
        /var/home/vrutkovs/src/github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/startupmonitorreadiness/readiness_checks_test.go:455 +0x1b4
github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/startupmonitorreadiness.setupServerClient.func1({0x252d8e0?, 0xc00077c540?}, 0x1ef57a0?)
        /var/home/vrutkovs/src/github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/startupmonitorreadiness/readiness_checks_test.go:583 +0x27
net/http.HandlerFunc.ServeHTTP(0xc00073a5fd?, {0x252d8e0?, 0xc00077c540?}, 0x47048e?)
        /usr/lib/sdk/golang/src/net/http/server.go:2122 +0x2f
net/http.serverHandler.ServeHTTP({0xc00045dc80?}, {0x252d8e0, 0xc00077c540}, 0xc000722a00)
        /usr/lib/sdk/golang/src/net/http/server.go:2936 +0x316
net/http.(*conn).serve(0xc000537710, {0x252e800, 0xc00045db60})
        /usr/lib/sdk/golang/src/net/http/server.go:1995 +0x612
created by net/http.(*Server).Serve
        /usr/lib/sdk/golang/src/net/http/server.go:3089 +0x5ed
    /var/home/vrutkovs/src/github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/startupmonitorreadiness/readiness_checks_test.go:573: unexpected message "failed sending request to kube-apiserver: Get \"http://127.0.0.1:41369/healthz?verbose=true&exclude=api-openshift-apiserver-available\": EOF", expected "/healthz?verbose=true&exclude=-openshift-apiserver-available\": EOF"
```
is what I get when a wrong message is expected.

I copied this test case from `readyz` to have an additional check for request parameters

pkg/operator/startupmonitorreadiness/readiness_checks.go
vrutkovs: Yeah, we may want to exclude this and `oauth-server` too - let me add those too

pkg/operator/startupmonitorreadiness/readiness_checks_test.go
p0lyn0mial: mhm, if nothing recovers from a panic does it mean that not ALL tests are run :( ?!

compare: https://go.dev/play/p/4dnPJLOClis vs https://go.dev/play/p/LYaqICe9Z1P

pkg/operator/startupmonitorreadiness/readiness_checks.go
```diff
@@ -377,3 +381,22 @@ func filterByNodeName(kasPods []corev1.Pod, currentNodeName string) []corev1.Pod
...
+	return []string{
+		// Speed up SNO install by declaring kube-apiserver healthy before openshift-apiserver is ready
```
p0lyn0mial, line 387: should we also mention that this is SAFE because the location of the openshift-apiserver is hardcoded in the 
kube-apiserver and it is not part of the configuration/secret?

vrutkovs: I don't think its useful here - iiuc any bogus value in "except" will be ignored , but I don't mind adding it

pkg/operator/startupmonitorreadiness/readiness_checks_test.go
p0lyn0mial: ping, I think we should fix these 2 tests.

pkg/operator/startupmonitorreadiness/readiness_checks.go
p0lyn0mial: we could add a comment here and at the source (https://github.com/openshift/kubernetes/blob/53f3066f47d3a594d6b93843916e8711ee673715/openshift-kube-apiserver/openshiftkubeapiserver/sdn_readyz_wait.go)

pkg/operator/startupmonitorreadiness/readiness_checks_test.go
vrutkovs: Right, let me figure out what's happening here

vrutkovs: Ah, there's `recover` in `net/http`: https://github.com/golang/go/blob/master/src/net/http/server.go#L1863

pkg/operator/startupmonitorreadiness/readiness_checks.go
vrutkovs: Sure, I'll add the comment there once this merges

p0lyn0mial:
LGTM

@vrutkovs are you waiting for any specific signal from the CI or do you want to merge this PR ?

vrutkovs:
single-node jobs should be passing, I'm a bit concerned that e2e-gcp-operator-single-node is not passing. Lets run it again

/test e2e-gcp-operator-single-node

p0lyn0mial:
It looks like `e2e-aws-operator-disruptive-single-node` is permanently broken, it looks like an issue with the fallback mechanism provided by the operator. It is hard to debug because we don't have logs from the startup monitor - that should be our first step.

vrutkovs:
Thanks, seems `e2e-gcp-operator-single-node` was not updated after 1.27 bump as well, so no regressions in this PR and its ready to be lgtm'ed

vrutkovs:
Thanks, seems `e2e-gcp-operator-single-node` was not updated after 1.27 bump as well, so no regressions in this PR and its ready to be lgtm'ed

vrutkovs:
e2e-gcp-operator currently permafails - needs https://github.com/openshift/cluster-kube-apiserver-operator/pull/1497

vrutkovs:
e2e-gcp-operator currently permafails - needs https://github.com/openshift/cluster-kube-apiserver-operator/pull/1497

PR #1493, opened 2023-05-15
OCPBUGS-2765: dont log tokens

ibihim:
# What

Bump library-go as a follow up on https://github.com/openshift/library-go/pull/1500.

# Why

The library-go update doesn't do anything if not used.

FYI: @soltysh, @stlaz


ibihim:
Not necessary anymore as a more up-to-date version of library-go got merged.

PR #1494, opened 2023-05-15
update probe exclusions to speed up install and single-node reaction

deads2k:
the bootstrap kube-apiserver and the single-node kube-apiserver will never be able to reach the openshift-apiserver or oauth-apiserver, so the probe check for them simply delays startup until the timeout is reached.  This updates PR excludes the reachability checks where appropriate to gain 60s.


bindata/assets/kube-apiserver/pod.yaml
```diff
@@ -103,15 +103,31 @@ spec:
         scheme: HTTPS
         port: 6443
         path: livez
-      initialDelaySeconds: 45
-      timeoutSeconds: 10
+      initialDelaySeconds: 0
```
deads2k, line 106: added startup probe to avoid the need for an extra long delay here.

bindata/assets/kube-apiserver/pod.yaml
```diff
@@ -103,15 +103,31 @@ spec:
...
+      periodSeconds: 10
+      timeoutSeconds: 5
```
deads2k, line 108: reduced timeout because we should not take 10s to answer probes

bindata/assets/kube-apiserver/pod.yaml
```diff
@@ -103,15 +103,31 @@ spec:
...
+      successThreshold: 1
+      failureThreshold: 3
```
deads2k, line 110: added default values for Abu's previous request on other apiservers.

deads2k:
oh, forgot the install is currently broken overall.  will recheck tomorrow

deads2k:
/hold

there's a type mismatch and in the slack thread, it might be more useful to actually disable these entirely for single-node so we end up more similar to microshift.

PR #1500, opened 2023-05-25
OCPBUGS-14008: Enable "send-retry-after-while-not-ready-once" on SNO

vrutkovs:
This commit updates default SNO apiserver parameters to include "send-retry-after-while-not-ready-once" parameter


pkg/operator/configobservation/apiserver/observe_send_retry_after_while_not_ready_once.go
```diff
@@ -0,0 +1,76 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/configobservation"
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+)
+
+const sendRetryAfterWhileNotReadyOnceKey = "send-retry-after-while-not-ready-once"
```
tkashem, line 16: I guess it's not possible, but can we make it static through config overriding or something? It should always be enabled for SNO.

pkg/operator/configobservation/apiserver/observe_send_retry_after_while_not_ready_once.go
```diff
@@ -0,0 +1,76 @@
...
+
+// ObserveSendRetryAfterWhileNotReadyOnce ensures that send-retry-after-while-not-ready-once is set for SNO clusters.
+func ObserveSendRetryAfterWhileNotReadyOnce(genericListers configobserver.Listers, _ events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, errs []error) {
+	defer func() {
+		// Prune the observed config so that it only contains apiServerArguments field.
+		ret = configobserver.Pruned(ret, []string{"apiServerArguments"})
```
tkashem, line 22: can we lower the scope to `send-retry-after-while-not-ready-once` only, similar to  `ObserveShutdownDelayDuration`?

pkg/operator/configobservation/apiserver/observe_send_retry_after_while_not_ready_once.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// read the current value
+	expectedApiServerArguments := make(map[string]interface{}, 0)
+	currentApiServerArguments, _, err := unstructured.NestedMap(existingConfig, "apiServerArguments")
+	if err != nil {
+		errs = append(errs, fmt.Errorf("unable to extract apiserver arguments from the existing config: %v", err))
+		// without a full set of apiserver arguments we can't proceed
+		return existingConfig, append(errs, err)
+	}
+	// shallow copy current apiserver arguments to expected
```
tkashem, line 41: nit: so on any error, we are not continuing, can we 
```
	if err != nil {
		  if !apierrors.IsNotFound(err) {
                  }
                 errs = append(errs, fmt.Errorf("unable to extract apiserver arguments from the existing config: %v", err))
                return existingConfig, errs
	}
```

why is NotFound special, can't we exit when there is any error?

pkg/operator/configobservation/apiserver/observe_send_retry_after_while_not_ready_once.go
```diff
@@ -0,0 +1,76 @@
...
+	for k, v := range currentApiServerArguments {
+		expectedApiServerArguments[k] = v
+	}
+
+	_, found, err := unstructured.NestedStringSlice(currentApiServerArguments, sendRetryAfterWhileNotReadyOnceKey)
+	if err != nil {
+		errs = append(errs, fmt.Errorf("unable to extract apiserver arguments from the existing config: %v", err))
+		// without a full set of apiserver arguments we can't proceed
+		return existingConfig, append(errs, err)
```
tkashem, line 50: do we need to add the `err` twice?

pkg/operator/configobservation/apiserver/observe_send_retry_after_while_not_ready_once.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	if infra.Status.ControlPlaneTopology == configv1.SingleReplicaTopologyMode {
+		if err = unstructured.SetNestedStringSlice(expectedApiServerArguments, []string{"true"}, sendRetryAfterWhileNotReadyOnceKey); err != nil {
+			errs = append(errs, fmt.Errorf("unable to extract apiserver arguments from the existing config: %v", err))
+			// without a full set of apiserver arguments we can't proceed
+			return existingConfig, append(errs, err)
```
tkashem, line 57: same: do we need to add the err twice?

pkg/operator/configobservation/apiserver/observe_send_retry_after_while_not_ready_once.go
vrutkovs: iiuc this would cause an additional rollout for SNO, but I'm not certain

pkg/operator/configobservation/apiserver/observe_send_retry_after_while_not_ready_once.go
```diff
@@ -0,0 +1,58 @@
...
+		errs = append(errs, fmt.Errorf("unable to extract send retry after while not ready setting from the existing config: %v", err))
+		// keep going, we are only interested in the observed value which will overwrite the current configuration anyway
+	}
+	if len(currentSendRetryAfterWhileNotReadyOnceSlice) > 0 {
+		currentSendRetryAfterWhileNotReadyOnce = currentSendRetryAfterWhileNotReadyOnceSlice[0]
+	}
+
+	// see if the current and the observed value differ
+	observedConfig := map[string]interface{}{}
+	if !reflect.DeepEqual(observedSendRetryAfterWhileNotReadyOnce, currentSendRetryAfterWhileNotReadyOnce) {
```
deads2k, line 49: under what conditions is the int equal to the string?  Seems like you need `observedSendRetryAfterWhileNotReadyOnce` to be a string.

deads2k: I think you're missing the "don't touch this" test case that would have caught this.

vrutkovs: Ah, it was comparing `bool` and `string` (and converted to string when a change was required). Test case added

PR #1506, opened 2023-06-02
OCPBUGS-14010: increase timeout for probes

tkashem:


bindata/assets/kube-apiserver/pod.yaml
```diff
@@ -102,16 +102,32 @@ spec:
       httpGet:
         scheme: HTTPS
         port: 6443
-        path: livez
+        path: healthz
```
deads2k, line 105: I think this one should stay livez

bindata/assets/kube-apiserver/pod.yaml
```diff
@@ -102,16 +102,32 @@ spec:
...
+      periodSeconds: 10
       timeoutSeconds: 10
+      successThreshold: 1
+      failureThreshold: 3
     readinessProbe:
       httpGet:
         scheme: HTTPS
         port: 6443
         path: readyz
       initialDelaySeconds: 10
```
deads2k, line 116: I don't think you need this delay.  Do the others have it?

PR #1510, opened 2023-06-19
OCPBUGS-13946: do not use one second timeout when asserting a webhook connection

p0lyn0mial:
Previously the dial timeout to a webook was set to one second which seems to be very aggressive and could cause failures which could put the operator into degraded state.

This PR reads the timeout value for a webhook from the spec or uses a default value of 10 seconds if it wasn't specified

### DIFF
--- a/pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
+++ b/pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
@@ -22,6 +22,9 @@ type webhookInfo struct {
 	Service               *serviceReference
 	CABundle              []byte
 	FailurePolicyIsIgnore bool
+	// TimeoutSeconds specifies the timeout for a webhook.
+	// After the timeout passes, the webhook call will be ignored or the API call will fail
+	TimeoutSeconds *int32
 }
 
 // serviceReference generically represents a service reference
@@ -49,7 +52,7 @@ func (c *webhookSupportabilityController) updateWebhookConfigurationDegraded(ctx
 				serviceMsgs = append(serviceMsgs, msg)
 				continue
 			}
-			err = c.assertConnect(ctx, webhook.Name, webhook.Service, webhook.CABundle)
+			err = c.assertConnect(ctx, webhook.Name, webhook.Service, webhook.CABundle, webhook.TimeoutSeconds)
 			if err != nil {
 				msg := fmt.Sprintf("%s: %s", webhook.Name, err)
 				if webhook.FailurePolicyIsIgnore {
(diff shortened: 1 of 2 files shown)


pkg/operator/webhooksupportabilitycontroller/degraded_webhook_admission.go
```diff
@@ -27,6 +27,7 @@ func (c *webhookSupportabilityController) updateMutatingAdmissionWebhookConfigur
 				Name:                  webhook.Name,
 				CABundle:              webhook.ClientConfig.CABundle,
 				FailurePolicyIsIgnore: webhook.FailurePolicy != nil && *webhook.FailurePolicy == admissionregistrationv1.Ignore,
+				TimeoutSeconds:        webhook.TimeoutSeconds,
```
benluddy, line 30: Does it make sense to perform defaulting here rather than as part of every check?

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
```diff
@@ -114,7 +121,7 @@ func (c *webhookSupportabilityController) assertConnect(ctx context.Context, web
 		case <-time.After(time.Duration(i) * time.Second):
 		}
 		dialer := &tls.Dialer{
-			NetDialer: &net.Dialer{Timeout: 1 * time.Second},
+			NetDialer: &net.Dialer{Timeout: timeout},
```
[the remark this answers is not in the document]
p0lyn0mial: The timeout applies to connection and TLS handshake as a whole.
See https://github.com/golang/go/blob/master/src/crypto/tls/tls.go#L123

pkg/operator/webhooksupportabilitycontroller/degraded_webhook_admission.go
p0lyn0mial: The defaulting is cheap. In the future we might consider adding some logs.
In addition to that it is consistent with defaulting the `port`.

benluddy: Right, I'm not worried about the cost, I think it would be a tidier separation of responsibility. Since we already translate the API object into an internal representation for use by the dial probes, it didn't make sense to me that the internal representation (i.e. `webhookInfo`) wasn't directly usable by the dial probe.

It's only my preference and I'm satisfied with consistency too.

benluddy:
Is this a good time to add test coverage for dial timeouts?

p0lyn0mial:
> Is this a good time to add test coverage for dial timeouts?

It turns out that unit test for dial timeouts would complicated, we would need to start a web server, a fake dns server and we would need to measure duration of the test.

I was also thinking about refactoring the code just to validate if a timeout value is applied but i don’t see a huge gain here.

I would like to merge this PR without timeout coverage.

PR #1511, opened 2023-06-19
OCPBUGS-14940: api_performance_dashboard: show apiserver_longrunning_requests metric

p0lyn0mial:
It looks like the previous metric used in the dashboard (`apiserver_longrunning_gauge`) metric was depreciated in `1.23` (upstream) - https://github.com/kubernetes/kubernetes/issues/103441

The new metric is `apiserver_longrunning_requests`(used in this PR)  - https://github.com/kubernetes/kubernetes/pull/103799/files

The old metric was removed in https://github.com/kubernetes/kubernetes/commit/731397086b763b9a951a414d37018f7e4f03b99c
and at the same time the new metrics was promoted to "STABLE" level.

### DIFF
--- a/manifests/0000_90_kube-apiserver-operator_05_api_performance_dashboard.yaml
+++ b/manifests/0000_90_kube-apiserver-operator_05_api_performance_dashboard.yaml
@@ -1109,7 +1109,7 @@ data:
           "steppedLine": false,
           "targets": [
             {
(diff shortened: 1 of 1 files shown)


dgrisonnet (approved):
/lgtm
/approve

p0lyn0mial:
For the record: I think that the perf-dashboard is also used by the `openshift-apiserver`. So the fix will be automatically applied to `oas` as well.

PR #1512, opened 2023-06-19
OCPBUGS-14496: manifests: fix the scope of the TechPreviewNoUpgrade alert

dgrisonnet:
The TechPreviewNoUpgrade alert used to have an expression that was only excluding the default feature set meaning that all other feature set were reported as causing the cluster to be unupgradeable. However, that is not the case for the LatencySensitive feature set. As such, this change limit the list of feature set causing this alert to fire to:

  - TechPreviewNoUpgrade
  - CustomNoUpgrade

Also, since the original name of this alert wasn't generic enough compare to its scope, it was renamed to UnsupportedFeatureSetEnabled.

### DIFF
--- a/manifests/0000_90_kube-apiserver-operator_03_servicemonitor.yaml
+++ b/manifests/0000_90_kube-apiserver-operator_03_servicemonitor.yaml
@@ -51,7 +51,7 @@ spec:
           Cluster has enabled Technology Preview features that cannot be undone and will prevent upgrades.
           The TechPreviewNoUpgrade feature set is not recommended on production clusters.
       expr: |
-        cluster_feature_set{name!="", namespace="openshift-kube-apiserver-operator"} == 0
+        cluster_feature_set{name=~"TechPreviewNoUpgrade|CustomNoUpgrade", namespace="openshift-kube-apiserver-operator"} == 0
       for: 10m
       labels:
         severity: warning


dgrisonnet:
Holding since the name change we need to be reflected in origin if we decide to go with it.
/hold

/assign @p0lyn0mial 
/cc @wking

manifests/0000_90_kube-apiserver-operator_03_servicemonitor.yaml
```diff
@@ -44,14 +44,14 @@ spec:
...
     rules:
-    - alert: TechPreviewNoUpgrade
+    - alert: UnsupportedFeatureSetEnabled
       annotations:
-        summary: Cluster has enabled tech preview features that will prevent upgrades.
+        summary: Cluster has enabled unsupported features that will prevent upgrades.
         description: >-
-          Cluster has enabled Technology Preview features that cannot be undone and will prevent upgrades.
-          The TechPreviewNoUpgrade feature set is not recommended on production clusters.
+          Cluster has enabled a feature set that cannot be undone and will prevent upgrades.
```
wking, line 51: Can `CustomNoUpgrade` be unset?  I forget where the guard that prevents unsetting `TechPreviewNoUpgrade` lives, but maybe it needs to be updated to consider `CustomNoUpgrade` as well?

manifests/0000_90_kube-apiserver-operator_03_servicemonitor.yaml
```diff
@@ -44,14 +44,14 @@ spec:
   groups:
   - name: cluster-version
     rules:
-    - alert: TechPreviewNoUpgrade
+    - alert: UnsupportedFeatureSetEnabled
```
deads2k, line 47: TechPreview is reported.

dgrisonnet:
I reverted the name change back to `TechPreviewNoUpgrade` so we won't need to change origin anymore.

/unhold

manifests/0000_90_kube-apiserver-operator_03_servicemonitor.yaml
```diff
@@ -51,7 +51,7 @@ spec:
           Cluster has enabled Technology Preview features that cannot be undone and will prevent upgrades.
           The TechPreviewNoUpgrade feature set is not recommended on production clusters.
       expr: |
-        cluster_feature_set{name!="", namespace="openshift-kube-apiserver-operator"} == 0
+        cluster_feature_set{name=~"TechPreviewNoUpgrade|CustomNoUpgrade", namespace="openshift-kube-apiserver-operator"} == 0
```
p0lyn0mial, line 54: alright, so any value other than `TechPreviewNoUpgrade` or `CustomNoUpgrade` won't prevent upgrades, correct?

dgrisonnet: Correct, or at the very least it doesn't prevent upgrades because a tech preview feature was enabled. If there comes an alerting use case for a new feature set, we could always append it to the existing alert or create a brand new alert just for it.

p0lyn0mial: how will we be able to detect that a new feature has been added that blocks upgrades?

dgrisonnet: it will be hard for us the api team to know, but openshift/api approvers should be aware of that since they would need to approve changes to https://github.com/openshift/api/blob/fd420c26854b53f62bd704b9cb8952a82c281403/config/v1/types_feature.go#L35-L50.

p0lyn0mial: @dgrisonnet @deads2k @mfojtik Is the process of adding a new feature documented? Can we rely on it? Maybe we can simply write a basic e2e test that will fail when a new flag blocking upgrades is added? Based on that, we could detect the change and update our alert.

vrutkovs: No, this will only update alert status. Upgrades will still be blocked by operator's Uprgadable=False status

p0lyn0mial:
/lgtm
/hold

@dgrisonnet feel free to cancel the hold. I'm fine with merging this PR and adding a test in a follow-up PR.

dgrisonnet:
Unholding to fix the issue at hand.

/unhold

PR #1513, opened 2023-06-20
Add DeprecatedAPIInUse alert

dgrisonnet:
Add an alert that will warn users using deprecated APIs that don't have a precise removal date. This is currently the case for CRDs since there is no way to set a removal time frame in the object.

### DIFF
--- a/bindata/assets/alerts/api-usage.yaml
+++ b/bindata/assets/alerts/api-usage.yaml
@@ -43,3 +43,21 @@ spec:
           labels:
             namespace: openshift-kube-apiserver
             severity: info
+        - alert: DeprecatedAPIInUse
+          annotations:
+            summary: Deprecated API that might be removed in the next version is being used.
+            description: >-
+              Deprecated API that might be removed in the next version is being used. Removing the workload that is using
+              the {{ $labels.group }}.{{ $labels.version }}/{{ $labels.resource }} API might be necessary for
+              a successful upgrade to the next cluster version with Kubernetes {{ $labels.removed_release }}.
+              Refer to `oc get apirequestcounts {{ $labels.resource }}.{{ $labels.version }}.{{ $labels.group }} -o yaml` to identify the workload.
+          expr: >-
+            group by (group,version,resource,removed_release) (apiserver_requested_deprecated_apis{removed_release=""})
+            * on (group,version,resource) group_left ()
+            sum by (group,version,resource) (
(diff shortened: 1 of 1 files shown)


bindata/assets/alerts/api-usage.yaml
```diff
@@ -43,3 +43,21 @@ spec:
...
+        - alert: DeprecatedAPIInUse
+          annotations:
+            summary: Deprecated API that might be removed in the next version is being used.
+            description: >-
+              Deprecated API that might be removed in the next version is being used. Removing the workload that is using
+              the {{ $labels.group }}.{{ $labels.version }}/{{ $labels.resource }} API might be necessary for
+              a successful upgrade to the next cluster version with Kubernetes {{ $labels.removed_release }}.
+              Refer to `oc get apirequestcounts {{ $labels.resource }}.{{ $labels.version }}.{{ $labels.group }} -o yaml` to identify the workload.
+          expr: >-
+            group by (group,version,resource,removed_release) (apiserver_requested_deprecated_apis{removed_release=""})
```
dgrisonnet, line 55: This is intentionally set to empty string to catch APIs that don't have a precise removal date which align with was is done for CRDs in k/k: https://github.com/kubernetes/kubernetes/blob/2e93c65eff5979e21178d9d7117916e49af29838/staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/customresource_handler.go#L355

p0lyn0mial: I'm not convinced if we should alert on APIs that don't have `removed_release` field set. I think this especially applies to CRDs. By default kube doesn't have any CRDs installed. In general CRDs are used as an extension mechanism and are not tied to any specific release. Am I right?

If the above is true I don't think we can change anything upstream.

We could prepare an improvement for downstream. 
Then we should consider if it would be a feature of OCP or just a mechanism used by internal components.

dgrisonnet: The current idea for the `removed_release` field is to have a patch in o/k that would populate it from an annotation in the CR. It would then be up to the operator to set it if needed, but for those who don't comply with our guidelines, it would still be worth having an alert catching the use of deprecated CRs even though we don't know when they will be removed if ever

dgrisonnet:
/remove-lifecycle stale

I need to write a Jira ticket to get the CR deprecation effort prioritized.

PR #1516, opened 2023-06-21
update probes for best practices and consistency

deads2k:
updates probes to match other apiservers and use startup probes which didn't exist when we started.

/assign @vrutkovs


deads2k:
spoke via slack. He wanted to see payload pass.  they're doing fine.   merging.

deads2k:
all required jobs passed.

PR #1521, opened 2023-06-28
OCPBUGS-15489: manifests: add new PrometheusRule for recording rules

vrutkovs:
In https://github.com/openshift/cluster-kube-apiserver-operator/pull/1484 recording rules were added to the PrometheusRule which was supposed to be removed by CVO (it has `release.openshift.io/delete: "true"` annotation). This PR moves recording rules to a new PrometheusRule - and fixes missing datapoints on apiservers dashboard.

After https://github.com/openshift/cluster-authentication-operator/pull/616 was merged openshift-oauth-apiserver is available on the dashboard, so this PR updates recording rules to include it as well.

This also removes unnecessary v2 suffix in dashboard name

### DIFF
--- a/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
+++ b/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
@@ -153,118 +153,118 @@ spec:
   - name: api-performance
     rules:
     - record: resource_verb:apiserver_request_duration_seconds_bucket:rate:1m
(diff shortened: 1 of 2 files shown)


p0lyn0mial:
> In https://github.com/openshift/cluster-kube-apiserver-operator/pull/1484 recording rules were added to the PrometheusRule which was supposed to be removed by CVO (it has release.openshift.io/delete: "true" annotation)

does it mean that at some point we can remove the old recording rule?

vrutkovs:
> does it mean that at some point we can remove the old recording rule?

Yes - https://issues.redhat.com/browse/OCPBUGS-15504, although I'm not sure if it can happen in 4.14 or needs to be deferred to 4.15

PR #1523, opened 2023-07-07
certrotation: rotate kube-apiserver-to-kubelet-signer when 80% of validity is over

p0lyn0mial:
we haven’t been paged because the `recovery-controller` will create a new certificate (when the old expires) and because the cert is actually auto-reloaded.

### DIFF
--- a/pkg/operator/certrotationcontroller/certrotationcontroller.go
+++ b/pkg/operator/certrotationcontroller/certrotationcontroller.go
@@ -171,10 +171,12 @@ func newCertRotationController(
 	certRotator = certrotation.NewCertRotationController(
 		"KubeAPIServerToKubeletClientCert",
 		certrotation.RotatedSigningCASecret{
-			Namespace:              operatorclient.OperatorNamespace,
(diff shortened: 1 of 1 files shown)


pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -174,7 +174,7 @@ func newCertRotationController(
 			Namespace:              operatorclient.OperatorNamespace,
 			Name:                   "kube-apiserver-to-kubelet-signer",
 			Validity:               1 * 365 * defaultRotationDay, // this comes from the installer
-			Refresh:                8 * 365 * defaultRotationDay, // this means we effectively do not rotate
+			Refresh:                292 * defaultRotationDay,
```
s-urbaniak, line 177: nit: small godoc that this is 80% of the validity range and that this invariant should be maintained

p0lyn0mial: pushed, ptal.

PR #1526, opened 2023-07-19
OCPBUGS-13635: make webhook connection failure a warning in log

mfojtik:
Since these logs are only for debugging and are informational, they should not be errors but just warnings.


dgrisonnet (approved):
/lgtm
/approve

PR #1527, opened 2023-07-20
OCPBUGS-16511: remove dependency on typed prometheus client

sanchezl:
- remove dependency on typed prometheus client
- go mod vendor


mfojtik:
👍 

/lgtm
/approve

PR #1531, opened 2023-07-27
exempt PSa from namespaces ignored by the SCCs

stlaz:
/assign @deads2k 
/assign @ibihim

### DIFF
--- a/bindata/assets/config/defaultconfig.yaml
+++ b/bindata/assets/config/defaultconfig.yaml
@@ -29,6 +29,9 @@ admission:
           # This configuration should never be exposed to cluster users as no
           # such guarantees are made for any other OpenShift SA/user.
           - system:serviceaccount:openshift-infra:build-controller
+          # the following namespaces are considered privileged in terms of SCCs, which
+          # are disabled there, too
+          namespaces: ["default", "kube-system", "kube-public"]
 apiServerArguments:
   allow-privileged:
     - "true"


bindata/assets/config/defaultconfig.yaml
```diff
@@ -29,6 +29,9 @@ admission:
           # This configuration should never be exposed to cluster users as no
           # such guarantees are made for any other OpenShift SA/user.
           - system:serviceaccount:openshift-infra:build-controller
+          # the following namespaces are considered privileged in terms of SCCs, which
+          # are disabled there, too
+          namespaces: ["default", "kube-system", "kube-public"]
```
deads2k, line 34: doing this will expose us to failing audits where pods don't conform to the actual permissions.  If instead a cluster-policy-controller sets the labels to privileged, we will 
1. have the same impact on pod create
2. pass audits
3. clearly expose the actual privilege level

stlaz: Actually, the global config should be the first place to look at for any such tool, as you need to know what the global policies are if you're to assess what the PSa permissions of a pod are. Not every namespace has to be labelled.

I would expect an audit to be interested in:
- what is the global restriction level? Why is it not restricted?
- are there any entities exempted from PSa admission? Why?
- are there any namespaces specifically set to privileged? Why?

deads2k: a cross vendor audit tool can only see the API. There is no option for "check the global config" because you cannot see it.

stlaz:
/close
this is now done in the cluster-policy-controller

PR #1537, opened 2023-08-14
API-1618: Add a new controller which automatically approves CSRs for master nodes

vrutkovs:
The controller would check CSRs and approve those which are issued by master nodes to be renewed. This would make sure that the cluster can recover after being shutdown for ~year, as cluster-kube-apiserver-operator can't proceed after kubelet certs have expired and there is no way user can approve CSRs as localhost and localhost-recovery.kubeconfig on the nodes have also expired.

CI job results:
* pass on all SNO tests
* HA suspend: 
```
monitoring                                 4.15.0-0.ci.test-2023-09-08-145040-ci-op-hjdgvlgb-latest   False       True          True       24m     reconciling Telemeter client Deployment failed: updating Deployment object failed: waiting for DeploymentRollout of openshift-monitoring/telemeter-client: context deadline exceeded
```
* HA shutdown:
kubelet won't start since `resolve-prepender` cannot be pulled:
```
Mar 06 16:40:38 test-infra-cluster-59284a39-master-0 resolv-prepender.sh[9525]: Trying to pull registry.build02.ci.openshift.org/ci-op-hjdgvlgb/stable@sha256:3ccc541f6644e7a4027380ac3f330cbae65277440b960c28e018a78ce74a7c72...
Mar 06 16:40:38 test-infra-cluster-59284a39-master-0 resolv-prepender.sh[9525]: Error: initializing source docker://registry.build02.ci.openshift.org/ci-op-hjdgvlgb/stable@sha256:3ccc541f6644e7a4027380ac3f330cbae65277440b960c28e018a78ce74a7c72: pinging container registry registry.build02.ci.openshift.org: Get "https://registry.build02.ci.openshift.org/v2/": tls: failed to verify certificate: x509: certificate has expired or is not yet valid: current time 2024-03-06T16:40:38Z is after 2023-10-10T07:36:58Z
```
https://github.com/openshift/machine-config-operator/pull/3907 should resove that

### DIFF
--- a/pkg/cmd/certregenerationcontroller/cmd.go
+++ b/pkg/cmd/certregenerationcontroller/cmd.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
@@ -123,10 +124,19 @@ func (o *Options) Run(ctx context.Context) error {
 		return err
 	}
 
+	kubeInformers := informers.NewSharedInformerFactory(kubeClient, 10*time.Minute)
+	masterCSRController := NewControlPlaneCSRController(
+		ctx,
+		kubeClient,
+		kubeInformers,
+		o.controllerContext.EventRecorder,
+	)
+
 	// We can't start informers until after the resources have been requested. Now is the time.
 	configInformers.Start(ctx.Done())
 	kubeAPIServerInformersForNamespaces.Start(ctx.Done())
 	dynamicInformers.Start(ctx.Done())
+	kubeInformers.Start(ctx.Done())
 
 	// FIXME: These are missing a wait group to track goroutines and handle graceful termination
 	// (@deads2k wants time to think it through)
@@ -137,6 +147,10 @@ func (o *Options) Run(ctx context.Context) error {
 
 	go func() {
 		caBundleController.Run(ctx)
+	}()
+
+	go func() {
+		masterCSRController.Run(ctx, 1)
 	}()
 
 	<-ctx.Done()
--- a/pkg/cmd/certregenerationcontroller/csrapprover.go
+++ b/pkg/cmd/certregenerationcontroller/csrapprover.go
@@ -1,0 +1,295 @@
+package certregenerationcontroller
+
+import (
+	"context"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+	"net"
+	"strings"
+	"time"
+
+	certapiv1 "k8s.io/api/certificates/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	certv1client "k8s.io/client-go/kubernetes/typed/certificates/v1"
+	"k8s.io/klog/v2"
+
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	csrv1listers "k8s.io/client-go/listers/certificates/v1"
+)
+
+const (
+	controllerName       = "control-plane-csr-approver"
+	bootstrapperUsername = "system:serviceaccount:openshift-machine-config-operator:node-bootstrapper"
+)
+
+var (
+	mcoUserGroups = sets.NewString(
+		"system:authenticated",
+		"system:serviceaccounts",
+		"system:serviceaccounts:openshift-machine-config-operator",
+	)
+	kubeletUserGroups = sets.NewString("system:authenticated", "system:nodes")
+)
+
+type ControlPlaneCSRController interface {
+	factory.Controller
+}
+
+type NodeAddresses struct {
+	DNSNames    sets.String
+	IPAddresses sets.String
+}
+
+type controller struct {
+	factory.Controller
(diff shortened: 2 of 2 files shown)


pkg/cmd/certregenerationcontroller/cmd.go
```diff
@@ -123,10 +124,22 @@ func (o *Options) Run(ctx context.Context) error {
 		return err
 	}
 
+	kubeInformers := informers.NewSharedInformerFactory(kubeClient, 10*time.Minute)
```
vrutkovs, line 127: Not sure if this is a good idea to piggyback on certregen controller, but I'd prefer to avoid running another binary and modifying static pod with a new container

vrutkovs:
Seems library-go's CSRApprover is not suitable for this task:
```
2024-02-21T14:21:37.840079140+00:00 stderr F I0221 14:21:37.840016       1 event.go:298] Event(v1.ObjectReference{Kind:"Namespace", Namespace:"openshift-kube-apiserver", Name:"openshift-kube-apiserver", UID:"", APIVersion:"v1", ResourceVersion:"", FieldPath:""}): type: 'Normal' reason: 'CSRDenial' The CSR "csr-66zz5" has been denied: IllegitimateRequester - requester cannot request certificates for themselves
2024-02-21T14:21:37.843745215+00:00 stderr F I0221 14:21:37.843725       1 csrapprover.go:247] ControlPlaneNodeApprover: working on CSR csr-66zz5
2024-02-21T14:21:37.843998723+00:00 stderr F I0221 14:21:37.843981       1 event.go:298] Event(v1.ObjectReference{Kind:"Namespace", Namespace:"openshift-kube-apiserver", Name:"openshift-kube-apiserver", UID:"", APIVersion:"v1", ResourceVersion:"", FieldPath:""}): type: 'Normal' reason: 'CSRApproval' The CSR "csr-66zz5" has been approved
2024-02-21T14:21:37.845789385+00:00 stderr F E0221 14:21:37.845756       1 base_controller.go:266] "WebhookAuthenticatorCertApprover_master-nodes-csr-approver-controller" controller failed to sync "csr-66zz5", err: Operation cannot be fulfilled on certificatesigningrequests.certificates.k8s.io "csr-66zz5": the object has been modified; please apply your changes to the latest version and try again
```
It expects CSR CN to be [different](https://github.com/openshift/library-go/blob/d7e7beca5baea998564ed16bd3d7e7dffbc7c01f/pkg/operator/csr/csr_approver.go#L115-L117) from spec.username

pkg/cmd/certregenerationcontroller/csrapprover.go
```diff
@@ -0,0 +1,298 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create new control plane node approver: %w", err)
+	}
+	return &ControlPlaneCSRController{
+		controller: csr.NewCSRApproverController(
+			controllerName,
+			nil,
+			kubeClient.CertificatesV1().CertificateSigningRequests(),
+			kubeInformers.Certificates().V1().CertificateSigningRequests(),
+			csr.NewLabelFilter(labels.NewSelector()),
```
deads2k, line 67: if you only want to approve the first kubelet client certificate, this needs to filter on
1. the username of the CSR requestor must be the bootstrapping service account
2. the CSR must be for the right signer
3. the CSR must be for a kubelet shaped client certificates

pkg/cmd/certregenerationcontroller/csrapprover.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+func (c *controller) Validate(csrObj *certapiv1.CertificateSigningRequest, x509CSR *x509.CertificateRequest) error {
+	if csrObj == nil || x509CSR == nil {
+		return fmt.Errorf("received a 'nil' CSR")
+	}
+
+	klog.Infof("ControlPlaneNodeApprover: working on CSR %s", csrObj.Name)
+	// Check username. It should either be a node bootstrapper or match one of known nodes
+	if csrObj.Spec.Username != bootstrapperUsername && !c.getUsernames().Has(csrObj.Spec.Username) {
```
p0lyn0mial, line 200: During my investigation on the cert-recovery story I found out that the `bootstrapperUsername` is a public asset. In theory any process running on a node can access it and issue a CSR. So, it looks like we cannot trust this asset and it looks like we need a different solution.

vrutkovs: Correct, there is no way for us to determine that this CSR genuinely comes from the node. However, we ensure that the approved cert will be valid for node IPs and DNSes only, it would be useless for any other node, so its not that dangerous to approve it

p0lyn0mial: > However, we ensure that the approved cert will be valid for node IPs and DNSes only, it would be useless for any other node, so its not that dangerous to approve it

I think that kube doesn't validate if a client certificate comes from a specific IP address. Anyone with the certificate can claim the identity.

vrutkovs: Correct, but this identity is usable on the particular master node. Also, this CSR will be approved only for refreshes, so a new node certs won't be approved. I can't imagine a way this cert to be useful for the attacker

p0lyn0mial: > Also, this CSR will be approved only for refreshes

How do you know the old certificate has expired ?

> Correct, but this identity is usable on the particular master node

How can you ensure that since there is no way to check on the server if a request comes from a specific machine ?

>  I can't imagine a way this cert to be useful for the attacker

If the signer for a CSR is public then any process on any machine can read it and send a CSR request, right ?

p0lyn0mial: > The certificate has a set of IPs and DNSNames its valid for. It can't be used from any other node

If any process can generate a certificate with exactly the same content (IPs, DNSNames) then how can we ensure anything ?

Note that in cryptography it doesn't matter which IP sent a request. What matters is presenting a valid secret, in this case, a certificate.

Can you point me to code that check if the IP address of a sender is associated with a certificate ?

vrutkovs:
This will never be secure enough, as kubelet can't prove to kube-apiserver that its indeed authenticated.

Instead we should use ACME protocol to exchange secret keys and use TLS bootstrap to issue a public key

PR #1542, opened 2023-09-05
OCPBUGS-18247: manifests: don't include recording rules when Console capability is not enabled

vrutkovs:
In order to avoid additional load on Prometheus the recording rules for kube-apiserver dashboard are not included when Console capability is not enablked . These are not used anywhere else, so it should not affect any other components.


vrutkovs:
/hold

testing if its indeed disabled on single node

vrutkovs:
/hold cancel

Verified via cluster bot that the cluster without Console capability doesn't have recording rules

dgrisonnet (approved):
/lgtm
/approve

Holding until merging to 4.14 is possible again
/hold

PR #1552, opened 2023-09-19
alerts: exclude iowait and steal CPU modes

dgrisonnet:
Exclude the iowait and steal CPU modes on top of the idle mode since these correspond to idle/wait states.

Per the iostat man page:

%idle
Show the percentage of time that the CPU or CPUs were idle and the system did not have an outstanding disk I/O request.

%iowait
Show the percentage of time that the CPU or CPUs were idle during which the system had an outstanding disk I/O request.

%steal
Show the percentage of time spent in involuntary wait by the virtual CPU or CPUs while the hypervisor was servicing another virtual processor.

### DIFF
--- a/bindata/assets/alerts/cpu-utilization.yaml
+++ b/bindata/assets/alerts/cpu-utilization.yaml
@@ -21,8 +21,16 @@ spec:
               To fix this, increase the CPU and memory on your control plane nodes.
           expr: |
             sum(
-              100 - (avg by (instance) (rate(node_cpu_seconds_total{mode="idle"}[1m])) * 100)
-              AND on (instance) label_replace( kube_node_role{role="master"}, "instance", "$1", "node", "(.+)" )
+              100 - (
+                avg by (instance) (
+                  sum without (mode) (
+                    rate(
+                      node_cpu_seconds_total{mode=~"idle|iowait|steal"}[1m]
+                    )
+                  )
+                ) * 100
+              )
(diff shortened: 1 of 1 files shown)


bindata/assets/alerts/cpu-utilization.yaml
```diff
@@ -21,8 +21,16 @@ spec:
               To fix this, increase the CPU and memory on your control plane nodes.
           expr: |
             sum(
-              100 - (avg by (instance) (rate(node_cpu_seconds_total{mode="idle"}[1m])) * 100)
-              AND on (instance) label_replace( kube_node_role{role="master"}, "instance", "$1", "node", "(.+)" )
+              100 - (
+                avg by (instance) (
+                  sum without (mode) (
+                    rate(
+                      node_cpu_seconds_total{mode=~"idle|iowait|steal"}[1m]
```
wking, line 28: Is there separate alert coverage for "high overall control-plane iowait, maybe get faster disks or scale down your consumption"?

dgrisonnet: I don't think we have one, but it would definitely make sense especially considering the history with I/O on Azure

wking: #1553 is poking around in this space now :)

dgrisonnet: yes I started a discussion with the etcd team in https://redhat-internal.slack.com/archives/C027U68LP/p1695210456306999, but we still need to figure out if alerting on iowait is the right thing to do

PR #1553, opened 2023-09-20
ETCD-187: add highOverallControlPlaneIOWait alert rule

Elbehery:
resolves https://issues.redhat.com/browse/ETCD-187

cc @wking @dgrisonnet @williamcaban @tjungblu

### DIFF
--- a/bindata/assets/alerts/cpu-utilization.yaml
+++ b/bindata/assets/alerts/cpu-utilization.yaml
@@ -69,3 +69,13 @@ spec:
           labels:
             namespace: openshift-kube-apiserver
             severity: critical
+        - alert: highOverallControlPlaneIOWait
+          annotations:
+            description: Control-plane nodes are spending more than 20% of CPU time on iowait mode.
+            summary: CPU utilization on master nodes is more than 20% on iowait mode.
+          expr: |
+            (sum(irate(node_cpu_seconds_total {mode="iowait"} [2m])) without (cpu)) / count(node_cpu_seconds_total) without (cpu) * 100 > 20
+            AND on (instance) label_replace( kube_node_role{role="master"}, "instance", "$1", "node", "(.+)" )
+          for: 1h
+          labels:
+            severity: warning


bindata/assets/alerts/cpu-utilization.yaml
```diff
@@ -69,3 +69,13 @@ spec:
           labels:
             namespace: openshift-kube-apiserver
             severity: critical
+        - alert: highOverallControlPlaneIOWait
+          annotations:
+            description: Control-plane nodes are spending more than 20% of CPU time on iowait mode.
+            summary: CPU utilization on master nodes is more than 20% on iowait mode.
+          expr: |
+            (sum(irate(node_cpu_seconds_total {mode="iowait"} [2m])) without (cpu)) / count(node_cpu_seconds_total) without (cpu) * 100 > 20
```
dgrisonnet, line 77: Hi @openshift/sre-alert-sme, we are trying to build an alert to detect disk/network pressure causing CPU latency. Based on your experience with OCP, would 20% of CPU time waiting for IO make sense to start alerting on?

Also what could be a meaningful threshold for a critical alert?

bindata/assets/alerts/cpu-utilization.yaml
```diff
@@ -69,3 +69,13 @@ spec:
           labels:
             namespace: openshift-kube-apiserver
             severity: critical
+        - alert: highOverallControlPlaneIOWait
+          annotations:
+            description: Control-plane nodes are spending more than 20% of CPU time on iowait mode.
```
wking, line 74: I think this is going to need a bit more to be actionable.  Do we have recommended advice for when `iowait` is high?  I expect we'll want some ideas about digging in to find what we're waiting for, and then thoughts about how to get more throughput from that thing.  For control-plane nodes, assuming that the bottlenecks will primarily be etcd vs. disk storage and various containerized things vs. network capacity is probably fairly safe, so a runbook with steps on how to check that?  Is etcd vs. disk latency something that's already covered by its own alerts?  I know we have `histogram_quantile` advice [in docs][1], but "we doc those recommendations but do not alert on them" and "we alert on those recommendations, and doc the recommendations, but  do not doc the alerts" would both surprise me as an intentional strategy.

[1]: https://docs.openshift.com/container-platform/4.13/scalability_and_performance/recommended-performance-scale-practices/recommended-etcd-practices.html#recommended-etcd-practices_recommended-etcd-practices

wking: aha, [existing alerts][1], `etcdMemberCommunicationSlow`, `etcdMemberCommunicationSlow`, and `etcdHighCommitDurations`.  I don't see any of those looking at read-throughput today, but that sort of thing may be a sufficient cover, and we might not need to cover `iowait` at the node level?  Or if we do cover it at the node level, the message could be "you seem to have over-provisioned your CPUs; if your workload is happy, you could maybe save some money by downsizing to cheaper CPUs" or some such, because your most efficient when your workload is CPU throttled and I/O throttled and right up on the available-memory limit all at the same time.  For actual CPU use and memory, we want [some overhead to handle reboots][2].  But I/O capacity is a harder resource to get a handle on, and I'm not aware of a way to say "we want some I/O overhead on nodes so they can burst if they have to work harder while handling a control-plane roll".

[1]: https://github.com/openshift/cluster-etcd-operator/blob/407c126614a1d15d14e3b9890b3a8f7eae150d72/manifests/0000_90_etcd-operator_03_prometheusrule.yaml#L40C14-L89
[2]: https://docs.openshift.com/container-platform/4.13/scalability_and_performance/recommended-performance-scale-practices/recommended-control-plane-practices.html#master-node-sizing_recommended-control-plane-practices

PR #1555, opened 2023-09-21
Remove noisy ExtremelyHighIndividualControlPlaneCPU warning alert

dgrisonnet:
The ExtremelyHighIndividualControlPlaneCPU had two different variation with different criticality.
- The warning alert was firing after 5 minutes of the control plane CPU usage being above 90%
- The critical alert firing after 60 minutes of the control plane CPU usage being above 90%

Because the timeline of a warning level alert in OCP was defined to be 60 minutes, having the warning level alert response timeline collide with the time we expect response from the critical alert makes it noisy.

On top of that, 5 minutes is very low considering that this alert most likely always fire during upgrades because of the pressure control plane nodes are suffering from post upgrade.

### DIFF
--- a/bindata/assets/alerts/cpu-utilization.yaml
+++ b/bindata/assets/alerts/cpu-utilization.yaml
@@ -34,25 +34,6 @@ spec:
         - alert: ExtremelyHighIndividualControlPlaneCPU
           annotations:
             summary: >-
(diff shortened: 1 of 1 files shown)


sdodson:
>On top of that, 5 minutes is very low considering that this alert most likely always fire during upgrades because of the pressure control plane nodes are suffering from post upgrade.

By this you mean the churn associated with the rolling update of worker nodes, evicting pods, rescheduling, etc?

PR #1562, opened 2023-09-27
Update "deprecated API in use" alert expressions for 1.28.

benluddy:


bindata/assets/alerts/api-usage.yaml
```diff
@@ -34,7 +34,7 @@ spec:
               a successful upgrade to the next EUS cluster version with Kubernetes {{ $labels.removed_release }}.
               Refer to `oc get apirequestcounts {{ $labels.resource }}.{{ $labels.version }}.{{ $labels.group }} -o yaml` to identify the workload.
           expr: >-
-            group by (group,version,resource,removed_release) (apiserver_requested_deprecated_apis{removed_release=~"1[.]2[89]"})
+            group by (group,version,resource,removed_release) (apiserver_requested_deprecated_apis{removed_release=~"1[.](29|30)"})
```
dgrisonnet, line 37: I think this should be 1.29 only since the next EUS will be 4.16 which will most likely be based on 1.29

benluddy: You're right, thanks.

dgrisonnet (approved):
/lgtm
/approve

PR #1563, opened 2023-09-29
skip when network status.serviceNetwork is nil

lance5890:
Skip when the network config status.serviceNetwork is nil

related issue: https://github.com/openshift/cluster-kube-apiserver-operator/issues/1564


lance5890:
@tkashem @p0lyn0mial

lance5890:
@vrutkovs

vrutkovs:
We don't use Github issues to track CKAO issues, please file a new one at https://issues.redhat.com/, project OCPBUGS, component "kube-apiserver"

PR #1566, opened 2023-10-10
OCPBUGS-20331: manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver: Rename to kube-apiserver-performance-recording-rules

wking:
Exactly like 7457da4fc0 (#1565), this pivot:

* Turns the original (group, namepspace, name) into a delete manifest, with no capability annotation (all clusters should delete this resource).
* Creates a (group, namespace, newName) manifest with the original spec content and capability annotation (only clusters with the cap enabled should have this resource).

which allows updates like:

1. Old release with the old (kind, group, namepspace, name) not annotated for a capability.
2. Request update to new release.
3. Outgoing cluster-version operator compares the outgoing manifests with the incoming manifests to decide if any capabilities need to be implicitly enabled:
    1. It sees that the old (kind, group, namepspace, name) isn't labeled with any capabilities, so no need to implicitly enable anything there.
    2. It sees that the new (kind, group, namespace, newName) is annotated for the capability, but it doesn't see anything with that k,g,n,n getting reconciled in the outgoing manifest set, so it thinks "just some new manifest for a capability I do not care about", and does not enable the annotated capability.

In a separate pull request than 7457da4fc0, because this one doesn't need to go back to 4.14.  The RecordingRule is from 04f36e7309 (#1521), which landed in 4.14, so 4.13 clusters do not have a matching (kind, group, namespace, name) manifest to worry about matching.

### DIFF
--- a/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
+++ b/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
@@ -119,7 +119,7 @@ spec:
 apiVersion: monitoring.coreos.com/v1
 kind: PrometheusRule
 metadata:
-  name: kube-apiserver-recording-rules
+  name: kube-apiserver-performance-recording-rules
   namespace: openshift-kube-apiserver
   annotations:
     include.release.openshift.io/self-managed-high-availability: "true"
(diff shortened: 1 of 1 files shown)


vrutkovs:
/close

Updated original PR, as it has better description. In case I'm unavailable feel free to reopen and add fixes

wking:
I'd like to rebase and add the same kind of pivot to #1551's `kube-apiserver-recording-rules` PrometheusRule.

/reopen

wking:
[`openshift-cloud-credential-operator` `TargetDown`][1] seems orthogonal:

/retest

[1]:https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-kube-apiserver-operator/1566/pull-ci-openshift-cluster-kube-apiserver-operator-master-e2e-aws-ovn/1712393544678772736

wking:
@vrutkovs pointed out that this PrometheusRule was new in 4.14, so we won't need to backport this  change to 4.14.  I've rerolled with 0b6040b -> 0412fb3 to sit in parallel with #1565 instead of on top of it.

dgrisonnet (approved):
/lgtm
/approve

sdodson:
Nevermind. No need for 4.14 backport per 
>@vrutkovs pointed out that this PrometheusRule was new in 4.14, so we won't need to backport this change to 4.14. I've rerolled with https://github.com/openshift/cluster-kube-apiserver-operator/commit/0b6040beb51aaf239e28f2a07f64c8d1c41a95fc -> https://github.com/openshift/cluster-kube-apiserver-operator/commit/0412fb31e03c5c278d1ba90a267507392c9a1fc8 to sit in parallel with https://github.com/openshift/cluster-kube-apiserver-operator/pull/1565 instead of on top of it.

PR #1568, opened 2023-10-25
Annotate managed certs

vrutkovs:


pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -302,6 +303,11 @@ func ManageClientCABundle(ctx context.Context, lister corev1listers.ConfigMapLis
 	if err != nil {
 		return nil, false, err
 	}
+	if requiredConfigMap.Annotations == nil {
```
dinhxuanvu, line 306: I was doing this in my PR on service-ca-operator as well but then I think it is better just to group all of ownership-related annotations operation to a func. Then, insert that func anywhere you need. Plus, it may be handy in the future to add more annotations settings in that func instead of having to coping the code everywhere.

vrutkovs: Good idea, we should have "set necessary annotations" func and "verify annotations" func

pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -133,6 +133,8 @@ func newCertRotationController(
 		certrotation.RotatedSigningCASecret{
 			Namespace:              operatorclient.OperatorNamespace,
 			Name:                   "aggregator-client-signer",
+			JiraComponent:          "kube-apiserver",
+			Description:            "Signer for the kube-apiserver to create client certificates for aggregated apiservers to recognize as a front-proxy.",
```
deads2k, line 137: pull descriptions to a separate PR, just do owners here.

In the descriptions PR, add information about which processes and flags these are attached to.  Or in cases where its a constituent indicate that it is combined with other certs to eventually be passed to kube-apiserver --foo.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -302,6 +303,11 @@ func ManageClientCABundle(ctx context.Context, lister corev1listers.ConfigMapLis
...
+		requiredConfigMap.Annotations = map[string]string{}
+	}
+	requiredConfigMap.Annotations[annotations.OpenShiftComponent] = "kube-apiserver"
+	requiredConfigMap.Annotations[annotations.OpenShiftDescription] = "CA for kube-apiserver clients"
```
deads2k, line 310: separate to another PR as other descriptions above.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -322,6 +331,11 @@ func manageKubeAPIServerCABundle(ctx context.Context, lister corev1listers.Confi
 	if err != nil {
 		return nil, false, err
 	}
+	if requiredConfigMap.Annotations == nil {
+		requiredConfigMap.Annotations = map[string]string{}
+	}
+	requiredConfigMap.Annotations[annotations.OpenShiftComponent] = "kube-apiserver"
+	requiredConfigMap.Annotations[annotations.OpenShiftDescription] = "CA for kube-apiserver server"
```
deads2k, line 338: remove description.

PR #1582, opened 2023-11-16
OCPBUGS-22117: alerts: add sno cpu usage alerts

qJkee:
Create an alert for SNO specifically for high cpu usage by control plane pods.

I did remove `runbook_url` intentionally. Can we reuse the regular runbook or you'd like to see a separate one for SNO? It does not fit 100% for SNO case

### DIFF
--- a/bindata/assets/alerts/cpu-utilization-sno.yaml
+++ b/bindata/assets/alerts/cpu-utilization-sno.yaml
@@ -1,0 +1,55 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: cpu-utilization
+  namespace: openshift-kube-apiserver
+spec:
+  groups:
+    - name: control-plane-cpu-utilization
+      rules:
+        - alert: ExtremelyHighIndividualControlPlaneCPU
+          annotations:
+            summary: >-
+              CPU utilization across plane pods is high; a control plane pods outage may cause a cascading failure;
+              check control plane logs for possible issues and/or add more CPU.
+            runbook_url: PASS
+            description: >-
+              Usually, when control plane pods start to utilize more CPU, this means that there are might be
+              a issue at them. It's recommended to check pods and their logs to verify that there are not issues.
+              To fix this, increase the CPU and memory on your control plane nodes.
+          expr: | 
+            sum(
+              rate(
+                container_cpu_system_seconds_total{namespace=~"openshift-etcd|openshift-kube-apiserver|openshift-kube-controller-manager|openshift-kube-scheduler|openshift-apiserver|openshift-controller-manager|openshift-authentication",image!=""
+                }[5m]
+              )
+            ) / sum(machine_cpu_cores) * 100 
+            > 50
+          for: 10m
+          labels:
+            namespace: openshift-kube-apiserver
+            severity: warning
+        - alert: ExtremelyHighIndividualControlPlaneCPU
+          annotations:
+            summary: >-
+              CPU utilization on a control plane pods is very high, more CPU pressure is likely to cause a failover; increase available CPU.
+            runbook_url: PASS
+            description: >-
+              Extreme CPU pressure can cause slow serialization and poor performance from the kube-apiserver and etcd.
+              When this happens, there is a risk of clients seeing non-responsive API requests which are issued again
+              causing even more CPU pressure.
+              It can also cause failing liveness probes due to slow etcd responsiveness on the backend.
+              If one kube-apiserver fails under this condition, chances are you will experience a cascade as the remaining
+              kube-apiservers are also under-provisioned. To fix this, increase the CPU and memory on your control plane nodes.
+          expr: |
+            sum(
+              rate(
+                container_cpu_system_seconds_total{namespace=~"openshift-etcd|openshift-kube-apiserver|openshift-kube-controller-manager|openshift-kube-scheduler|openshift-apiserver|openshift-controller-manager|openshift-authentication",image!=""
+                }[5m]
+              )
+            ) / sum(machine_cpu_cores) * 100
+            > 70
+          for: 5m
+          labels:
+            namespace: openshift-kube-apiserver
+            severity: critical
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -170,6 +170,12 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 	var notOnSingleReplicaTopology resourceapply.ConditionalFunction = func() bool {
 		return infrastructure.Status.ControlPlaneTopology != configv1.SingleReplicaTopologyMode
 	}
+
+	// onSingleReplicaTopology represents SNO deployments
+	var onSingleReplicaTopology resourceapply.ConditionalFunction = func() bool {
+		return infrastructure.Status.ControlPlaneTopology == configv1.SingleReplicaTopologyMode
+	}
+
 	var never resourceapply.ConditionalFunction = func() bool { return false }
 	staticResourceController := staticresourcecontroller.NewStaticResourceController(
 		"KubeAPIServerStaticResources",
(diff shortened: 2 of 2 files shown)


bindata/assets/alerts/cpu-utilization-sno.yaml
```diff
@@ -0,0 +1,82 @@
...
+              It's recommended to check pods and their logs to verify that there are not issues.
+              If all is ok, then increase the CPU on the node and increase alert threshold.
+          expr: |
+            sum(
+              rate(
+                container_cpu_system_seconds_total{namespace=~"openshift-etcd|openshift-kube-apiserver|openshift-kube-controller-manager|openshift-kube-scheduler|openshift-apiserver|openshift-controller-manager|openshift-authentication",image!=""
+                }[5m]
+              )
+            )
+            > 3
```
simonpasquier, line 28: This is assuming that the number of vCPUs allocated for the platform components is fixed for all setups?

bindata/assets/alerts/cpu-utilization-sno.yaml
```diff
@@ -0,0 +1,82 @@
...
+              rate(
+                container_cpu_system_seconds_total{namespace=~"openshift-etcd|openshift-kube-apiserver|openshift-kube-controller-manager|openshift-kube-scheduler|openshift-apiserver|openshift-controller-manager|openshift-authentication",image!=""
+                }[5m]
+              )
+            )
+            > 3.4
+          for: 5m
+          labels:
+            namespace: openshift-kube-apiserver
+            severity: alert
```
simonpasquier, line 57: not a valid severity level

bindata/assets/alerts/cpu-utilization-sno.yaml
```diff
@@ -0,0 +1,82 @@
...
+        - alert: ExtremelyHighIndividualControlPlaneCPU
```
simonpasquier, line 58: rather than having a new alert name, I'd keep the same name and leverage different severities

bindata/assets/alerts/cpu-utilization-sno.yaml
simonpasquier: Your choice but having 2 alerts on the same metric at the same level (warning) would lead to alert fatigue IMHO.
The valid values are critical, warning or info:
https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/alerting-consistency.md#style-guide

bindata/assets/alerts/cpu-utilization-sno.yaml
simonpasquier: > it can be always modified

how? the value is hardcoded in the expression so you'll change for all or nothing?

bindata/assets/alerts/cpu-utilization-sno.yaml
simonpasquier: In case of SNO clusters, you probably want only 2 `ExtremelyHighIndividualControlPlaneCPU` alerting rules (one warning and one critical). The `HighOverallControlPlaneCPU` alerting rule which aggregates across all control-plane nodes  isn't relevant since 1 cluster = 1 node.

bindata/assets/alerts/cpu-utilization-sno.yaml
simonpasquier: The primary use case for `AlertingRule` is for the cluster admins to setup new alerting rules that aren't covered by the product.
For out-of-the-box alerting rules, it's preferred to have adaptive thresholds if possible.

bindata/assets/alerts/cpu-utilization-sno.yaml
```diff
@@ -0,0 +1,55 @@
...
+  namespace: openshift-kube-apiserver
+spec:
+  groups:
+    - name: control-plane-cpu-utilization
+      rules:
+        - alert: ExtremelyHighIndividualControlPlaneCPU
+          annotations:
+            summary: >-
+              CPU utilization across plane pods is high; a control plane pods outage may cause a cascading failure;
+              check control plane logs for possible issues and/or add more CPU.
```
simonpasquier, line 14: I would move the `; a control plane pods outage may ...; check control plane logs ...` part to the description.

bindata/assets/alerts/cpu-utilization-sno.yaml
```diff
@@ -0,0 +1,55 @@
...
+            runbook_url: PASS
+            description: >-
+              Usually, when control plane pods start to utilize more CPU, this means that there are might be
+              a issue at them. It's recommended to check pods and their logs to verify that there are not issues.
+              To fix this, increase the CPU and memory on your control plane nodes.
```
simonpasquier, line 19: it says "control plane nodes" but there's only 1 node.

bindata/assets/alerts/cpu-utilization-sno.yaml
simonpasquier: I thought that it was possible to pin the management workload to a subset of the cores. In which case sum(machine_cpu_cores) can be larger than the number of cores allocated for control plane components?

https://docs.openshift.com/container-platform/4.14/scalability_and_performance/enabling-workload-partitioning.html
https://docs.openshift.com/container-platform/4.14/scalability_and_performance/ztp_far_edge/ztp-reference-cluster-configuration-for-vdu.html#ztp-sno-du-enabling-workload-partitioning_sno-configure-for-vdu

simonpasquier: But when it's enabled, is there a way to know how many cores are allocated?

simonpasquier: Ideally you want to build an alert expression that takes into account the variability of workload partitioning and another metric reporting the number of cores being allocated is the way to go indeed. Having said, I'm not a maintainer of this repository so I don't feel strongly whether you should do it now or later.

wking:
I suspect [the original 4.15.0 check][1] is stale now that 4.15 has forked off the dev branch, and the bug should be retargeted to 4.16.0.

/jira refresh

[1]: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1582#issuecomment-1814661653

wking:
I adjusted the bug; hopefully good again now:

/jira refresh

bindata/assets/alerts/cpu-utilization-sno.yaml
```diff
@@ -0,0 +1,55 @@
...
+          expr: | 
+            sum(
+              rate(
+                container_cpu_system_seconds_total{namespace=~"openshift-etcd|openshift-kube-apiserver|openshift-kube-controller-manager|openshift-kube-scheduler|openshift-apiserver|openshift-controller-manager|openshift-authentication",image!=""
+                }[5m]
+              )
+            ) / sum(machine_cpu_cores) * 100
```
wking, line 26: `controlPlaneTopology: SingleReplica` means the control plane isn't highly available, but doesn't mean that the there are no compute nodes, does it?  I expect in practice most clusters with `controlPlaneTopology: SingleReplica` have no attached compute nodes, but the change that they might makes me a bit nervous about dropping the `kube_node_role{role="master"}` entirely.

The 60% threshold is also from [these docs][1]:

> On a large and dense cluster with three control plane nodes, the CPU and memory usage will spike up when one of the nodes is stopped, rebooted, or fails. The failures can be due to unexpected issues with power, network, underlying infrastructure, or intentional cases where the cluster is restarted after shutting it down to save costs. The remaining two control plane nodes must handle the load in order to be highly available, which leads to increase in the resource usage. This is also expected during upgrades because the control plane nodes are cordoned, drained, and rebooted serially to apply the operating system updates, as well as the control plane Operators update. To avoid cascading failures, keep the overall CPU and memory resource usage on the control plane nodes to at most 60% of all available capacity to handle the resource usage spikes. Increase the CPU and memory on the control plane nodes accordingly to avoid potential downtime due to lack of resources.

and as we adjust the associated alerts to account for single-node control planes (which makes sense to me), I think we probably also want to adjust those docs, to keep the docs and alerts in sync.

[1]: https://docs.openshift.com/container-platform/4.14/scalability_and_performance/recommended-performance-scale-practices/recommended-control-plane-practices.html#master-node-sizing_recommended-control-plane-practices

PR #1583, opened 2023-11-16
manifests: set owning component for TLS artifacts

vrutkovs:


bindata/bootkube/manifests/configmap-csr-controller-ca.yaml
```diff
@@ -4,6 +4,8 @@ kind: ConfigMap
 metadata:
   name: csr-controller-ca
   namespace: openshift-config-managed
+  annotations:
+    "openshift.io/owning-component": "kube-apiserver"
```
deads2k, line 8: I claim this is owned by the kube-controller-manager which writes content to this location here: https://github.com/openshift/cluster-kube-controller-manager-operator/blob/master/pkg/operator/targetconfigcontroller/targetconfigcontroller.go#L724

While you may wish to try moving this part of rendering, do so as a followup because it's reasonably likely something will break.

bindata/assets/kube-apiserver/trusted-ca-cm.yaml
```diff
@@ -3,5 +3,7 @@ kind: ConfigMap
 metadata:
   namespace: openshift-kube-apiserver
   name: trusted-ca-bundle
+  annotations:
+    "openshift.io/owning-component": "kube-apiserver"
```
deads2k, line 7: this is actually owned by the cluster network operator, right?

PR #1584, opened 2023-11-17
NO-JIRA: Add ownership for the admin kubeconfig

deads2k:
This comes from the installer, but it's not user provided. It is logically owned by the kube-apiserver and something must manage the equivalent when the admin.kubeconfig expires.  We're close to that point.


vrutkovs (approved):
/lgtm

dinhxuanvu (commented):
/lgtm

PR #1586, opened 2023-11-22
OCPNODE-1890: Bump k8s api to v0.29.0

sairameshv:
cc: @soltysh


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/openshift/cluster-kube-apiserver-operator
 
-go 1.20
+go 1.21.3
```
soltysh, line 3: Same as before go1.20

soltysh (changes requested):

go.mod
sairameshv: Addressed

soltysh (approved):
/lgtm
/approve

PR #1587, opened 2023-11-22
OCPBUGS-24005: webhookcontroller: report when a webhook resource is missing a caBundle provided by the service-ca-operator

p0lyn0mial:
This is clearly a gap in the platform. Perhaps the best way would be to introduce a readiness condition `kas` would honour. On the other hand the issue is not that severe since most of the time it will appear during a webhook installation. I've checked and the cert-manager works exactly in the same way.

Initially I was thinking about waiting for a caBundle to be populated but realised it would hide the issue. I've changed the PR to report the issue instead.

### DIFF
--- a/pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
+++ b/pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
@@ -16,6 +16,12 @@ import (
 	"k8s.io/klog/v2"
 )
 
+const (
+	// injectCABundleAnnotationName is the annotation used by the
+	// service-ca-operator to indicate which resources it should inject the CA into.
+	injectCABundleAnnotationName = "service.beta.openshift.io/inject-cabundle"
+)
+
 // webhookInfo generically represents a webhook
 type webhookInfo struct {
 	Name                  string
@@ -25,6 +31,9 @@ type webhookInfo struct {
 	// TimeoutSeconds specifies the timeout for a webhook.
 	// After the timeout passes, the webhook call will be ignored or the API call will fail
 	TimeoutSeconds *int32
+	// HasServiceCaAnnotation indicates whether a webhook
+	// resource has been annotated for CABundle injection by the service-ca-operator
+	HasServiceCaAnnotation bool
 }
 
 // serviceReference generically represents a service reference
@@ -52,7 +61,7 @@ func (c *webhookSupportabilityController) updateWebhookConfigurationDegraded(ctx
 				serviceMsgs = append(serviceMsgs, msg)
 				continue
 			}
-			err = c.assertConnect(ctx, webhook.Name, webhook.Service, webhook.CABundle, webhook.TimeoutSeconds)
+			err = c.assertConnect(ctx, webhook.Name, webhook.Service, webhook.CABundle, webhook.HasServiceCaAnnotation, webhook.TimeoutSeconds)
 			if err != nil {
 				msg := fmt.Sprintf("%s: %s", webhook.Name, err)
 				if webhook.FailurePolicyIsIgnore {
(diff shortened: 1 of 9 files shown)


pkg/operator/webhooksupportabilitycontroller/degraded_webhook_conversion.go
```diff
@@ -48,5 +56,24 @@ func hasCRDConversionWebhookConfiguration(crd *apiextensionsv1.CustomResourceDef
...
+	if waitForServiceCABundle(crd.Annotations, clientConfig.CABundle) {
+		return false
+	}
 	return true
 }
+
+// waitForServiceCABundle returns true if a resource has the service-ca
+// annotation and the caBundle hasn't been yet provided.
+func waitForServiceCABundle(annotations map[string]string, caBundle []byte) bool {
+	if hasServiceCaAnnotation(annotations) {
```
tkashem, line 68: do we need to check for the annotation, shouldn't it be enough to just check for the length of the caBundle?

pkg/operator/webhooksupportabilitycontroller/degraded_webhook_conversion.go
```diff
@@ -48,5 +56,24 @@ func hasCRDConversionWebhookConfiguration(crd *apiextensionsv1.CustomResourceDef
...
+		return len(caBundle) == 0
+	}
+	return false
+}
+
+func hasServiceCaAnnotation(annotations map[string]string) bool {
+	if annotations == nil {
+		return false
+	}
```
benluddy, line 77:
```suggestion
```

pkg/operator/webhooksupportabilitycontroller/degraded_webhook_conversion.go
p0lyn0mial: I think we do, at least when the resource is annotated, in that case we rely on the service-ca to provide the bundle.

p0lyn0mial: In general, a caBundle is optional and when it is empty we use CAs provided by the underlying OS.

p0lyn0mial:
> TODO: check what would kas do in such a situation. I think there might be a gap in the platfrom.

I haven't found any additional checks in `kas` that would prevent the server from passing request to a "not fully" configured webhook. I think that the assumption here is that administrators/controllers provide a complete configuration. In OCP that step requires an input from the service-ca. 

The service-ca requires a non-empty `WebhookClientConfig` so that it can inject a caBundle. In order to have a non-empty`WebhookClientConfig` you have to specify a service ref or an url address otherwise the extention-apiserver [won't allow you to add a conversion configuration ](https://github.com/kubernetes/kubernetes/blob/5edb27aa3820f15a235e7f4a5807eb20538cae94/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation/validation.go#L550).

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
```diff
@@ -130,9 +139,13 @@ func (c *webhookSupportabilityController) assertConnect(ctx context.Context, web
 		var conn net.Conn
 		conn, err = dialer.DialContext(ctx, "tcp", net.JoinHostPort(host, port))
 		if err != nil {
+			err = fmt.Errorf("failed to connect to webhook %q via service %q: %v", webhookName, net.JoinHostPort(host, port), err)
+			if caBundleProvidedByServiceCA && len(caBundle) == 0 {
+				err = fmt.Errorf("%v. NOTE that the caBundle is provided by the service-ca-operator but was empty. Please check the service-ca's logs", err)
+			}
```
sanchezl, line 145: I think we should perform this check before attempting the connection. There is little chance that the connection would succeed anyway. This would allow for a simpler error message for the status and less noise in the webhook logs for failed connections that would never have succeeded without the CA bundle.

sanchezl (changes requested):

pkg/operator/webhooksupportabilitycontroller/degraded_webhook.go
p0lyn0mial: OK, I've updated the PR, PTAL.

sanchezl (approved):
/lgtm

PR #1594, opened 2023-11-29
graceful conn termination during webhook check

sanchezl:
Gracefully close TLS connections from the webhook checker to prevent webhooks from logging messages such as:
```
http: TLS handshake error from 127.0.0.1:38954: EOF
```


sanchezl:
No longer needed.

PR #1605, opened 2023-12-18
OCPBUGS-25345: sort errs to void flapping messages

sanchezl:
- bump(*) pickup latest library-go
- bump(*) update vendor


sanchezl:
Already pulled into master.

PR #1618, opened 2024-01-08
NO-JIRA: extend node-system-admin-client validity to 2 years

vrutkovs:
This is necessary to test multi year time skew, so that CSRs could be approved using localhost-recovery kubeconfig


dgrisonnet (approved):
/lgtm

PR #1621, opened 2024-01-09
NO-ISSUE: prevent update status conflicts 

deads2k:
This reduces the 

> Operation cannot be fulfilled on kubeapiservers.operator.openshift.io

messages in the log.  I found 50-100 in runs prior to the update here.

After this update had 5.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/openshift/api v0.0.0-20231219140051-ddc590a81acb
 	github.com/openshift/build-machinery-go v0.0.0-20230228230858-4cd708338479
 	github.com/openshift/client-go v0.0.0-20231218155125-ff7d9f9bf415
-	github.com/openshift/library-go v0.0.0-20240108202620-5674ec6ced1c
+	github.com/openshift/library-go v0.0.0-20240110142250-07ee1a48ec84
 	github.com/pkg/profile v1.5.0 // indirect
 	github.com/prometheus/client_golang v1.16.0
 	github.com/spf13/cobra v1.7.0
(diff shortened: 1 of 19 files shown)


p0lyn0mial:
This PR is pulling https://github.com/openshift/library-go/pull/1658 and it looks like the fix was to issue a live GET after running into a conflict on an update. Why not issue a `PATCH` request which should help resolving conflicts, right ?

deads2k:
> This PR is pulling [openshift/library-go#1658](https://github.com/openshift/library-go/pull/1658) and it looks like the fix was to issue a live GET after running into a conflict on an update. Why not issue a `PATCH` request which should help resolving conflicts, right ?

By passing a function, the operator gets to decide whether the conflict is meaningful enough to skip an update (decide not to update).  I don't remember if we used this capability or not.

Also, a json patch cannot be predicated based on another field.  So the patch for conditions would rely on order, which will fail when a condition is added or removed.  Recall semantic merge patch isn't supported for CRDs.

Finally, server-side-apply has arrived since we added this style update, but because every field a fieldmanager sets must be set on every SSA call, we have update every controller to set every condition instead of the ones it cares about. While this could be done, it is not a drop-in replacement (this PR is) and will need to be phased in if someone wishes to undertake it, meaning this method will remain.

p0lyn0mial:
> Also, a json patch cannot be predicated based on another field. So the patch for conditions would rely on order, which will fail when a condition is added or removed. Recall semantic merge patch isn't supported for CRDs.

okay, so the list would have to be replaced and not merged

> Finally, server-side-apply has arrived since we added this style update, but because every field a fieldmanager sets must be set on every SSA call, we have update every controller to set every condition instead of the ones it cares about. While this could be done, it is not a drop-in replacement (this PR is) and will need to be phased in if someone wishes to undertake it, meaning this method will remain.

+1


> This reduces the

>> Operation cannot be fulfilled on kubeapiservers.operator.openshift.io

> messages in the log. I found 50-100 in runs prior to the update here.

Please help me understand the details of this issue and the provided fix. Does it mean that our caches were/are stale? Does it mean that the update frequency is very high? Finally, does it mean that after the fix, we will issue more requests to the server to get the live object?

deads2k:
> Please help me understand the details of this issue and the provided fix. Does it mean that our caches were/are stale?

Yes, cache was stale.

> Does it mean that the update frequency is very high?

It's high *sometimes*.  During upgrade and configuration changes many controllers try to issue status updates for individual conditions.

> Finally, does it mean that after the fix, we will issue more requests to the server to get the live object?

Yes, but recall that we're comparing repeatedly failing UpdateStatus (usually failing more than once), to a single extra get.  So if we look at the behavior, we're comparing 5 updates to 2 updates and one get.  So it's fewer calls overall.

p0lyn0mial:
ok, let's try this PR out.

/lgtm

PR #1632, opened 2024-01-24
OCPBUGS-24005: when skipping a webhook check because of missing CA log the name of the webhook

p0lyn0mial:
for easier troubleshooting when a check is skipped because `service-ca` hasn't provided a CA for the webook we should also log the name of the webhook.


sanchezl (approved):
/lgtm

PR #1633, opened 2024-01-24
OCPBUGS-27842: Add sno section to alert description

qJkee:
Add additional information to HighOverallControlPlaneCPU alert description regarding SNO clusters.

### DIFF
--- a/bindata/assets/alerts/cpu-utilization.yaml
+++ b/bindata/assets/alerts/cpu-utilization.yaml
@@ -10,15 +10,18 @@ spec:
         - alert: HighOverallControlPlaneCPU
           annotations:
             summary: >-
-              CPU utilization across all three control plane nodes is higher than two control plane nodes can sustain; a single control plane node outage may
-              cause a cascading failure; increase available CPU.
+              CPU utilization across all control plane nodes is more than 60% of the total available CPU. Control plane node outage may cause a cascading failure; increase available CPU.
             runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-kube-apiserver-operator/ExtremelyHighIndividualControlPlaneCPU.md
             description: >-
-              Given three control plane nodes, the overall CPU utilization may only be about 2/3 of all available capacity.
+              On a multi-node cluster with three control plane nodes, the overall CPU utilization may only be about 2/3 of all available capacity.
               This is because if a single control plane node fails, the remaining two must handle the load of the cluster in order to be HA.
-              If the cluster is using more than 2/3 of all capacity, if one control plane node fails, the remaining two are likely to
-              fail when they take the load.
+              If the cluster is using more than 2/3 of all capacity, if one control plane node fails, the remaining two are likely to fail when they take the load.
               To fix this, increase the CPU and memory on your control plane nodes.
+              
+              On a single node OpenShift (SNO) cluster, this alert will also fire if the 2/3 of the CPU cores of the node are in use by any workload. This level of CPU utlization
+              of an SNO cluster is probably not a problem under most circumstances, but high levels of utilization may result in degraded performance.
+              To manage this alert or silence it in case of false positives see the following link: 
+              https://docs.openshift.com/container-platform/latest/monitoring/managing-alerts.html
           expr: |
             sum(
               100 - (avg by (instance) (rate(node_cpu_seconds_total{mode="idle"}[1m])) * 100)


bindata/assets/alerts/cpu-utilization.yaml
```diff
@@ -14,11 +14,15 @@ spec:
...
             description: >-
-              Given three control plane nodes, the overall CPU utilization may only be about 2/3 of all available capacity.
+              On a multi-node cluster with three control plane nodes, the overall CPU utilization may only be about 2/3 of all available capacity.
               This is because if a single control plane node fails, the remaining two must handle the load of the cluster in order to be HA.
-              If the cluster is using more than 2/3 of all capacity, if one control plane node fails, the remaining two are likely to
-              fail when they take the load.
+              If the cluster is using more than 2/3 of all capacity, if one control plane node fails, the remaining two are likely to fail when they take the load.
               To fix this, increase the CPU and memory on your control plane nodes.
+              
+              On a SNO cluster, this alert will also fire if the 2/3 of the CPU cores of on the SNO are in use by any workload. This level of CPU utlization
```
simonpasquier, line 22: (suggestion) define SNO acronym first
```suggestion
              On a single node OpenShift (SNO) cluster, this alert will also fire if the 2/3 of the CPU cores of the node are in use by any workload. This level of CPU utlization
```

bindata/assets/alerts/cpu-utilization.yaml
```diff
@@ -14,11 +14,15 @@ spec:
...
+              of an SNO cluster is probably not a problem under most circumstances, but high levels of utilization may result in degraded performance.
+              To manage this alert and turn off it in case of false positives see the following link: 
+              https://docs.openshift.com/container-platform/4.14/monitoring/managing-alerts.html
```
simonpasquier, line 25: Are we going to update the 4.14 path? One option would be to use `https://docs.openshift.com/container-platform/latest/monitoring/managing-alerts.html` but the page might move in the future.

bindata/assets/alerts/cpu-utilization.yaml
```diff
@@ -14,11 +14,15 @@ spec:
...
+              On a multi-node cluster with three control plane nodes, the overall CPU utilization may only be about 2/3 of all available capacity.
               This is because if a single control plane node fails, the remaining two must handle the load of the cluster in order to be HA.
-              If the cluster is using more than 2/3 of all capacity, if one control plane node fails, the remaining two are likely to
-              fail when they take the load.
+              If the cluster is using more than 2/3 of all capacity, if one control plane node fails, the remaining two are likely to fail when they take the load.
               To fix this, increase the CPU and memory on your control plane nodes.
+              
+              On a SNO cluster, this alert will also fire if the 2/3 of the CPU cores of on the SNO are in use by any workload. This level of CPU utlization
+              of an SNO cluster is probably not a problem under most circumstances, but high levels of utilization may result in degraded performance.
+              To manage this alert and turn off it in case of false positives see the following link:
```
simonpasquier, line 24: "turn off" is a bit vague: are we advising to silence the alert or dropping it with the AlertRelabeling CRD?

simonpasquier (commented):
The summary annotation only needs an update.

bindata/assets/alerts/cpu-utilization.yaml
vrutkovs: I think we want to suggest to silence it

simonpasquier (commented):
I think that the summary annotation needs an update too. It mentions "three control plane nodes".

bindata/assets/alerts/cpu-utilization.yaml
```diff
@@ -11,14 +11,18 @@ spec:
           annotations:
             summary: >-
               CPU utilization across all three control plane nodes is higher than two control plane nodes can sustain; a single control plane node outage may
-              cause a cascading failure; increase available CPU.
+              cause a cascading failure; For Single Node OpenShift (SNO) clusters, monitor CPU usage, as high levels may impact performance. increase available CPU.
```
simonpasquier, line 14: From my POV the summary is a bit lengthy. From https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/alerting-consistency.md:

>  Think of summary as the first line of a commit message, or an email subject line. It should be brief but informative. The description is the longer, more detailed explanation of the alert.

````
CPU utilization across all control plane nodes is more than 60% of the total available CPU.

simonpasquier: sgtm

simonpasquier (commented):
/lgtm

sdodson:
Could we not have an alert that's tuned for SNO instead?

PR #1655, opened 2024-03-12
NO-JIRA: Bump library-go to latest master

vrutkovs:
Bump to pull in https://github.com/openshift/library-go/pull/1689


p0lyn0mial:
is kas-o the only operator we need to bump in order to pick up the fix ?

PR #1659, opened 2024-04-02
OCPBUGS-31384: use RotatedSigningCASecret controller in update only mode

tkashem:
UseSecretUpdateOnly is intended as a short term hack for a very specific use case,
and it works in tandem with a particular carry patch applied to the openshift kube-apiserver.
(https://github.com/openshift/kubernetes/pull/1924)

we will remove this when we migrate all of the affected secret
objects to their intended type: https://issues.redhat.com/browse/API-1800

in short tls secrets used by this operator are reconciled by multiple controllers  at the same time without any coordination. the issue is that the secret's crypto material can be regenerated, which has serious consequences for the platform as it can break external clients and the cluster itself.

xref: https://github.com/openshift/library-go/pull/1705
xref: https://github.com/openshift/kubernetes/pull/1924


pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -144,6 +144,10 @@ func newCertRotationController(
 			Lister:                 kubeInformersForNamespaces.InformersFor(operatorclient.OperatorNamespace).Core().V1().Secrets().Lister(),
 			Client:                 kubeClient.CoreV1(),
 			EventRecorder:          eventRecorder,
+
+			// we will remove this when we migrate all of the affected secret
+			// objects to their intended type: https://issues.redhat.com/browse/API-1800
+			UseSecretUpdateOnly: true,
```
p0lyn0mial, line 150: we should also apply it to the `RotatedSelfSignedCertKeySecret` controller, right ?

p0lyn0mial:
it also requires https://github.com/openshift/kubernetes/pull/1929

p0lyn0mial:
`e2e-gcp-operator` fails on `TestCertRotationStompOnBadType` for which i've opened https://github.com/openshift/kubernetes/pull/1932

in general due to the platform's long history (spanning several years)
and the complexity of ensuring that resources were consistently created with only one type
I think that we should relax the restrictions on allowed type mutation transitions.

p0lyn0mial:
issues like `event happened 4009 times, something is wrong: node/ip-10-0-102-140.us-east-2.compute.internal hmsg/e82226461c - reason/ConfigMapUpdated Updated ConfigMap/kube-control-plane-signer-ca -n openshift-kube-apiserver-operator: result=reject` will be fixed by https://github.com/openshift/cluster-kube-apiserver-operator/pull/1661

p0lyn0mial:
/retest-required


https://github.com/openshift/cluster-kube-apiserver-operator/pull/1661 merge which should unblock all jobs except `e2e-gcp-operator` which will be unblocked by https://github.com/openshift/kubernetes/pull/1932

PR #1652, opened 2024-03-05
NO-JIRA: certrotationcontroller: set AutoRegenerateAfterOfflineExpiry

vrutkovs:
This annotation ensures that the component certificates will be automatically regenerated when the cluster is offline (suspended or shutdown).
https://github.com/openshift/release/pull/49238 will ensure these tests are being run

PR #1661, opened 2024-04-04
NO-ISSUE: Revert "certrotationcontroller: set AutoRegenerateAfterOfflineExpiry for generated certificates

p0lyn0mial:
This reverts commit 6f3faa4beae08550e95f0fc124f49b9e6baca52c introduced in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1652

We have multiple controllers reconciling the same resource at the same time without any coordination.
For example the `kube-control-plane-signer-ca` configmap is synced by 4 controllers (actually it is 8 because we have 2 processes).

Since the mentioned PR added a distinct annotation (`AutoRegenerateAfterOfflineExpiry`) for the same configmap it caused a hot update loop since the configmap was different.

The conflict error can be seen [even on a successful run ](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-kube-apiserver-operator/1577/pull-ci-openshift-cluster-kube-apiserver-operator-master-e2e-aws-ovn/1771293936002797568) but usually it causes [the CI jobs to fail.](https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-cluster-kube-apiserver-operator-master-e2e-aws-ovn).

Since [the issue is common](https://search.dptools.openshift.org/?search=kube-control-plane-signer-ca&maxAge=48h&context=1&type=bug%2Bissue%2Bjunit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job) I decided to revert as it might even block CI payloads from merging. 

It should also unblock https://github.com/openshift/cluster-kube-apiserver-operator/pull/1659

We should stop adding more changes to the cert rotation controllers until we resolve the race.

### DIFF
--- a/pkg/operator/certrotationcontroller/certrotationcontroller.go
+++ b/pkg/operator/certrotationcontroller/certrotationcontroller.go
@@ -134,8 +134,7 @@ func newCertRotationController(
 			Namespace: operatorclient.OperatorNamespace,
 			Name:      "aggregator-client-signer",
 			AdditionalAnnotations: certrotation.AdditionalAnnotations{
-				JiraComponent:                    "kube-apiserver",
-				AutoRegenerateAfterOfflineExpiry: "https://github.com/openshift/cluster-kube-apiserver-operator/pull/1631,'operator conditions openshift-apiserver'",
+				JiraComponent: "kube-apiserver",
 			},
 			Validity:               30 * rotationDay,
 			Refresh:                15 * rotationDay,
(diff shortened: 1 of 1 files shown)


p0lyn0mial:
`ci/prow/k8s-e2e-gcp` and `ci/prow/e2e-gcp-operator` failed on cluster installation, logs are missing and it looks like loki wasn't set up either.

PR #1664, opened 2024-04-11
WRKLDS-1015: tolerate node-role.kubernetes.io/control-plane:NoExecute

knelasevero:
Required to be able to advise customers to follow workflow described in https://github.com/openshift/enhancements/pull/1583. If they want to taint control-plane nodes to avoid any other thing to run there (while enforcing it further with validating admission policies), control plane workloads need to tolerate that taint.


p0lyn0mial:
have you forgot to apply the toleration to `kube-apiserver`'s manifest ?

p0lyn0mial:
> have you forgot to apply the toleration to kube-apiserver's manifest ?

@ingvagabund pointed me to https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/bindata/assets/kube-apiserver/pod.yaml#L268-L269 which means that `kas` tolerates all taints.

/lgtm
/approve

PR #1665, opened 2024-04-12
NO-JIRA: Re-apply "certrotationcontroller: set AutoRegenerateAfterOfflineExpiry for generated certificates"

vrutkovs:
Reverts openshift/cluster-kube-apiserver-operator#1661

https://github.com/openshift/cluster-kube-apiserver-operator/pull/1721 has landed, so the controllers should no longer hotloop


dinhxuanvu (approved):
/lgtm

PR #1666, opened 2024-04-15
bump(k8s): update k8s version to v1.30.0

dinhxuanvu:
Pinned openshift/api, library-go, client-go changes


soltysh:
/hold
this should not merge into 4.16 accidentally

pkg/operator/configobservation/apienablement/observe_runtime_config.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 var DefaultGroupVersionsByFeatureGate = map[configv1.FeatureGateName][]schema.GroupVersion{
-	"ValidatingAdmissionPolicy": {{Group: "admissionregistration.k8s.io", Version: "v1beta1"}},
+	"ValidatingAdmissionPolicy": {{Group: "admissionregistration.k8s.io", Version: "v1"}},
```
benluddy, line 18: You can remove this entirely, we don't need to pass `--runtime-config` because this group/version will be enabled by default.

soltysh (approved):
/lgtm

PR #1673, opened 2024-05-06
[release-4.15] OCPBUGS-33315: add SNO control plane high cpu usage alert

qJkee:
Create separate SNO alert for control plane high cpu usage. This alert is also aware of workload partitioning and adjusts threshold if workload partitioning is enabled.

(cherry picked from commit d92b46166259ef43918e9f7cd6bc62f09a806d1f)

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -33,6 +33,7 @@ import (
 	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/targetconfigcontroller"
 	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/terminationobserver"
 	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/webhooksupportabilitycontroller"
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/workloadpartitioning"
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
 	"github.com/openshift/library-go/pkg/operator/apiserver/controller/auditpolicy"
 	"github.com/openshift/library-go/pkg/operator/certrotation"
@@ -200,7 +201,6 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 			"assets/kube-apiserver/storage-version-migration-prioritylevelconfiguration.yaml",
 			"assets/alerts/api-usage.yaml",
 			"assets/alerts/audit-errors.yaml",
-			"assets/alerts/cpu-utilization.yaml",
 			"assets/alerts/kube-apiserver-requests.yaml",
 			"assets/alerts/kube-apiserver-slos-basic.yaml",
 			"assets/alerts/podsecurity-violations.yaml",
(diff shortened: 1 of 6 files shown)


wangke19:
/hold
Pre-merge testing.

wangke19:
The associated bug has been pre-merge tested, got the expected results, detail see https://issues.redhat.com/browse/OCPBUGS-33315?focusedId=24685806&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-24685806

pkg/operator/starter.go
```diff
@@ -213,10 +213,15 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		operatorClient,
 		controllerContext.EventRecorder,
 	).
-		WithConditionalResources(bindata.Asset, []string{"assets/alerts/kube-apiserver-slos-extended.yaml"}, notOnSingleReplicaTopology, nil).
+		WithConditionalResources(bindata.Asset, []string{"assets/alerts/kube-apiserver-slos-extended.yaml", "assets/alerts/cpu-utilization.yaml"}, notOnSingleReplicaTopology, nil).
 		WithConditionalResources(bindata.Asset, []string{"assets/alerts/kube-apiserver-slos.yaml"}, never, nil). // TODO remove in 4.13
 		AddKubeInformers(kubeInformersForNamespaces)
 
+	err = workloadpartitioning.CreateSNOAlert(ctx, dynamicClient, infrastructure.Status, controllerContext.EventRecorder)
```
p0lyn0mial, line 220: why hasn't it be implemented as a controller ?
/cc @dgrisonnet (since you reviewed the original pr)

p0lyn0mial, line 220: why wasn’t `infrastructure.Status` taken from a lister, but only directly from the server?

p0lyn0mial: /cc @dgrisonnet

p0lyn0mial: I think it should be a controller because it would allow you to reconcile any potential changes to that resource and it would prevent the operator from exiting on error. Does it make sense ?

p0lyn0mial: I see, using the client directly is problematic because on error it will take the whole operator down :(

p0lyn0mial: I see two issues here. The first is that the alert can be changed, and then the controller should repair such a resource. The second is that we shouldn’t send direct requests to the server but rather read data from the cache.
I’m sorry but after an internal discussion with the team, we decided that we should revert the original PR.

/cc @tkashem @vrutkovs @benluddy

vrutkovs: Yeah, definitely not the best idea.

`StaticResourceController`, which should be used there,  can template manifests and watches the resulting reosurces

vrutkovs:
/hold

We should have spotted these issues in 4.16 PR, but at least lets not break 4.15

pkg/operator/starter.go
p0lyn0mial: @vrutkovs I don't see how the [StaticResourceController](https://github.com/openshift/cluster-kube-apiserver-operator/blob/d597184dccf293101d38e543587d40af99aca52a/vendor/github.com/openshift/library-go/pkg/operator/staticresourcecontroller/static_resource_controller.go#L96) could be used to change the resource. It would be nice if you could share an example.

p0lyn0mial:
/hold

until a new PR will be merged (xref: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1674)

pkg/operator/starter.go
vrutkovs: Easiest way is to [replace the variable in the asset](https://github.com/openshift/cluster-csi-snapshot-controller-operator/blob/66048040bb40915f0903a0e7de0bbc6f4b4faef5/pkg/operator/starter.go#L165) and

tkashem: given that the code freeze is imminent soon, my recommendation is to revert this PR and then we open a new PR that achieves the same goal using a controller pattern

PR #1675, opened 2024-05-10
OCPBUGS-33522: add a controller that reconciles SCCs' volumes

stlaz:
The storage team added "ephemeral" and "csi" volumes to all SCCs
somewhere around OCP version 4.13. However, since the past us set
SCCs as "create-only" (due to the legacy permission system),
meaning the CVO won't reconcile the SCCs, this change
never made it to upgraded clusters' SCCs. People that lack SCC update
privileges were installing odd operators they found on the internets
in hope to fix their SCC issues.

Introduce a controller that adds these volumes to all SCCs that the
CVO does not automatically reconcile.

/assign @deads2k 
/cc @openshift/openshift-team-auth-maintainers


pkg/operator/sccreconcilecontroller/sccreconcilecontroller.go
```diff
@@ -0,0 +1,79 @@
...
+		ResyncEvery(5*time.Minute).
+		WithFilteredEventsInformersQueueKeyFunc(
+			factory.ObjectNameToKey,
+			factory.NamesFilter(createOnlySCCs...),
+			sccInformer.Informer(),
+		).
+		ToController("SCCReconcileController", recorder.WithComponentSuffix("scc-reconcile-controller")), nil
+}
+
+func (c *sccReconcileController) sync(ctx context.Context, controllerContext factory.SyncContext) error {
```
deads2k, line 61: great work in this function.  Details for other readers to observe
1. zero apiserver access until work is necessary, so zero-http requests in steady state
2. zero memory copy until we know modification is needed.
3. copy before mutation

deads2k:
/lgtm
/approve
/hold

holding for other auth team members to review if they like.  @stlaz unhold at your discretion.

PR #1676, opened 2024-05-10
OCPEDGE-902: add SNO control plane high cpu usage alert

qJkee:
Create separate SNO alert for control plane high cpu usage. This alert is also aware of workload partitioning and adjusts the threshold if workload partitioning is enabled.

This PR contains improvements suggested in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1673

### DIFF
--- a/bindata/assets/alerts/cpu-utilization-sno.yaml
+++ b/bindata/assets/alerts/cpu-utilization-sno.yaml
@@ -1,0 +1,47 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: cpu-utilization
+  namespace: openshift-kube-apiserver
+spec:
+  groups:
+    - name: control-plane-cpu-utilization
+      rules:
+        - alert: HighOverallControlPlaneCPU
+          annotations:
+            summary: >-
+              CPU utilization across control plane pods is more than 60% of total CPU. High CPU usage usually means that something goes wrong.
+            runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-kube-apiserver-operator/ExtremelyHighIndividualControlPlaneCPU.md
+            description: >-
+              This level of CPU utlization of an control plane is probably not a problem under most circumstances, but high levels of utilization may indicate
+              problems with cluster or control plane pods. To manage this alert or modify threshold it in case of false positives see the following link:
+              https://docs.openshift.com/container-platform/latest/monitoring/managing-alerts.html
+          expr: |
+            sum(rate(container_cpu_usage_seconds_total{namespace=~"openshift-.*",image!=""}[4m])) / ${CPU-COUNT} * 100 > 60
+          for: 10m
+          labels:
+            namespace: openshift-kube-apiserver
+            severity: warning
+        - alert: ExtremelyHighIndividualControlPlaneCPU
+          annotations:
+            summary: >-
+              CPU utilization across control plane pods is more than 90% of total CPU. High CPU usage usually means that something goes wrong.
+            runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-kube-apiserver-operator/ExtremelyHighIndividualControlPlaneCPU.md
+            description: >-
+              This level of CPU utlization of an control plane is probably not a problem under most circumstances, but high levels of utilization may indicate
+              problems with cluster or control plane pods. When workload partitioning is enabled,
+              Extreme CPU pressure can cause slow serialization and poor performance from the kube-apiserver and etcd.
+              When this happens, there is a risk of clients seeing non-responsive API requests which are issued again
+              causing even more CPU pressure.
+              It can also cause failing liveness probes due to slow etcd responsiveness on the backend.
+              If one kube-apiserver fails under this condition, chances are you will experience a cascade as the remaining
+              kube-apiservers are also under-provisioned.
+              To fix this, increase the CPU and memory on your control plane nodes.
+              To manage this alert or modify threshold it in case of false positives see the following link: 
+              https://docs.openshift.com/container-platform/latest/monitoring/managing-alerts.html
+          expr: |
+            sum(rate(container_cpu_usage_seconds_total{namespace=~"openshift-.*",image!=""}[4m])) / ${CPU-COUNT} * 100 > 90
+          for: 1h
+          labels:
+            namespace: openshift-kube-apiserver
+            severity: critical
--- a/pkg/operator/highcpuusagealertcontroller/highcpuusagealert_controller.go
+++ b/pkg/operator/highcpuusagealertcontroller/highcpuusagealert_controller.go
@@ -1,0 +1,166 @@
+package highcpuusagealertcontroller
+
+import (
+	"bytes"
+	"context"
+	"strconv"
+	"time"
+
+	configv1 "github.com/openshift/api/config/v1"
+	configv1informers "github.com/openshift/client-go/config/informers/externalversions/config/v1"
+	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
+	"github.com/openshift/cluster-kube-apiserver-operator/bindata"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/dynamic/dynamicinformer"
+	"k8s.io/utils/cpuset"
+)
+
+// default and taken from the docs
+const defaultCoresNum = 8
+
+var performanceGroup = schema.GroupVersionResource{Group: "performance.openshift.io", Version: "v2", Resource: "performanceprofiles"}
+
+type highCPUUsageAlertController struct {
+	client               dynamic.Interface
+	infraLister          configlistersv1.InfrastructureLister
+	clusterVersionLister configlistersv1.ClusterVersionLister
+}
+
+func NewHighCPUUsageAlertController(
+	configInformer configv1informers.Interface,
+	dynamicInformersForTargetNamespace dynamicinformer.DynamicSharedInformerFactory,
+	client dynamic.Interface,
+	recorder events.Recorder,
+) factory.Controller {
+	c := &highCPUUsageAlertController{
+		client:               client,
+		infraLister:          configInformer.Infrastructures().Lister(),
+		clusterVersionLister: configInformer.ClusterVersions().Lister(),
+	}
+
+	prometheusAlertInformerForTargetNamespace := dynamicInformersForTargetNamespace.ForResource(schema.GroupVersionResource{
+		Group:    "monitoring.coreos.com",
+		Version:  "v1",
+		Resource: "prometheusrules",
+	})
+
+	return factory.New().
+		WithInformers(configInformer.Infrastructures().Informer(), configInformer.ClusterVersions().Informer(), prometheusAlertInformerForTargetNamespace.Informer()).
+		WithSync(c.sync).ResyncEvery(10*time.Minute).
+		ToController("highCPUUsageAlertController", recorder.WithComponentSuffix("high-cpu-usage-alert-controller"))
+}
+
+func (c *highCPUUsageAlertController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	infra, err := c.infraLister.Get("cluster")
+	if err != nil {
+		return err
+	}
+
+	var alertRaw []byte
+
+	if infra.Status.InfrastructureTopology != configv1.SingleReplicaTopologyMode {
+		// we moved creation of the alert here because the static resource controller was constantly
+		// deleting the alert and was fighting with this controller
+		alertRaw, err = bindata.Asset("assets/alerts/cpu-utilization.yaml")
+		if err != nil {
+			return err
+		}
+	} else {
+		clusterVersion, err := c.clusterVersionLister.Get("version")
+		if err != nil {
+			return err
+		}
+
+		alertRaw, err = snoAlert(ctx, c.client, clusterVersion.Status.Capabilities.EnabledCapabilities, infra.Status.CPUPartitioning)
+		if err != nil {
+			return err
+		}
+	}
+
+	alertObj, err := resourceread.ReadGenericWithUnstructured(alertRaw)
+	if err != nil {
+		return err
+	}
+
+	_, _, err = resourceapply.ApplyPrometheusRule(ctx, c.client, syncCtx.Recorder(), alertObj.(*unstructured.Unstructured))
+	return err
+}
+
+func snoAlert(ctx context.Context, client dynamic.Interface, enabledCapabilities []configv1.ClusterVersionCapability, cpuMode configv1.CPUPartitioningMode) ([]byte, error) {
+	cores := defaultCoresNum
+
+	// if NodeTuning capability disabled, there are no PerformanceProfile, so we proceed
+	// with default value.
+	if sets.New(enabledCapabilities...).Has(configv1.ClusterVersionCapabilityNodeTuning) && cpuMode == configv1.CPUPartitioningAllNodes {
+		foundCores, found, err := performanceProfileControlPlaneCores(ctx, client)
+		if err != nil {
+			return nil, err
+		}
+		// set cores from PerformanceProfile if expectedToFindCores
+		// if not, proceed with default values
+		if found {
+			cores = foundCores
+		}
+	}
+
+	fileData, err := bindata.Asset("assets/alerts/cpu-utilization-sno.yaml")
+	if err != nil {
+		return nil, err
+	}
+	fileData = bytes.ReplaceAll(fileData, []byte(`${CPU-COUNT}`), []byte(strconv.Itoa(cores)))
+
+	return fileData, nil
+}
+
+// performanceProfileControlPlaneCores returns cores allocated for control plane pods via
+// PerformanceProfile object. Bool value indicates if PerformanceProfile is expectedToFindCores for master node
+func performanceProfileControlPlaneCores(ctx context.Context, client dynamic.Interface) (int, bool, error) {
+	// fetch resource directly instead of using an informer because
+	// NodeTuning capability can be disabled at start and enabled later
+	obj, err := client.Resource(performanceGroup).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return 0, false, err
+	}
+
+	for _, pf := range obj.Items {
+		nodeSelector, found, err := unstructured.NestedStringMap(pf.Object, "spec", "nodeSelector")
+		if err != nil {
+			return 0, false, err
+		}
+		if !found {
+			continue
+		}
+		if _, ok := nodeSelector["node-role.kubernetes.io/master"]; !ok {
+			continue
+		}
+
+		reservedCPU, found, err := unstructured.NestedString(pf.Object, "spec", "cpu", "reserved")
+		if err != nil {
+			return 0, false, err
+		}
+		if !found {
+			continue
+		}
+
+		cores, err := coresInCPUSet(reservedCPU)
+		if err != nil {
+			return 0, false, err
+		}
+		return cores, true, nil
+	}
+
+	return 0, false, nil
+}
+
+func coresInCPUSet(set string) (int, error) {
+	cpuMap, err := cpuset.Parse(set)
+	return cpuMap.Size(), err
+}
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -25,6 +25,7 @@ import (
 	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/configobservation/configobservercontroller"
 	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/configobservation/node"
 	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/connectivitycheckcontroller"
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/highcpuusagealertcontroller"
 	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/kubeletversionskewcontroller"
 	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/nodekubeconfigcontroller"
 	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/operatorclient"
@@ -66,6 +67,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/dynamic/dynamicinformer"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/ptr"
@@ -115,7 +117,7 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		"openshift-apiserver",
 	)
 	configInformers := configv1informers.NewSharedInformerFactory(configClient, 10*time.Minute)
-	operatorClient, dynamicInformers, err := genericoperatorclient.NewStaticPodOperatorClient(controllerContext.KubeConfig, operatorv1.GroupVersion.WithResource("kubeapiservers"))
+	operatorClient, dynamicInformersForAllNamespaces, err := genericoperatorclient.NewStaticPodOperatorClient(controllerContext.KubeConfig, operatorv1.GroupVersion.WithResource("kubeapiservers"))
 	if err != nil {
 		return err
 	}
@@ -175,6 +177,8 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		WithEventHandler(operatorclient.TargetNamespace, "LateConnections", terminationobserver.ProcessLateConnectionEvents).
 		ToController(kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace), kubeClient.CoreV1(), controllerContext.EventRecorder)
 
+	// TODO: use informer instead of direct api call
+	// Also, in the future there is a plan to make infrastructure type dynamic
 	infrastructure, err := configClient.ConfigV1().Infrastructures().Get(ctx, "cluster", metav1.GetOptions{})
 	if err != nil {
 		return err
@@ -182,6 +186,7 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 	var notOnSingleReplicaTopology resourceapply.ConditionalFunction = func() bool {
 		return infrastructure.Status.ControlPlaneTopology != configv1.SingleReplicaTopologyMode
 	}
+
 	staticResourceController := staticresourcecontroller.NewStaticResourceController(
 		"KubeAPIServerStaticResources",
 		bindata.Asset,
@@ -208,7 +213,6 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 			"assets/kube-apiserver/storage-version-migration-prioritylevelconfiguration.yaml",
 			"assets/alerts/api-usage.yaml",
 			"assets/alerts/audit-errors.yaml",
-			"assets/alerts/cpu-utilization.yaml",
 			"assets/alerts/kube-apiserver-requests.yaml",
 			"assets/alerts/kube-apiserver-slos-basic.yaml",
 			"assets/alerts/podsecurity-violations.yaml",
@@ -223,6 +227,14 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 	).
 		WithConditionalResources(bindata.Asset, []string{"assets/alerts/kube-apiserver-slos-extended.yaml"}, notOnSingleReplicaTopology, nil).
 		AddKubeInformers(kubeInformersForNamespaces)
+
+	dynamicInformersForTargetNamespace := dynamicinformer.NewFilteredDynamicSharedInformerFactory(dynamicClient, 12*time.Hour, operatorclient.TargetNamespace, nil)
+
+	highCpuUsageAlertController := highcpuusagealertcontroller.NewHighCPUUsageAlertController(
+		configInformers.Config().V1(),
+		dynamicInformersForTargetNamespace,
+		dynamicClient,
+		controllerContext.EventRecorder)
 
 	targetConfigReconciler := targetconfigcontroller.NewTargetConfigController(
 		os.Getenv("IMAGE"),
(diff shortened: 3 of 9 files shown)


pkg/operator/highcpuusagealertcontroller/performanceprofile.go
```diff
@@ -0,0 +1,66 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/utils/cpuset"
+)
+
+// default and taken from the docs
+const defaultCoresNum = 8
+
+var performanceGroup = schema.GroupVersionResource{Group: "performance.openshift.io", Version: "v2", Resource: "performanceprofiles"}
```
p0lyn0mial, line 16: Will this API/GVK be always available on an SNO platform ?

Why aren’t the performance profiles part of openshift/api?

p0lyn0mial (commented):
I will do a deeper review later, first I would like to sort out which client we should use for performanceprofiles.

pkg/operator/highcpuusagealertcontroller/performanceprofile.go
p0lyn0mial: > Why aren’t the performance profiles part of openshift/api?

ok, from the internal discussion we had on Slack, the profile should be part of the openshift/api since it is a public API (used by users). It is a technical debt, and will eventually be moved there.

p0lyn0mial: Will the [CRD](https://github.com/openshift/cluster-node-tuning-operator/blob/master/manifests/20-performance-profile.crd.yaml) be installed on an `SNO` cluster when the `cpuPartitioningMode` is set to `AllNodes`. Or could it happen that the API/CRD won't be available at all?

p0lyn0mial: When is the `cluster-node-tuning-operator` actually installed on the cluster?

p0lyn0mial: Is the operator installed by CVO or manually be end users ?

p0lyn0mial: I've found https://github.com/openshift/enhancements/blob/master/enhancements/workload-partitioning/wide-availability-workload-partitioning.md#workflow-description which indicates that the operator and the API/CRD is actually installed/created during cluster creation.

pkg/operator/highcpuusagealertcontroller/performanceprofile.go
```diff
@@ -0,0 +1,66 @@
...
+	Spec struct {
+		CPU *struct {
+			Reserved *string `json:"reserved"`
+		} `json:"cpu"`
+		NodeSelector map[string]string `json:"nodeSelector"`
+	} `json:"spec"`
+}
+
+func controlPlaneCores(ctx context.Context, client dynamic.Interface) (int, error) {
+	obj, err := client.Resource(performanceGroup).List(ctx, metav1.ListOptions{})
```
p0lyn0mial, line 32: Okay, after talking with the team, we are okay with using a dynamic client. That also means we should list resources from a lister, not directly from the API. So, you will have to start an informer and plug it into this controller.

pkg/operator/starter.go
```diff
@@ -224,6 +226,14 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		WithConditionalResources(bindata.Asset, []string{"assets/alerts/kube-apiserver-slos-extended.yaml"}, notOnSingleReplicaTopology, nil).
 		AddKubeInformers(kubeInformersForNamespaces)
 
+	performanceProfileInformerFactory := highcpuusagealertcontroller.PerformanceProfileDynamicInformer(dynamicClient)
```
p0lyn0mial, line 229: we already have an instance of `dynamicInformers` defined for this operator. Could we use it instead of creating a new one ?

pkg/operator/highcpuusagealertcontroller/cpuPartitioningAlertRenderer.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+type HighCPUUsageController struct {
+	client                   dynamic.Interface
+	infraLister              configlistersv1.InfrastructureLister
+	performanceProfileLister cache.GenericLister
+}
+
+func NewHighCPUUsageAlertController(infraInformer configv1informers.InfrastructureInformer, client dynamic.Interface, genericInformer dynamicinformer.DynamicSharedInformerFactory, recorder events.Recorder) factory.Controller {
+	performanceInformer := genericInformer.ForResource(performanceGroup)
```
p0lyn0mial, line 31: In the previous comment you said the the informer can be disabled by setting a capability.
Thus we should make this informer optional.
It should be not created when the capability was set.

pkg/operator/highcpuusagealertcontroller/cpuPartitioningAlertRenderer.go
```diff
@@ -0,0 +1,93 @@
+package highcpuusagealertcontroller
```
p0lyn0mial, line 1: Please check the file naming convention used for controllers in this operator and change the file name.

pkg/operator/highcpuusagealertcontroller/cpuPartitioningAlertRenderer.go
```diff
@@ -0,0 +1,93 @@
...
+	"k8s.io/client-go/dynamic/dynamicinformer"
+	"k8s.io/client-go/tools/cache"
+
+	v1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/cluster-kube-apiserver-operator/bindata"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"k8s.io/client-go/dynamic"
+)
+
+type HighCPUUsageController struct {
```
p0lyn0mial, line 24: make this type private.

bindata/assets/alerts/cpu-utilization-sno.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: monitoring.coreos.com/v1
```
p0lyn0mial, line 1: I have noticed that we already have `cpu-utilization.yaml` alert [defined](https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/bindata/assets/alerts/cpu-utilization.yaml).
Is there a way we could reuse it instead of creating a new one ?

p0lyn0mial: I have also noticed that `cpu-utilization.yaml` is not used by this operator, is that correct ?
I have also noticed that you [changed](https://github.com/openshift/cluster-kube-apiserver-operator/commit/29d75eafb79519cf468c6bbc8b4568ed7468be29) the alert to describe the behaviour on an SNO cluster.

I'm confused, why do we need `2` separate alerts ?
/cc @dgrisonnet

p0lyn0mial (commented):
I have stopped reviewing because I would like to understand whey do we have `2` separate alerts for high cpu utilisation first.

pkg/operator/highcpuusagealertcontroller/performanceprofile.go
p0lyn0mial: @qJkee could you please also create a bug/story for moving `performanceprofiles` to `openshift/api` repo ?

bindata/assets/alerts/cpu-utilization-sno.yaml
p0lyn0mial: I'm not sure if we can, because before we merge this PR we need to make sure that `cpu-utilization.yaml` will not be created on an SNO cluster. Also, after this change your controller will be reconciling the alter only on SNO clusters. You can work on both PRs at the same time.

pkg/operator/highcpuusagealertcontroller/cpuPartitioningAlertRenderer.go
p0lyn0mial: OK, so after our discussion this morning about the capability being enabled at runtime, and after taking more time to think it through and weighing some trade-offs, I think we could:

- ensure the alert is reconciled every 5 or 10 minutes.
- check whether the capability is enabled or disabled.
- if the capability is enabled, list data directly from the server (as you previously suggested).
- if the capability is disabled, use the default value.
- document the previous 3 steps
- reconcile the alert.

WDYT ?

p0lyn0mial: > Do I hear it right that are suggesting to use dynamic client instead of dynamic informer for listing 

yeah, but we should check if the capability is enabled (from the lister).

the other idea I had (other than creating the controller/informer dynamically) would be to terminate the operator when the capability was enabled at runtime - but that would be more complicated.

please document why we have decided to use the dynamic client.

pkg/operator/highcpuusagealertcontroller/performanceprofile.go
```diff
@@ -0,0 +1,69 @@
...
+type performanceProfile struct {
+	Spec struct {
+		CPU *struct {
+			Reserved *string `json:"reserved"`
+		} `json:"cpu"`
+		NodeSelector map[string]string `json:"nodeSelector"`
+	} `json:"spec"`
+}
+
+func performanceProfileControlPlaneCores(ctx context.Context, cl dynamic.Interface) (int, error) {
```
p0lyn0mial, line 32: you could change the signature of this method to return a boolean to know if the value was found or not so we could apply the default value.

pkg/operator/highcpuusagealertcontroller/highcpuusagealert_controller.go
```diff
@@ -0,0 +1,91 @@
...
+	alertObj, err := resourceread.ReadGenericWithUnstructured(alertRaw)
+	if err != nil {
+		return err
+	}
+
+	_, _, err = resourceapply.ApplyPrometheusRule(ctx, c.client, syncCtx.Recorder(), alertObj.(*unstructured.Unstructured))
+	return err
+}
+
+func (c *highCPUUsageController) snoAlert(ctx context.Context, cpuMode v1.CPUPartitioningMode) ([]byte, error) {
```
p0lyn0mial, line 66: you could pass `EnabledCapabilities` to this function for easier unit testing.

pkg/operator/highcpuusagealertcontroller/highcpuusagealert_controller.go
```diff
@@ -0,0 +1,91 @@
...
+	infra, err := c.infraLister.Get("cluster")
+	if err != nil {
+		return err
+	}
+
+	// we care only about SNO clusters
+	if infra.Status.InfrastructureTopology != v1.SingleReplicaTopologyMode {
+		return nil
+	}
+
```
p0lyn0mial, line 51: By passing `EnabledCapabilities` to `snoAlert` we would have all preconditions listed and checked before reconciling the alert.

pkg/operator/highcpuusagealertcontroller/performanceprofile.go
```diff
@@ -0,0 +1,69 @@
...
+	// fetch resource directly instead of using an informer because
+	// NodeTuning capability can be disabled at start and enabled later
+	obj, err := cl.Resource(performanceGroup).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return defaultCoresNum, nil
```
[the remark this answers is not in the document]
p0lyn0mial: worth checking but I think that an empty result is different from `NotFound` error.

pkg/operator/highcpuusagealertcontroller/highcpuusagealert_controller.go
```diff
@@ -0,0 +1,91 @@
...
+func NewHighCPUUsageAlertController(configInformer configv1informers.Interface, client dynamic.Interface, recorder events.Recorder) factory.Controller {
+	c := &highCPUUsageController{
+		client:               client,
+		infraLister:          configInformer.Infrastructures().Lister(),
+		clusterversionLister: configInformer.ClusterVersions().Lister(),
+	}
+
+	return factory.New().
+		WithInformers(configInformer.Infrastructures().Informer(), configInformer.ClusterVersions().Informer()).
```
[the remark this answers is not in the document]
p0lyn0mial: why could use a typed client since `prometheusrules` is served by an aggregated API, WDYT?

pkg/operator/highcpuusagesnoalertcontroller/performanceprofile_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+func Test_performanceProfileControlPlaneCores(t *testing.T) {
+	type args struct {
+		ctx context.Context
+		cl  dynamic.Interface
+	}
+	tests := []struct {
+		name       string
+		args       args
+		coresFound int
```
p0lyn0mial, line 24: this is our expectation, please rename to `expectedCores` or `expectedCoreCount`

pkg/operator/highcpuusagesnoalertcontroller/performanceprofile_test.go
```diff
@@ -0,0 +1,114 @@
...
+		found      bool
```
p0lyn0mial, line 25: this could be `expectToFindCores`

pkg/operator/highcpuusagesnoalertcontroller/performanceprofile_test.go
```diff
@@ -0,0 +1,114 @@
...
+		wantErr    bool
```
p0lyn0mial, line 26: and for symmetry we could rename this field to `expectErr`

pkg/operator/highcpuusagesnoalertcontroller/performanceprofile.go
```diff
@@ -0,0 +1,65 @@
...
+		if errors.IsNotFound(err) {
+			return 0, false, nil
+		}
+		return 0, false, err
+	}
+
+	for _, pf := range obj.Items {
+		nodeSelector, found, err := unstructured.NestedStringMap(pf.Object, "spec", "nodeSelector")
+		if err != nil {
+			return 0, false, err
```
p0lyn0mial, line 35: please cover this path in a unit test.

pkg/operator/highcpuusagesnoalertcontroller/performanceprofile.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+		if !found {
+			continue
```
p0lyn0mial, line 38: please cover this path in a unit test.

pkg/operator/highcpuusagesnoalertcontroller/performanceprofile.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+		if _, ok := nodeSelector["node-role.kubernetes.io/master"]; !ok {
+			continue
+		}
+
+		reservedCPU, found, err := unstructured.NestedString(pf.Object, "spec", "cpu", "reserved")
+		if err != nil {
+			return 0, false, err
```
p0lyn0mial, line 46: please cover this path in a unit test.

pkg/operator/highcpuusagesnoalertcontroller/performanceprofile.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+		if !found {
+			continue
```
p0lyn0mial, line 49: please cover this path in a unit test.

pkg/operator/highcpuusagesnoalertcontroller/performanceprofile.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+
+		cores, err := coresInCPUSet(reservedCPU)
+		if err != nil {
+			return 0, false, err
```
p0lyn0mial, line 54: please cover this path in a unit test.

pkg/operator/highcpuusagesnoalertcontroller/highcpuusagesnoalert_controller_test.go
```diff
@@ -0,0 +1,90 @@
...
+			}
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("snoAlert() got = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+func alertWithGivenCores(cores int) []byte {
+	fileData, err := bindata.Asset("assets/alerts/cpu-utilization-sno.yaml")
```
p0lyn0mial, line 85: we usually compare against pre-computed files, for example https://github.com/openshift/cluster-authentication-operator/blob/master/pkg/operator/workload/sync_openshift_oauth_apiserver_test.go#L71

pkg/operator/starter.go
```diff
@@ -175,13 +176,20 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		WithEventHandler(operatorclient.TargetNamespace, "LateConnections", terminationobserver.ProcessLateConnectionEvents).
 		ToController(kubeInformersForNamespaces.InformersFor(operatorclient.TargetNamespace), kubeClient.CoreV1(), controllerContext.EventRecorder)
 
-	infrastructure, err := configClient.ConfigV1().Infrastructures().Get(ctx, "cluster", metav1.GetOptions{})
-	if err != nil {
-		return err
-	}
 	var notOnSingleReplicaTopology resourceapply.ConditionalFunction = func() bool {
-		return infrastructure.Status.ControlPlaneTopology != configv1.SingleReplicaTopologyMode
```
p0lyn0mial, line 183: let's not change this function in this PR, we cannot hide errors, we need a better strategy.

pkg/operator/highcpuusagesnoalertcontroller/highcpuusagesnoalert_controller.go
```diff
@@ -0,0 +1,104 @@
...
+	client dynamic.Interface,
+	recorder events.Recorder,
+) factory.Controller {
+	c := &highCPUUsageSNOController{
+		client:               client,
+		infraLister:          configInformer.Infrastructures().Lister(),
+		clusterVersionLister: configInformer.ClusterVersions().Lister(),
+	}
+
+	prometheusAlertInformer := dynamicInformer.ForResource(schema.GroupVersionResource{Group: "monitoring.coreos.com", Version: "v1", Resource: "prometheusrules"})
```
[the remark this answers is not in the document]
p0lyn0mial: okay, since `resourceapply.ApplyPrometheusRule` doesn't work with a typed client i'm actually okay with using a dynamic informer for symmetry. 

I think that we should create the informer only for `openshift-kube-apiserver` namespace.

bindata/assets/alerts/cpu-utilization-sno.yaml
```diff
@@ -0,0 +1,47 @@
...
+              This level of CPU utlization of an control plane is probably not a problem under most circumstances, but high levels of utilization may indicate
+              problems with cluster or control plane pods. To manage this alert or modify threshold it in case of false positives see the following link:
+              https://docs.openshift.com/container-platform/latest/monitoring/managing-alerts.html
+          expr: |
+            sum(rate(container_cpu_usage_seconds_total{namespace=~"openshift-.*",image!=""}[5m])) / ${CPU-COUNT} * 100 > 60
+          for: 10m
+          labels:
+            namespace: openshift-kube-apiserver
+            severity: warning
+        - alert: ExtremelyHighIndividualControlPlaneCPU
```
wangke19, line 25: As I known we have [this alert from 4.10](https://issues.redhat.com/browse/API-1363), there is one case https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-47633 for this, it is special for SNO?

pkg/operator/highcpuusagesnoalertcontroller/performanceprofile.go
```diff
@@ -0,0 +1,65 @@
...
+
+// performanceProfileControlPlaneCores returns cores allocated for control plane pods via
+// PerformanceProfile object. Bool value indicates if PerformanceProfile is found for master node
+func performanceProfileControlPlaneCores(ctx context.Context, cl dynamic.Interface) (int, bool, error) {
+	// fetch resource directly instead of using an informer because
+	// NodeTuning capability can be disabled at start and enabled later
+	obj, err := cl.Resource(performanceGroup).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return 0, false, nil
```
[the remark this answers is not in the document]
p0lyn0mial: if the api exists and there are no performance profiles created we should get an empty list not an error.

pkg/operator/highcpuusagesnoalertcontroller/performanceprofile.go
p0lyn0mial: since we use a dynamic client, I just want to ensure the code won't break when something changes. As you have already prepared the unit tests, adding a new test case with modified input won't take long.

pkg/operator/starter.go
p0lyn0mial: instead of hiding the error we could change the controller that accepts this method to report it.
the controller could also wait until the preconditions will be fulfilled before creating/deleting the resources.

p0lyn0mial: yes, revert the change and add the TODO.

pkg/operator/starter.go
```diff
@@ -221,9 +229,17 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		operatorClient,
 		controllerContext.EventRecorder,
 	).
-		WithConditionalResources(bindata.Asset, []string{"assets/alerts/kube-apiserver-slos-extended.yaml"}, notOnSingleReplicaTopology, nil).
+		WithConditionalResources(bindata.Asset, []string{"assets/alerts/kube-apiserver-slos-extended.yaml", "assets/alerts/cpu-utilization.yaml"}, notOnSingleReplicaTopology, nil).
 		AddKubeInformers(kubeInformersForNamespaces)
 
+	dynamicInformersForKubeApiServerNamespace := dynamicinformer.NewFilteredDynamicSharedInformerFactory(dynamicClient, 12*time.Hour, "openshift-kube-apiserver", nil)
```
p0lyn0mial, line 235: please rename to `dynamicInformersForTargetNamespace`.

pkg/operator/starter.go
```diff
@@ -221,9 +225,17 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		operatorClient,
 		controllerContext.EventRecorder,
 	).
-		WithConditionalResources(bindata.Asset, []string{"assets/alerts/kube-apiserver-slos-extended.yaml"}, notOnSingleReplicaTopology, nil).
+		WithConditionalResources(bindata.Asset, []string{"assets/alerts/kube-apiserver-slos-extended.yaml", "assets/alerts/cpu-utilization.yaml"}, notOnSingleReplicaTopology, nil).
```
p0lyn0mial, line 228: Since this will be addressed by https://github.com/openshift/cluster-kube-apiserver-operator/pull/1680 we don't have to change it in this PR, right ?

pkg/operator/starter.go
```diff
@@ -221,9 +225,17 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
...
 		AddKubeInformers(kubeInformersForNamespaces)
 
+	dynamicInformersForTargetNamespace := dynamicinformer.NewFilteredDynamicSharedInformerFactory(dynamicClient, 12*time.Hour, "openshift-kube-apiserver", nil)
```
p0lyn0mial, line 231: please use `operatorclient.TargetNamespace`

pkg/operator/starter.go
```diff
@@ -221,9 +225,17 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
...
+
+	highCpuUsageAlertController := highcpuusagesnoalertcontroller.NewHighCPUUsageSNOAlertController(
```
p0lyn0mial, line 233: nit: rename to `highCpuUsageSNOAlertController`

pkg/operator/highcpuusagesnoalertcontroller/highcpuusagesnoalert_controller.go
```diff
@@ -0,0 +1,104 @@
...
+	client dynamic.Interface,
+	recorder events.Recorder,
+) factory.Controller {
+	c := &highCPUUsageSNOController{
+		client:               client,
+		infraLister:          configInformer.Infrastructures().Lister(),
+		clusterVersionLister: configInformer.ClusterVersions().Lister(),
+	}
+
+	prometheusAlertInformer := dynamicInformer.ForResource(monitoringv1.SchemeGroupVersion.WithResource("prometheusrules"))
```
p0lyn0mial, line 42: Is this the correct GVR ? I don't think so.

I have checked the operator logs from a CI run and it thinks the same:
```
W0604 13:14:07.950220       1 reflector.go:547] k8s.io/client-go@v0.30.0/tools/cache/reflector.go:232: failed to list monitoring.openshift.io/v1, Resource=prometheusrules: the server could not find the requested resource
E0604 13:14:07.950260       1 reflector.go:150] k8s.io/client-go@v0.30.0/tools/cache/reflector.go:232: Failed to watch monitoring.openshift.io/v1, Resource=prometheusrules: failed to list monitoring.openshift.io/v1, Resource=prometheusrules: the server could not find the requested resource
```

pkg/operator/highcpuusagesnoalertcontroller/highcpuusagesnoalert_controller.go
```diff
@@ -0,0 +1,104 @@
...
+
+	return factory.New().
+		WithInformers(configInformer.Infrastructures().Informer(), configInformer.ClusterVersions().Informer(), prometheusAlertInformer.Informer()).
+		WithSync(c.sync).ResyncEvery(10*time.Minute).
+		ToController("highCPUUsageSNOController", recorder.WithComponentSuffix("high-cpu-usage-controller"))
```
p0lyn0mial, line 47: nit: rename to `high-cpu-usage-sno-controller`

pkg/operator/highcpuusagesnoalertcontroller/highcpuusagesnoalert_controller.go
```diff
@@ -0,0 +1,104 @@
...
+
+type highCPUUsageSNOController struct {
+	client               dynamic.Interface
+	infraLister          configlistersv1.InfrastructureLister
+	clusterVersionLister configlistersv1.ClusterVersionLister
+}
+
+func NewHighCPUUsageSNOAlertController(
+	configInformer configv1informers.Interface,
+	dynamicInformer dynamicinformer.DynamicSharedInformerFactory,
```
p0lyn0mial, line 32: rename to `dynamicInformersForTargetNamespace`

pkg/operator/highcpuusagesnoalertcontroller/highcpuusagesnoalert_controller.go
p0lyn0mial, line 42: please rename it to `prometheusAlertInformerForTargetNamespace`

pkg/operator/highcpuusagesnoalertcontroller/performanceprofile.go
```diff
@@ -0,0 +1,61 @@
...
+	"context"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/utils/cpuset"
+)
+
+// default and taken from the docs
```
p0lyn0mial, line 13: why not to move the content of this file to `highcpuusagesnoalert_controller` ?
do we need a separate file for just one function ?

pkg/operator/highcpuusagesnoalertcontroller/performanceprofile_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+		expectedCores       int
+		expectedToFindCores bool
+		expectedErr         bool
+	}{
+		{
+			name:                "no performanceProfile",
+			expectedToFindCores: false,
+			expectedErr:         false,
+		}, {
```
p0lyn0mial, line 29: formatting, please add a new line.

pkg/operator/highcpuusagesnoalertcontroller/performanceprofile_test.go
```diff
@@ -0,0 +1,168 @@
...
+			coresFound, found, err := performanceProfileControlPlaneCores(context.Background(), clientWithObjects(tt.clusterObjects...))
+			if (err != nil) != tt.expectedErr {
+				t.Errorf("performanceProfileControlPlaneCores() error = %v, expectedErr %v", err, tt.expectedErr)
+				return
+			}
+			if coresFound != tt.expectedCores {
+				t.Errorf("performanceProfileControlPlaneCores() got = %v, want %v", coresFound, tt.expectedCores)
+			}
+			if found != tt.expectedToFindCores {
+				t.Errorf("performanceProfileControlPlaneCores() got1 = %v, want %v", found, tt.expectedToFindCores)
```
p0lyn0mial, line 79: could you rename to `t.Errorf("performanceProfileControlPlaneCores() foundCores = %v, expectedToFindCores = %v", found, tt.expectedToFindCores)` ?

pkg/operator/highcpuusagesnoalertcontroller/performanceprofile_test.go
```diff
@@ -0,0 +1,168 @@
...
+			name:                "no cpu set in performance profile",
+			clusterObjects:      []runtime.Object{noCPUSetInvalidPerformanceProfile()},
+			expectedToFindCores: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			coresFound, found, err := performanceProfileControlPlaneCores(context.Background(), clientWithObjects(tt.clusterObjects...))
+			if (err != nil) != tt.expectedErr {
+				t.Errorf("performanceProfileControlPlaneCores() error = %v, expectedErr %v", err, tt.expectedErr)
```
p0lyn0mial, line 72: please update the error msg to  `expectedErr = %v` (same below)

pkg/operator/highcpuusagesnoalertcontroller/highcpuusagesnoalert_controller_test.go
```diff
@@ -0,0 +1,90 @@
...
+			goldenFile: "./testdata/alert_2_cores.yaml",
+			wantErr:    false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := snoAlert(context.Background(), clientWithObjects(tt.args.clusterObjects...), tt.args.enabledCapabilities, tt.args.cpuMode)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("snoAlert() error = %v, expectedErr %v", err, tt.wantErr)
```
p0lyn0mial, line 65: please update the error msg to `expectedErr = %v` (same below)

pkg/operator/highcpuusagesnoalertcontroller/highcpuusagesnoalert_controller_test.go
```diff
@@ -0,0 +1,90 @@
...
+				return
```
p0lyn0mial, line 66: change to `t.Fatalf` and remove the `return` statement.

pkg/operator/highcpuusagesnoalertcontroller/performanceprofile_test.go
```diff
@@ -0,0 +1,168 @@
...
+			"metadata": map[string]interface{}{
+				"name": "performanceProfile" + strconv.Itoa(rand.Int()),
+			},
+			"spec": map[string]interface{}{
+				"nodeSelector": map[string]interface{}{
+					"node-role.kubernetes.io/master": "",
+				},
+				"cpu": map[string]interface{}{
+					"isolated": "0-13",
+					"reserved": "14+15",
```
p0lyn0mial, line 128: is this profile invalid because of this entry ?

p0lyn0mial:
/lgtm
/hold

@qJkee wants to run an e2e test. Feel free to cancel `hold` once you are ready. 
@qJkee please don't forget to merge https://github.com/openshift/cluster-kube-apiserver-operator/pull/1680

PR #1678, opened 2024-05-14
Add "check-api-lb" subcommand

vrutkovs:
This command is used by bootstrap to verify that bootstrap instance is ready to be shutdown


pkg/cmd/checkapilb/cmd.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+// Run contains the logic of the check-api-server command.
+func (r *checkAPILBOpts) Run() error {
+	pollCtx, pollCancel := context.WithTimeout(context.Background(), timeout)
+	defer pollCancel()
+
+	attempt := 0
+	return wait.PollImmediateUntil(time.Second, func() (done bool, err error) {
+		retryCtx, retryCancel := context.WithTimeout(context.Background(), time.Second)
```
rphillips, line 96: A 1 second context timeout seems too short.

vrutkovs: Right, I can make it 5 secs just to be safe

patrickdillon:
we need to check the `ControlPlaneTopology` in cluster infrastructure to distinguish single node installs, in which case we don't want to block waiting for 2 endpoints

PR #1680, opened 2024-05-16
OCPEDGE-1102: Revert high cpu usage alert description

qJkee:
As discussed at https://github.com/openshift/cluster-kube-apiserver-operator/pull/1676#discussion_r1601703298, reverting SNO section on alert description and making deployment available only at non SNO clusters

### DIFF
--- a/bindata/assets/alerts/cpu-utilization.yaml
+++ b/bindata/assets/alerts/cpu-utilization.yaml
@@ -10,18 +10,15 @@ spec:
         - alert: HighOverallControlPlaneCPU
           annotations:
             summary: >-
-              CPU utilization across all control plane nodes is more than 60% of the total available CPU. Control plane node outage may cause a cascading failure; increase available CPU.
+              CPU utilization across all three control plane nodes is higher than two control plane nodes can sustain; a single control plane node outage may
+              cause a cascading failure; increase available CPU.
             runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-kube-apiserver-operator/ExtremelyHighIndividualControlPlaneCPU.md
             description: >-
-              On a multi-node cluster with three control plane nodes, the overall CPU utilization may only be about 2/3 of all available capacity.
+              Given three control plane nodes, the overall CPU utilization may only be about 2/3 of all available capacity.
               This is because if a single control plane node fails, the remaining two must handle the load of the cluster in order to be HA.
-              If the cluster is using more than 2/3 of all capacity, if one control plane node fails, the remaining two are likely to fail when they take the load.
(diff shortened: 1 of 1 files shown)


pkg/operator/starter.go
```diff
@@ -221,7 +220,7 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 		operatorClient,
 		controllerContext.EventRecorder,
 	).
-		WithConditionalResources(bindata.Asset, []string{"assets/alerts/kube-apiserver-slos-extended.yaml"}, notOnSingleReplicaTopology, nil).
+		WithConditionalResources(bindata.Asset, []string{"assets/alerts/kube-apiserver-slos-extended.yaml", "assets/alerts/cpu-utilization.yaml"}, notOnSingleReplicaTopology, nil).
```
p0lyn0mial, line 223: give that the `assets/alerts/cpu-utilization.yaml` alert was created on previous clusters and that now it will be created only on SNO cluster should we also delete it for clusters upgrading from prior versions ?

bindata/assets/alerts/cpu-utilization.yaml
```diff
@@ -10,18 +10,15 @@ spec:
         - alert: HighOverallControlPlaneCPU
           annotations:
             summary: >-
-              CPU utilization across all control plane nodes is more than 60% of the total available CPU. Control plane node outage may cause a cascading failure; increase available CPU.
+              CPU utilization across all three control plane nodes is higher than two control plane nodes can sustain; a single control plane node outage may
```
p0lyn0mial, line 13: @dgrisonnet are you ok with the changes ?

pkg/operator/starter.go
p0lyn0mial: Better check :) - It looks like we should be good https://github.com/openshift/cluster-kube-apiserver-operator/blob/d597184dccf293101d38e543587d40af99aca52a/vendor/github.com/openshift/library-go/pkg/operator/staticresourcecontroller/static_resource_controller.go#L167

p0lyn0mial: Could you please add a TODO where `notOnSingleReplicaTopology` is defined?

Based on our conversation from the morning the `TODO` should say that in the future the infra object might/will be changed at runtime thus the function must read the ~current state from the cluster.

p0lyn0mial:
@qJkee once the CI jobs finish could you please manually check if the alert was created on a standard cluster and wasn't on an SNO cluster? Thanks!

bindata/assets/alerts/cpu-utilization.yaml
dgrisonnet: yes it sounds fine to revert back to where it was before

PR #1685, opened 2024-05-21
OCPBUGS-35343: make shutdown-delay-duration configurable

tkashem:


pkg/cmd/render/render.go
```diff
@@ -97,6 +106,7 @@ func (r *renderOpts) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&r.clusterConfigFile, "cluster-config-file", r.clusterConfigFile, "Openshift Cluster API Config file.")
 	fs.StringVar(&r.clusterAuthFile, "cluster-auth-file", r.clusterAuthFile, "Openshift Cluster Authentication API Config file.")
 	fs.StringVar(&r.infraConfigFile, "infra-config-file", "", "File containing infrastructure.config.openshift.io manifest.")
+	fs.DurationVar(&r.shutdownDelayDuration, "shutdown-delay-duration", r.shutdownDelayDuration, "shutdown-delay-duration argument for the bootstrap kube-apiserver.")
```
p0lyn0mial, line 109: why not to add it directly to the bootstrap-api-server's manifest ? - https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/bindata/bootkube/bootstrap-manifests/kube-apiserver-pod.yaml#L46

p0lyn0mial: oh, cool!
should we also set `shutdown-send-retry-after` ?

pkg/cmd/render/render.go
```diff
@@ -192,7 +206,7 @@ func (r *renderOpts) Run() error {
 		BindAddress:                   "0.0.0.0:6443",
 		BindNetwork:                   "tcp4",
 		TerminationGracePeriodSeconds: 135, // bit more than 70s (minimal termination period) + 60s (apiserver graceful termination)
-		ShutdownDelayDuration:         "",  // do not override
+		ShutdownDelayDuration:         r.shutdownDelayDuration.String(),
```
p0lyn0mial, line 209: BTW: Do we want to expose it as a command line arg ? If yes then we have to coordinate updating it with `terminationGracePeriodSeconds`. Otherwise `kubelet` might terminate `kas` process before graceful termination completes.

pkg/cmd/render/render.go
```diff
@@ -62,6 +66,11 @@ func newRenderCommand(testOverrides ...func(*renderOpts)) *cobra.Command {
 		lockHostPath:   "/var/run/kubernetes/lock",
 		etcdServerURLs: []string{"https://127.0.0.1:2379"},
 		etcdServingCA:  "root-ca.crt",
+
+		// by default, we are giving the load balancer 20s to remove the
+		// bootstrap kube-apiserver from its pool after TERM signal is
+		// sent to the kube-apiserver on the bootstrap node.
+		shutdownDelayDuration: 20 * time.Second,
```
p0lyn0mial, line 73: What is the default value for `shutdownDelayDuration` ?
It looks like it is `0`, does it mean that the bootstrap api server didn't give any time to LBs to react ?

vrutkovs:
/retitle OCPBUGS-35343: make shutdown-delay-duration configurable

pkg/cmd/render/render.go
vrutkovs: iiuc its ) for SNO and 129 for AWS:
https://github.com/openshift/cluster-kube-apiserver-operator/blob/37df1b1f80d3be6036b9e31975ac42fcb21b6447/pkg/operator/configobservation/apiserver/observe_termination_duration.go#L37-L46

PR #1686, opened 2024-05-21
AUTH-521: add disabled syncer as reason to CFE for PSA

ibihim:
## What

- Add a new reason that is being set if a customer disabled the label syncer.
- Check only namespaces, where we at least own one label.

## Why

We want to:

- identify who disabled it and will have a failing workload and
- "properly" calculate PSA violations and not "believe" every label we find

### DIFF
--- a/pkg/operator/podsecurityreadinesscontroller/conditions.go
+++ b/pkg/operator/podsecurityreadinesscontroller/conditions.go
@@ -5,6 +5,7 @@ import (
 	"sort"
 	"strings"
 
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 
@@ -13,9 +14,12 @@ import (
 )
 
 const (
-	PodSecurityCustomerType     = "PodSecurityCustomerEvaluationConditionsDetected"
-	PodSecurityOpenshiftType    = "PodSecurityOpenshiftEvaluationConditionsDetected"
-	PodSecurityRunLevelZeroType = "PodSecurityRunLevelZeroEvaluationConditionsDetected"
+	PodSecurityCustomerType       = "PodSecurityCustomerEvaluationConditionsDetected"
+	PodSecurityOpenshiftType      = "PodSecurityOpenshiftEvaluationConditionsDetected"
+	PodSecurityRunLevelZeroType   = "PodSecurityRunLevelZeroEvaluationConditionsDetected"
+	PodSecurityDisabledSyncerType = "PodSecurityDisabledSyncerEvaluationConditionsDetected"
+
+	labelSyncControlLabel = "security.openshift.io/scc.podSecurityLabelSync"
 )
 
 var (
@@ -28,24 +32,31 @@ var (
 )
 
 type podSecurityOperatorConditions struct {
-	violatingOpenShiftNamespaces    []string
-	violatingRunLevelZeroNamespaces []string
-	violatingCustomerNamespaces     []string
+	violatingOpenShiftNamespaces      []string
+	violatingRunLevelZeroNamespaces   []string
+	violatingCustomerNamespaces       []string
+	violatingDisabledSyncerNamespaces []string
 }
 
-func (c *podSecurityOperatorConditions) addViolation(name string) {
-	if runLevelZeroNamespaces.Has(name) {
-		c.violatingRunLevelZeroNamespaces = append(c.violatingRunLevelZeroNamespaces, name)
+func (c *podSecurityOperatorConditions) addViolation(ns *corev1.Namespace) {
+	if runLevelZeroNamespaces.Has(ns.Name) {
+		c.violatingRunLevelZeroNamespaces = append(c.violatingRunLevelZeroNamespaces, ns.Name)
 		return
 	}
 
-	isOpenShift := strings.HasPrefix(name, "openshift")
+	isOpenShift := strings.HasPrefix(ns.Name, "openshift")
 	if isOpenShift {
-		c.violatingOpenShiftNamespaces = append(c.violatingOpenShiftNamespaces, name)
+		c.violatingOpenShiftNamespaces = append(c.violatingOpenShiftNamespaces, ns.Name)
 		return
 	}
 
-	c.violatingCustomerNamespaces = append(c.violatingCustomerNamespaces, name)
+	if ns.Labels[labelSyncControlLabel] == "false" {
+		// This is the only case in which the controller wouldn't enforce the pod security standards.
+		c.violatingDisabledSyncerNamespaces = append(c.violatingDisabledSyncerNamespaces, ns.Name)
+		return
+	}
+
+	c.violatingCustomerNamespaces = append(c.violatingCustomerNamespaces, ns.Name)
 }
 
 func makeCondition(conditionType string, namespaces []string) operatorv1.OperatorCondition {
@@ -76,5 +87,6 @@ func (c *podSecurityOperatorConditions) toConditionFuncs() []v1helpers.UpdateSta
 		v1helpers.UpdateConditionFn(makeCondition(PodSecurityCustomerType, c.violatingCustomerNamespaces)),
 		v1helpers.UpdateConditionFn(makeCondition(PodSecurityOpenshiftType, c.violatingOpenShiftNamespaces)),
 		v1helpers.UpdateConditionFn(makeCondition(PodSecurityRunLevelZeroType, c.violatingRunLevelZeroNamespaces)),
+		v1helpers.UpdateConditionFn(makeCondition(PodSecurityDisabledSyncerType, c.violatingDisabledSyncerNamespaces)),
 	}
 }
--- a/pkg/operator/podsecurityreadinesscontroller/podsecurityreadinesscontroller.go
+++ b/pkg/operator/podsecurityreadinesscontroller/podsecurityreadinesscontroller.go
@@ -79,7 +79,7 @@ func (c *PodSecurityReadinessController) sync(ctx context.Context, syncCtx facto
 				return err
 			}
 			if isViolating {
-				conditions.addViolation(ns.Name)
+				conditions.addViolation(&ns)
 			}
 
 			return nil
--- a/pkg/operator/podsecurityreadinesscontroller/podsecurityreadinesscontroller.go
+++ b/pkg/operator/podsecurityreadinesscontroller/podsecurityreadinesscontroller.go
@@ -4,12 +4,10 @@ import (
 	"context"
 	"time"
 
-	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
-	applyconfiguration "k8s.io/client-go/applyconfigurations/core/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/util/retry"
@@ -24,11 +22,6 @@ import (
 const (
 	checkInterval = 240 * time.Minute // Adjust the interval as needed.
 )
-
-var podSecurityAlertLabels = []string{
-	psapi.AuditLevelLabel,
-	psapi.WarnLevelLabel,
-}
 
 // PodSecurityReadinessController checks if namespaces are ready for Pod Security Admission enforcement.
 type PodSecurityReadinessController struct {
@@ -46,19 +39,12 @@ func NewPodSecurityReadinessController(
 ) (factory.Controller, error) {
 	warningsHandler := &warningsHandler{}
 
-	kubeClientCopy := rest.CopyConfig(kubeConfig)
-	kubeClientCopy.WarningHandler = warningsHandler
-	// We don't want to overwhelm the apiserver with requests. On a cluster with
-	// 10k namespaces, we would send 10k + 1 requests to the apiserver.
-	kubeClientCopy.QPS = 2
-	kubeClientCopy.Burst = 2
-	kubeClient, err := kubernetes.NewForConfig(kubeClientCopy)
+	kubeClient, err := newWarningAwareKubeClient(warningsHandler, kubeConfig)
 	if err != nil {
 		return nil, err
 	}
 
-	selector := labels.NewSelector()
-	labelsRequirement, err := labels.NewRequirement(psapi.EnforceLevelLabel, selection.DoesNotExist, []string{})
+	selector, err := nonEnforcingSelector()
 	if err != nil {
 		return nil, err
 	}
@@ -67,7 +53,7 @@ func NewPodSecurityReadinessController(
 		operatorClient:    operatorClient,
 		kubeClient:        kubeClient,
 		warningsHandler:   warningsHandler,
-		namespaceSelector: selector.Add(*labelsRequirement).String(),
+		namespaceSelector: selector,
 	}
 
 	return factory.New().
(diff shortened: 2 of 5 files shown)


pkg/operator/podsecurityreadinesscontroller/labelsyncer.go
```diff
@@ -0,0 +1,103 @@
...
+
+	extractedPerManager, err := newLabelsToManager(ns)
+	if err != nil {
+		klog.Errorf("ns extraction failed: %v", err)
+		return false
+	}
+
+	var owningAtLeastOneLabel bool
+	for _, labelName := range []string{
+		psapi.EnforceLevelLabel, psapi.EnforceVersionLabel,
```
stlaz, line 38: are the versions any use for the purpose of CFE?

pkg/operator/podsecurityreadinesscontroller/labelsyncer.go
```diff
@@ -0,0 +1,103 @@
...
+func isNSControlled(ns *corev1.Namespace) bool {
+	if ns.Labels[labelSyncControlLabel] == "true" {
+		return true
+	}
+
+	if strings.HasPrefix(ns.Name, "openshift") {
+		return false
+	}
+
+	extractedPerManager, err := newLabelsToManager(ns)
```
stlaz, line 30: why do you care about any other managers?

pkg/operator/podsecurityreadinesscontroller/conditions.go
```diff
@@ -28,24 +30,30 @@ var (
...
-	if runLevelZeroNamespaces.Has(name) {
-		c.violatingRunLevelZeroNamespaces = append(c.violatingRunLevelZeroNamespaces, name)
+func (c *podSecurityOperatorConditions) addViolation(ns *corev1.Namespace) {
+	if runLevelZeroNamespaces.Has(ns.Name) {
+		c.violatingRunLevelZeroNamespaces = append(c.violatingRunLevelZeroNamespaces, ns.Name)
 		return
 	}
 
-	isOpenShift := strings.HasPrefix(name, "openshift")
+	isOpenShift := strings.HasPrefix(ns.Name, "openshift")
```
stlaz, line 45: you shoulnd't need to treat the prefix differently in `isNSControlled()` if you return here

pkg/operator/podsecurityreadinesscontroller/labelsyncer.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+var (
+	alertLabels = []string{
+		psapi.WarnLevelLabel,
+		psapi.AuditLevelLabel,
+	}
+)
+
+func isNSControlled(ns *corev1.Namespace) bool {
```
stlaz, line 29: What is this actually supposed to do? The `HasAll` in return is making this quite confusing.

Why do you care about the warn label?
Why do you ignore the enforce label?

pkg/operator/podsecurityreadinesscontroller/labelsyncer.go
```diff
@@ -0,0 +1,108 @@
...
+			ownedLabels = ownedLabels.Union(managedLabels)
+		}
+	}
+
+	// Verify that alert labels, that are used to check for violation are owned by us.
+	return ownedLabels.HasAll(alertLabels...)
+}
+
+func isAlertLabel(label string) bool {
+	for _, l := range alertLabels {
```
stlaz, line 66: map?

pkg/operator/podsecurityreadinesscontroller/labelsyncer.go
```diff
@@ -0,0 +1,108 @@
...
+		if label == l {
+			return true
+		}
+	}
+
+	return false
+}
+
+func isSyncController(name string) bool {
+	return len(name) == 0 ||
```
stlaz, line 76: empty name also means true?

stlaz (changes requested):

pkg/operator/podsecurityreadinesscontroller/labelsyncer.go
```diff
@@ -0,0 +1,108 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog/v2"
+	psapi "k8s.io/pod-security-admission/api"
+)
+
+const (
+	syncerControllerName  = "pod-security-admission-label-synchronization-controller"
+	policyControllerName  = "cluster-policy-controller"
```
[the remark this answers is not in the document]
ibihim: Why?

pkg/operator/podsecurityreadinesscontroller/labelsyncer.go
ibihim: We are ignoring use cases, where enforce label is set as we don't do that. We assume that the customer took ownership of it.
We want to ensure that we are responsible for audit and warn labels, if the `labelSyncControlLabel` is not set.

pkg/operator/podsecurityreadinesscontroller/labelsyncer.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	// The customer explicitly tells us to not manage the namespace.
+	if ns.Labels[labelSyncControlLabel] == "false" {
+		return false
+	}
+
+	// Check who is managing the labels.
+	ownedLabels := sets.New[string]()
+	for _, fieldEntry := range ns.ManagedFields {
```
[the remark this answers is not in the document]
ibihim: SSA?

pkg/operator/podsecurityreadinesscontroller/labelsyncer.go
```diff
@@ -0,0 +1,108 @@
...
+		psapi.WarnLevelLabel,
+		psapi.AuditLevelLabel,
+	}
+)
+
+func isNSControlled(ns *corev1.Namespace) bool {
+	// The customer explicitly tells us to manage the namespace.
+	if ns.Labels[labelSyncControlLabel] == "true" {
+		return true
+	}
```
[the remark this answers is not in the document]
ibihim: tell me more

pkg/operator/podsecurityreadinesscontroller/labelsyncer.go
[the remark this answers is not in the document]
stlaz: because this field manager does nothing with the labels you care about

pkg/operator/podsecurityreadinesscontroller/labelsyncer.go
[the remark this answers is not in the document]
stlaz: the label being set to true will still be ignored with platform namespaces

pkg/operator/podsecurityreadinesscontroller/labelsyncer.go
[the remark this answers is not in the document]
stlaz: Server-side apply, the reason we have the managed fields. You want to use (or experiment with using) something like this: https://github.com/openshift/cluster-policy-controller/blob/eaea543f4c845a7b65705f12e162cc121bb12f88/pkg/psalabelsyncer/podsecurity_label_sync_controller.go#L261

pkg/operator/podsecurityreadinesscontroller/labelsyncer.go
stlaz: > We are ignoring use cases, where enforce label is set as we don't do that.

right, but if user sets the enforced label, you don't need to worry about that namespace triggering violations, right?
> Those labels are being used to search for potential upcoming violations

Used by whom? Why do we care if a user gets a warning when they try to create a pod?

pkg/operator/podsecurityreadinesscontroller/labelsyncer.go
stlaz: use map

pkg/operator/podsecurityreadinesscontroller/labelsyncer.go
stlaz: but what does that mean?

pkg/operator/podsecurityreadinesscontroller/labelsyncer.go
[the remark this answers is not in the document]
ibihim: > Now we handle the labels differently (updating only owned fields, setting owner and so on). This is pre-4.14.

pkg/operator/podsecurityreadinesscontroller/podsecurityreadinesscontroller.go
ibihim, line 1: Discussion:
- WE should only care for psa level comparison, if we are the owners.

pkg/operator/podsecurityreadinesscontroller/labelsyncer.go
ibihim: Discussion: We should only consider labels we set and most probably only audit.

pkg/operator/podsecurityreadinesscontroller/violation.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	return targetLevel
+}
+
+func getSyncerLabels(managedFields []metav1.ManagedFieldsEntry) sets.Set[string] {
+	ownedLabels := sets.New[string]()
+
+	for _, entry := range managedFields {
+		if entry.Manager != "" && entry.Manager != syncerControllerName {
```
ibihim, line 90: It might be that we only care about `syncerControllerName` cases as this is the controller setting it... 😅

pkg/operator/podsecurityreadinesscontroller/violation.go
```diff
@@ -0,0 +1,130 @@
...
+
+		if managedLabels.Len() != 0 {
+			ownedLabels = ownedLabels.Union(managedLabels)
+		}
+	}
+
+	return ownedLabels
+}
+
+func getLabels(fieldsEntry metav1.ManagedFieldsEntry) (sets.Set[string], error) {
```
ibihim, line 108: copy pasted from cluster-policy-controller's psa logic

pkg/operator/podsecurityreadinesscontroller/podsecurityreadinesscontroller_test.go
```diff
@@ -189,3 +261,49 @@ func TestPodSecurityViolationController(t *testing.T) {
 		})
 	}
 }
+
+func TestNamespaceSelector(t *testing.T) {
```
ibihim, line 265: we are indirectly testing namespace selector logic here, but makes me feel safer.

pkg/operator/podsecurityreadinesscontroller/violation.go
```diff
@@ -0,0 +1,130 @@
...
+		psapi.WarnLevelLabel:  {},
+		psapi.AuditLevelLabel: {},
+	}
+)
+
+func (c *PodSecurityReadinessController) isNamespaceViolating(ctx context.Context, ns *corev1.Namespace) (bool, error) {
+	viableLabels := getSyncerLabels(ns.ManagedFields).Intersection(alertLabels)
+	if viableLabels.Len() == 0 {
+		// If there are no labels managed by the syncer, we can't make a decision.
+		return false, nil
```
ibihim, line 33: This one is tricky, but I would not to prefer to assume that the customer set reasonable values.

pkg/operator/podsecurityreadinesscontroller/podsecurityreadinesscontroller.go
```diff
@@ -114,58 +100,23 @@ func (c *PodSecurityReadinessController) sync(ctx context.Context, syncCtx facto
...
-		psapi.EnforceLevelLabel: string(targetLevel),
-	})
-
-	_, err := c.kubeClient.CoreV1().
-		Namespaces().
-		Apply(ctx, nsApply, metav1.ApplyOptions{
-			DryRun:       []string{metav1.DryRunAll},
-			FieldManager: "pod-security-readiness-controller",
-		})
+func getNamespaceSelector() (string, error) {
```
liouk, line 103: - func signature hides what the ns selector will select
- `get` prefixes do not feel very go-like

How about something like `psaEnforcedNamespaceSelector()`? or passing arguments, like `namespaceLabelSelector(psapi.EnforceLevelLabel, selection.DoesNotExist)`?

pkg/operator/podsecurityreadinesscontroller/violation.go
```diff
@@ -0,0 +1,72 @@
...
+	"k8s.io/klog/v2"
+	psapi "k8s.io/pod-security-admission/api"
+)
+
+var podSecurityAlertLabels = []string{
+	psapi.AuditLevelLabel,
+	psapi.WarnLevelLabel,
+}
+
+func (c *PodSecurityReadinessController) isNamespaceViolating(ctx context.Context, ns *corev1.Namespace) (bool, error) {
```
liouk, line 29: Similarly as previous comment, `is` prefix not very common in go; how about `namespaceViolating()`?

pkg/operator/podsecurityreadinesscontroller/violation.go
```diff
@@ -2,29 +2,58 @@ package podsecurityreadinesscontroller
...
-var podSecurityAlertLabels = []string{
-	psapi.AuditLevelLabel,
-	psapi.WarnLevelLabel,
-}
+const (
+	syncerControllerName = "pod-security-admission-label-synchronization-controller"
+)
+
+var (
+	alertLabels = sets.Set[string]{
```
liouk, line 23: Simpler definition with the same lib:
```go
alertLabels = sets.New(psapi.WarnLevelLabel, psapi.AuditLevelLabel)
```

pkg/operator/podsecurityreadinesscontroller/violation.go
```diff
@@ -47,26 +76,55 @@ func (c *PodSecurityReadinessController) isNamespaceViolating(ctx context.Contex
...
+	return targetLevel
+}
 
-	_, err := c.kubeClient.CoreV1().
-		Namespaces().
-		Apply(ctx, nsApply, metav1.ApplyOptions{
-			DryRun:       []string{metav1.DryRunAll},
-			FieldManager: "pod-security-readiness-controller",
-		})
+func getSyncerLabels(managedFields []metav1.ManagedFieldsEntry) sets.Set[string] {
```
liouk, line 86: Same comment about `get` prefix.

pkg/operator/podsecurityreadinesscontroller/conditions.go
```diff
@@ -28,24 +32,31 @@ var (
...
+	violatingOpenShiftNamespaces      []string
+	violatingRunLevelZeroNamespaces   []string
+	violatingCustomerNamespaces       []string
+	violatingDisabledSyncerNamespaces []string
 }
 
-func (c *podSecurityOperatorConditions) addViolation(name string) {
-	if runLevelZeroNamespaces.Has(name) {
-		c.violatingRunLevelZeroNamespaces = append(c.violatingRunLevelZeroNamespaces, name)
+func (c *podSecurityOperatorConditions) addViolation(ns *corev1.Namespace) {
```
liouk, line 41: Why did you choose to pass `*corev1.Namespace` instead of its name? Curious, since no other fields of the NS are used anywhere.

ibihim:
```go
if ns.Labels[labelSyncControlLabel] == "false" {
```

https://github.com/openshift/cluster-kube-apiserver-operator/pull/1686/files#diff-4b300130e48b2b1df6907499643909f1f95fc3eb7064fac655ff704a61695e10R53

pkg/operator/podsecurityreadinesscontroller/podsecurityreadinesscontroller.go
ibihim: It is a pure man's constant 😄

pkg/operator/podsecurityreadinesscontroller/violation.go
ibihim: From the standard library: [IsUpper](https://pkg.go.dev/unicode#IsUpper)

pkg/operator/podsecurityreadinesscontroller/conditions.go
liouk: Ah right, missed that one :+1:

pkg/operator/podsecurityreadinesscontroller/labelsyncer.go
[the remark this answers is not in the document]
ibihim: Hmm, maybe, if we ignore the empty space case for manager...

pkg/operator/podsecurityreadinesscontroller/labelsyncer.go
ibihim: It might be valid to consider only labels by the controller.

pkg/operator/podsecurityreadinesscontroller/podsecurityreadinesscontroller_test.go
```diff
@@ -37,6 +98,7 @@ func TestPodSecurityViolationController(t *testing.T) {
 						psapi.AuditLevelLabel: "restricted",
```
ibihim, line 98: I think we should use hardcoded strings in here, just to be sure.

dinhxuanvu (commented):
/lgtm

PR #1689, opened 2024-05-23
OCPBUGS-33963: Create one-shot migrations for the flowcontrol group.

benluddy:
The v1alpha1 version of the flowcontrol.apiserver.k8s.io group is removed in 4.16. Clusters installed prior to 4.8 may still have flowcontrol resources that were persisted using v1alpha1. If the storage version of those resources is not migrated before updating to 4.16, they will not be decodable. This ultimately prevents kube-apiserver instances from becoming ready.


benluddy:
This migration needs to happen _before_ initiating an update to 4.16.

/cherry-pick release-4.16 release-4.15

sanchezl (approved):
/lgtm

PR #1693, opened 2024-05-29
OCPBUGS-34544: Disable PersistentVolumeLabel admission plugin

jsafrane:
PersistentVolumeLabel admission plugin needs a valid cloud config. kube-apiserver does not get it in OCP 4.17, and thus it rejects valid in-tree GCE PersistentVolumes.

The admission plugin was deprecated upstream a long time ago, it was removed from Kubernetes 1.31 (OCP 4.18) and it cannot work without a cloud provider in kube-apiserver anyway.

cc @openshift/storage

### DIFF
--- a/bindata/assets/config/defaultconfig.yaml
+++ b/bindata/assets/config/defaultconfig.yaml
@@ -65,7 +65,6 @@ apiServerArguments:
     - NodeRestriction
     - OwnerReferencesPermissionEnforcement
     - PersistentVolumeClaimResize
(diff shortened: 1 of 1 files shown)


jsafrane:
k8s-e2e-gcp-serial with failing SELinuxMount metrics is not related to this PR - it most probably needs kubelet 1.30, which is not yet present in RHCOS.
I am not able to judge the other errors

elmiko:
thanks for posting this @jsafrane 

/lgtm

PR #1704, opened 2024-06-19
NO-JIRA: nodekubeconfigcontroller: set ownership component for node kubeconfigs

vrutkovs:
Kubeconfigs are parsed as TLS artifact registry sources, so secrets containing kubeconfigs should have necessary annotations

### DIFF
--- a/pkg/operator/nodekubeconfigcontroller/nodekubeconfigcontroller.go
+++ b/pkg/operator/nodekubeconfigcontroller/nodekubeconfigcontroller.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/openshift/api/annotations"
 	operatorv1 "github.com/openshift/api/operator/v1"
 	configv1informers "github.com/openshift/client-go/config/informers/externalversions/config/v1"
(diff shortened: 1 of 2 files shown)


pkg/operator/nodekubeconfigcontroller/nodekubeconfigcontroller.go
```diff
@@ -147,6 +148,10 @@ func ensureNodeKubeconfigs(ctx context.Context, client coreclientv1.CoreV1Interf
 		requiredSecret.StringData[k] = data
 	}
 
+	requiredSecret.Annotations = map[string]string{
```
p0lyn0mial, line 151: this will completely wipe out already set annotations. change to:
1. first check if the map is nil
2. if nil then allocate memory
3. add the new annotation.

vrutkovs: I expect `ApplySecret` to preserve the existing ones. But its a good idea not to rely on that, I agree

p0lyn0mial: If I add an annotation to the manifest stored in the assets this code will remove it so we have to fix it.

PR #1713, opened 2024-07-18
ETCD-636: add etcd backup cert

Elbehery:
This PR adds a certificate to be used by `etcd backup server` 

resolves https://issues.redhat.com/browse/ETCD-636

cc @soltysh @p0lyn0mial @openshift/openshift-team-etcd

### DIFF
--- a/pkg/operator/certrotationcontroller/certrotationcontroller.go
+++ b/pkg/operator/certrotationcontroller/certrotationcontroller.go
@@ -848,6 +848,64 @@ func newCertRotationController(
 	)
 	ret.certRotators = append(ret.certRotators, certRotator)
 
+	// add etcd cert
+	certRotator = certrotation.NewCertRotationController(
+		"etcdBackupServerCert",
+		certrotation.RotatedSigningCASecret{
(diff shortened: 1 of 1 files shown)


vrutkovs:
This looks correct, but it should use controllers from https://github.com/openshift/library-go/pull/1722, otherwise it would race on signer updates

pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -848,6 +848,64 @@ func newCertRotationController(
 	)
 	ret.certRotators = append(ret.certRotators, certRotator)
 
+	// add etcd cert
+	certRotator = certrotation.NewCertRotationController(
+		"etcdBackupServerCert",
```
p0lyn0mial, line 853: I'm not sure if this certificate should by managed by `kas-o`

Alternative would be to:
1. create your own CA and use it to generate the client cert
2. use CSR api to get a client certificate

Both `1` and `2` can be done in `etcd-o` operator.

PR #1732, opened 2024-09-11
WIP: OCPBUGS-41778: increase kube-apiserver failureThreshold

liouk:
This PR improves the liveness/readiness checks of the oauth-apiserver.

In particular:
- set `etcd-healthcheck-timeout` and `etcd-readycheck-timeout` to 1sec less than the respective probe timeouts
- replace liveness probe path `healthz` with `livez`, and exclude `etcd` from the probe
- increase the readiness probe failure threshold to 3

Note that there is no need to update `shutdown-delay-duration` and `terminationGracePeriodSeconds` as their values are already large enough to account for the increase in the readiness probe failure threshold (see the [defaultconfig.yaml](https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/bindata/assets/config/defaultconfig.yaml#L167-L168) and the [targetconfigcontroller.go](https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/pkg/operator/targetconfigcontroller/targetconfigcontroller.go#L432-L436)).


bindata/bootkube/bootstrap-manifests/kube-apiserver-pod.yaml
```diff
@@ -80,7 +80,9 @@ spec:
         port: 6443
         path: readyz
       initialDelaySeconds: 10
```
p0lyn0mial, line 82: should we also exclude the etcd check from the livez path ?

p0lyn0mial: actually why not to make the probes to look exactly the same for the bootstrap server and a standard server ?

bindata/assets/kube-apiserver/pod.yaml
```diff
@@ -117,12 +117,12 @@ spec:
       periodSeconds: 5
       timeoutSeconds: 10
       successThreshold: 1
-      failureThreshold: 1
+      failureThreshold: 3
```
p0lyn0mial, line 120: With these settings, we don't need to change the shutdown delay (70s) or the terminationGracePeriodSeconds (135s).

p0lyn0mial:
/lgtm

holding off to give others time to review
/hold

bindata/assets/kube-apiserver/pod.yaml
```diff
@@ -117,12 +117,12 @@ spec:
...
     startupProbe:
       httpGet:
         scheme: HTTPS
         port: 6443
-        path: healthz
+        path: livez
```
benluddy, line 125: Should the startup probe exclude the etcd check as the liveness probe does?

bindata/bootkube/bootstrap-manifests/kube-apiserver-pod.yaml
```diff
@@ -71,16 +71,32 @@ spec:
...
+      initialDelaySeconds: 0
+      periodSeconds: 5
+      timeoutSeconds: 10
+      successThreshold: 1
+      failureThreshold: 3
+    startupProbe:
+      httpGet:
+        scheme: HTTPS
+        port: 6443
+        path: livez
```
benluddy, line 94: And here?

p0lyn0mial:
@deads2k FYI `ci/prow/e2e-aws-ovn-upgrade` failed on (might be a bug in the test):
```
missing acquiring stage for namespace/openshift-machine-api node/ip-10-0-90-109.ec2.internal pod/control-plane-machine-set-operator-7dc4465f5c-zz6pd uid/5704ff7b-ac83-46b1-a07c-f54a02bce1a5 container/control-plane-machine-set-operator}
```

p0lyn0mial:
@deads2k `ci/prow/e2e-aws-ovn-upgrade` has failed for the second time in a row on `missing acquiring stage` this time for `namespace/openshift-cluster-storage-operator` It might be an issue with the test, given how late we are in the release cycle and that the failing test is not caused by the changes introduced in this PR are you willing to manually override the failing job?

PR #1741, opened 2024-09-26
starter.go: make apply function private

p0lyn0mial:


p0lyn0mial:
turns out these functions are used in multiple places.

/close

PR #1743, opened 2024-09-26
OCPBUGS-42083: Don't rollout revision until three etcd endpoints are listed

vrutkovs:
Prevent apiserver from pointing to a single etcd endpoint (IP and localhost), which will lose connection when the single etcd is being updated

### DIFF
--- a/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
+++ b/pkg/operator/targetconfigcontroller/targetconfigcontroller.go
@@ -48,7 +48,8 @@ type TargetConfigController struct {
 	kubeClient      kubernetes.Interface
 	configMapLister corev1listers.ConfigMapLister
 
-	isStartupMonitorEnabledFn func() (bool, error)
+	isStartupMonitorEnabledFn    func() (bool, error)
+	notOnSingleReplicaTopologyFn func() bool
 }
 
 func NewTargetConfigController(
@@ -58,15 +59,17 @@ func NewTargetConfigController(
 	kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces,
 	kubeClient kubernetes.Interface,
 	isStartupMonitorEnabledFn func() (bool, error),
+	notOnSingleReplicaTopologyFn func() bool,
 	eventRecorder events.Recorder,
 ) factory.Controller {
 	c := &TargetConfigController{
-		targetImagePullSpec:       targetImagePullSpec,
-		operatorImagePullSpec:     operatorImagePullSpec,
-		operatorClient:            operatorClient,
-		kubeClient:                kubeClient,
-		configMapLister:           kubeInformersForNamespaces.ConfigMapLister(),
-		isStartupMonitorEnabledFn: isStartupMonitorEnabledFn,
+		targetImagePullSpec:          targetImagePullSpec,
+		operatorImagePullSpec:        operatorImagePullSpec,
+		operatorClient:               operatorClient,
+		kubeClient:                   kubeClient,
+		configMapLister:              kubeInformersForNamespaces.ConfigMapLister(),
+		isStartupMonitorEnabledFn:    isStartupMonitorEnabledFn,
+		notOnSingleReplicaTopologyFn: notOnSingleReplicaTopologyFn,
 	}
 
 	return factory.New().WithInformers(
@@ -100,7 +103,8 @@ func (c TargetConfigController) sync(ctx context.Context, syncContext factory.Sy
 	}
 
 	// block until config is observed and specific paths are present
-	if err := isRequiredConfigPresent(operatorSpec.ObservedConfig.Raw); err != nil {
+	isNotOnSingleReplicaTopology := c.notOnSingleReplicaTopologyFn()
+	if err := isRequiredConfigPresent(operatorSpec.ObservedConfig.Raw, isNotOnSingleReplicaTopology); err != nil {
 		syncContext.Recorder().Warning("ConfigMissing", err.Error())
 		return err
 	}
@@ -116,7 +120,7 @@ func (c TargetConfigController) sync(ctx context.Context, syncContext factory.Sy
 	return nil
 }
 
-func isRequiredConfigPresent(config []byte) error {
+func isRequiredConfigPresent(config []byte, isNotSingleNode bool) error {
 	if len(config) == 0 {
 		return fmt.Errorf("no observedConfig")
 	}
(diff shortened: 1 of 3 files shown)


pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -148,6 +148,16 @@ func isRequiredConfigPresent(config []byte) error {
 		if configValString, ok := configVal.(string); ok && len(configValString) == 0 {
 			return fmt.Errorf("%v empty in config", strings.Join(requiredPath, "."))
 		}
+
+		if requiredPath[0] == "etcd-servers" {
```
dgrisonnet, line 152: I'd be curious to see in the tests if this also gated rev 1, though I doubt it will be the case. Worst case scenario we can always add that check as a precondition to the revision controller.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -148,6 +148,16 @@ func isRequiredConfigPresent(config []byte) error {
...
+			configValSlice, ok := configVal.([]interface{})
+			if !ok {
+				return fmt.Errorf("%v is not a slice", strings.Join(requiredPath, "."))
+			}
+			if len(configValSlice) < 3 {
```
dgrisonnet, line 157: This needs to be gated for SNO as I assume it would only get 2 endpoints. Also, shouldn't we aim for 4 endpoints since we include localhost?

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
deads2k, line 152: remember to only do this for HA clusters, not single node.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
vrutkovs: yes, adding an SNO gate too.

>shouldn't we aim for 4 endpoints since we include localhost?

Three endpoints - localhost, local IP and etcd IP on any other master - should be sufficient to prevent hangup for 10 mins.
We may never have all 4 endpoints on HA with assisted installer - one of masters is used as a bootstrap, so we shouldn't rely on having all 3 masters available during bootstrap.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
vrutkovs: Right, `requiredPath[0]` is `apiServerArguments`, updated the code and added unit tests

vrutkovs:
/test e2e-azure-ovn
/payload-aggregate periodic-ci-openshift-release-master-nightly-4.17-e2e-azure-ovn-kube-apiserver-rollout 10

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
p0lyn0mial: I haven’t read the code, but I’m wondering if we only want to stop new revisions during the installation time. If not, what happens when a customer takes a master node out of the pool for service? Does that mean the kas-o will stop installing new revisions? If so, that’s bad, right?

p0lyn0mial: Also, recently there was a change to the RevisionController (library-go) which permits passing a precondition function. New revisions will be allowed by the controller once the precondition fulfils. Maybe you could use it.

vrutkovs: >If not, what happens when a customer takes a master node out of the pool for service?

Multiple scenarios here:
* The node may be physically down for maintenance, but it would still be listed as an endpoint. 

* Also, while we require three endpoints - its in fact two distinct etcd servers (local IP and localhost are considered single etcd server but two distinct endpoints). So this code won't prevent one master from being torn down during CPMS scale up / down

* However there is an etcd restore procedure where the cluster is reduced to a single node to restore to backup and later being scale up to 3. We specifically don't want rolling out new revisions until all three masters are available to ensure all three etcd servers are in sync.

TODO: run e2e to do cpms scale up / etcd disruption to ensure that it works as expected

vrutkovs: >New revisions will be allowed by the controller once the precondition fulfils

`WithRevisionControllerPrecondition` might be an alternative to this solution, yes

p0lyn0mial: > WithRevisionControllerPrecondition might be an alternative to this solution, yes

Not a blocking comment just wanted to let you know the mechanism exist so that you may consider using it.

So, when an etcd endpoint is removed from the list?

p0lyn0mial: Also, should we add a similar mechanism for the aggregated API servers ?

vrutkovs: >just wanted to let you know the mechanism exist so that you may consider using it

Right, I think current one is more suited for this situation though - we already have a config validation function so its only natural to extend it. This way we prevent the config from being generated not merely rolled out.

However if this method won't work in some scenarios `WithRevisionControllerPrecondition` would be a good candidate to try.

>should we add a similar mechanism for the aggregated API servers 

Its not necessary imo, the issue is happening early on bootstrap when openshift-apiserver is not even created. 

If kube-apiserver can reliably write to etcd we'll notice slow etcd rollout via "static pod didn't rollout in 3 mins" test - and a similar solution can be applied to openshift-apiservers

vrutkovs:
/test e2e-azure-ovn
/payload-aggregate periodic-ci-openshift-release-master-nightly-4.17-e2e-azure-ovn-kube-apiserver-rollout 10

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
dgrisonnet: My comment was about checking in the test run if the config for revision 1 got the required list of etcd-servers or if it bypassed this check. We need to make sure that rev 1 gets the minimum number of endpoints, otherwise we might run into the same problem as before with the first kube-apiserver pod.

vrutkovs: Ah, got it. Well, we can't issue a revision if config is not considered correct, so I think we have this covered.

Not sure we already have revision controller unit tests

vrutkovs:
/payload-aggregate periodic-ci-openshift-release-master-nightly-4.17-e2e-azure-ovn-kube-apiserver-rollout 10

deads2k:
The problem this is targeting just failed another payload.  I *think* the operator job is failing in general.  I'm overriding and will notify the team in slack.

/override ci/prow/e2e-gcp-operator

PR #1747, opened 2024-09-27
API-1835: update to use the latest revision controller

deads2k:
proof of https://github.com/openshift/library-go/pull/1801

/hold for real bump


bertinatto (commented):
/lgtm

PR #1750, opened 2024-10-07
NO-JIRA: Update "deprecated API in use" alert expressions for 1.31.

p0lyn0mial:

PR #1751, opened 2024-10-07
pull https://github.com/openshift/library-go/pull/1819

p0lyn0mial:


p0lyn0mial:
ok, i manually checked the audit-logs [form](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-kube-apiserver-operator/1751/pull-ci-openshift-cluster-kube-apiserver-operator-master-e2e-gcp-operator/1843191514499387392/artifacts/e2e-gcp-operator/gather-audit-logs/artifacts/) and requests to `/livez` are no longer persisted in the audit logs.

However I noticed the following entries:
```
{"kind":"Event","apiVersion":"audit.k8s.io/v1","level":"Metadata","auditID":"fb6a08f8-3656-4435-899a-74dee6416bf1","stage":"ResponseComplete","requestURI":"/readyz","verb":"get","user":{"username":"system:apiserver","uid":"45190fa5-6708-486e-9522-40785dec4762","groups":["system:masters"]},"sourceIPs":["::1"],"userAgent":"kube-apiserver/v1.31.1 (linux/amd64) kubernetes/8540bd4","responseStatus":{"metadata":{},"code":500},"requestReceivedTimestamp":"2024-10-07T09:40:42.005730Z","stageTimestamp":"2024-10-07T09:40:42.006133Z","annotations":{"authorization.k8s.io/decision":"allow","authorization.k8s.io/reason":""}}
```

which indicates that the requests from the self-client are not matching the audit policy. I am not sure why the self-client queries the /readyz endpoint, but we could update the audit policy to exclude requests from this user to the health check endpoints.

deads2k:
> which indicates that the requests from the self-client are not matching the audit policy. I am not sure why the self-client queries the /readyz endpoint, but we could update the audit policy to exclude requests from this user to the health check endpoints.

Don't exclude the entire user, we need to know what that user does.  Weird that it escapes policy though.

p0lyn0mial:
/close

will pull the fix into https://github.com/openshift/cluster-kube-apiserver-operator/pull/1750

PR #1757, opened 2024-10-21
NO-JIRA: nodekubeconfigcontroller: add auto-regenerate-after-expiry annotation to node kubeconfig configmap

vrutkovs:
`node-kubeconfigs` contains configmaps contain lb-ext/lb-int/localhost/localhost-recovery kubeconfigs, which get updated every two years. The configmap should have annotation to mark it as being updated offline


vrutkovs:
We don't test if these are being regenerated during cluster shutdown for 2 years

PR #1759, opened 2024-10-24
OCPBUGS-43777: certrotationcontroller: run tests which runs deployment and creates projects

vrutkovs:
Tests we run after cert rotation should ensure that pod gets created from deployment, scheduled on the node and openshift-apiserver can create projects to validate that all component certificates have been regenerated. The test names are included in
certificates.openshift.io/auto-regenerate-after-offline-expiry annotation


pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -324,7 +324,7 @@ func newCertRotationController(
 			Name:      "service-network-serving-certkey",
 			AdditionalAnnotations: certrotation.AdditionalAnnotations{
 				JiraComponent:                    "kube-apiserver",
-				AutoRegenerateAfterOfflineExpiry: "https://github.com/openshift/cluster-kube-apiserver-operator/pull/1631,'operator conditions kube-apiserver'",
+				AutoRegenerateAfterOfflineExpiry: "https://github.com/openshift/cluster-kube-apiserver-operator/pull/1631,'[sig-apps] Deployment RollingUpdateDeployment should delete old pods and create new ones [Conformance] [Suite:openshift/conformance/parallel/minimal] [Suite:k8s]'",
```
deads2k, line 327: I don't understand how this one tests the service network.  

1. test to external LB to create deployment
2. kcm uses internal LB to create replicaset and pod
3. scheduler uses internal LB to scheduler pod
4. kubelet uses internal LB to get pods and report available status
5. test uses external LB to confirm available replicas

which hop uses the service network?

pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -646,7 +646,7 @@ func newCertRotationController(
 			Name:      "control-plane-node-admin-client-cert-key",
 			AdditionalAnnotations: certrotation.AdditionalAnnotations{
 				JiraComponent:                    "kube-apiserver",
-				AutoRegenerateAfterOfflineExpiry: "https://github.com/openshift/cluster-kube-apiserver-operator/pull/1631,'operator conditions kube-apiserver'",
+				AutoRegenerateAfterOfflineExpiry: "https://github.com/openshift/cluster-kube-apiserver-operator/pull/1631,'[sig-apps] Deployment RollingUpdateDeployment should delete old pods and create new ones [Conformance] [Suite:openshift/conformance/parallel/minimal] [Suite:k8s]'",
```
deads2k, line 649: I don't think this test uses the kubeconfig that this certificate is embedded into. this is for being able to use `oc` if you ssh to the node.

pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -698,7 +698,7 @@ func newCertRotationController(
 			Name:      "check-endpoints-client-cert-key",
 			AdditionalAnnotations: certrotation.AdditionalAnnotations{
 				JiraComponent:                    "kube-apiserver",
-				AutoRegenerateAfterOfflineExpiry: "https://github.com/openshift/cluster-kube-apiserver-operator/pull/1631,'operator conditions kube-apiserver'",
+				AutoRegenerateAfterOfflineExpiry: "https://github.com/openshift/cluster-kube-apiserver-operator/pull/1631,'[sig-apps] Deployment RollingUpdateDeployment should delete old pods and create new ones [Conformance] [Suite:openshift/conformance/parallel/minimal] [Suite:k8s]'",
```
deads2k, line 701: I don't think that this test confirms that the check endpoints controller (which may or may not still exist) can function.

pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -723,7 +723,7 @@ func newCertRotationController(
 			Name:      "node-system-admin-signer",
 			AdditionalAnnotations: certrotation.AdditionalAnnotations{
 				JiraComponent:                    "kube-apiserver",
-				AutoRegenerateAfterOfflineExpiry: "https://github.com/openshift/cluster-kube-apiserver-operator/pull/1631,'operator conditions kube-apiserver'",
+				AutoRegenerateAfterOfflineExpiry: "https://github.com/openshift/cluster-kube-apiserver-operator/pull/1631,'[sig-apps] Deployment RollingUpdateDeployment should delete old pods and create new ones [Conformance] [Suite:openshift/conformance/parallel/minimal] [Suite:k8s]'",
```
deads2k, line 726: same as client-cert comment above.

p0lyn0mial:
I can tag this PR to get it merged, as it simply updates some descriptions, but I'm not sure if that is the intended goal.

I think the goal is to ensure we actually have a list of tests that can validate whether certificates have been regenerated properly. To do that, I would have to carefully check each test listed here.

If you want me to confirm the tests, I would suggest changing just one item. That would make it much easier for me to check and make some progress.

vrutkovs:
The plan here is:
* have this PR merged to add annotations
* merge https://github.com/openshift/release/pull/58130 - it already has been rehearsed to ensure the tests are passing
* Update https://github.com/openshift/origin/pull/29213 with new nightly data to include annotations in TLS registry

pkg/operator/certrotationcontroller/certrotationcontroller.go
p0lyn0mial: @vrutkovs, if you could provide an explanation, that would be very helpful. 

I also think that adding a comment with an explanation for each certificate/test pair would be greatly beneficial for future us and for easier reviewing.

vrutkovs: Ah, sorry, previously this was using
>[sig-apps] Deployment RollingUpdateDeployment should delete old pods and create new ones [Conformance] [Suite:openshift/conformance/parallel/minimal] [Suite:k8s]

but deployment rollout doesn't use service network, instead its

> [sig-network] Services should serve a basic endpoint from pods [Conformance] [Suite:openshift/conformance/parallel/minimal] [Suite:k8s]

now.

I'm not sure its worth commenting tests (we probably should have an easy to understand alias), but there will be an update for certificate descriptions: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1763 - would it be sufficient? Do you want me to add comments on what these tests do in the code (and not touch the annotations)?

pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -135,7 +135,7 @@ func newCertRotationController(
 			Name:      "aggregator-client-signer",
```
p0lyn0mial, line 135: is this the signer used to sign `--proxy-client-cert` ?

pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -150,7 +150,7 @@ func newCertRotationController(
 			Name:      "kube-apiserver-aggregator-client-ca",
```
p0lyn0mial, line 150: is this the CA that is added to `--requestheader-client-ca-file` ?

pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -162,7 +162,7 @@ func newCertRotationController(
 			Name:      "aggregator-client",
```
p0lyn0mial, line 162: is this `--proxy-client-cert` ?

p0lyn0mial (commented):
okay, let's start with the aggregation layer. 

Could you remind me of the purpose of the certificates and explain how the`[sig-cli] oc adm new-project [apigroup:project.openshift.io][apigroup:authorization.openshift.io] [Suite:openshift/conformance/parallel]` test is going to validate each certificate?

pkg/operator/certrotationcontroller/certrotationcontroller.go
vrutkovs: Yes

pkg/operator/certrotationcontroller/certrotationcontroller.go
vrutkovs: Correct

pkg/operator/certrotationcontroller/certrotationcontroller.go
vrutkovs: Yes (that's the entire CA btw)

pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -134,8 +134,9 @@ func newCertRotationController(
 			Namespace: operatorclient.OperatorNamespace,
 			Name:      "aggregator-client-signer",
 			AdditionalAnnotations: certrotation.AdditionalAnnotations{
-				JiraComponent:                    "kube-apiserver",
-				AutoRegenerateAfterOfflineExpiry: "https://github.com/openshift/cluster-kube-apiserver-operator/pull/1631,'operator conditions openshift-apiserver'",
+				JiraComponent: "kube-apiserver",
+				// This test would ensure kube-apiserver request would be passed to openshift-apiserver
+				AutoRegenerateAfterOfflineExpiry: "https://github.com/openshift/cluster-kube-apiserver-operator/pull/1631,'[sig-cli] oc adm new-project [apigroup:project.openshift.io][apigroup:authorization.openshift.io] [Suite:openshift/conformance/parallel]'",
```
hasbro17, line 139: > // AutoRegenerateAfterOfflineExpiry contains a link to PR and an e2e test name which verifies
	// that TLS artifact is correctly regenerated after it has expired

Would the PR not be https://github.com/openshift/release/pull/58130 where we add and rehearse these tests to the cert rotation jobs?

vrutkovs: According to https://github.com/openshift/origin/blob/master/pkg/cmd/update-tls-artifacts/generate-owners/tlsmetadata/autoregenerate_after_expiry/requirement.go#L23 it needs to be the PR which added the annotation, not necessarily the one which runs these tests

PR #1760, opened 2024-10-28
AUTH-541: OIDC structured auth config

liouk:
This PR does the following:
- adds unit tests for the `AuthMetadata` config observer
- adds a new config observer for direct OIDC; the config observer:
   - detects auth type `OIDC` and the structured auth config CM in `openshift-config-managed` (created by the CAO in [cluster-authentication-operator#713](https://github.com/openshift/cluster-authentication-operator/pull/713))
   - copies it to `openshift-kube-apiserver` where it will be used as a revisioned configmap and synced to a static file on all KAS nodes
   - enables OIDC config via the `--authentication-config` CLI flag of the KAS pods
- modifies the `AuthMetadata` and `WebhookTokenAuthenticator` config observers to delete their respective resources when auth type `OIDC` is detected

Enhancement: https://github.com/openshift/enhancements/pull/1632

### DIFF
--- a/pkg/operator/configobservation/auth/external_oidc.go
+++ b/pkg/operator/configobservation/auth/external_oidc.go
@@ -1,0 +1,143 @@
+package auth
+
+import (
+	"fmt"
+	"path"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/api/features"
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/configobservation"
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/operatorclient"
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/klog/v2"
+)
+
+const (
+	SourceAuthConfigCMNamespace = "openshift-config-managed"
+	AuthConfigCMName            = "auth-config"
+	authConfigKeyName           = "auth-config.json"
+)
+
+var (
+	authConfigPath = []string{"apiServerArguments", "authentication-config"}
+)
+
+func NewObserveExternalOIDC(featureGateAccessor featuregates.FeatureGateAccess) configobserver.ObserveConfigFunc {
+	return (&externalOIDC{
+		featureGateAccessor: featureGateAccessor,
+	}).ObserveExternalOIDC
+}
+
+type externalOIDC struct {
+	featureGateAccessor featuregates.FeatureGateAccess
+}
+
+// ObserveExternalOIDC observes the authentication.config/cluster resource
+// and if the type field is set to OIDC, it configures an external OIDC provider
+// to the KAS pods by setting the --authentication-config apiserver argument. It also
+// takes care of synchronizing the structured auth config file into the apiserver's namespace
+// so that it gets mounted as a static file on each node.
+func (o *externalOIDC) ObserveExternalOIDC(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, _ []error) {
+	defer func() {
+		ret = configobserver.Pruned(ret, authConfigPath)
+	}()
+
+	if !o.featureGateAccessor.AreInitialFeatureGatesObserved() {
+		// if we haven't observed featuregates yet, return the existing
+		return existingConfig, nil
+	}
+
+	featureGates, err := o.featureGateAccessor.CurrentFeatureGates()
+	if err != nil {
+		return existingConfig, []error{err}
+	}
+
+	if !featureGates.Enabled(features.FeatureGateExternalOIDC) {
+		return existingConfig, nil
+	}
+
+	listers := genericListers.(configobservation.Listers)
+	auth, err := listers.AuthConfigLister.Get("cluster")
+	if errors.IsNotFound(err) {
+		recorder.Eventf("ObserveExternalOIDC", "authentications.config.openshift.io/cluster: not found")
+		klog.Warningf("authentications.config.openshift.io/cluster: not found")
+		return existingConfig, nil
+	} else if err != nil {
+		return existingConfig, []error{err}
+	}
+
+	targetAuthConfig, err := listers.ConfigMapLister().ConfigMaps(operatorclient.TargetNamespace).Get(AuthConfigCMName)
+	if err != nil && !errors.IsNotFound(err) {
+		return existingConfig, []error{err}
+	}
+
+	if auth.Spec.Type != configv1.AuthenticationTypeOIDC {
+		// empty source name/namespace effectively deletes target configmap
+		if err := genericListers.ResourceSyncer().SyncConfigMap(
+			resourcesynccontroller.ResourceLocation{Namespace: operatorclient.TargetNamespace, Name: AuthConfigCMName},
+			resourcesynccontroller.ResourceLocation{Namespace: "", Name: ""},
+		); err != nil {
+			return existingConfig, []error{err}
+		}
+
+		if targetAuthConfig != nil {
+			recorder.Eventf("ObserveExternalOIDC", "OIDC auth configmap %s/%s exists; requested deletion", operatorclient.TargetNamespace, AuthConfigCMName)
+		}
+
+		return nil, nil
+	}
+
+	// auth type is OIDC
+
+	sourceAuthConfig, err := validateSourceConfigMap(listers)
+	if err != nil {
+		return existingConfig, []error{err}
+
+	} else if sourceAuthConfig == nil {
+		klog.Warningf("configmap %s/%s not found; skipping configuration of OIDC", SourceAuthConfigCMNamespace, AuthConfigCMName)
+		return existingConfig, nil
+	}
+
+	if err := genericListers.ResourceSyncer().SyncConfigMap(
+		resourcesynccontroller.ResourceLocation{Namespace: operatorclient.TargetNamespace, Name: AuthConfigCMName},
+		resourcesynccontroller.ResourceLocation{Namespace: sourceAuthConfig.Namespace, Name: sourceAuthConfig.Name},
+	); err != nil {
+		return existingConfig, []error{err}
+	}
+
+	if targetAuthConfig == nil {
+		recorder.Eventf("ObserveExternalOIDC", "OIDC auth configmap %s/%s does not exist; requested sync", operatorclient.TargetNamespace, AuthConfigCMName)
+	}
+
+	observedConfig := make(map[string]interface{})
+	if err := unstructured.SetNestedStringSlice(observedConfig, []string{path.Join("/etc/kubernetes/static-pod-resources/configmaps/", AuthConfigCMName, authConfigKeyName)}, authConfigPath...); err != nil {
+		return existingConfig, []error{err}
+	}
+
+	return observedConfig, nil
+}
+
+func validateSourceConfigMap(listers configobservation.Listers) (*corev1.ConfigMap, error) {
+	sourceAuthConfig, err := listers.ConfigMapLister().ConfigMaps(SourceAuthConfigCMNamespace).Get(AuthConfigCMName)
+	if errors.IsNotFound(err) {
+		return nil, nil
+	} else if err != nil {
+		return nil, fmt.Errorf("failed to get configmap %s/%s: %v", SourceAuthConfigCMNamespace, AuthConfigCMName, err)
+	}
+
+	if data, found := sourceAuthConfig.Data[authConfigKeyName]; !found {
+		return nil, fmt.Errorf("configmap %s/%s is invalid: key '%s' missing", SourceAuthConfigCMNamespace, AuthConfigCMName, authConfigKeyName)
+
+	} else if len(data) == 0 {
+		return nil, fmt.Errorf("configmap %s/%s is invalid: key '%s' has empty value", SourceAuthConfigCMNamespace, AuthConfigCMName, authConfigKeyName)
+	}
+
+	return sourceAuthConfig, nil
+}
--- a/pkg/operator/configobservation/auth/webhook_authenticator.go
+++ b/pkg/operator/configobservation/auth/webhook_authenticator.go
@@ -11,6 +11,7 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 
+	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/library-go/pkg/operator/configobserver"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
@@ -51,6 +52,7 @@ func ObserveWebhookTokenAuthenticator(genericListers configobserver.Listers, rec
 
 	auth, err := listers.AuthConfigLister.Get("cluster")
 	if errors.IsNotFound(err) {
+		recorder.Eventf("ObserveWebhookTokenAuthenticator", "authentications.config.openshift.io/cluster: not found")
 		return observedConfig, nil
 	} else if err != nil {
 		return existingConfig, append(errs, err)
@@ -62,7 +64,7 @@ func ObserveWebhookTokenAuthenticator(genericListers configobserver.Listers, rec
 	}
 
 	observedWebhookConfigured := len(webhookSecretName) > 0
-	if observedWebhookConfigured {
+	if observedWebhookConfigured && auth.Spec.Type != configv1.AuthenticationTypeOIDC {
 		// retrieve the secret from config and validate it, don't proceed on failure
 		kubeconfigSecret, err := listers.ConfigSecretLister().Secrets("openshift-config").Get(webhookSecretName)
 		if err != nil {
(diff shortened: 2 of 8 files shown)


xingxingxia:
Tried to install test cluster using Cluser-bot, but installations always fail. Per https://github.com/openshift/installer/blob/640266e/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template#L92 , I executed it against the payload built from our PRs, it shows 1.30.5 which seems incorrect:
```
$ oc adm release info -o 'jsonpath={.displayVersions.kubernetes.Version}' "${MIRROR_FLAG}" registry.buildxxx.xxxx.org/ci-ln-wqp6qzk/release:latest
1.30.5
```

However I executed against other QE's on going tested 4.18 pre-merge payload , they show correct 1.31.2:
```
$ oc adm release info -o 'jsonpath={.displayVersions.kubernetes.Version}' "${MIRROR_FLAG}" registry.buildxxx.xxxx.org/ci-ln-1zzh6w2/release:latest
1.31.2
```

liouk:
Investigating issue found in https://issues.redhat.com/browse/OCPBUGS-44592, will hold.

/hold

liouk:
> Investigating issue found in https://issues.redhat.com/browse/OCPBUGS-44592, will hold.

Issue [fixed](https://github.com/openshift/cluster-kube-apiserver-operator/compare/ef36c41e20bc737a2cfd17230bc2c41385b1ab39..5cdfb5e78892398989fcc6bc3efa7f09030e081c), cancelling hold.

/hold cancel

pkg/operator/configobservation/auth/external_oidc.go
```diff
@@ -0,0 +1,159 @@
...
+		return existingConfig, []error{err}
+	}
+
+	if auth.Spec.Type != configv1.AuthenticationTypeOIDC {
+		if _, err := listers.ConfigMapLister().ConfigMaps(operatorclient.TargetNamespace).Get(AuthConfigCMName); errors.IsNotFound(err) {
+			return nil, nil
+
+		} else if err != nil {
+			return existingConfig, []error{fmt.Errorf("failed to get configmap %s/%s: %v", operatorclient.TargetNamespace, AuthConfigCMName, err)}
+		}
```
benluddy, line 78: Will the following call to `syncConfigMap` do effectively the same thing? What's the advantage of short-circuiting here?

pkg/operator/configobservation/auth/external_oidc.go
```diff
@@ -0,0 +1,159 @@
...
+
+		return nil, nil
+	}
+
+	// auth type is OIDC
+
+	sourceAuthConfig, err := listers.ConfigMapLister().ConfigMaps(SourceAuthConfigCMNamespace).Get(AuthConfigCMName)
+	if errors.IsNotFound(err) {
+		klog.Warningf("configmap %s/%s not found; skipping configuration of OIDC", SourceAuthConfigCMNamespace, AuthConfigCMName)
```
benluddy, line 92: For my understanding, if the target CM exists and the source CM is missing, the resource syncer will eventually delete the target CM asynchronously, right?

pkg/operator/configobservation/auth/external_oidc.go
```diff
@@ -0,0 +1,159 @@
...
+	if err := validateAuthConfigCM(sourceAuthConfig); err != nil {
+		return existingConfig, []error{fmt.Errorf("configmap %s/%s is invalid: %v", SourceAuthConfigCMNamespace, AuthConfigCMName, err)}
+	}
+
+	targetAuthConfig, err := listers.ConfigMapLister().ConfigMaps(operatorclient.TargetNamespace).Get(AuthConfigCMName)
+	if err != nil && !errors.IsNotFound(err) {
+		return existingConfig, []error{err}
+	}
+
+	if targetAuthConfig == nil || !equality.Semantic.DeepEqual(targetAuthConfig.Data, sourceAuthConfig.Data) {
```
benluddy, line 108: I would expect the syncer to be responsible for deciding whether or not it needs to do a write. Why test here?

pkg/operator/configobservation/auth/external_oidc.go
```diff
@@ -0,0 +1,159 @@
...
+		recorder.Eventf("ObserveExternalOIDC", "synced configmap %s/%s to %s/%s", sourceNamespace, sourceName, operatorclient.TargetNamespace, AuthConfigCMName)
+	}
+
+	return nil
+}
+
+func validateAuthConfigCM(cm *corev1.ConfigMap) error {
+	if cm == nil {
+		return fmt.Errorf("configmap is nil")
+	}
```
benluddy, line 142: Nit: The caller is already handling "not found", so we never expect this to be called with a nil pointer. Would it be fine to let this panic on a nil pointer dereference since that would violate our assumptions about how it is used?

pkg/operator/configobservation/auth/external_oidc.go
```diff
@@ -0,0 +1,159 @@
...
+	authConfigRaw, ok := cm.Data["auth-config.json"]
+	if !ok {
+		return fmt.Errorf("missing required 'auth-config.json' key")
+	}
+
+	if len(authConfigRaw) == 0 {
+		return fmt.Errorf("value of key 'auth-config.json' is empty")
+	}
+
+	var authConfig apiserver.AuthenticationConfiguration
```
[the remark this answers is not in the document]
liouk: https://pkg.go.dev/k8s.io/apiserver/pkg/apis/apiserver/load only contains `Load*()` funcs for the `AuthorizationConfiguration` configuration type, instead of `AuthenticationConfiguration` which we need here.

pkg/operator/configobservation/configobservercontroller/observe_config_controller.go
```diff
@@ -38,7 +39,7 @@ type ConfigObserver struct {
 	factory.Controller
 }
 
-func NewConfigObserver(operatorClient v1helpers.StaticPodOperatorClient, kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces, configInformer configinformers.SharedInformerFactory, operatorInformer operatorv1informers.SharedInformerFactory, resourceSyncer resourcesynccontroller.ResourceSyncer, featureGateAccessor featuregates.FeatureGateAccess, eventRecorder events.Recorder, groupVersionsByFeatureGate map[configv1.FeatureGateName][]schema.GroupVersion) *ConfigObserver {
```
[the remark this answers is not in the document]
liouk: Good catch, leftover from previous changes I'd presume.

pkg/operator/configobservation/auth/webhook_authenticator.go
```diff
@@ -62,7 +63,7 @@ func ObserveWebhookTokenAuthenticator(genericListers configobserver.Listers, rec
 	}
 
 	observedWebhookConfigured := len(webhookSecretName) > 0
-	if observedWebhookConfigured {
+	if observedWebhookConfigured && auth.Spec.Type != configv1.AuthenticationTypeOIDC {
```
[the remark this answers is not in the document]
liouk: That comment is misleading indeed -- this relies on a validation admission plugin: https://github.com/openshift/kubernetes/blob/8540bd493e8a42ffb61a3f470f71290488332b1c/openshift-kube-apiserver/admission/customresourcevalidation/authentication/validate_authentication.go#L103-L130

pkg/operator/configobservation/auth/external_oidc.go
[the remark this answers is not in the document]
benluddy: Oh, sorry, missed that. Then you'll need to set up the codec yourself here.

pkg/operator/configobservation/auth/external_oidc.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	if !featureGates.Enabled(features.FeatureGateExternalOIDC) {
+		return existingConfig, nil
+	}
+
+	listers := genericListers.(configobservation.Listers)
+	auth, err := listers.AuthConfigLister.Get("cluster")
+	if errors.IsNotFound(err) {
+		klog.Warningf("authentications.config.openshift.io/cluster: not found")
```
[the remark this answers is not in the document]
liouk: I followed the example of auth_metadata and webhook_authenticator config observers that don't record events for that case; I'll add events to all three of them, since that's what we usually do.

pkg/operator/configobservation/auth/external_oidc.go
liouk: I wanted to avoid another sync when we know we don't have that CM anymore, mainly because the syncer seems to be [doing a Get API call](https://github.com/liouk/cluster-kube-apiserver-operator/blob/5cdfb5e78892398989fcc6bc3efa7f09030e081c/vendor/github.com/openshift/library-go/pkg/operator/resource/resourceapply/core.go#L472).

pkg/operator/configobservation/auth/external_oidc.go
liouk: Yes, that's correct, This check exists here to guard from the case where type has been set to OIDC but CAO isn't ready yet and the source CM hasn't appeared yet -- no point in trying to configure OIDC until it does.

pkg/operator/configobservation/auth/external_oidc.go
liouk: As per my previous comment, I wanted to avoid an extra API call that the syncer does.

xingxingxia:
From test result perspective, based on good pre-merge test results in https://issues.redhat.com/browse/OCPBUGS-44592?focusedId=26134688&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-26134688 , adding below label:
/label qe-approved

pkg/operator/configobservation/auth/external_oidc.go
```diff
@@ -0,0 +1,137 @@
...
+		return existingConfig, nil
+
+	} else if err != nil {
+		return existingConfig, []error{fmt.Errorf("failed to get configmap %s/%s: %v", SourceAuthConfigCMNamespace, AuthConfigCMName, err)}
+	}
+
+	authConfigRaw := sourceAuthConfig.Data[authConfigKeyName]
+	if len(authConfigRaw) == 0 {
+		return existingConfig, []error{fmt.Errorf("configmap %s/%s is invalid: key '%s' missing or value empty", SourceAuthConfigCMNamespace, AuthConfigCMName, authConfigKeyName)}
+	}
```
everettraven, line 101: Nit: Make "missing" and "empty value" two distinct errors
```suggestion
	authConfigRaw, ok := sourceAuthConfig.Data[authConfigKeyName]
	if !ok {
		return existingConfig, []error{fmt.Errorf("configmap %s/%s is invalid: key '%s' missing", SourceAuthConfigCMNamespace, AuthConfigCMName, authConfigKeyName)}
	}
	if len(authConfigRaw) == 0 {
		return existingConfig, []error{fmt.Errorf("configmap %s/%s is invalid: key '%s' has empty value", SourceAuthConfigCMNamespace, AuthConfigCMName, authConfigKeyName)}
	}
```

pkg/operator/configobservation/auth/external_oidc.go
everettraven: My understanding is that the resource syncer is already configured with a configmap client that is backed by a cache.

Following the crumbs, it looks like the [resource sync controller we setup for the config observer](https://github.com/openshift/cluster-kube-apiserver-operator/blob/1c44cc51cbbd5e565c3ecf13b277f3a0fbfe20a4/pkg/operator/starter.go#L190) is [instantiated](https://github.com/openshift/cluster-kube-apiserver-operator/blob/1c44cc51cbbd5e565c3ecf13b277f3a0fbfe20a4/pkg/operator/starter.go#L169) using the local [resourcesynccontroller.NewResourceSyncController](https://github.com/openshift/cluster-kube-apiserver-operator/blob/1c44cc51cbbd5e565c3ecf13b277f3a0fbfe20a4/pkg/operator/resourcesynccontroller/resourcesynccontroller.go#L13) function, that configures a [library-go ResourceSyncController](https://github.com/openshift/cluster-kube-apiserver-operator/blob/1c44cc51cbbd5e565c3ecf13b277f3a0fbfe20a4/pkg/operator/resourcesynccontroller/resourcesynccontroller.go#L19) using a [caching ConfigMap getter](https://github.com/openshift/cluster-kube-apiserver-operator/blob/1c44cc51cbbd5e565c3ecf13b277f3a0fbfe20a4/pkg/operator/resourcesynccontroller/resourcesynccontroller.go#L24) which is implemented in library go [here](https://github.com/openshift/library-go/blob/0f76e23726cdb7c320fbd964f47a38ab8d69e2a4/pkg/operator/v1helpers/core_getters.go#L25)

From what I can tell, this extra guard check is not necessary the way things are currently configured. That being said, someone could come along and change it to no longer be a cached configmap getter and then it would be live calls. But I doubt that will be something that sneaks in and shouldn't be the concern of this controller.

pkg/operator/configobservation/auth/external_oidc.go
everettraven: As mentioned in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1760/files#r1931117915, I don't think the syncer is actually performing an extra API call. It certainly _looks_ like it does but we configure the configmap getter as a caching configmap getter already.

pkg/operator/configobservation/auth/external_oidc.go
everettraven: I meant to put https://github.com/openshift/cluster-kube-apiserver-operator/pull/1760/files#r1931117915 here, but got lost in my review pane :facepalm:

everettraven: My understanding is that the resource syncer is already configured with a configmap client that is backed by a cache.

Following the crumbs, it looks like the [resource sync controller we setup for the config observer](https://github.com/openshift/cluster-kube-apiserver-operator/blob/1c44cc51cbbd5e565c3ecf13b277f3a0fbfe20a4/pkg/operator/starter.go#L190) is [instantiated](https://github.com/openshift/cluster-kube-apiserver-operator/blob/1c44cc51cbbd5e565c3ecf13b277f3a0fbfe20a4/pkg/operator/starter.go#L169) using the local [resourcesynccontroller.NewResourceSyncController](https://github.com/openshift/cluster-kube-apiserver-operator/blob/1c44cc51cbbd5e565c3ecf13b277f3a0fbfe20a4/pkg/operator/resourcesynccontroller/resourcesynccontroller.go#L13) function, that configures a [library-go ResourceSyncController](https://github.com/openshift/cluster-kube-apiserver-operator/blob/1c44cc51cbbd5e565c3ecf13b277f3a0fbfe20a4/pkg/operator/resourcesynccontroller/resourcesynccontroller.go#L19) using a [caching ConfigMap getter](https://github.com/openshift/cluster-kube-apiserver-operator/blob/1c44cc51cbbd5e565c3ecf13b277f3a0fbfe20a4/pkg/operator/resourcesynccontroller/resourcesynccontroller.go#L24) which is implemented in library go [here](https://github.com/openshift/library-go/blob/0f76e23726cdb7c320fbd964f47a38ab8d69e2a4/pkg/operator/v1helpers/core_getters.go#L25)

From what I can tell, this extra guard check is not necessary the way things are currently configured. That being said, someone could come along and change it to no longer be a cached configmap getter and then it would be live calls. But I doubt that will be something that sneaks in and shouldn't be the concern of this controller.

liouk:
Holding, contains fixups.

/hold

pkg/operator/configobservation/auth/external_oidc.go
```diff
@@ -0,0 +1,126 @@
...
+	} else if err != nil {
+		return existingConfig, []error{fmt.Errorf("failed to get configmap %s/%s: %v", SourceAuthConfigCMNamespace, AuthConfigCMName, err)}
+	}
+
+	if data, found := sourceAuthConfig.Data[authConfigKeyName]; !found {
+		return existingConfig, []error{fmt.Errorf("configmap %s/%s is invalid: key '%s' missing", SourceAuthConfigCMNamespace, AuthConfigCMName, authConfigKeyName)}
+
+	} else if len(data) == 0 {
+		return existingConfig, []error{fmt.Errorf("configmap %s/%s is invalid: key '%s' has empty value", SourceAuthConfigCMNamespace, AuthConfigCMName, authConfigKeyName)}
+	}
```
everettraven, line 107: Nit: Wrap this in a separate function call that performs pre-sync validations?

everettraven (approved):
Only have a minor nit - will leave it up to you if you think the change is worth making. Other than that, the actual code LGTM

everettraven (approved):
/lgtm

liouk:
Squashed all fixups.

/hold cancel

tkashem:
/label acknowledge-critical-fixes-only
/hold

(feel free to cancel when you think it's ready to be merged)

dinhxuanvu:
The serial test looks perma-failed for this PR. It hasn't passed a single time since the last commit. Retesting may not yield a better result.
You may want to take a deeper look to see what the issue may be @liouk.

everettraven:
/hold

Need time to investigate the failures that seem to be related to the kube-apiserver rollout. As @dinhxuanvu mentioned, this test is permafailing.

pkg/operator/configobservation/auth/external_oidc.go
```diff
@@ -0,0 +1,138 @@
...
+type externalOIDC struct {
+	featureGateAccessor featuregates.FeatureGateAccess
+}
+
+// ObserveExternalOIDC observes the authentication.config/cluster resource
+// and if the type field is set to OIDC, it configures an external OIDC provider
+// to the KAS pods by setting the --authentication-config apiserver argument. It also
+// takes care of synchronizing the structured auth config file into the apiserver's namespace
+// so that it gets mounted as a static file on each node.
+func (o *externalOIDC) ObserveExternalOIDC(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
```
everettraven, line 46: It looks like this might config observation function is triggering a non-deterministic config error as seen in: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-kube-apiserver-operator/1760/pull-ci-openshift-cluster-kube-apiserver-operator-master-e2e-aws-ovn-serial/1895151853344133120/artifacts/e2e-aws-ovn-serial/gather-extra/artifacts/pods/openshift-kube-apiserver-operator_kube-apiserver-operator-6cdbc46584-g95qs_kube-apiserver-operator.log (search "non-deterministic config observation")

I went to look at examples of other config observers and saw a pattern like this at the top of the methods: 
https://github.com/openshift/cluster-kube-apiserver-operator/blob/56e73468f10d59550b3dcf2ac23b69068489c40e/pkg/operator/configobservation/auth/podsecurityadmission.go#L83-L85

It looks like config observers are meant to only return configuration paths that they may interact with and returning the entire thing will muck with the config merging process that happens in https://github.com/openshift/library-go/blob/31d9e3b0ae4b8dd00fa030560ea9501dd0e81549/pkg/operator/configobserver/config_observer_controller.go#L160-L176

This `Prune()` function was introduced in https://github.com/openshift/library-go/pull/744

I think updating this method to have:
```suggestion
func (o *externalOIDC) ObserveExternalOIDC(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, []error) {
	authConfigPath := []string{apiServerArgumentsPath, argAuthConfig}
	defer func() {
		ret = configobserver.Pruned(ret, authConfigPath)
	}()
```

should resolve this issue.

everettraven (approved):
/lgtm

everettraven (approved):
/lgtm

PR #1764, opened 2024-11-07
API-1835: migrate startup monitor conditon to ssa 

p0lyn0mial:
requires https://github.com/openshift/library-go/pull/1870


bertinatto (commented):
/lgtm

Issue #1564, opened 2023-09-30
Skip generate cert when network config status.serviceNetwork is nil 


lance5890:
1. When the network config status.serviceNetwork is nil , the problems is shown as follows:
```
Sep 20 15:35:41 master1 hyperkube[8624]: E0920 15:35:41.603133    8624 remote_runtime.go:222] "RunPodSandbox from runtime service failed" err="rpc error: code = Unknown desc = failed to create pod network sandbox k8s_karmada-aggregated-apiserver-7f6c7d8dd5-gm7gb_karmada-system_ff9e40b2-245a-4b7c-95d4-917bf94cf3a6_0(b1b5310d3645d82e7c4f468cf48bd512a1f226c6de7d57ee312a8bc9fbec6c01): error adding pod karmada-system_karmada-aggregated-apiserver-7f6c7d8dd5-gm7gb to CNI network \"multus-cni-network\": plugin type=\"multus\" name=\"multus-cni-network\" failed (add): [karmada-system/karmada-aggregated-apiserver-7f6c7d8dd5-gm7gb/ff9e40b2-245a-4b7c-95d4-917bf94cf3a6:k8s-pod-network]: error adding container to network \"k8s-pod-network\": error getting ClusterInformation: Get \"https://21.101.0.1:443/apis/crd.projectcalico.org/v1/clusterinformations/default\": x509: cannot validate certificate for 21.101.0.1 because it doesn't contain any IP SANs"

```
2. and the network cr is shown as follows:
```
oc get network cluster -oyaml
apiVersion: config.openshift.io/v1
kind: Network
metadata:
  creationTimestamp: "2023-09-20T07:10:47Z"
  generation: 1
  name: cluster
  resourceVersion: "715"
  uid: bb3082a0-cebb-4a64-b662-db7ce2e76837
spec:
  apiAddress: 10.255.71.88
  clusterNetwork:
  - cidr: 21.100.0.0/16
    hostPrefix: 24
  externalIP:
    policy: {}
  serviceNetwork:
  - 21.101.0.0/16
status: {}
```
3. and the service-network-serving-certkey secret is shown as follows:
```
oc get secret service-network-serving-certkey -n openshift-kube-apiserver -oyaml
apiVersion: v1
data:
  tls.crt: 
  tls.key: 
kind: Secret
metadata:
  annotations:
[...]

PR #1776, opened 2024-12-10
OCPBUGS-45943: skip generating certs when networkConfig.status.ServiceNetwork is nil

lance5890:
Skip when the network config status.serviceNetwork is nil

related issue: https://github.com/openshift/cluster-kube-apiserver-operator/issues/1564


lance5890:
@vrutkovs

vrutkovs (approved):
/approve
/lgtm

PR #1796, opened 2025-02-03
OCPBUGS-49773: Bump library-go for static pod controller apply fixes.

benluddy:


benluddy:
/hold

https://github.com/openshift/cluster-kube-apiserver-operator/pull/1795

PR #1810, opened 2025-02-14
CNTRLPLANE-79: Disable oauth admission plugins

liouk:
When auth type is not OAuth (i.e. `OIDC` or `None`), we must disable the plugins relevant for RoleBindingRestrictions, as this API relies on the User & Group APIs, which are part of the OAuth apiserver.

This PR does the following:
- adds a generic config observer for admission plugins, which manages the `apiServerArguments.enable-admission-plugins` and `apiServerArguments.disable-admission-plugins` fields of the KAS config
- adds two special merge rules for the fields mentioned above
- removes RoleBindingRestriction related plugins from `defaultconfig.yaml`; these are now managed by the admission plugins config observer

**_Why a generic admission plugin config observer and not specific to the RBR plugins?_**

There are mainly two reasons behind this choice:
1. Follow the config observer pattern where each observer manages configuration fields as a whole and not partially. Thus, this observer will manage the two fields as a whole and not only a subset of their potential contents.
2. Provide a generic structure and central point that can be reused in the future for other admission plugins.

**_Why do we need special merge rules?_**

The `resourcemerge` which is used to merge all config sources into one configmap, does not merge contents of slices; if a field is a slice, it will replace the contents completely. Defining special merge rules for the enabled/disabled admission plugins allows us to merge together lists of plugins that come from various sources (e.g. the default config and the admission plugin config observer).

**_Why remove the RoleBindingRestrictions from the default config?_**

Anything that is part of the default config will _not_ be passed as an input to a config observer. Therefore, in order to disable the plugins, we need to both remove them from the enabled list, but also add them to the disabled list as well. The simplest way to do that is to have the config observer decide whether to enable or disable these plugins.

### DIFF
--- a/pkg/operator/configobservation/apiserver/observe_admission_plugins.go
+++ b/pkg/operator/configobservation/apiserver/observe_admission_plugins.go
@@ -1,0 +1,99 @@
+package apiserver
+
+import (
+	"fmt"
+	"slices"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/configobservation"
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+type pluginCheckerFunc func(listers configobservation.Listers) (enabled, disabled []string, err error)
+
+var (
+	enableAdmissionPluginsPath  = []string{"apiServerArguments", "enable-admission-plugins"}
+	disableAdmissionPluginsPath = []string{"apiServerArguments", "disable-admission-plugins"}
+
+	pluginCheckers = []pluginCheckerFunc{
+		roleBindingRestrictionPluginChecker,
+	}
+)
+
+// ObserveAdmissionPlugins manages the apiServerArguments.enable-admission-plugins and
+// apiServerArguments.disable-admission-plugins fields of the configuration. It defines a list of
+// plugin checkers which check the state of specific plugins, and add them to the enabled or disabled
+// list as required. This observer will overwrite any pre-existing values of the two fields in the existingConfig.
+func ObserveAdmissionPlugins(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]any) (ret map[string]any, _ []error) {
+	defer func() {
+		ret = configobserver.Pruned(ret, enableAdmissionPluginsPath, disableAdmissionPluginsPath)
+	}()
+
+	if len(pluginCheckers) == 0 {
+		return existingConfig, nil
+	}
+
+	listers := genericListers.(configobservation.Listers)
+	enabledSet := sets.New[string]()
+	disabledSet := sets.New[string]()
+
+	for _, pluginChecker := range pluginCheckers {
+		enabled, disabled, err := pluginChecker(listers)
+		if err != nil {
+			return existingConfig, []error{err}
+		}
+
+		enabledSet.Insert(enabled...)
+		disabledSet.Insert(disabled...)
+	}
+
+	if intersection := enabledSet.Intersection(disabledSet); intersection.Len() > 0 {
+		return existingConfig, []error{fmt.Errorf("plugins cannot be enabled and disabled at the same time: %v", intersection.UnsortedList())}
+	}
+
+	observedConfig := map[string]any{}
+
+	if enabledSet.Len() > 0 {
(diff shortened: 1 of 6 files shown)


pkg/operator/configobservation/apiserver/observe_admission_plugins.go
```diff
@@ -0,0 +1,92 @@
...
+		ret = configobserver.Pruned(ret, enableAdmissionPluginsPath, disableAdmissionPluginsPath)
+	}()
+
+	if len(pluginCheckers) == 0 {
+		return existingConfig, nil
+	}
+
+	listers := genericListers.(configobservation.Listers)
+	enabledSet := sets.NewString()
+	disabledSet := sets.NewString()
```
sanchezl, line 41: There are now generic versions of these. Just be aware of the sorting the values if needed (e.g. `slices.Sorted(slices.Values(enabledSet.UnsortedList()))`)

sanchezl (changes requested):
Can you comment on how the config overrides will work with the special merge? Otherwise lgtm.

pkg/operator/configobservation/apiserver/observe_admission_plugins.go
liouk: Thanks, good to know! Updated to use the new generic versions of the sets.

pkg/operator/configobservation/apiserver/observe_admission_plugins.go
```diff
@@ -0,0 +1,95 @@
...
+		disabledSet.Insert(disabled...)
+	}
+
+	if intersection := enabledSet.Intersection(disabledSet); intersection.Len() > 0 {
+		return existingConfig, []error{fmt.Errorf("plugins cannot be enabled and disabled at the same time: %v", intersection.UnsortedList())}
+	}
+
+	observedConfig := map[string]any{}
+
+	if enabledSet.Len() > 0 {
```
everettraven, line 60: if the length of the enabled set is 0, should this ensure that the returned configuration is that this nested slice is actually empty? Is that the same as what just not setting it does? Should there be a test for the case where a previous config observer may have set this and this overrides it to empty when nothing is returned as being in the enabled set?

pkg/operator/configobservation/apiserver/observe_admission_plugins.go
```diff
@@ -0,0 +1,95 @@
...
+
+	pluginCheckers = []pluginCheckerFunc{
+		roleBindingRestrictionPluginChecker,
+	}
+)
+
+// ObserveAdmissionPlugins manages the apiServerArguments.enable-admission-plugins and
+// apiServerArguments.disable-admission-plugins fields of the configuration. It defines a list of
+// plugin checkers which check the state of specific plugins, and add them to the enabled or disabled
+// list as required. This observer will overwrite any pre-existing values of the two fields in the existingConfig.
```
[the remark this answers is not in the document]
liouk: The standard set of enabled admission plugins gets enabled by `defaultconfig.yaml` and any overrides that are in place; such plugins will not appear in the `existingConfig`, as this contains the value of the `kubeapiservers/cluster.ObservedConfig` field, which is populated by configobservation in the first place. This is also why we can't simply have the configobserver simply drop these plugins from any pre-existing configuration, and we need a special merge.

Also, the goal is to provide a central management place for `enable/disable-admission-plugins` so that other observers are discouraged from modifying these fields. Unfortunately the config observers are processed in a random order, and merging the output of config observers would overwrite slices, therefore there can always be a config observer that takes over these fields; in the end there's no guarantee that any other config observer will respect the output of this one.

pkg/operator/configobservation/apiserver/observe_admission_plugins.go
liouk: I aimed to preserve a similar behavior as the existing setup; currently, for example, the `--disable-admission-plugins` field is not set in the default config or overrides, and this results in the field missing from the merged configuration. I wrote this config observer in a similar way; if there's no plugins to enable/disable, don't set the flag at all.

liouk:
@sanchezl 
> Can you comment on how the config overrides will work with the special merge? Otherwise lgtm.

The config overrides are [one of the configs being merged](https://github.com/openshift/cluster-kube-apiserver-operator/pull/1810/files#diff-0af15f182aea9898e784fc589175ef4c2411c9391bb5a1f41327ba7c3e2172f5R286-R296), which means that the special merge will honor them as well. Basically, the special merge will execute each time the selected path is encountered in one of the configs being merged.

I added a [unit test](https://github.com/openshift/cluster-kube-apiserver-operator/pull/1810/files#diff-cde314791a8ec62ee3dcda3351c2fdc297c59cf3a924da9a9a3d5131cdbe21e3R427-R482) to illustrate the usage of `resourcemerge.MergePrunedConfigMap()` in the presence of these merge rules.

PR #1815, opened 2025-02-20
MON-4129: slos: accomodate for Prometheus v3 "le" normalization

machine424:
ensure all series involved in the different queries change
    during the integer->float transition so that rate calculation remains consistent across all series.
    
    If apiserver_request_sli_duration_seconds_bucket{le="1"} had a last value of 15 and
    then apiserver_request_sli_duration_seconds_bucket{le=~"1.0"} reappeared with 20, the
    rate calculated over a range where both {le="1"} and {le="1.0"} overlap will not
    account for the 20−15=5 difference, as the two series are distinct. But
    apiserver_request_sli_duration_seconds_count's rate will still take
    that 5 jump into account as the series doesn't change.
    
    Replace apiserver_request_sli_duration_seconds_count with
    apiserver_request_sli_duration_seconds_bucket{le=~"60(.0)?"}
    since they should be equal given that the timeout is 60s and cannot be customized.
    
    This change is temporary to avoid silencing alerts or having to reset/forget historical integer buckets during the transition.
    
    Later, we'll revert back to using apiserver_request_sli_duration_seconds_count.

### DIFF
--- a/bindata/assets/alerts/kube-apiserver-slos-basic.yaml
+++ b/bindata/assets/alerts/kube-apiserver-slos-basic.yaml
@@ -43,18 +43,18 @@ spec:
         (
           (
             # too slow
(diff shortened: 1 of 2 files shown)


dgrisonnet (approved):
/lgtm
/approve

PR #1816, opened 2025-02-26
MON-4129: slos: move to float buckets as Prometheus v3 normalized integer->float

machine424:
no longer considering the old integer buckets is intentional

ensure all series involved in the different queries change during the integer->float transition so that rate calculation remains consistent across all series.

If apiserver_request_sli_duration_seconds_bucket{le="1"} had a last value of 15 and then apiserver_request_sli_duration_seconds_bucket{le="1.0"} reappeared with 20, the rate of {le="1.0"} over a range that covers the transition  will not account for the 20−15=5 difference, as the two {le="1"} and {le="1.0"} series are distinct. But apiserver_request_sli_duration_seconds_count's rate will still take that 5 jump into account as the series doesn't change.

Replace apiserver_request_sli_duration_seconds_count with apiserver_request_sli_duration_seconds_bucket{le="60.0"} since they should be equal given that the timeout is 60s and cannot be customized.

This change is temporary to avoid silencing alerts during the transition.

Later, we'll revert back to using apiserver_request_sli_duration_seconds_count.

### DIFF
--- a/bindata/assets/alerts/kube-apiserver-slos-basic.yaml
+++ b/bindata/assets/alerts/kube-apiserver-slos-basic.yaml
@@ -43,18 +43,18 @@ spec:
         (
           (
             # too slow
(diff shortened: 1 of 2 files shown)


machine424:
/hold
until prometheus v3 is merged

PR #1825, opened 2025-03-18
ioutil deprecation

sebrandon1:
As of Go 1.16, `ioutil` was deprecated.  Read more about it [here](https://go.dev/doc/go1.16#ioutil).

Similar to:
- https://github.com/openshift-kni/cnf-features-deploy/pull/1777


dinhxuanvu (commented):
/lgtm

PR #1827, opened 2025-03-19
MON-4160: operator: add topology metrics

jan--f:
This commit adds two new metrics providing the two topology modes, control plane and infrastructure respectively.


vrutkovs (approved):
/approve
/lgtm

PR #1829, opened 2025-04-04
MON-4161: add `cluster:controlplane_topology:sum` recording rule

jan--f:
This very simple recording rule is used as a telemetry signal. This is best done via a recording rule so that non-relevant labels can be aggregated and thus reduce the cardinality in our telemetry storage.

/cc @vrutkovs


vrutkovs (approved):
/approve
/lgtm

vrutkovs (approved):

PR #1832, opened 2025-04-23
OCPBUGS-55274: Change validity / refresh for some certs which should not be affected by dev cycle rotation change

vrutkovs:
Some certificates should be excepted from rotation being set sizty times faster in development mode - i.e. localhost-recovery.kubeconfig


pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -136,10 +142,13 @@ func newCertRotationController(
 
 	if featureGates.Enabled(features.FeatureShortCertRotation) {
 		monthPeriod = 2 * time.Hour
+		devRotationExceptionYear = monthPeriod
```
hasbro17, line 145: Did you mean:
```suggestion
		devRotationExceptionMonth = monthPeriod
```

PR #1833, opened 2025-04-25
NO-JIRA: Use new PSS annotation

jacobsee:
Don't attempt to pick the stricter of the warn or audit PSS labels to determine whether the namespace is violating - look directly at the new MinimallySufficientPodSecurityStandard label instead.

### DIFF
--- a/pkg/operator/podsecurityreadinesscontroller/conditions.go
+++ b/pkg/operator/podsecurityreadinesscontroller/conditions.go
@@ -18,8 +18,12 @@ const (
 	PodSecurityOpenshiftType      = "PodSecurityOpenshiftEvaluationConditionsDetected"
 	PodSecurityRunLevelZeroType   = "PodSecurityRunLevelZeroEvaluationConditionsDetected"
 	PodSecurityDisabledSyncerType = "PodSecurityDisabledSyncerEvaluationConditionsDetected"
+	PodSecurityInconclusiveType   = "PodSecurityInconclusiveEvaluationConditionsDetected"
 
 	labelSyncControlLabel = "security.openshift.io/scc.podSecurityLabelSync"
+
+	violationReason    = "PSViolationsDetected"
+	inconclusiveReason = "PSViolationDecisionInconclusive"
 )
 
 var (
@@ -36,6 +40,7 @@ type podSecurityOperatorConditions struct {
 	violatingRunLevelZeroNamespaces   []string
 	violatingCustomerNamespaces       []string
 	violatingDisabledSyncerNamespaces []string
+	inconclusiveNamespaces            []string
 }
 
 func (c *podSecurityOperatorConditions) addViolation(ns *corev1.Namespace) {
@@ -59,16 +64,31 @@ func (c *podSecurityOperatorConditions) addViolation(ns *corev1.Namespace) {
 	c.violatingCustomerNamespaces = append(c.violatingCustomerNamespaces, ns.Name)
 }
 
-func makeCondition(conditionType string, namespaces []string) operatorv1.OperatorCondition {
+func (c *podSecurityOperatorConditions) addInconclusive(ns *corev1.Namespace) {
+	c.inconclusiveNamespaces = append(c.inconclusiveNamespaces, ns.Name)
+}
+
+func makeCondition(conditionType, conditionReason string, namespaces []string) operatorv1.OperatorCondition {
+	var messageFormatter string
+
+	switch conditionReason {
+	case violationReason:
+		messageFormatter = "Violations detected in namespaces: %v"
+	case inconclusiveReason:
+		messageFormatter = "Could not evaluate violations for namespaces: %v"
+	default:
+		messageFormatter = "Unexpected condition for namespace: %v"
+	}
+
 	if len(namespaces) > 0 {
 		sort.Strings(namespaces)
 		return operatorv1.OperatorCondition{
 			Type:               conditionType,
 			Status:             operatorv1.ConditionTrue,
 			LastTransitionTime: metav1.Now(),
-			Reason:             "PSViolationsDetected",
+			Reason:             conditionReason,
 			Message: fmt.Sprintf(
-				"Violations detected in namespaces: %v",
+				messageFormatter,
 				namespaces,
 			),
 		}
@@ -84,9 +104,10 @@ func makeCondition(conditionType string, namespaces []string) operatorv1.Operato
 
 func (c *podSecurityOperatorConditions) toConditionFuncs() []v1helpers.UpdateStatusFunc {
 	return []v1helpers.UpdateStatusFunc{
-		v1helpers.UpdateConditionFn(makeCondition(PodSecurityCustomerType, c.violatingCustomerNamespaces)),
-		v1helpers.UpdateConditionFn(makeCondition(PodSecurityOpenshiftType, c.violatingOpenShiftNamespaces)),
-		v1helpers.UpdateConditionFn(makeCondition(PodSecurityRunLevelZeroType, c.violatingRunLevelZeroNamespaces)),
-		v1helpers.UpdateConditionFn(makeCondition(PodSecurityDisabledSyncerType, c.violatingDisabledSyncerNamespaces)),
+		v1helpers.UpdateConditionFn(makeCondition(PodSecurityCustomerType, violationReason, c.violatingCustomerNamespaces)),
+		v1helpers.UpdateConditionFn(makeCondition(PodSecurityOpenshiftType, violationReason, c.violatingOpenShiftNamespaces)),
+		v1helpers.UpdateConditionFn(makeCondition(PodSecurityRunLevelZeroType, violationReason, c.violatingRunLevelZeroNamespaces)),
+		v1helpers.UpdateConditionFn(makeCondition(PodSecurityDisabledSyncerType, violationReason, c.violatingDisabledSyncerNamespaces)),
+		v1helpers.UpdateConditionFn(makeCondition(PodSecurityInconclusiveType, inconclusiveReason, c.inconclusiveNamespaces)),
 	}
 }
--- a/pkg/operator/podsecurityreadinesscontroller/conditions_test.go
+++ b/pkg/operator/podsecurityreadinesscontroller/conditions_test.go
@@ -9,7 +9,7 @@ import (
 )
 
 func TestCondition(t *testing.T) {
-	t.Run("with namespaces", func(t *testing.T) {
+	t.Run("with violating namespaces", func(t *testing.T) {
 		namespaces := []string{"namespace1", "namespace2"}
 		expectedCondition := operatorv1.OperatorCondition{
 			Type:    PodSecurityCustomerType,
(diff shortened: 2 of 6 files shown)


pkg/operator/podsecurityreadinesscontroller/violation.go
```diff
@@ -54,6 +51,28 @@ func (c *PodSecurityReadinessController) isNamespaceViolating(ctx context.Contex
 	return len(c.warningsHandler.PopAll()) > 0, nil
 }
 
+func determineEnforceLabelForNamespace(ns *applyconfiguration.NamespaceApplyConfiguration) (string, error) {
```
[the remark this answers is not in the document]
everettraven: @ibihim Might also have some thoughts and or additional context to share here

pkg/operator/podsecurityreadinesscontroller/violation.go
```diff
@@ -54,6 +51,28 @@ func (c *PodSecurityReadinessController) isNamespaceViolating(ctx context.Contex
...
+	if _, ok := ns.Labels[securityv1.MinimallySufficientPodSecurityStandard]; ok {
```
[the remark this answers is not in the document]
jacobsee: Thanks for the catch!

pkg/operator/podsecurityreadinesscontroller/violation.go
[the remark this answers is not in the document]
jacobsee: That's fair. Falling back to current behavior in the absence of the annotation seemed "safest" to me if we somehow find a situation in which the label syncer isn't updated to set that annotation, but this component is updated... but that does seem unlikely.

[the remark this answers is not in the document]
jacobsee: Updated

pkg/operator/podsecurityreadinesscontroller/conditions.go
```diff
@@ -59,16 +70,45 @@ func (c *podSecurityOperatorConditions) addViolation(ns *corev1.Namespace) {
...
+}
+
+func makeCondition(conditionType, conditionReason string, namespaces []string) operatorv1.OperatorCondition {
+	var messageFormatter string
+
+	if conditionReason == violationReason {
+		messageFormatter = "Violations detected in namespaces: %v"
+	} else if conditionReason == inconclusiveReason {
+		messageFormatter = "Could not evaluate violations for namespaces: %v"
+	}
```
everettraven, line 101: What should happen if the condition reason provided is one we don't know about?

When doing the `fmt.Sprintf()` with an empty format string, the message would end up looking something like:
```
%!(EXTRA []string=[foo bar baz])
```

This isn't very helpful to users.

We could either:
- Use a `switch` statement with a `default` case that sets a sane message for any condition reason we don't have specific handling for
- Have the message/message format itself passed as a parameter

pkg/operator/podsecurityreadinesscontroller/conditions_test.go
```diff
@@ -9,16 +9,44 @@ import (
...
+			t.Errorf("expected condition message %s, got %s", expectedCondition.Message, condition.Message)
+		}
+	})
+
+	t.Run("with inconclusive namespaces", func(t *testing.T) {
+		namespaces := []string{"namespace1", "namespace2"}
+		expectedCondition := operatorv1.OperatorCondition{
+			Type:    PodSecurityCustomerViolationType,
+			Status:  operatorv1.ConditionTrue,
+			Reason:  "PSViolationsDetected",
```
everettraven, line 45: Should this be the new `PSViolationDecisionInconclusive` reason?

pkg/operator/podsecurityreadinesscontroller/conditions.go
```diff
@@ -14,12 +14,19 @@ import (
 )
 
 const (
-	PodSecurityCustomerType       = "PodSecurityCustomerEvaluationConditionsDetected"
-	PodSecurityOpenshiftType      = "PodSecurityOpenshiftEvaluationConditionsDetected"
-	PodSecurityRunLevelZeroType   = "PodSecurityRunLevelZeroEvaluationConditionsDetected"
-	PodSecurityDisabledSyncerType = "PodSecurityDisabledSyncerEvaluationConditionsDetected"
+	PodSecurityCustomerViolationType          = "PodSecurityCustomerEvaluationViolationConditionsDetected"
```
ibihim, line 17: Is there a good reason, why we change the name of that? It would mean that we need to change the metrics we are looking for in Thanos.

pkg/operator/podsecurityreadinesscontroller/podsecurityreadinesscontroller.go
```diff
@@ -87,6 +87,8 @@ func (c *PodSecurityReadinessController) sync(ctx context.Context, syncCtx facto
 		if err != nil {
 			klog.V(2).ErrorS(err, "namespace:", ns.Name)
 
+			conditions.addInconclusive(&ns)
```
ibihim, line 90: I like the idea of logging an error as a CFE condition.

pkg/operator/podsecurityreadinesscontroller/violation.go
```diff
@@ -2,42 +2,36 @@ package podsecurityreadinesscontroller
...
-	if len(viableLabels) == 0 {
-		// If there are no labels managed by the syncer, we can't make a decision.
-		return false, nil
+	var enforceLabel string
+
+	if _, ok := nsApplyConfig.Annotations[securityv1.MinimallySufficientPodSecurityStandard]; ok {
+		// Pick the MinimallySufficientPodSecurityStandard if it exists
+		enforceLabel = nsApplyConfig.Annotations[securityv1.MinimallySufficientPodSecurityStandard]
+	} else {
+		return false, fmt.Errorf("unable to determine if the namespace is violating because the MinimallySufficientPodSecurityStandard annotation wasn't found")
```
ibihim, line 30: Nit:
```go
	enforceLabel, ok := nsApplyConfig.Annotations[securityv1.MinimallySufficientPodSecurityStandard]
        if !ok {
		return false, fmt.Errorf("")
	}
```

ibihim: I am not explicitly sure if we would like to try to keep the previous logic as a backup. At least in the context of an upgrade.

pkg/operator/podsecurityreadinesscontroller/conditions.go
```diff
@@ -14,12 +14,19 @@ import (
...
+	PodSecurityOpenshiftViolationType         = "PodSecurityOpenshiftEvaluationViolationConditionsDetected"
+	PodSecurityRunLevelZeroViolationType      = "PodSecurityRunLevelZeroEvaluationViolationConditionsDetected"
+	PodSecurityDisabledSyncerViolationType    = "PodSecurityDisabledSyncerEvaluationViolationConditionsDetected"
+	PodSecurityCustomerInconclusiveType       = "PodSecurityCustomerEvaluationInconclusiveConditionsDetected"
```
ibihim, line 21: Why not just 1 inconclusive type?

pkg/operator/podsecurityreadinesscontroller/conditions.go
```diff
@@ -59,16 +70,45 @@ func (c *podSecurityOperatorConditions) addViolation(ns *corev1.Namespace) {
...
+	c.inconclusiveCustomerNamespaces = append(c.inconclusiveCustomerNamespaces, ns.Name)
+}
+
+func makeCondition(conditionType, conditionReason string, namespaces []string) operatorv1.OperatorCondition {
+	var messageFormatter string
+
+	if conditionReason == violationReason {
+		messageFormatter = "Violations detected in namespaces: %v"
+	} else if conditionReason == inconclusiveReason {
+		messageFormatter = "Could not evaluate violations for namespaces: %v"
```
ibihim, line 100: Hm, we might need to read the [CFE documentation](https://github.com/openshift/enhancements/blob/995b620cb04c030bf62c908e188472fe7031a704/dev-guide/cluster-fleet-evaluation.md) says about those messages. E.g. I remember that we need to be careful with those messages under certain circumstances.

pkg/operator/podsecurityreadinesscontroller/conditions.go
jacobsee: We could do that if the distinction between types of namespaces is not as important for inconclusive determinations as it is for violations - no opposition here.

jacobsee: I do like this better - consolidated to 1 inconclusive type

pkg/operator/podsecurityreadinesscontroller/conditions.go
jacobsee: I've consolidated down to 1 inconclusive type and put these names back the way they were to keep things simpler.

pkg/operator/podsecurityreadinesscontroller/conditions.go
jacobsee: I might be missing something, but I'm not seeing a format for the condition messages defined somewhere... For violations, this should format exactly the same as it always has - only the inconclusive part is new.

pkg/operator/podsecurityreadinesscontroller/violation.go
```diff
@@ -54,6 +50,28 @@ func (c *PodSecurityReadinessController) isNamespaceViolating(ctx context.Contex
 	return len(c.warningsHandler.PopAll()) > 0, nil
 }
 
+func determineEnforceLabelForNamespace(ns *applyconfiguration.NamespaceApplyConfiguration) (string, error) {
+	if _, ok := ns.Annotations[securityv1.MinimallySufficientPodSecurityStandard]; ok {
+		// Pick the MinimallySufficientPodSecurityStandard if it exists
```
ibihim, line 55: Do you think it would be better to phrase the reasoning, that we discussed in chat? Something in the sense like "should be the go-to route, but we keep the older one for upgrade scenarios as of now / other approaches can be removed once annotation is there".

ibihim (commented):
One last request, could you clean up the commit messages / amount of commits? :)

jacobsee:
@ibihim Yep absolutely, cleaner now 👍

pkg/operator/podsecurityreadinesscontroller/podsecurityreadinesscontroller_test.go
```diff
@@ -79,6 +63,24 @@ func TestPodSecurityViolationController(t *testing.T) {
 		},
 	}
 
+	syncFields := []metav1.ManagedFieldsEntry{{
+		Manager: "pod-security-admission-label-synchronization-controller",
+		FieldsV1: &metav1.FieldsV1{
+			Raw: []byte(`{
+				"f:metadata": {
+					"f:annotations": {
+				    "f:security.openshift.io/MinimallySufficientPodSecurityStandard": {}
```
everettraven, line 72: Indenting on this off, or is this just me?

pkg/operator/podsecurityreadinesscontroller/conditions.go
jacobsee: I think we might be mixing up the `Reason` and the `Message` fields.

Before this PR, we were using `PSViolationsDetected` for the condition's `Reason` and `Violations detected in namespaces: %v` for the condition's `Message` (for all types of violations).

Now, `Reason` can be `PSViolationsDetected` or `PSViolationDecisionInconclusive` and
`Message` can be a few different strings... Are you saying that we now need to break `Reason` out by violation type?

pkg/operator/podsecurityreadinesscontroller/violation.go
```diff
@@ -54,6 +50,31 @@ func (c *PodSecurityReadinessController) isNamespaceViolating(ctx context.Contex
 	return len(c.warningsHandler.PopAll()) > 0, nil
 }
 
+func determineEnforceLabelForNamespace(ns *applyconfiguration.NamespaceApplyConfiguration) (string, error) {
+	if _, ok := ns.Annotations[securityv1.MinimallySufficientPodSecurityStandard]; ok {
+		// This should generally exist and will be the only supported method of determining
+		// the enforce level going forward - however, we're keeping the label fallback for
+		// now to account for any workloads not yet annotated using a new enough version of
+		// the syncer, such as during upgrade scenarios.
+		return ns.Annotations[securityv1.MinimallySufficientPodSecurityStandard], nil
```
everettraven, line 59: Non-blocking nit: Could avoid a second map access with something like:
```suggestion
	if label, ok := ns.Annotations[securityv1.MinimallySufficientPodSecurityStandard]; ok {
		// This should generally exist and will be the only supported method of determining
		// the enforce level going forward - however, we're keeping the label fallback for
		// now to account for any workloads not yet annotated using a new enough version of
		// the syncer, such as during upgrade scenarios.
		return label, nil
```

everettraven (approved):
Changes look fine to me.

Had a couple minor comments, but nothing that I think should block this if CI is happy.

/lgtm

ibihim (approved):

PR #1834, opened 2025-04-28
OCPBUGS-55013: SCC: add hostmount-anyuid-v2

haircommander:
the intent of hostmount-anyuid is to allow a pod access to paths on the host. The problem is, hostPath volumes aren't selinux relabeled, and there are paths that the default selinux type `container_t` cannot access.

This breaks expectation, and makes pods that rely on hostmount-anyuid brittle to selinux changes on the host (see https://issues.redhat.com/browse/OCPBUGS-55013)

Instead of relaxing permissions of all paths on the host, or increasing the ability of container_t, we should trust pods that are granted access to this already powerful SCC to use its power fully.

### DIFF
--- a/bindata/bootkube/scc-manifests/0000_20_kube-apiserver-operator_00_scc-hostmount-anyuid-v2.yaml
+++ b/bindata/bootkube/scc-manifests/0000_20_kube-apiserver-operator_00_scc-hostmount-anyuid-v2.yaml
@@ -1,0 +1,47 @@
+allowHostDirVolumePlugin: true
+allowHostIPC: false
+allowHostNetwork: false
+allowHostPID: false
+allowHostPorts: false
+allowPrivilegeEscalation: true
+allowPrivilegedContainer: false
+allowedCapabilities:
+apiVersion: security.openshift.io/v1
+defaultAddCapabilities:
+fsGroup:
+  type: RunAsAny
+groups: []
+kind: SecurityContextConstraints
+metadata:
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    release.openshift.io/create-only: "true"
+    kubernetes.io/description: |-
+      hostmount-anyuid-v2 provides all the features of the
+      restricted SCC but allows host mounts any UID, and any SELinux label by a pod.  This is primarily
+      used by the persistent volume recycler. WARNING: this SCC allows host file
+      system access as any UID, including UID 0.  Grant with caution.
+  name: hostmount-anyuid-v2
+priority:
+readOnlyRootFilesystem: false
+requiredDropCapabilities:
+- MKNOD
+runAsUser:
+  type: RunAsAny
+seLinuxContext:
+  type: RunAsAny
+supplementalGroups:
+  type: RunAsAny
+volumes:
+- configMap
+- csi
+- downwardAPI
+- emptyDir
+- ephemeral
+- hostPath
+- nfs
+- persistentVolumeClaim
+- projected
+- secret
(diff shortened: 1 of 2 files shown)


bindata/bootkube/scc-manifests/0000_20_kube-apiserver-operator_00_scc-hostmount-anyuid-v2.yaml
```diff
@@ -0,0 +1,49 @@
...
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    release.openshift.io/create-only: "true"
+    kubernetes.io/description: |-
+      hostmount-anyuid-v2 provides all the features of the
+      restricted SCC but allows host mounts any UID, and any SELinux label by a pod.  This is primarily
+      used by the persistent volume recycler. WARNING: this SCC allows host file
+      system access as any UID, including UID 0.  Grant with caution.
+  name: hostmount-anyuid
```
aleskandro, line 26: Hey @haircommander , thanks for the quick follow up! Should this be 

```suggestion
  name: hostmount-anyuid-v2
```
?

cc @lwan-wanglin

bindata/bootkube/scc-manifests/0000_20_kube-apiserver-operator_00_scc-hostmount-anyuid-v2.yaml
```diff
@@ -0,0 +1,49 @@
...
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    release.openshift.io/create-only: "true"
+    kubernetes.io/description: |-
+      hostmount-anyuid-v2 provides all the features of the
+      restricted SCC but allows host mounts any UID, and any SELinux label by a pod.  This is primarily
+      used by the persistent volume recycler. WARNING: this SCC allows host file
+      system access as any UID, including UID 0.  Grant with caution.
```
ibihim, line 25: Well, in good practice of the v2s, we should mention that we are by-passing SELinux context as well.

bindata/bootkube/scc-manifests/0000_20_kube-apiserver-operator_00_scc-hostmount-anyuid-v2.yaml
```diff
@@ -0,0 +1,49 @@
...
+requiredDropCapabilities:
+- MKNOD
+runAsUser:
+  type: RunAsAny
+seLinuxContext:
+  type: RunAsAny
+supplementalGroups:
+  type: RunAsAny
+users:
+- system:serviceaccount:openshift-infra:pv-recycler-controller
```
sjenning, line 38: we are moving away from specifying subjects directly in the SCC and it seems that `pv-recycler-controller` works just find with with old SCC, so let start out clean in v2.

ibihim, line 38: Curious if we can drop this.

bindata/bootkube/scc-manifests/0000_20_kube-apiserver-operator_00_scc-hostmount-anyuid-v2.yaml
sjenning: I think this is captured with `and any SELinux label by a pod`

ibihim-unsafe (changes requested):
Looks good. Please create a PR for [e2e tests](https://github.com/openshift/origin) as well.

ibihim (changes requested):
Looks good. Please create a PR for [e2e tests](https://github.com/openshift/origin) as well.

ibihim-unsafe (approved):

Prashanth684:
thanks @ibihim . @vrutkovs could you approve please?

PR #1846, opened 2025-05-21
NO-JIRA: Enable watch termination grace period

benluddy:
This should mostly eliminate aligned informer re-lists during kube-apiserver rollout.

### DIFF
--- a/bindata/assets/config/defaultconfig.yaml
+++ b/bindata/assets/config/defaultconfig.yaml
@@ -163,6 +163,8 @@ apiServerArguments:
     - 70s # give SDN some time to converge: 30s for iptable lock contention, 25s for the second try and some seconds for AWS to update ELBs
   shutdown-send-retry-after:
   - "true"
+  shutdown-watch-termination-grace-period:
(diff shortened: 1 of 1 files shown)


bindata/assets/config/defaultconfig.yaml
```diff
@@ -163,6 +163,8 @@ apiServerArguments:
     - 70s # give SDN some time to converge: 30s for iptable lock contention, 25s for the second try and some seconds for AWS to update ELBs
   shutdown-send-retry-after:
   - "true"
+  shutdown-watch-termination-grace-period:
+    - 25s
```
p0lyn0mial, line 167: does the server drain watch requests, or does it wait for them to terminate naturally?

p0lyn0mial: ah, ok, it rate limits # of watchers that will be closed. 

does setting this flag affect (slow down) the shutdown process? I just want to make sure no other adjustments are required.

p0lyn0mial:
/retitle NO-JIRA: Enable watch termination grace period

PR #1856, opened 2025-06-13
NO-JIRA: remove unused yaml

lance5890:
These yamls only works when upgrade from 4.14, no need in the newest 4.20 versions

next steps of https://github.com/openshift/cluster-kube-apiserver-operator/pull/1565

### DIFF
--- a/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
+++ b/manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
@@ -248,12 +248,3 @@ spec:
       expr: sum(rate(apiserver_tls_handshake_errors_total{apiserver=~"openshift-apiserver|kube-apiserver|openshift-oauth-apiserver"}[5m])) by (apiserver)
     - record: resource:apiserver_storage_objects:max
       expr: max(apiserver_storage_objects{apiserver=~"openshift-apiserver|kube-apiserver|openshift-oauth-apiserver"}) by (apiserver, resource)
----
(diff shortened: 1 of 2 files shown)


manifests/0000_90_kube-apiserver-operator_04_servicemonitor-apiserver.yaml
```diff
@@ -248,12 +248,3 @@ spec:
...
       expr: max(apiserver_storage_objects{apiserver=~"openshift-apiserver|kube-apiserver|openshift-oauth-apiserver"}) by (apiserver, resource)
----
-apiVersion: monitoring.coreos.com/v1
-kind: PrometheusRule
-metadata:
-  name: kube-apiserver-recording-rules
-  namespace: openshift-kube-apiserver
-  annotations:
-    include.release.openshift.io/self-managed-high-availability: "true"
-    release.openshift.io/delete: "true"
```
lance5890, line 259: only works when upgrade from 4.14

manifests/0000_90_kube-apiserver-operator_05_api_performance_dashboard.yaml
```diff
@@ -3329,14 +3329,3 @@ data:
...
     }
----
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: grafana-dashboard-api-performance
-  namespace: openshift-config-managed
-  annotations:
-    include.release.openshift.io/self-managed-high-availability: "true"
-    release.openshift.io/delete: "true"
```
lance5890, line 3340: also has the "release.openshift.io/delete" annotation

PR #1866, opened 2025-07-07
OCPBUGS-58232: Don't wait for 2 etcd members when the cluster is TNF installed with assisted-installer

giladravid16:
Currently unless the cluster is an SNO, we wait until etcd is HA.
However when installing a TNF cluster with assisted-installer, there is only going to be one CP node until the very end of the installation because TNF clusters only have 2 CP nodes and assisted-installer uses one of the CP nodes as the bootstrap. So in this case we can't wait for etcd to be HA.
We can know if a cluster is TNF based on it's control plane topology is DualReplica, and we can know if the cluster is installed with assisted-installer by checking if the openshift-etcd namespace has the annotation openshift.io/delayed-bootstrap.

Closes [OCPBUGS-58232](https://issues.redhat.com/browse/OCPBUGS-58232)

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -208,6 +208,21 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 	var notOnSingleReplicaTopology resourceapply.ConditionalFunction = func() bool {
 		return infrastructure.Status.ControlPlaneTopology != configv1.SingleReplicaTopologyMode
 	}
+	isMultipleEtcdEndpointsRequired := true
+	switch infrastructure.Status.ControlPlaneTopology {
+	case configv1.SingleReplicaTopologyMode:
+		isMultipleEtcdEndpointsRequired = false
+	case configv1.DualReplicaTopologyMode:
+		etcdNamespace, err := kubeClient.CoreV1().Namespaces().Get(ctx, "openshift-etcd", metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		_, hasDelayedBootstrapAnnotation := etcdNamespace.Annotations["openshift.io/delayed-bootstrap"]
+		isMultipleEtcdEndpointsRequired = !hasDelayedBootstrapAnnotation
+	}
+	var requireMultipleEtcdEndpoints resourceapply.ConditionalFunction = func() bool {
+		return isMultipleEtcdEndpointsRequired
+	}
 
 	staticResourceController := staticresourcecontroller.NewStaticResourceController(
 		"KubeAPIServerStaticResources",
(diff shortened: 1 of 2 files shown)


pkg/operator/starter.go
```diff
@@ -208,6 +208,15 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 	var notOnSingleReplicaTopology resourceapply.ConditionalFunction = func() bool {
 		return infrastructure.Status.ControlPlaneTopology != configv1.SingleReplicaTopologyMode
 	}
+	etcdNamespace, err := kubeClient.CoreV1().Namespaces().Get(ctx, "openshift-etcd", metav1.GetOptions{})
```
vrutkovs, line 211: We don't need to fetch the namespace if `ControlPlaneTopology` is not `DualReplicaTopologyMode`, right?

vrutkovs (changes requested):

pkg/operator/starter.go
giladravid16: I changed it to only get the namespace if it's dual replica

pkg/operator/starter.go
```diff
@@ -208,6 +208,17 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
...
+	if infrastructure.Status.ControlPlaneTopology == configv1.DualReplicaTopologyMode {
+		etcdNamespace, err := kubeClient.CoreV1().Namespaces().Get(ctx, "openshift-etcd", metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		_, isDelayedDualReplica = etcdNamespace.Annotations["openshift.io/delayed-bootstrap"]
+	}
+	var requireMultipleEtcdEndpoints resourceapply.ConditionalFunction = func() bool {
+		return !(infrastructure.Status.ControlPlaneTopology == configv1.SingleReplicaTopologyMode || isDelayedDualReplica)
+	}
```
vrutkovs, line 221: I'm not happy about `!(infrastructure.Status.ControlPlaneTopology == configv1.SingleReplicaTopologyMode || isDelayedDualReplica)`, lets try this
```suggestion
	isMultipleEtcdEndpointsRequired := true
	switch infrastructure.Status.ControlPlaneTopology {
	case configv1.SingleReplicaTopologyMode:
		isMultipleEtcdEndpointsRequired = false
	case configv1.DualReplicaTopologyMode:
		etcdNamespace, err := kubeClient.CoreV1().Namespaces().Get(ctx, "openshift-etcd", metav1.GetOptions{})
		if err != nil {
			return err
		}
		_, isMultipleEtcdEndpointsRequired = etcdNamespace.Annotations["openshift.io/delayed-bootstrap"]
	}
	var requireMultipleEtcdEndpoints resourceapply.ConditionalFunction = func() bool {
		return isMultipleEtcdEndpointsRequired
	}
```

vrutkovs (changes requested):

PR #1871, opened 2025-07-17
NO-JIRA: Update the deprecated API usage alerts for 1.33.

benluddy:
OpenShift 4.20 is an EUS release based on Kubernetes 1.33, so the EUS alert needs to fire for Kubernetes 1.34 and 1.35 removals.


dgrisonnet (approved):
/lgtm

PR #1872, opened 2025-07-18
OCPBUGS-59527: certrotationcontroller: extend node-system-admin-signer lifetime

vrutkovs:
This signer is used to issue the certificate for node-system-admin-client. The target certificate is expected to last 2 years, however it cannot be valid longer than its signer. As a result, signer validity is extended to 3 years, being refreshed at 80% of its lifetime (2.5 years)

This change is required to test [1 year shutdowns](https://github.com/openshift/release/pull/64965) (instead of just 360d)

### DIFF
--- a/pkg/operator/certrotationcontroller/certrotationcontroller.go
+++ b/pkg/operator/certrotationcontroller/certrotationcontroller.go
@@ -812,10 +812,10 @@ func newCertRotationController(
 				AutoRegenerateAfterOfflineExpiry: "https://github.com/openshift/cluster-kube-apiserver-operator/pull/1631,'[Conformance][sig-api-machinery][Feature:APIServer] local kubeconfig \"localhost-recovery.kubeconfig\" should be present on all masters and work [apigroup:config.openshift.io] [Suite:openshift/conformance/parallel/minimal]'",
 				Description:                      "Signer for the per-master-debugging-client.",
 			},
-			Validity: devRotationExceptionYear,
+			Validity: 3 * devRotationExceptionYear,
 			// Refresh set to 80% of the validity.
 			// This range is consistent with most other signers defined in this pkg.
-			Refresh:                devRotationExceptionTenMonth,
+			Refresh:                3 * devRotationExceptionTenMonth,
 			RefreshOnlyWhenExpired: refreshOnlyWhenExpired,
 			Informer:               kubeInformersForNamespaces.InformersFor(operatorclient.OperatorNamespace).Core().V1().Secrets(),
 			Lister:                 kubeInformersForNamespaces.InformersFor(operatorclient.OperatorNamespace).Core().V1().Secrets().Lister(),


pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -812,10 +812,10 @@ func newCertRotationController(
 				AutoRegenerateAfterOfflineExpiry: "https://github.com/openshift/cluster-kube-apiserver-operator/pull/1631,'[Conformance][sig-api-machinery][Feature:APIServer] local kubeconfig \"localhost-recovery.kubeconfig\" should be present on all masters and work [apigroup:config.openshift.io] [Suite:openshift/conformance/parallel/minimal]'",
 				Description:                      "Signer for the per-master-debugging-client.",
 			},
-			Validity: devRotationExceptionYear,
+			Validity: 3 * devRotationExceptionYear,
```
p0lyn0mial, line 815: I think the issue was introduced in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1618/files. (before that PR signer was valid for longer than the target cert)
This means we should probably consider changing/fixing previous versions as well. 

However, I have a general question. I don’t understand why we keep updating the validity of the certificate. Is this only needed to pass the E2E test? I was under the impression that the platform should be able to recover from expired certificates. If that’s the case, we wouldn’t need to keep increasing the validity period.

vrutkovs: >This means we should probably consider changing/fixing previous versions as well

Yes, we may want to backport it to 4.16.

> I don’t understand why we keep updating the validity of the certificate. Is this only needed to pass the E2E test? I was under the impression that the platform should be able to recover from expired certificates

The platform can regenerate certificates if node CSRs are approved. In order to approve them, the user needs a kubeconfig, which has not yet expired after the node was started after sleep. Previously, before 4.16, we supported only 90 days of suspend, so certificate in localhost-recovery.kubeconfig was valid for 120 days. Later on we started supporting 1 year of shutdown for fresh SNOs, so the certificate was extended to 2 years. Unfortunately, due to other issues (inmemory loopback cert), we never properly tested 1 year suspend - it has always been 360 days. This is why this issue was left undiscovered for quite some time.

So this change is not just an e2e convenience, users will hit this problem if they suspend their SNOs for an entire year too

pkg/operator/certrotationcontroller/certrotationcontroller.go
```diff
@@ -812,10 +812,10 @@ func newCertRotationController(
...
 			// Refresh set to 80% of the validity.
 			// This range is consistent with most other signers defined in this pkg.
-			Refresh:                devRotationExceptionTenMonth,
+			Refresh:                3 * devRotationExceptionTenMonth,
```
p0lyn0mial, line 818: since we have a local controller running on each machine for certificate recovery, how is it possible that this certificate/signer is not being regenerated?

vrutkovs: We're running certrotation controller in RefreshOnlyWhenExpired mode to update node-system-admin-client secret, but we're not running nodekubeconfigcontroller to generate new kubeconfigs on disk. 

Should we run it - yes, maybe, after all we already run resourcesync controller after all. However, it would make recovery procedure more unpredictable - i.e. the controller may delay kubeconfig generation and users will attempt to approve CSR with expired kubeconfig. Also, it will allow customers use it in unsupported configurations - i.e. after the cluster was shutdown for 3 years - I don't know the implications of that yet

pkg/operator/certrotationcontroller/certrotationcontroller.go
p0lyn0mial: What is the procedure for shutting down a cluster? For example, if I turn my cluster off in month 23 and then turn it back on in month 35 (1 year later), when is the next time I can turn the cluster off again? 
Is this process documented somewhere? 
How can customers validate that the cert was successfully rotated before shutting down the cluster again?

vrutkovs: >What is the procedure for shutting down a cluster? 

https://docs.redhat.com/en/documentation/openshift_container_platform/4.16/html/backup_and_restore/graceful-shutdown-cluster

> For example, if I turn my cluster off in month 23 and then turn it back on in month 35 (1 year later), when is the next time I can turn the cluster off again?

Not supported, only freshly installed cluster is allowed to be shut down for 1 year. Any cluster is allowed to be suspended or shutdown for 3 months though, see https://issues.redhat.com/browse/API-1828

>How can customers validate that the cert was successfully rotated before shutting down the cluster again?

They would check operator status, which is expected to validate that. In e2e, we also run a set of tests to make sure any functionality is not lost

pkg/operator/certrotationcontroller/certrotationcontroller.go
p0lyn0mial, line 818: So this is on a fresh cluster: what happens if I turn off my cluster in month 29 and turn it back on after another 12 months? I think the signer will have expired, since it will be 41 months old and its validity is 36 months.

The refresh is set to 80% of its validity period, which is 30 months.

Once the signer expires, the kas will no longer be able to validate the client cert, and users won't be able to approve CSRs.

p0lyn0mial: ok, how will customers know when it's safe to shut down a cluster after a fresh installation?

pkg/operator/certrotationcontroller/certrotationcontroller.go
p0lyn0mial, line 815: Also, why are the certs valid for 2 and 3 years? How did we calculate that? (Just trying to understand the math)

pkg/operator/certrotationcontroller/certrotationcontroller.go
wangke19: Here is the impact of this PR change to `Validity/Refresh`.
Mode | Before (Validity/Refresh) | After PR | Impact
-- | -- | -- | --
Default | 1 year / 292d | 3 years / 876d | ✅ Less rotation, ops-friendly
Development Mode | 1 year / 292d | 3 years / 876d | ✅ Longer-lived debug certs, consistent
ShortCert Only | 3h / 2.5h | 9h / 7.5h | ✅ Still short, but 3× longer
DevMode + ShortCert|3h / 2.5h | 9h / 7.5h|✅ Same as above
</tbody>

vrutkovs: This would affect not just the signer, but the target too: previously they were 1 year each, now signer would last 3 years and target would last 2 years

p0lyn0mial:
one more thing, does changing the validity affect existing clusters ? i think yes, do we have a e2e test that verifies that ?

vrutkovs:
>one more thing, does changing the validity affect existing clusters ? i think yes, do we have a e2e test that verifies that ?

They will be regenerated when the new refresh period kicks in, so in two years. We use these kubeconfigs to approve CSRs in shutdown/suspend tests

>and one more question, these certificates are used only during recovery scenarios and are not used by any component, right ?

Correct, these changes apply for `localhost-recovery.kubeconfig`. I don't think we have a way of verifying that no components use this kubeconfig (it needs to be a privileged component though)

p0lyn0mial:
how are we going to fix clusters that were shut down after installation and whose `node-system-admin-signer` expired after 1 year

vrutkovs:
>we could put garbage in the content and look for fireworks.

It gets refreshed when content mismatches the expected, so its not trivial.

>how are we going to fix clusters that were shut down after installation and whose node-system-admin-signer expired after 1 year

Users can assemble a custom kubeconfig using 10 year certificates for control plane using on disk content. We don't advertise it as it may be misused, but these clusters are certainly recoverable

p0lyn0mial:
> It gets refreshed when content mismatches the expected, so its not trivial.

i know it is an extra ask but maybe we could hack it, turn of the regeneration, put some garbage and see what breaks?

vrutkovs:
Lets see - https://github.com/openshift/cluster-kube-apiserver-operator/pull/1883 should make it expire in 1 hour and get refreshed after 10 years

vrutkovs:
I don't mind making it, but I'd rather keep it private - this may give customers an impression that you can bend the support rules and suspend clusters whenever for whatever period

p0lyn0mial:
> I don't mind making it, but I'd rather keep it private

ok, please consider creating a gdoc with instructions on how to recover from the issue. It would help the person working on a potential customer case.

p0lyn0mial:
> **Lets see - https://github.com/openshift/cluster-kube-apiserver-operator/pull/1883 should make it expire in 1 hour and get refreshed after 10 years

could you take a look at the results from that job and check if nothing else broke ?

vrutkovs:
>could you take a look at the results from that job and check if nothing else broke ?

Yes, apart from explicit test using those kubeconfig I don't see any regressions

PR #1873, opened 2025-07-24
OCPBUGS-44842: Set not-before/not-after annotations

vrutkovs:
Pull in https://github.com/openshift/library-go/pull/1971 to make sure most secrets created by controller would have refresh-period annotation set.

### DIFF
--- a/pkg/operator/nodekubeconfigcontroller/nodekubeconfigcontroller.go
+++ b/pkg/operator/nodekubeconfigcontroller/nodekubeconfigcontroller.go
@@ -29,10 +29,10 @@ const workQueueKey = "key"
 type NodeKubeconfigController struct {
 	operatorClient v1helpers.StaticPodOperatorClient
 
-	kubeClient          kubernetes.Interface
-	configMapLister     corev1listers.ConfigMapLister
-	secretLister        corev1listers.SecretLister
-	infrastuctureLister configv1listers.InfrastructureLister
+	kubeClient           kubernetes.Interface
+	configMapLister      corev1listers.ConfigMapLister
+	secretLister         corev1listers.SecretLister
+	infrastructureLister configv1listers.InfrastructureLister
 }
 
 func NewNodeKubeconfigController(
@@ -43,11 +43,11 @@ func NewNodeKubeconfigController(
 	eventRecorder events.Recorder,
 ) factory.Controller {
 	c := &NodeKubeconfigController{
-		operatorClient:      operatorClient,
-		kubeClient:          kubeClient,
-		configMapLister:     kubeInformersForNamespaces.ConfigMapLister(),
-		secretLister:        kubeInformersForNamespaces.SecretLister(),
-		infrastuctureLister: infrastuctureInformer.Lister(),
+		operatorClient:       operatorClient,
+		kubeClient:           kubeClient,
+		configMapLister:      kubeInformersForNamespaces.ConfigMapLister(),
+		secretLister:         kubeInformersForNamespaces.SecretLister(),
+		infrastructureLister: infrastuctureInformer.Lister(),
 	}
 
 	return factory.New().WithInformers(
@@ -85,7 +85,7 @@ func (c NodeKubeconfigController) sync(ctx context.Context, syncContext factory.
 		c.kubeClient.CoreV1(),
 		c.secretLister,
 		c.configMapLister,
-		c.infrastuctureLister,
+		c.infrastructureLister,
 		syncContext.Recorder(),
 	)
 	if err != nil {
--- a/pkg/operator/nodekubeconfigcontroller/nodekubeconfigcontroller.go
+++ b/pkg/operator/nodekubeconfigcontroller/nodekubeconfigcontroller.go
@@ -14,6 +14,7 @@ import (
 	"github.com/openshift/cluster-kube-apiserver-operator/bindata"
 	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/operatorclient"
 	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/certrotation"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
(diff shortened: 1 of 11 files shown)


pkg/operator/nodekubeconfigcontroller/nodekubeconfigcontroller.go
```diff
@@ -112,6 +116,22 @@ func ensureNodeKubeconfigs(ctx context.Context, client coreclientv1.CoreV1Interf
 		return fmt.Errorf("system:admin client private key missing from secret %s/node-system-admin-client", operatorclient.OperatorNamespace)
 	}
 
+	// Ensure secret key matches the certificate
+	_, err = tls.X509KeyPair(systemAdminClientCert, systemAdminClientKey)
```
p0lyn0mial, line 120: could you point me to a location where this cert is created?

vrutkovs: We could read it from the secret, but that would require certrotationcontroller to set those first - I'm not entirely sure we can rely on that

wangke19:
Failed [e2e-aws-ovn-serial](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-kube-apiserver-operator/1873/pull-ci-openshift-cluster-kube-apiserver-operator-main-e2e-aws-ovn-serial/1948367251623120896) depends on https://github.com/openshift/release/pull/67513

pkg/operator/nodekubeconfigcontroller/nodekubeconfigcontroller.go
vrutkovs: sure, updated

pkg/operator/nodekubeconfigcontroller/nodekubeconfigcontroller_test.go
```diff
@@ -85,7 +88,34 @@ func (l *secretLister) Get(name string) (*corev1.Secret, error) {
 	return l.client.CoreV1().Secrets(l.namespace).Get(context.Background(), name, metav1.GetOptions{})
 }
 
+const privateKey = `
```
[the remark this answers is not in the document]
vrutkovs: We still need them to verify kubeconfig file is generated correctly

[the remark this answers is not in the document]
p0lyn0mial: I think we only need to check if the expected value is in the correct place. The content itself doesn’t matter that’s why it worked with the previously set value. Since we’re not unpacking the content, storing a smaller value makes the code a bit more readable. Does it make sense ?

pkg/operator/nodekubeconfigcontroller/nodekubeconfigcontroller.go
```diff
@@ -152,6 +153,13 @@ func ensureNodeKubeconfigs(ctx context.Context, client coreclientv1.CoreV1Interf
 		requiredSecret.Annotations = map[string]string{}
 	}
 	requiredSecret.Annotations[annotations.OpenShiftComponent] = "kube-apiserver"
+	// Copy not-before/not-after annotations from systemAdminClientCert
```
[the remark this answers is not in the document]
vrutkovs: CA may contain several certs, so their not-before/not-after is more complicated - should it be nearest notAfter or latest notAfter? In any case, CA's are a collection of signers, so we should focus on signer/targets first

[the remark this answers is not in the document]
p0lyn0mial: ok

p0lyn0mial:
@wangke19 is there anything else we should consider before merging this pr ?

pkg/operator/nodekubeconfigcontroller/nodekubeconfigcontroller.go
```diff
@@ -152,6 +153,13 @@ func ensureNodeKubeconfigs(ctx context.Context, client coreclientv1.CoreV1Interf
...
+	if len(systemAdminCredsSecret.Annotations[certrotation.CertificateNotBeforeAnnotation]) > 0 {
+		requiredSecret.Annotations[certrotation.CertificateNotBeforeAnnotation] = systemAdminCredsSecret.Annotations[certrotation.CertificateNotBeforeAnnotation]
+	}
+	if len(systemAdminCredsSecret.Annotations[certrotation.CertificateNotAfterAnnotation]) > 0 {
+		requiredSecret.Annotations[certrotation.CertificateNotAfterAnnotation] = systemAdminCredsSecret.Annotations[certrotation.CertificateNotAfterAnnotation]
+	}
```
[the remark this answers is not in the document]
vrutkovs: somehow it doesn't crash in my unit tests, but lets add this just to be safe

wangke19:
> @wangke19 is there anything else we should consider before merging this pr ?

lgtm

pkg/operator/nodekubeconfigcontroller/nodekubeconfigcontroller.go
```diff
@@ -152,6 +153,15 @@ func ensureNodeKubeconfigs(ctx context.Context, client coreclientv1.CoreV1Interf
 		requiredSecret.Annotations = map[string]string{}
 	}
 	requiredSecret.Annotations[annotations.OpenShiftComponent] = "kube-apiserver"
+	// Copy not-before/not-after annotations from systemAdminClientCert
+if systemAdminCredsSecret.Annotations != nil {
```
p0lyn0mial, line 157: in go reading from `nil` map works, no ?
xref: https://go.dev/play/p/hrmvBX3L2Q1

p0lyn0mial: `if len(systemAdminCredsSecret.Annotations[certrotation.CertificateNotBeforeAnnotation]) > 0` also works, right ?

wangke19: yes, the original code snippet is safe as-is regarding nil map reads.

vrutkovs: >len(systemAdminCredsSecret.Annotations[certrotation.CertificateNotBeforeAnnotation]) > 0

That would never reset an empty annotation if it was ever set, wouldn't it? Setting it to empty string is a common method to make controller set it again

p0lyn0mial: thanks

PR #1875, opened 2025-07-28
OCPBUGS-50489: Add missing service ports to apiserver service

shirmoran:
Add ports 6080 (insecure-readyz) and 17697 (check-endpoints) to the apiserver service. These ports were already exposed by the pods but missing from the service definition, preventing EndpointSlice creation for communication matrix documentation (for mor details see [bug](https://issues.redhat.com/browse/OCPBUGS-50489) description)

### DIFF
--- a/bindata/assets/kube-apiserver/svc.yaml
+++ b/bindata/assets/kube-apiserver/svc.yaml
@@ -11,3 +11,12 @@ spec:
   - name: https
     port: 443
     targetPort: 6443
+    protocol: TCP
+  - name: insecure-readyz
+    port: 6080
+    targetPort: 6080
+    protocol: TCP
+  - name: check-endpoints
+    port: 17697
+    targetPort: 17697
+    protocol: TCP


vrutkovs:
/hold

Who's communicating with kube-apiserver via insecure-readyz and check-endpoint ports?

shirmoran:
> /hold
> 
> Who's communicating with kube-apiserver via insecure-readyz and check-endpoint ports?

@vrutkovs insecure-readyz and check-endpoint ports are open externally, and for that they need a service to enable the creation of endpointslices (or communication matrix documentation purposes). Not sure if those ports should be exposed, but if they are exposed they should be documented.

shirmoran:
> > for that they need a service to enable the creation of endpointslices (or communication matrix documentation purposes).
> 
> `check-endpoints` has hostPort set so it doesn't need a service. Same goes for `insecure-readyz` - LB is pointing directly to the pod, so neither of those needs service entries. So this is why I'm asking - who's communicating with those pods via a service? Am I missing some consumer - or service entries are required because communication matrix is not aware of those methods?

We are working on a new requirement for all open ports to be documented with Endpointslices (see [enhancment proposal](https://github.com/openshift/enhancements/blob/master/enhancements/network/communication-flows-matrix-ingress.md) of the communication matrix). This is why every port opened externally should have a service (even if it is used for documentation purposes only)  which creates an EPS automatically.

vrutkovs:
> This is why every port opened externally should have a service (even if it is used for documentation purposes only)

Oh,  I see, okay.
/hold cancel

vrutkovs (approved):
/approve
/lgtm

PR #1882, opened 2025-07-31
NO-JIRA: Add negative tests of GOAWAY-chance

wangke19:
Add negative tests of GOAWAY-chance

### DIFF
--- a/pkg/operator/configobservation/apiserver/observe_goaway_chance_test.go
+++ b/pkg/operator/configobservation/apiserver/observe_goaway_chance_test.go
@@ -88,6 +88,30 @@ func TestObserveGoawayChance(t *testing.T) {
 				"goaway-chance": []interface{}{"0"},
 			}},
 		},
+
+		// scenario 8 - Invalid goaway-chance value (negative)
+		{
+			name:                 "invalid: negative value",
+			controlPlaneTopology: configv1.HighlyAvailableTopologyMode,
+			existingKubeAPIConfig: map[string]interface{}{"apiServerArguments": map[string]interface{}{
+				"goaway-chance": []interface{}{"-0.1"},
+			}},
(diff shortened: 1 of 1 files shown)


pkg/operator/configobservation/apiserver/observe_goaway_chance_test.go
```diff
@@ -88,6 +88,30 @@ func TestObserveGoawayChance(t *testing.T) {
...
+
+		// scenario 9 - Invalid goaway-chance value (greater than 0.2)
+		{
+			name:                 "invalid: value greater than 0.2",
+			controlPlaneTopology: configv1.HighlyAvailableTopologyMode,
+			existingKubeAPIConfig: map[string]interface{}{"apiServerArguments": map[string]interface{}{
+				"goaway-chance": []interface{}{"0.3"},
+			}},
+			expectedKubeAPIConfig: map[string]interface{}{"apiServerArguments": map[string]interface{}{
+				"goaway-chance": []interface{}{"0.001"},
```
vrutkovs, line 112: Do we have it reset to `0.001` or it would crashloop too? Perhaps we should add `expectedError` here too?

wangke19: In this test case, a negative goaway-chance value in an HA cluster is automatically sanitized to the default of "0.001" by the ObserveGoawayChance function. The goal is to ensure invalid values are corrected automatically without crashing or failing the operator logic. `crashloop` is not result we expected.

Adding `expectedError` would make your tests more complex but wouldn’t add meaningful verification under the current logic.
Here is test results:
```
/usr/local/go/bin/go test -timeout 30s -run ^TestObserveGoawayChance$ github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/configobservation/apiserver

ok  	github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/configobservation/apiserver	0.019s
```

PR #1883, opened 2025-07-31
WIP Expire node-system-admin-client and node-system-admin-signer

vrutkovs:


vrutkovs:
[Test using these kubeconfigs](https://github.com/openshift/origin/blob/main/test/extended/apiserver/kubeconfigs.go) now fails, but other components are okay

PR #1869, opened 2025-07-11
Revert "OCPBUGS-56551: certregenerationcontroller: start configInformers again"

neisw:
Reverts openshift/cluster-kube-apiserver-operator#1853

4.19 backport [cluster-kube-apiserver-operator#1861](https://github.com/openshift/cluster-kube-apiserver-operator/pull/1861) was [reverted](https://github.com/openshift/cluster-kube-apiserver-operator#1867) due to pathological event failures during minor upgrade.  Confirmed with @vrutkovs we should revert here as well while the fix is reworked.

/cc @vrutkovs

PR #1812, opened 2025-02-19
OCPBUGS-55217: Optimistically update Kube Server and Client CA bundles

vrutkovs:
Instead of re-creating configmap from scratch every time this function should attempt to use existing configmap and replace the contents only. This would prevent extra configmap updates when metadata changes.

Bump library-go to update https://github.com/openshift/library-go/pull/1936

PR #1891, opened 2025-08-06
Revert "Revert "OCPBUGS-56551: certregenerationcontroller: start configInformers again""

vrutkovs:
Reverts openshift/cluster-kube-apiserver-operator#1869

Now that CA Bundles are created optimistically in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1812, we no longer get flooded with create/update CA bundle updates - so we can start informers again to make cert rotation work after suspend/shutdown


vrutkovs:
I ran:
* cert rotation suspend tests to ensure starting informers makes sidecar controller regenerate certs as expected
* payload upgrade 4.19 -> 4.20 to make sure we're not spamming with events anymore (this is why this was reverted the first time)

PR #1896, opened 2025-08-14
NO-JIRA: Use BoundServceAccountTokenVolume de default

lance5890:


lance5890:
> Could you create a jira ticket with a more thorough description? We certainly need to discuss it first in order to avoid regressions - and consider backporting it too

I will create a Jira ticket for this. The change has been tested in my local cluster, and I'll share more details soon. Do you think this PR deserves an `/ok-to-test to` verify if it works in the ocp CI environment?

vrutkovs:
Ah, yes, sure
/ok-to-test

lance5890:
> Could you create a jira ticket with a more thorough description? We certainly need to discuss it first in order to avoid regressions - and consider backporting it too

the related pr https://github.com/openshift/cluster-version-operator/pull/660  will block the upgrade

lance5890:
with the https://github.com/openshift/cluster-version-operator/pull/1264 merged, now this is not blocked too@tchap

PR #1908, opened 2025-08-27
OCPBUGS-45571: sccreconcilecontroller: Handle scc not found

tchap:
Fix scc not found errors not being handled and spamming logs.


vrutkovs:
/hold 

How is this "fixing" error? I think its just hiding it, isn't it? If this error occurs it means some component has required `key` scc

PR #1909, opened 2025-08-28
OCPBUGS-60997: targetconfigcontroller: make sure extension-apiserver-authentication has necessary annotations

vrutkovs:
configmap kube-system/extension-apiserver-authentication is created by kube-apiserver, but it doesn't have ownership metadata. This commit updates target config controller to set necessary metadata (ownership and description)


pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -507,6 +513,30 @@ func ensureKubeAPIServerTrustedCA(ctx context.Context, client coreclientv1.CoreV
...
+		}
+		return err
+	}
+
+	// update if annotations modified by the user
+	if !reflect.DeepEqual(cm.Annotations, required.Annotations) {
+		cm.Annotations = required.Annotations
+		updatedCM, err := cmCLient.Update(ctx, cm, metav1.UpdateOptions{})
+		resourcehelper.ReportUpdateEvent(recorder, updatedCM, err)
+		return err
```
lance5890, line 534: this line 534 `return err` maybe redundant as line 537 can return this err too

vrutkovs: Yeah, good idea

lance5890: yes, it might be shadowed when use the same err, my bad

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -237,6 +237,11 @@ func createTargetConfig(ctx context.Context, c TargetConfigController, recorder
 		errors = append(errors, fmt.Errorf("%q: %v", "configmap/trusted-ca-bundle", err))
 	}
 
+	err = ensureKubeAPIServerExtensionAuthenticationCA(ctx, c.kubeClient.CoreV1(), recorder)
```
wangke19, line 240: The new logic is not covered by unit tests. Adding a unit test for ensureKubeAPIServerExtensionAuthenticationCA would be beneficial to verify its behavior under different conditions
      (e.g., when the configmap doesn't exist, when it has no annotations, or when it has incorrect annotations) and to protect against future regressions.

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -507,6 +512,39 @@ func ensureKubeAPIServerTrustedCA(ctx context.Context, client coreclientv1.CoreV
...
+			modified = true
+		}
+	}
+
+	if modified {
+		if _, err := cmClient.Update(ctx, cm, metav1.UpdateOptions{}); err != nil {
+		    recorder.Warningf("AnnotationUpdateFailed", "Failed to update annotations on configmap kube-system/extension-apiserver-authentication: %v", err)
+		 }
+		return nil
+	}
```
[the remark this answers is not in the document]
vrutkovs: Fixed, also needed `return err` when err != nil

PR #1915, opened 2025-08-29
NO-JIRA: Add priority field to prevent early shutdown

CoreyCook8:
Based on the issue described here : https://github.com/kubernetes/kubernetes/issues/133442

priorityClassName is currently ignored by Kubelet for static pod files so setting this value has no impact on the gracefulShutdown order causing the static pods to start to be killed as soon as shutdown begins.

To prevent this we must set priority explicitly

## Summary by CodeRabbit

* **Chores**
  * Increased the kube-apiserver Pod’s scheduling priority to improve reliability under node pressure and ensure critical control-plane availability.
  * No functional changes to existing settings; behavior remains the same aside from higher scheduling precedence.

### DIFF
--- a/bindata/assets/kube-apiserver/pod.yaml
+++ b/bindata/assets/kube-apiserver/pod.yaml
@@ -288,6 +288,7 @@ spec:
       readOnlyRootFilesystem: true
   terminationGracePeriodSeconds: {{.GracefulTerminationDuration}}
   hostNetwork: true
+  priority: 2000001000
   priorityClassName: system-node-critical
   tolerations:
   - operator: "Exists"


benluddy:
/lgtm

Thanks for opening this PR and calling out this issue!

There's a concise summary here for the record: https://github.com/kubernetes/kubernetes/issues/133535#issuecomment-3189884729. It's the numeric priority that Kubelet uses to determine shutdown priority, and the priority admission plugin is responsible for setting a numeric priority based on a priority class name. Static pods aren't created through API admission, so they don't get the numeric priority.

I don't know if we absolutely need kube-apiserver to be shut down later on an HA topology, since other components on the same node should continue to work while their local kube-apiserver process rolls out, but it does seem good to shrink the window of time we're running with reduced redundancy, even if it ends up making node shutdown take a bit longer. Maybe we'll see smoother rollouts on single-node with this?

benluddy:
/retitle NO-JIRA: Add priority field to prevent early shutdown

PR #1919, opened 2025-09-11
OCPBUGS-46086: Always set service-account-jwks-uri to LB URL even with custom issuer

ShazaAldawamneh:
Previously, when a custom service account issuer was set, the
service-account-jwks-uri argument was not configured in the KubeAPIServer,
causing the JWKS URI to default to the node IP. This led to TLS errors
because the node IP is not included in the certificate SAN.

This commit updates observedConfig to always set
service-account-jwks-uri to the API LB URL regardless of whether the
issuer is default or custom. Unit tests have been updated to validate
this behavior.

Fixes: TLS SAN issues for clients accessing JWKS URI with custom issuers.

### DIFF
--- a/pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
+++ b/pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
@@ -125,21 +125,74 @@ func observedConfig(existingConfig map[string]interface{},
 		"api-audiences":          apiServerArgumentValue,
 	}
 
-	// If the issuer is not set in KAS, we rely on the config-overrides.yaml to set both
-	// the issuer and the api-audiences but configure the jwks-uri to point to
-	// the LB so that it does not default to KAS IP which is not included in the serving certs
+	// If the issuer is not set in KAS, we rely on config-overrides.yaml to provide both
+	// the service-account-issuer and api-audiences. In that case, we still configure
+	// service-account-jwks-uri to point at the API server load balancer endpoint so we
+	// don’t fall back to the KAS IP, which is not included in the serving cert SANs.
+	//
+	// The service-account-jwks-uri is configured based on the issuer type:
+	//
+	// Default issuer:
+	//   We set service-account-jwks-uri to the API server load balancer endpoint
+	//   (<apiServerURL>/openid/v1/jwks) so clients don’t fall back to the KAS IP,
+	//   which is not covered by the serving certificate SANs.
+	//
+	// Custom issuer:
+	//   A custom issuer is expected to be an OIDC issuer (a URL). It should fall into
+	//   one of two categories:
+	//
+	//   1) Issuer is the external cluster URL. In this case, the API server already
+	//      serves both the discovery document and public keys at:
+	//        - <issuer>/.well-known/openid-configuration
+	//        - <issuer>/openid/v1/jwks
+	//      This is effectively the same behavior as the default issuer case.
+	//
+	//   2) Issuer is any other URL. In this case, we expect the user to host both the
+	//      OIDC discovery document and JWKS outside of the cluster at:
+	//        - <issuer>/.well-known/openid-configuration
+	//        - <issuer>/openid/v1/jwks
+	//
+	//   Any other JWKS location would require a future enhancement (RFE) and is not
+	//   currently supported.
+
 	if observedActiveIssuer == defaultServiceAccountIssuerValue {
+		// Default issuer → use API LB
 		infrastructureConfig, err := getInfrastructureConfig("cluster")
 		if err != nil {
 			return existingConfig, append(errs, err)
 		}
-		if apiServerExternalURL := infrastructureConfig.Status.APIServerURL; len(apiServerExternalURL) == 0 {
+
+		apiServerExternalURL := infrastructureConfig.Status.APIServerURL
+		if len(apiServerExternalURL) == 0 {
 			return existingConfig, append(errs, fmt.Errorf("APIServerURL missing from infrastructure/cluster"))
-		} else {
-			apiServerArguments["service-account-jwks-uri"] = []interface{}{apiServerExternalURL + "/openid/v1/jwks"}
+		}
+
+		apiServerArguments["service-account-jwks-uri"] = []interface{}{
+			apiServerExternalURL + "/openid/v1/jwks",
+		}
+	} else {
+		// Custom issuer → only set jwks-uri if the issuer is a valid HTTPS URL,
+		// as required by the OpenID Discovery 1.0 spec. If it's not, we skip setting
+		// the flag and log a warning. Per the kube-apiserver docs, a non-HTTPS issuer
+		// will also cause ServiceAccountIssuerDiscovery to be disabled.
+		parsed, parseErr := url.Parse(observedActiveIssuer)
+		switch {
+		case parseErr != nil:
+			klog.Warningf("custom service account issuer %q could not be parsed as a URL: %v; "+
+				"ServiceAccountIssuerDiscovery will be disabled and service-account-jwks-uri will not be configured. "+
+				"It is highly recommended that the issuer be a valid HTTPS URL per the OpenID Discovery spec.",
+				observedActiveIssuer, parseErr)
+		case parsed.Scheme != "https":
+			klog.Warningf("custom service account issuer %q has scheme %q instead of \"https\"; "+
+				"ServiceAccountIssuerDiscovery will be disabled and service-account-jwks-uri will not be configured. "+
+				"It is highly recommended that the issuer be a valid HTTPS URL per the OpenID Discovery spec.",
+				observedActiveIssuer, parsed.Scheme)
+		default:
+			apiServerArguments["service-account-jwks-uri"] = []interface{}{
+				observedActiveIssuer + "/openid/v1/jwks",
+			}
 		}
 	}
-
 	return map[string]interface{}{"apiServerArguments": apiServerArguments}, errs
 
 }
--- a/pkg/operator/configobservation/auth/auth_serviceaccountissuer_test.go
+++ b/pkg/operator/configobservation/auth/auth_serviceaccountissuer_test.go
@@ -31,87 +31,128 @@ func TestObservedConfig(t *testing.T) {
 		issuer                 string
 		trustedIssuers         []string
 		existingIssuer         string
+		existingJWKSURI        string
 		authError              error
 		infraError             error
 		expectedIssuer         string
 		expectedTrustedIssuers []string
 		expectedChange         bool
-		expectInternalJWKI     bool
+		expectedJWKSURI        string
 	}{
 		{
-			name:               "no issuer, no previous issuer means we default",
-			existingIssuer:     "",
-			issuer:             defaultServiceAccountIssuerValue,
-			expectedIssuer:     defaultServiceAccountIssuerValue,
-			expectInternalJWKI: true,
+			name:            "no issuer, no previous issuer means we default",
+			existingIssuer:  "",
+			existingJWKSURI: "",
+			issuer:          defaultServiceAccountIssuerValue,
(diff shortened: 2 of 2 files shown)


pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
```diff
@@ -128,16 +128,14 @@ func observedConfig(existingConfig map[string]interface{},
...
-		} else {
-			apiServerArguments["service-account-jwks-uri"] = []interface{}{apiServerExternalURL + "/openid/v1/jwks"}
-		}
+	infrastructureConfig, err := getInfrastructureConfig("cluster")
+	if err != nil {
+		return existingConfig, append(errs, err)
+	}
+	if apiServerExternalURL := infrastructureConfig.Status.APIServerURL; len(apiServerExternalURL) == 0 {
+		return existingConfig, append(errs, fmt.Errorf("APIServerURL missing from infrastructure/cluster"))
+	} else {
```
sjenning: `else` is not needed since we are returning in the `if`

pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
```diff
@@ -128,16 +128,20 @@ func observedConfig(existingConfig map[string]interface{},
...
-		}
+	// Always configure service-account-jwks-uri to point to the API LB
+	infrastructureConfig, err := getInfrastructureConfig("cluster")
+	if err != nil {
+		return existingConfig, append(errs, err)
+	}
+
+	apiServerExternalURL := infrastructureConfig.Status.APIServerURL
+	if len(apiServerExternalURL) == 0 {
+		return existingConfig, append(errs, fmt.Errorf("APIServerURL missing from infrastructure/cluster"))
```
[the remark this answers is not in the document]
ShazaAldawamneh: Thanks for raising this @benluddy good point.

Do you think it would be better to avoid failing the entire observed config if status.apiServerURL is empty, and instead just skip setting service-account-jwks-uri while still configuring service-account-issuer and api-audiences?

I want to make sure we don’t block issuer configuration on clusters where APIServerURL might not be populated.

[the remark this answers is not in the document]
everettraven: Anything outside of the scope of ^ relating to setting the JWKS URI to a different location than `issuer + /openid/v1/jwks` would be an RFE IMO. It's not currently a use case we would support based on our existing implementation.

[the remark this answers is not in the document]
ShazaAldawamneh: Thanks Bryce, this really helps clarify the tradeoffs.

Your proposed logic (default issuer = existing behavior, custom issuer = issuer + /openid/v1/jwks) makes sense and avoids hard-depending on apiServerURL.
Before I change the implementation, can you confirm you'd see that approach over always forcing the LB URL better? @everettraven

pkg/operator/configobservation/auth/auth_serviceaccountissuer_test.go
```diff
@@ -151,19 +152,27 @@ func TestObservedConfig(t *testing.T) {
...
+				expectedJWKSURI = "https://lb.example.com/openid/v1/jwks"
+			default:
+				expectedJWKSURI = tc.issuer + "/openid/v1/jwks"
 			}
+			// Check that JWKS URI is correctly set
+			uri, ok := unmarshalledConfig.APIServerArguments["service-account-jwks-uri"]
+			// Always expect JWKS URI to be set now
+			require.True(t, ok, "expected service-account-jwks-uri to be set")
+			require.Equal(t, kubecontrolplanev1.Arguments{expectedJWKSURI}, uri)
+
```
everettraven: With the recent implementation changes, I think the expectations we want to check in our tests has changed a bit.

These are the scenarios I think we need to make sure we've effectively captured:
- No existing configuration of the `service-account-jwks-uri`, default issuer, we successfully retrieve the API server load balancer URL, `service-account-jwks-uri` is successfully set to `{loadBalancerURL}/openid/v1/jwks`
- No existing configuration of the `service-account-jwks-uri`, default issuer, we _do not_ successfully retrieve the API server load balancer URL, `service-account-jwks-uri` should NOT be set.
- No existing configuration of the `service-account-jwks-uri`, custom issuer, `service-account-jwks-uri` should be set to `{issuer}/openid/v1/jwks`.
- Existing configuration of the `service-account-jwks-uri`, default issuer, we successfully retrieve the API server load balancer URL, `service-account-jwks-uri` is successfully set to `{loadBalancerURL}/openid/v1/jwks`
- Existing configuration of the `service-account-jwks-uri`, default issuer, we _do not_ successfully retrieve the API server load balancer URL, `service-account-jwks-uri` is the same as what was present in the existing config
- Existing configuration of the `service-account-jwks-uri`, custom issuer, `service-account-jwks-uri` should be set to `{issuer}/openid/v1/jwks`.

pkg/operator/configobservation/auth/auth_serviceaccountissuer_test.go
```diff
@@ -151,19 +152,27 @@ func TestObservedConfig(t *testing.T) {
...
+				// infra error → fallback to default service account issuer
+				expectedJWKSURI = "https://kubernetes.default.svc/openid/v1/jwks"
+			case tc.issuer == defaultServiceAccountIssuerValue:
+				expectedJWKSURI = "https://lb.example.com/openid/v1/jwks"
+			default:
+				expectedJWKSURI = tc.issuer + "/openid/v1/jwks"
 			}
+			// Check that JWKS URI is correctly set
+			uri, ok := unmarshalledConfig.APIServerArguments["service-account-jwks-uri"]
+			// Always expect JWKS URI to be set now
```
everettraven: This is no longer true. We are preserving the logic for the default issuer case which means that if there was no previous configuration that successfully resolved a `service-account-jwks-uri` argument and we are _still_ not able to successfully resolve it (i.e an infra error) we would not expect this argument to be present in the configuration.

pkg/operator/configobservation/auth/auth_serviceaccountissuer_test.go
```diff
@@ -151,19 +152,27 @@ func TestObservedConfig(t *testing.T) {
...
-				t.Errorf("expected no service-account-jwks-uri to be set, it is %+v", uri.String())
+			// Determine expected JWKS URI based on what observedConfig will do
+			var expectedJWKSURI string
+			switch {
+			case tc.authError != nil:
+				// auth error → keep existing issuer
+				expectedJWKSURI = tc.existingIssuer + "/openid/v1/jwks"
+			case tc.infraError != nil:
+				// infra error → fallback to default service account issuer
+				expectedJWKSURI = "https://kubernetes.default.svc/openid/v1/jwks"
```
everettraven: This case would be whatever the previous configuration value was

pkg/operator/configobservation/auth/auth_serviceaccountissuer_test.go
```diff
@@ -151,19 +152,27 @@ func TestObservedConfig(t *testing.T) {
...
+			case tc.issuer == defaultServiceAccountIssuerValue:
+				expectedJWKSURI = "https://lb.example.com/openid/v1/jwks"
+			default:
+				expectedJWKSURI = tc.issuer + "/openid/v1/jwks"
 			}
```
rh-roman: What do you think about making `expectedJWKSURI` be one of the test parameters. Something like

```
		{
			name:           "custom issuer, no previous issuer",
			existingIssuer: "",
			issuer:         "https://custom.com",
			expectedIssuer: "https://custom.com",
            expectedJWKSURI: "https://custom.com/openid/v1/jwks",
			expectedChange: true,
		},
```
That removes all this logic from the test body and makes it easier to see expected output in test declaration. Also gives you a place to leave a comment about why it's the expected URI (if you want).

pkg/operator/configobservation/auth/auth_serviceaccountissuer_test.go
```diff
@@ -83,33 +85,44 @@ func TestObservedConfig(t *testing.T) {
...
+			issuer:          defaultServiceAccountIssuerValue,
+			infraError:      expectedErrInfra,
+			expectedIssuer:  defaultServiceAccountIssuerValue,
+			expectedJWKSURI: "https://lb.example.com/openid/v1/jwks", // no previous + infra error -> do NOT set JWKS
+		},
+		{
+			name:            "default issuer, no previous issuer, infra getter error",
+			existingIssuer:  "",
+			issuer:          defaultServiceAccountIssuerValue,
+			expectedIssuer:  defaultServiceAccountIssuerValue,
```
everettraven: You wouldn't have an expected issuer here because there was no previously existing issuer, and because we return the existing configuration in that case, it would be an empty issuer.

pkg/operator/configobservation/auth/auth_serviceaccountissuer_test.go
```diff
@@ -151,19 +163,17 @@ func TestObservedConfig(t *testing.T) {
...
-			}
-			if !tc.expectInternalJWKI && ok {
-				t.Errorf("expected no service-account-jwks-uri to be set, it is %+v", uri.String())
+			if tc.expectedJWKSURI != "" {
+				require.True(t, ok, "expected service-account-jwks-uri to be set")
+				require.Equal(t, kubecontrolplanev1.Arguments{tc.expectedJWKSURI}, uri)
+			} else {
+				require.False(t, ok, "did not expect service-account-jwks-uri to be set")
 			}
+
```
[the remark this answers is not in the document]
ShazaAldawamneh: Good catch. I've removed the explicit check for the JWKS URI and am now relying on the final config comparison. It simplifies the test logic quite a bit. Thanks!

pkg/operator/configobservation/auth/auth_serviceaccountissuer.go
```diff
@@ -125,18 +125,55 @@ func observedConfig(existingConfig map[string]interface{},
...
+	//
+	// The service-account-jwks-uri is configured based on the issuer type:
+	//
+	// Default issuer:
+	//   We set service-account-jwks-uri to the API server load balancer endpoint
+	//   (<apiServerURL>/openid/v1/jwks) so clients don’t fall back to the KAS IP,
+	//   which is not covered by the serving certificate SANs.
+	//
+	// Custom issuer:
+	//   A custom issuer is expected to be an OIDC issuer (a URL). It should fall into
```
benluddy: Do we need to validate this assumption? If it's not, we pass a non-URI string for `--service-account-jwks-uri` and I _think_ an option validation error will prevent kube-apiserver from starting.

everettraven: We can certainly add tighter guardrails here, and I would encourage that to be done at least on the API side for admission time validation.

Depending on whether or not this needs to be backported, and how far back it needs to be backported, the API side validations should automatically ratchet, making it so that existing invalid instances aren't totally invalidated but that new attempts to set/update the issuer values to a non-URL issuer would be rejected.

We could also add a verification step here if folks wanted stronger validation along the chain, but that would mean we would need to be careful in how we implement it to prevent suddenly breaking the configuration syncing logic for end-users that may have incorrectly configured this field.

I'm not strongly opinionated one way or another, but in an ideal world we would do both. If we had to pick one or the other, I'd probably lean towards ratcheting the API validation so that we don't risk breaking the configuration process here.

Worst case scenario seems like workload identity for cloud-provider interactions doesn't work quite properly but that the cluster itself remains fully operational. I suspect that a mishap here would be stumbled upon fairly quickly during configuration of this functionality - unless the users configuring it are not doing any testing to ensure they've configured things appropriately. All that to say, if we have bandwidth to appropriately ratchet validation without breaking users lets do it - otherwise it seems like a pretty low severity thing to me.

PR #1922, opened 2025-09-17
OCPBUGS-15430: move alerting rules from CMO over

tjungblu:


tjungblu:
@verified by @tjungblu 

installed correctly

image

PR #1923, opened 2025-09-18
enable resource v1beta2 api if DynamicResourceAllocation is enabled

tkashem:
enable resource v1beta2 api if the DRA feature gate DynamicResourceAllocation is enabled. v1beta2 has been added in 1.33


tkashem:
Note: it's late for 4.20.0, but we should backport it to the next 4.20.z stream

/cherry-pick release-4.20

tkashem:
/hold

(rebase may be necessary after https://github.com/openshift/cluster-kube-apiserver-operator/pull/1932 merges)

PR #1927, opened 2025-09-22
OCPSTRAT-2371: MutatingAdmissionPolicy e2es depend on both v1alpha1 and v1beta1 of admissionregistration.k8s.io being served in k8 1.34

bertinatto:
The API started being served in v1.32.0: https://github.com/kubernetes/kubernetes/blob/release-1.34/pkg/features/kube_features.go#L1964

/assign @benluddy 
CC @jacobsee

### DIFF
--- a/pkg/operator/configobservation/apienablement/observe_runtime_config.go
+++ b/pkg/operator/configobservation/apienablement/observe_runtime_config.go
@@ -17,8 +17,16 @@ import (
 
 var defaultGroupVersionsByFeatureGate = map[configv1.FeatureGateName][]groupVersionByOpenshiftVersion{
 	"MutatingAdmissionPolicy": {
-		{KubeVersionRange: semver.MustParseRange("< 1.34.0"), GroupVersion: schema.GroupVersion{Group: "admissionregistration.k8s.io", Version: "v1alpha1"}},
-		{KubeVersionRange: semver.MustParseRange(">= 1.34.0"), GroupVersion: schema.GroupVersion{Group: "admissionregistration.k8s.io", Version: "v1beta1"}},
+		// Both v1alpha1 and v1beta1 versions must be served pre-GA because e2e tests exercise both APIs.
+		// A GA OpenShift release could inadvertently serve these versions if MutatingAdmissionPolicy
+		// gets added to the default featureSet in openshift/api as part of transitioning from
+		// (feature off, v1beta1 off) to (feature on, v1 on).
+		// To prevent that, version ranges below include min and max bounds.
+		// TODO: Update version ranges when rebasing to k8s v1.35+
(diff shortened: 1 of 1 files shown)


bertinatto:
/payload-job-with-prs periodic-ci-openshift-release-master-ci-4.21-e2e-aws-ovn-techpreview openshift/kubernetes#2148

pkg/operator/configobservation/apienablement/observe_runtime_config.go
```diff
@@ -17,8 +17,13 @@ import (
 
 var defaultGroupVersionsByFeatureGate = map[configv1.FeatureGateName][]groupVersionByOpenshiftVersion{
 	"MutatingAdmissionPolicy": {
-		{KubeVersionRange: semver.MustParseRange("< 1.34.0"), GroupVersion: schema.GroupVersion{Group: "admissionregistration.k8s.io", Version: "v1alpha1"}},
-		{KubeVersionRange: semver.MustParseRange(">= 1.34.0"), GroupVersion: schema.GroupVersion{Group: "admissionregistration.k8s.io", Version: "v1beta1"}},
+		// MutatingAdmissionPolicy was introduced as TechPreviewNoUpgrade in OCP 4.20 (k8s v1.33) with v1alpha1 API.
+		// OCP 4.21 (k8s v1.34) added v1beta1 API support alongside v1alpha1.
```
benluddy, line 21: Please mention that the upstream E2E feature tests exercise both versions, so pre-GA we must serve both.

pkg/operator/configobservation/apienablement/observe_runtime_config.go
```diff
@@ -17,8 +17,13 @@ import (
...
+		// Both APIs are expected to be removed when the feature goes GA.
```
benluddy, line 22: They won't be removed right away. What we must prevent is a GA release _of OpenShift_ that serves either of these versions. That could happen if MutatingAdmissionPolicy is added to the default featureset in openshift/api in the future as part of a pivot from (feature default-off, v1beta1 default-off) to (feature default-on, v1 default-on).

pkg/operator/configobservation/apienablement/observe_runtime_config.go
```diff
@@ -17,8 +17,13 @@ import (
...
+		// GA is tentatively assumed for k8s v1.35, but the KEP does not specify a version yet: https://github.com/kubernetes/enhancements/issues/3962
+		// TODO: Update these ranges when rebasing to k8s v1.35+
```
benluddy, line 24: To be specific: if we have a v1 version of the resources required for MutatingAdmissionPolicy in 1.35, then we can remove all references to MutatingAdmissionPolicy here. If we _don't_ get v1 resources, then these ranges should be bumped to "<1.36.0" and we'll make another decision during the following rebase.

PR #1933, opened 2025-10-06
NO-ISSUE: pkg/operator/kubeletversionskewcontroller: Explicit next OpenShift version in skew message

wking:
Explicitly name the next OpenShift version, to reduce the chances that cluster administrators think the message applies to a currently in-progress update, when `Upgradeable` is only talking about future updates.


wking:
Rebased around f6a1c91112 with 4be403c47 -> 968ba6961

PR #1935, opened 2025-10-08
scc: give unauthenticated users access to restricted-v3

haircommander:

PR #1944, opened 2025-10-20
CNTRLPLANE-1544: scc: Grant authenticated users use of restricted-v3

tchap:
Taking over https://github.com/openshift/cluster-kube-apiserver-operator/pull/1935

There is a missing `ClusterRoleBinding` that should have been added with `restricted-v3`.

Related tests change: https://github.com/openshift/origin/pull/30384


tchap:
/verified by "[sig-auth][Feature:OpenShiftAuthorization] The default cluster RBAC policy should have correct RBAC rules"

PR #1949, opened 2025-10-24
OCPBUGS-65755: Stop serving VolumeAttributesClass v1beta1

bertinatto:
Version v1 is available in Kubernetes 1.34.

/assign @benluddy


jsafrane:
/lgtm
from storage point of view, everyone should use v1

PR #1979, opened 2025-12-02
OCPBUGS-66136: certregenerationcontroller: Add cache sync timeout

tchap:
This critical controller can get stuck on cache sync when the 
certificates expire. Crashing when the initial cache sync times out 
should improve the situation.

There is now a hard-coded 2-minute timeout after which the controllers call `os.Exit(1)` from within their `Run` functions when caches don't manage to sync in time. This also improves the visibility of the issue as the container doesn't seem to be running just fine.

---

This change is based on seeing a client case where the recovery controller got stuck for 16 days before it got terminated.


tchap:
/jira refresh

Gosh, closed the PR by accident, reopening...

PR #1983, opened 2025-12-12
OCPBUGS-69758:CNTRLPLANE-2222:Migrate go test cert-rotation-tests.go to OTE

gangwgr:
[CNTRLPLANE-2222](https://issues.redhat.com//browse/CNTRLPLANE-2222):Migrate go test cert-rotation-tests.go to OTE

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/google/go-cmp v0.7.0
 	github.com/imdario/mergo v0.3.8
 	github.com/miekg/dns v1.1.61
+	github.com/onsi/ginkgo/v2 v2.21.0
 	github.com/openshift-eng/openshift-tests-extension v0.0.0-20250804142706-7b3ab438a292
 	github.com/openshift/api v0.0.0-20251015095338-264e80a2b6e7
 	github.com/openshift/build-machinery-go v0.0.0-20250530140348-dc5b2804eeee
@@ -78,7 +79,6 @@ require (
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.3-0.20250322232337-35a7c28c31ee // indirect
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
-	github.com/onsi/ginkgo/v2 v2.21.0 // indirect
 	github.com/onsi/gomega v1.35.1 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect


test/e2e/certrotation_test.go
```diff
@@ -1,146 +1,11 @@
...
-	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator"
-	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/operatorclient"
-	test "github.com/openshift/cluster-kube-apiserver-operator/test/library"
-	configv1helpers "github.com/openshift/library-go/pkg/config/clusteroperator/v1helpers"
-	"github.com/openshift/library-go/pkg/operator/genericoperatorclient"
-	"github.com/openshift/library-go/pkg/operator/v1helpers"
-)
+import "testing"
 
 func TestCertRotationTimeUpgradeable(t *testing.T) {
```
p0lyn0mial: let's add a comment:

```
// This test calls the shared function which
// can be called from both standard Go tests and Ginkgo 
//
// This situation is temporary until we test the new e2e-gcp-operator-serial-ote job.
// Eventually all tests will be run only as part of the OTE framework.
```

test/e2e/certrotation_test.go
```diff
@@ -1,146 +1,11 @@
...
-			return true, nil
-		}
-		t.Logf("\nCertRotationTimeUpgradeable: %#v\nUpgradeable: %#v", certRotationCondition, upgradeableCondition)
-		return false, nil
-	})
-	require.NoError(t, err)
+	testCertRotationTimeUpgradeable(t)
 }
 
 func TestCertRotationStompOnBadType(t *testing.T) {
```
p0lyn0mial: and here.

go.mod
```diff
@@ -36,6 +36,8 @@ require (
 	sigs.k8s.io/kube-storage-version-migrator v0.0.6-0.20230721195810-5c8923c5ff96
 )
 
+require github.com/onsi/ginkgo/v2 v2.21.0
```
p0lyn0mial: could this be part of the require block above ?

cmd/cluster-kube-apiserver-operator-tests-ext/main.go
```diff
@@ -19,6 +19,8 @@ import (
 	"k8s.io/klog/v2"
 
 	"github.com/openshift/cluster-kube-apiserver-operator/pkg/version"
+	// Import test packages to register Ginkgo tests
+	_ "github.com/openshift/cluster-kube-apiserver-operator/test/e2e"
```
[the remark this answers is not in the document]
p0lyn0mial: just a personal preference, similar to https://github.com/openshift/cluster-kube-apiserver-operator/blob/main/pkg/dependencymagnet/doc.go

in the future we will also reference the other tests: https://github.com/openshift/cluster-kube-apiserver-operator/tree/main/test

test/e2e/certrotation_test.go
gangwgr: Updated

gangwgr:
@wangke19 It is dev code test we are not changing anything nothing should be changed. And we also should not change. If we are doing changing it will impact cases. Better should not change in original case code when it is running fine. It is just copied from existing gotest,  nothing should be change

test/e2e/certrotation.go
```diff
@@ -0,0 +1,157 @@
...
+	g.It("[Operator][Serial] TestCertRotationTimeUpgradeable", func() {
+		testCertRotationTimeUpgradeable(g.GinkgoTB())
+	})
+	g.It("[Operator][Serial] TestCertRotationStompOnBadType", func() {
+		testCertRotationStompOnBadType(g.GinkgoTB())
+	})
+})
+
+func testCertRotationTimeUpgradeable(t testing.TB) {
+	kubeConfig, err := testlibrary.NewClientConfigForTest()
```
p0lyn0mial: could you adda a new tmp commit with t.Fail() just to see if the expected CI jobs will fail?

PR #1997, opened 2026-01-12
NO-JIRA: Add API server stabilization wait to serviceaccountissuer tests

gangwgr:
This PR improves the reliability of the serviceaccountissuer e2e tests by ensuring the API server has stabilized after configuration changes before proceeding with validation.

  ## Changes

  - Add `WaitForAPIServerToStabilizeOnTheSameRevision` calls after each `setServiceAccountIssuer` operation to ensure the API server has rolled out the configuration change
  - Increase test timeouts to `[Timeout:30m]` to accommodate the additional wait time
  - Import `testlibraryapi` from `library-go/test/library/apiserver` for the wait helper

  ## Motivation

  The serviceaccountissuer tests were failing intermittently because they were attempting to verify configuration changes before the API server had fully processed and stabilized on the new configuration. This change ensures we wait for the rollout to complete before validating the expected issuer values.
  
  Fixed Files

  1. /test/e2e/serviceaccountissuer.go (Ginkgo tests)

  Before: Three separate g.It test blocks that could be run individually
  After: Single combined test that runs all three phases sequentially:
  - "[Operator][Serial][Timeout:30m] serviceaccountissuer lifecycle test"
  - Uses g.By() to mark each phase

  2. /test/e2e/serviceaccountissuer_test.go (Standard Go tests)

  Before: Three separate Test* functions
  After: Single combined test TestServiceAccountIssuer that runs all three phases sequentially:
  - Phase 1: Setting first serviceaccountissuer
  - Phase 2: Setting second serviceaccountissuer (verifies first is retained as trusted)
  - Phase 3: Resetting to default serviceaccountissuer

  Why This Fix Was Needed

  The tests depend on sequential execution because:
  1. The first test sets https://first.foo.bar as the issuer
  2. The second test expects BOTH https://second.foo.bar (new) AND https://first.foo.bar (trusted for 24h)
  3. The third test resets to default

  Running them individually would fail because the second test needs the first issuer to already exist.

### DIFF
--- a/test/e2e/serviceaccountissuer.go
+++ b/test/e2e/serviceaccountissuer.go
@@ -18,6 +18,7 @@ import (
 
 	g "github.com/onsi/ginkgo/v2"
 	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/operatorclient"
+	testlibraryapi "github.com/openshift/library-go/test/library/apiserver"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


wangke19:
@gangwgr When the e2e-gcp-operator-serial-ote runs passed, yon can unhold.

test/e2e/serviceaccountissuer.go
```diff
@@ -82,6 +82,8 @@ func testServiceAccountIssuerDefaultIssuer(t testing.TB) {
 	setServiceAccountIssuer(t, authConfigClient, "")
 	err = pollForOperandIssuer(t, kubeClient, []string{"https://kubernetes.default.svc"})
 	require.NoError(t, err, "pollForOperandIssuer failed")
+	// Wait for API server to stabilize after configuration change
+	testlibraryapi.WaitForAPIServerToStabilizeOnTheSameRevision(t, kubeClient.Pods(operatorclient.TargetNamespace))
```
p0lyn0mial: is this really required ? (it wasn't before)

PR #1999, opened 2026-01-12
OCPBUGS-73561: Update openshift/library-go to get new default TLS configuration

sanchezl:
Manual cherry-pick of https://github.com/openshift/cluster-kube-apiserver-operator/pull/1986.


p0lyn0mial:
> I can, but it only results in a single line diff where we lose https://github.com/openshift/library-go/pull/2061. Thoughts?

then it is not worth it. thanks for checking. 

i assume we didn't find any errors while testing the changes, xref: https://github.com/openshift/library-go/pull/2051#discussion_r2527945157

/lgtm
/label backport-risk-assessed

PR #2002, opened 2026-01-16
NO-JIRA:Refactor serviceaccountissuer ginkgo test

gangwgr:
Summary of Changes to serviceaccountissuer.go

 New Pattern

  g.Describe("serviceaccountissuer", g.Serial, g.Ordered, func() {
      // 1. Setup once in BeforeAll
      var kubeClient, authConfigClient ...
      g.BeforeAll(func() {
          // Initialize clients once
      })

      // 2. Define test cases as data
      testCases := []struct{
          name string
          issuer string
          expectedIssuers []string
      }{
          {name: "first test", issuer: "...", expectedIssuers: ...},
          {name: "second test", issuer: "...", expectedIssuers: ...},
          {name: "third test", issuer: "...", expectedIssuers: ...},
      }

      // 3. Loop to dynamically create g.It() blocks
      for _, tc := range testCases {
          tc := tc // capture range variable
          g.It(tc.name, func() {
              // Test logic using tc.issuer and tc.expectedIssuers
          })
      }
  })

  Benefits

  -3 separate test cases in output (not just 1)
  - Shared setup via BeforeAll (runs once)
  - Sequential execution via g.Serial + g.Ordered
  - 
  What Changed

  1. Removed the 3 separate helper functions (testServiceAccountIssuerFirstIssuer, etc.)
  2. Added BeforeAll for shared client setup
  3. Added test cases array with data-driven approach
  4. Loop dynamically generates g.It() blocks for each test case
  5. Inline test logic directly in the g.It() using test case data

### DIFF
--- a/test/e2e/serviceaccountissuer.go
+++ b/test/e2e/serviceaccountissuer.go
@@ -18,7 +18,6 @@ import (
 
 	g "github.com/onsi/ginkgo/v2"
 	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/operatorclient"
-	testlibraryapi "github.com/openshift/library-go/test/library/apiserver"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


p0lyn0mial (commented):
I think that we should also revert:

https://github.com/gangwgr/cluster-kube-apiserver-operator/blob/4b617439b0600389f56ebe6053ba88e9c7c6ad86/test/e2e/serviceaccountissuer.go#L29

and 

https://github.com/openshift/cluster-kube-apiserver-operator/commit/aca76e9f3fa7f81194186a70b3d9d4d940d64b78#r174833973
(this could use g.Ordered)

PR #2017, opened 2026-01-27
OCPBUGS-68365: revert dev cert rotation

sanchezl:
revert dev cert rotation in release-4.22 branch


ricardomaraschini:
@sanchezl I think the time to unhold this has arrived.

/unhold

ricardomaraschini:
Here we seem to have a deadlock: to get this merged we need to make the jira ticket depending on a jira ticket targeting 4.23 but this fix is only needed on 4.22. I am following what has been done on [4.21](https://github.com/openshift/cluster-kube-apiserver-operator/pull/1984) and removing the `jira/invalid-bug` label:

/remove-label jira/invalid-bug

ricardomaraschini:
Alright, we are getting there. Now I need to verify this so I need a release.

/verified later @ricardomaraschini

PR #2097, opened 2026-04-09
API-1768:e2e network policy tests

kaleemsiddiqu:
tests for network policy

## Summary by CodeRabbit

* **Tests**
  * Added parallel and disruptive operator test suites
  * Added network policy conformance tests with reconciliation verification
  * Added network policy enforcement tests covering cross-namespace, metrics, and port scenarios
  * Added E2E helpers for creating/validating NetworkPolicies, pods, connectivity, and a cluster-operator availability polling helper

* **Chores**
  * Promoted Gomega to a direct module requirement


kaleemsiddiqu:
/retitle [API-1768](https://redhat.atlassian.net/browse/API-1768):e2e network policy tests

kaleemsiddiqu:
/verified by @kaleemsiddiqu 
executed these tests locally.

```
$ ./cluster-kube-apiserver-operator-tests-ext list|grep "NetworkPolicy"
    "name": "[sig-api-machinery] kube-apiserver operator [Operator][NetworkPolicy] should ensure kube-apiserver NetworkPolicies are defined [Suite:openshift/cluster-kube-apiserver-operator/conformance/parallel]",
    "name": "[sig-api-machinery] kube-apiserver operator [Serial][Operator][NetworkPolicy] should restore kube-apiserver NetworkPolicies after delete or mutation[Timeout:30m] [Suite:openshift/cluster-kube-apiserver-operator/conformance/serial]",
    "name": "[sig-api-machinery] kube-apiserver operator [Operator][NetworkPolicy] should enforce NetworkPolicy allow/deny basics in a test namespace",
    "name": "[sig-api-machinery] kube-apiserver operator [Operator][NetworkPolicy] should enforce kube-apiserver-operator NetworkPolicies",
    "name": "[sig-api-machinery] kube-apiserver operator [Operator][NetworkPolicy] should enforce cross-namespace ingress traffic",
    "name": "[sig-api-machinery] kube-apiserver operator [Operator][NetworkPolicy] should allow metrics but block other ports",
    "name": "[sig-api-machinery] kube-apiserver operator [Operator][NetworkPolicy] should allow metrics ingress from any namespace",
$ 
```

kaleemsiddiqu:
/verified by @kaleemsiddiqu 
Local run was successful for this.

cmd/cluster-kube-apiserver-operator-tests-ext/main.go
```diff
@@ -96,6 +96,27 @@ func prepareOperatorTestsRegistry() (*oteextension.Registry, error) {
...
+	extension.AddSuite(oteextension.Suite{
+		Name:        "openshift/cluster-kube-apiserver-operator/operator/parallel",
+		Parallelism: 1,
+		Qualifiers: []string{
+			`!name.contains("[Serial]") && name.contains("[Operator]")`,
+		},
+	})
+
+	// The following suite runs tests that are disruptive to the cluster.
+	extension.AddSuite(oteextension.Suite{
```
gangwgr, line 88: try with existing serial suite

kaleemsiddiqu: addressed

test/library/cluster_operator.go
```diff
@@ -14,6 +14,30 @@ import (
 	clusteroperatorhelpers "github.com/openshift/library-go/pkg/config/clusteroperator/v1helpers"
 )
 
+// WaitForClusterOperatorAvailableNotProgressingNotDegraded waits for a ClusterOperator to report
```
gangwgr, line 17: why we need this func?

cmd/cluster-kube-apiserver-operator-tests-ext/main.go
```diff
@@ -96,6 +96,17 @@ func prepareOperatorTestsRegistry() (*oteextension.Registry, error) {
 		},
 	})
 
+	// The following suite runs tests that verify the operator's behaviour.
+	// This suite is executed only on pull requests targeting this repository.
+	// Tests tagged with [Parallel] and [Operator] are included in this suite.
+	extension.AddSuite(oteextension.Suite{
+		Name:        "openshift/cluster-kube-apiserver-operator/operator/parallel",
```
gangwgr: we don't have parallel  job here you can remove this
add serial tag in case.

test/library/cluster_operator.go
kaleemsiddiqu: a function to check cluster operators status before start of tests and same check after reconciliation/deletion of network policies

cmd/cluster-kube-apiserver-operator-tests-ext/main.go
kaleemsiddiqu: All tests except following are running in parallel , so added it.

"[Operator][NetworkPolicy][Disruptive][Serial] should restore kube-apiserver NetworkPolicies after delete or mutation[Timeout:30m]

test/library/cluster_operator.go
gangwgr: Not sure policy trigger kas restart

cmd/cluster-kube-apiserver-operator-tests-ext/main.go
gangwgr: we don't  have job parallel job here, it will not run any job

test/e2e/network_policy.go
```diff
@@ -0,0 +1,114 @@
...
+	defaultDenyAllPolicyName       = "default-deny"
+	operatorAllowPolicyName        = "allow-all-egress-and-metrics-ingress"
+	operandAllowPolicyName         = "allow-all-egress"
+)
+
+var _ = g.Describe("[sig-api-machinery] kube-apiserver operator", func() {
+	g.It("[Operator][NetworkPolicy] should ensure kube-apiserver NetworkPolicies are defined", func() {
+		testKubeAPIServerNetworkPolicies()
+	})
+	g.It("[Operator][NetworkPolicy][Disruptive][Serial] should restore kube-apiserver NetworkPolicies after delete or mutation[Timeout:30m]", func() {
```
gangwgr: we don't need disruptive tag

kaleemsiddiqu: There already existed a tag in main.go 
`name.contains("[Operator]") && name.contains("[Serial]") || name.contains("[Disruptive]")`

and since nature of this test is disruptive(deletion,reconciliation) , so added that tag to match existing tag.

test/library/cluster_operator.go
kaleemsiddiqu: Its just to be on safer side that it policy trigger does not cause issue with cluster operators.

test/e2e/network_policy.go
gangwgr: we remove this tag [Disruptive] tag in latest change you need rebase  and try without it first

kaleemsiddiqu: done.

cmd/cluster-kube-apiserver-operator-tests-ext/main.go
kaleemsiddiqu: moved to serial run.

test/e2e/network_policy.go
```diff
@@ -0,0 +1,112 @@
...
+const (
+	kubeAPIServerNamespace         = "openshift-kube-apiserver"
+	kubeAPIServerOperatorNamespace = "openshift-kube-apiserver-operator"
+	defaultDenyAllPolicyName       = "default-deny"
+	operatorAllowPolicyName        = "allow-all-egress-and-metrics-ingress"
+	operandAllowPolicyName         = "allow-all-egress"
+)
+
+var _ = g.Describe("[sig-api-machinery] kube-apiserver operator", func() {
+	g.It("[Operator][NetworkPolicy] should ensure kube-apiserver NetworkPolicies are defined", func() {
```
xueqzhan: Can we use "[Feature:NetworkPolicy]" like in "[[sig-api-machinery][Feature:NetworkPolicy] Config Operator NetworkPolicy should verify config namespace NetworkPolicy enforcement [apigroup:config.openshift.io]"? [This way](https://github.com/openshift-eng/ci-test-mapping/blob/fe1ab8ef4f2963e533713239c691fd2317977398/README.md?plain=1#L103), your test is already mapped to NetworkPolicy feature (capability) in component readiness.

test/e2e/network_policy_enforcement.go
```diff
@@ -0,0 +1,204 @@
...
+	expectConnectivity(ctx, kubeClient, "default", map[string]string{"test": "client"}, kasOperatorIPs, 8443, true)
+
+	g.By("Testing metrics port 8443 from openshift-etcd with custom app label: should be denied")
+	expectConnectivity(ctx, kubeClient, "openshift-etcd", map[string]string{"test": "client"}, kasOperatorIPs, 8443, false)
+
+	g.By("Testing port-based blocking: unauthorized ports are still blocked")
+	expectConnectivity(ctx, kubeClient, "openshift-monitoring", map[string]string{"app.kubernetes.io/name": "prometheus"}, kasOperatorIPs, 9999, false)
+
+	g.By("Testing multiple unauthorized ports are still blocked by default-deny")
+	for _, port := range []int32{80, 443, 8080, 22, 3306, 9090} {
```
sanchezl: These port-deny checks cannot distinguish between "connection refused" (no process listening — nothing to do with NetworkPolicy) and "connection timed out" (packets dropped by NetworkPolicy). The operator pod only listens on 8443, so probing ports 80, 443, 8080, 22, 3306, 9090, 9999, and 12345 will always get "connection refused" — even with no NetworkPolicies applied.

To validate that NetworkPolicy is actually blocking traffic, `expectConnectivity` (or `readConnectivityResult`) would need to distinguish the two failure modes. For example, `agnhost connect --timeout=5s` returns immediately on "connection refused" but hangs until timeout when packets are dropped by policy. Checking the exit timing or error message would let the test assert that policy enforcement is the cause of the failure, not just "nothing is listening."

test/e2e/network_policy_enforcement.go
```diff
@@ -0,0 +1,204 @@
...
+	})
+	o.Expect(err).NotTo(o.HaveOccurred())
+	o.Expect(pods.Items).NotTo(o.BeEmpty(), "should have at least one operator pod")
+	kasOperatorIPs := podIPs(&pods.Items[0])
+
+	g.By("Testing allow-to-metrics policy: monitoring namespace can access metrics -> operator:8443")
+	expectConnectivity(ctx, kubeClient, "openshift-monitoring", map[string]string{"app.kubernetes.io/name": "prometheus"}, kasOperatorIPs, 8443, true)
+
+	g.By("Testing metrics policy: etcd namespace with custom app label should be denied")
+	expectConnectivity(ctx, kubeClient, "openshift-etcd", map[string]string{"test": "metrics-client"}, kasOperatorIPs, 8443, false)
```
sanchezl: Testing cross-namespace connectivity from the operator's perspective is reasonable — the operator cares that Prometheus can scrape its metrics, etc. However, the test name ("should allow metrics ingress from allowed namespaces only") implies the operator's own policies do namespace-based ingress filtering. They don't — `allow-all-egress-and-metrics-ingress` allows ingress from *all* sources on port 8443 (validated by `requireIngressAllowAll` in network_policy.go).

The etcd denial here is caused by `openshift-etcd`'s own policies in cluster-etcd-operator: a `default-deny` policy blocks all egress, and the `allow-all-egress` policy only permits egress from pods labeled `app: guard|installer|pruner|cluster-backup-cronjob|tnf-job`. The test pod (labeled `{"test": "metrics-client"}`) doesn't match that allowlist, so its egress is blocked before it ever reaches the operator.

Consider renaming to something like "should be blocked by source namespace egress policies when connecting to metrics port" to reflect what's actually being validated. This also applies to the similar assertion in `testMetricsOpenButOtherPortsBlocked` (line 160).

test/e2e/network_policy_enforcement.go
```diff
@@ -0,0 +1,204 @@
...
+	ctx := context.Background()
+	kubeConfig := getClientConfigForTest()
+	kubeClient, err := kubernetes.NewForConfig(kubeConfig)
+	o.Expect(err).NotTo(o.HaveOccurred())
+
+	namespace := "openshift-kube-apiserver-operator"
+
+	g.By("Creating test service account in monitoring namespace for connectivity tests")
+	ensureTestServiceAccount(ctx, kubeClient, "openshift-monitoring")
+
```
sanchezl: `netpolicy-test-sa` is created in system namespaces (`openshift-monitoring`, `openshift-etcd`, `openshift-console`, `default`) across the enforcement tests but is never cleaned up. The test-created namespace gets `DeferCleanup` (line 33) but these service accounts in system namespaces persist after the test run.

Consider adding `DeferCleanup` to delete the service account from each system namespace, or a shared cleanup that removes all `netpolicy-test-sa` service accounts at test completion.

test/e2e/helpers.go
```diff
@@ -0,0 +1,828 @@
...
+// requireIngressAllowAll asserts that a NetworkPolicy allows ingress from any source on the specified port.
+func requireIngressAllowAll(policy *networkingv1.NetworkPolicy, port int32) {
+	g.GinkgoHelper()
+	if !hasIngressAllowAll(policy.Spec.Ingress, port) {
+		g.Fail(fmt.Sprintf("%s/%s: expected ingress allow-all on port %d", policy.Namespace, policy.Name, port))
+	}
+}
+
+// requireEgressAllowAllTCP asserts that a NetworkPolicy has an egress allow-all TCP rule.
+// The rule should have no destination restrictions (empty To field) and allow TCP traffic.
```
sanchezl, line 141: The `if` guard makes the success path skip the assertion entirely — when `hasEgressAllowAllTCP` returns true, no assertion runs. On failure, it calls `hasEgressAllowAllTCP` a second time redundantly. This should just be:

```go
func requireEgressAllowAllTCP(policy *networkingv1.NetworkPolicy) {
	g.GinkgoHelper()
	o.Expect(hasEgressAllowAllTCP(policy.Spec.Egress)).To(o.BeTrue(),
		fmt.Sprintf("NetworkPolicy %s/%s should have egress allow-all TCP rule", policy.Namespace, policy.Name))
}
```

test/library/library.go
```diff
@@ -15,7 +15,7 @@ import (
 
 var (
 	WaitPollInterval = time.Second
-	WaitPollTimeout  = 10 * time.Minute
+	WaitPollTimeout  = 20 * time.Minute
```
sanchezl: This doubles `WaitPollTimeout` from 10m to 20m for *all* tests that use this constant, not just network policy tests. This could slow down failure detection for existing tests.

If network policy reconciliation needs a longer timeout, consider using a local constant in the test file instead:
```go
const networkPolicyReconcileTimeout = 20 * time.Minute
```

sanchezl (changes requested):

test/e2e/network_policy_enforcement.go
```diff
@@ -0,0 +1,204 @@
...
+	kasOperatorIPs := podIPs(&pods.Items[0])
+
+	g.By("Testing metrics port 8443 is accessible: default namespace -> kube-apiserver-operator:8443")
+	expectConnectivity(ctx, kubeClient, "default", map[string]string{"test": "client"}, kasOperatorIPs, 8443, true)
+
+	g.By("Testing port-based blocking: unauthorized ports are still blocked")
+	expectConnectivity(ctx, kubeClient, "openshift-monitoring", map[string]string{"app.kubernetes.io/name": "prometheus"}, kasOperatorIPs, 9999, false)
+
+	g.By("Testing multiple unauthorized ports are still blocked by default-deny")
+	for _, port := range []int32{80, 443, 8080, 22, 3306, 9090} {
```
sanchezl: The port-deny tests probe ports 80, 443, 8080, 22, 3306, 9090, 9999, and 12345 on the operator pod, but the operator only listens on port 8443 (the library-go default via `configdefaults.BindAddress`). Nothing is listening on any of these ports, so the connection attempts will get "connection refused", not "connection timed out" from NetworkPolicy blocking.

The new `ConnectivityResult` code (line 618 in helpers.go) now explicitly asserts that failed connections must be timeouts (`result.Timeout`), not refused. Since these ports will return `CONN_REFUSED`, these assertions will fail:

```go
o.Expect(result.Timeout || result.Success).To(o.BeTrue(), ...)
```

The port numbers (80, 443, 8080, 9090) look like they were chosen as common ports to block, but they're more typical of kube-apiserver ports, not kube-apiserver-*operator* ports. Since the operator only serves on 8443, port-based policy enforcement isn't meaningfully testable against the real operator pod — there's only one listening port.

The generic test in the temp namespace (`testGenericNetworkPolicyEnforcement`) is the right place for port-deny testing since it controls both the server and the policies. Consider removing the port-deny checks from the operator-targeting tests. Same issue applies to port 12345 (line 111) and port 9999 (line 163).

test/e2e/network_policy_enforcement.go
kaleemsiddiqu: added a new logic to test based on connectivity output. changes added in helpers.go

test/e2e/network_policy.go
kaleemsiddiqu: tests updated.

test/e2e/network_policy_enforcement.go
kaleemsiddiqu: Dropped etcd tests as these taken care of in etcd network policy tests.

test/e2e/network_policy_enforcement.go
kaleemsiddiqu: added the defer cleanup function for it.

test/e2e/helpers.go
kaleemsiddiqu: Updated.

test/library/library.go
kaleemsiddiqu: This updated too.

test/e2e/network_policy_enforcement.go
kaleemsiddiqu: Dropped these other ports checks and just kept 8080 in generic test

kaleemsiddiqu:
/verified by @kaleemsiddiqu 
local run successful

kaleemsiddiqu:
/verified by @kaleemsiddiqu 
Local execution of tests successful.

test/e2e/network_policy_enforcement.go
```diff
@@ -0,0 +1,195 @@
...
+		testGenericNetworkPolicyEnforcement()
+	})
+	g.It("[Operator][Feature:NetworkPolicy] should enforce kube-apiserver-operator NetworkPolicies", func() {
+		testKubeAPIServerOperatorNetworkPolicyEnforcement()
+	})
+	g.It("[Operator][Feature:NetworkPolicy] should enforce cross-namespace ingress traffic", func() {
+		testCrossNamespaceIngressEnforcement()
+	})
+	g.It("[Operator][Feature:NetworkPolicy] should allow metrics but block other ports", func() {
+		testMetricsOpenButOtherPortsBlocked()
```
sanchezl: This test now provides no value over `testMetricsIngressOpenAccess()` (it ends up testing a subset of the same thing). The following overlap (same namespace/pod/port):
```
    // testMetricsOpenButOtherPortsBlocked
	g.By("Testing metrics port 8443 is accessible: default namespace -> kube-apiserver-operator:8443")

    // testMetricsIngressOpenAccess
	g.By("Testing allow-to-metrics policy: default namespace can access metrics -> operator:8443")

```

As per your comment, I understand why only the `testGenericNetworkPolicyEnforcement()` could really test the deny path. 

Delete `testMetricsOpenButOtherPortsBlocked()`. If you want to do it as a follow-up, create the followup PR and remove the hold on this PR.

sanchezl (changes requested):

test/e2e/network_policy_enforcement.go
kaleemsiddiqu: Removed testMetricsOpenButOtherPortsBlocked()
But now all labels needed back. After CI run is successful, please add those.

kaleemsiddiqu:
/verified by @kaleemsiddiqu 
local run successful with new changes.

kaleemsiddiqu:
/verified by @kaleemsiddiqu
local run successful with new changes.

sanchezl:
/verified by "pull-ci-openshift-cluster-kube-apiserver-operator-main-e2e-gcp-operator-serial-ote"

PR #2098, opened 2026-04-09
OCPBUGS-62792: Fix self-signed serving cert using 30ns lifetime instead of 30 days

jubittajohn:
Bump library-go to consume the changes in https://github.com/openshift/library-go/pull/2155.

Before https://github.com/openshift/cluster-kube-apiserver-operator/pull/2058, the check-endpoints container lacked the `--config` flag, causing it to fall back to library-go's self-signed certificate generated with `MakeServerCert(..., 30)` where the function signature changed from int days to time.Duration in 4.19, resulting in 30 nanoseconds (~1 second validity) instead of 30 days. This causes the certificate  generated by the `kube-apiserver-check-endpoints` container to be valid for only 1 second.                                        

This bug is suppressed in main as [PR #2058](https://github.com/openshift/cluster-kube-apiserver-operator/pull/2058) (PQC feature work) added the --config flag which loads namedCertificates including `localhost-serving-cert-certkey`, so SNI routing now serves the properly managed 12-hour/30-day certificate, completely bypassing the buggy self-signed cert code—but since #2058 won't be backported as PQC feature work, we need to backport just the check-endpoints fix to 4.19 to resolve the customer issue.   (cc: @ricardomaraschini )

## Summary by CodeRabbit

* **Chores**
  * Bumped an internal library dependency to a newer release. No user-visible API or behavior changes; maintenance-only update.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/openshift/api v0.0.0-20260317165824-54a3998d81eb
 	github.com/openshift/build-machinery-go v0.0.0-20250530140348-dc5b2804eeee
 	github.com/openshift/client-go v0.0.0-20260317180604-743f664b82d1
-	github.com/openshift/library-go v0.0.0-20260409165127-c57da2bf5720
+	github.com/openshift/library-go v0.0.0-20260413093329-d2db42c961e1
 	github.com/pkg/profile v1.7.0 // indirect
 	github.com/prometheus/client_golang v1.23.2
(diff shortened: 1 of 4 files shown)


sanchezl (approved):
/lgtm

jubittajohn:
/close

superseded by this https://github.com/openshift/cluster-kube-apiserver-operator/pull/2107 PR that bumps `library-go` to a later commit which includes the desired changes in this PR.

PR #2101, opened 2026-04-13
CNTRLPLANE-3228: Implement 5.0 EUS schedule in KAS-O

tjungblu:
Implements the new EUS schedule (every third release instead of every second) in the kubelet skew guard controller.

---

> [!NOTE]
> **Medium Risk**
> Changes upgrade-gating logic that determines when clusters are marked non-upgradeable based on kubelet/API server minor skew, especially for upcoming 5.x releases. Incorrect skew math could incorrectly block upgrades or allow unsupported node versions.
> 
> **Overview**
> Updates the kubelet skew guard to compute minimum supported kubelet lag based on OpenShift major/minor: keeps 4.y’s odd/even (-1/-2) rules but switches 5.y+ to a *mod-3* policy (-1/-2/-3) to match the new EUS every-third-release cadence.
> 
> Reworks and expands unit tests to cover 5.0–5.2 scenarios, explicit API server version inputs, additional unsupported/edge cases (multiple nodes, ahead-of-API), and adds a dedicated test validating `minSupportedKubeletSkewForOpenShiftVersion()` across 4.22–5.3.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dc8e55a8569460ef18565af21db86b64517af1e0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated kubelet compatibility thresholds per OpenShift minor: OpenShift 4.x uses an alternating -2/-1 policy, while OpenShift 5.x+ uses a -1/-2/-3 pattern; tightened next-upgrade boundary so equality is treated as unsupported.

* **Tests**
  * Expanded and reworked tests to cover varied kubelet/API server version scenarios, semver-based kubelet derivation, and to validate the new skew rules across multiple minors.

### DIFF
--- a/pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
+++ b/pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
@@ -3,7 +3,6 @@ package kubeletversionskewcontroller
 import (
 	"context"
 	"fmt"
-	"regexp"
 	"sort"
 	"strings"
 
@@ -35,11 +34,27 @@ const (
 // KubeletVersionSkewController sets Upgradeable=False if the kubelet
 // version on a node prevents upgrading to a supported OpenShift version.
 //
-// For odd OpenShift versions, kubelet versions 0 or 1 minor
-// versions behind the API server version are supported.
+// For OpenShift 4.y, odd minors allow kubelet skew of one minor behind the
+// API server; even minors allow two minors behind (see minSupportedKubeletSkewForOpenShiftVersion).
 //
-// For even OpenShift versions, kubelet versions 0, 1, or 2
-// minor versions behind the API server version are supported.
+// The supported kubelet version skew is determined by how many
+// releases have elapsed since the last EUS cycle boundary. It
+// increases by one per release, reaching its maximum at the next
+// EUS boundary.
+//
+// OpenShift 4.x has a 2-release EUS cycle (4.20, 4.22, ...):
+//
+//	4.21: up to 1 version behind
+//	4.22: up to 2 versions behind
+//
+// OpenShift 5.x has a 3-release EUS cycle (5.2, 5.5, 5.8, ...):
+//
+//	5.0, 5.3: up to 1 version behind
+//	5.1, 5.4: up to 2 versions behind
+//	5.2, 5.5: up to 3 versions behind
+//
+// Note: 4.23 is the last 4.x release and is designated EUS, but
+// its supported skew is still -1 (one release from 4.22).
 type KubeletVersionSkewController interface {
 	factory.Controller
 }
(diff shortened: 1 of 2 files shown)


pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
sanchezl:
```suggestion
// The supported kubelet version skew is determined by how many                                                                                                             
// releases have elapsed since the last EUS cycle boundary. It
// increases by one per release, reaching its maximum at the next                                                                                                           
// EUS boundary.
//                                                                                                                                                                          
// OpenShift 4.x has a 2-release EUS cycle (4.20, 4.22, ...):
//   4.21: up to 1 version behind                                                                                                                                           
//   4.22: up to 2 versions behind
//                                                                                                                                                                          
// OpenShift 5.x has a 3-release EUS cycle (5.2, 5.5, 5.8, ...):
//   5.0, 5.3: up to 1 version behind                                                                                                                                       
//   5.1, 5.4: up to 2 versions behind
//   5.2, 5.5: up to 3 versions behind                                                                                                                                      
//                                                                                                                                                                          
// Note: 4.23 is the last 4.x release and is designated EUS, but
// its supported skew is still -1 (one release from 4.22).                                                                                                                  
```

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
```diff
@@ -68,11 +67,14 @@ func NewKubeletVersionSkewController(
 }
 
 func minSupportedKubeletSkewForOpenShiftVersion(v semver.Version) int {
```
sanchezl, line 85: In my [review of #2010](https://github.com/openshift/cluster-kube-apiserver-operator/pull/2010#discussion_r2723312851), I recommended a phased approach:
1. **4.22:** Wording fix (merged)
2. **4.23:** Hardcode `nextOpenShiftVersion := semver.Version{Major: 5, Minor: 0}`
3. **5.0:** Revert to `Minor + 1`, switch to `v.Minor % 3`

I had some incorrect assumptions then:
- I assumed 4.23 would upgrade to 5.0, but it actually upgrades to **5.1**, so the hardcoded `{5, 0}` would have been wrong.
- I didn't account for 4.23 being a **fast-forward of 5.0/main**, which means they must share the same code.

However, the underlying principle still holds: don't ship code that's wrong for the version it runs on. The current PR replaces mod-2 with mod-3 entirely, which would produce incorrect results if fast-forwarded to 4.23 (e.g. `23 % 3 = 2 → -3` instead of the correct `23 % 2 = 1 → -1`).

The fix: add a `switch v.Major` guard in `minSupportedKubeletSkewForOpenShiftVersion`, mod-2 for Major 4, mod-3 for Major 5. This works because 4.23 and 5.0 produce **identical outputs**:

| | 4.23 (mod-2) | 5.0 (mod-3) |
|---|---|---|
| `minSupportedSkew` | `23%2=1` → **-1** | `0%3=0` → **-1** |
| `minSupportedSkewNextVersion` | `24%2=0` → **-2** | `1%3=1` → **-2** |

Same k8s base (1.36), same outputs, same behavior. No feature gate, no revert needed.

This change also can't be deferred to 5.1. Without the mod-3 guard, the current mod-2 code on 5.0 computes `minSupportedSkewNextVersion = 1%2=1 → -1`, which would **incorrectly block 5.0→5.1 upgrades** for clusters with k8s 1.35 kubelets (the normal state after upgrading through the 4.22 EUS path).

Separately, the `nextOpenShiftVersion = {Major, Minor+1}` calculation computes 4.24 instead of 5.1 for 4.23, but since the outputs match, that cleanup can be deferred to 5.1.

sanchezl (changes requested):

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
tjungblu: wonderful clarification, thanks a ton! I've adjusted the PR accordingly. 

The test diff became a bit larger now as I've tried to keep the existing tests around to exercise the paths for the 4.x branches.

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
```diff
@@ -35,10 +34,27 @@ const (
...
+//
+// OpenShift 5.x has a 3-release EUS cycle (5.2, 5.5, 5.8, ...):
+//
+//	5.0, 5.3: up to 1 version behind
+//	5.1, 5.4: up to 2 versions behind
+//	5.2, 5.5: up to 3 versions behind
+//
+// Note: 4.23 is the last 4.x release and is designated EUS, but
+// its supported skew is still -1 (one release from 4.22).
 // minor versions behind the API server version are supported.
```
sanchezl: Delete this last line (orphaned from previous text).

sanchezl (changes requested):

pkg/operator/kubeletversionskewcontroller/kubelet_version_skew_controller.go
tjungblu: fixed!

sanchezl (approved):
/lgtm

PR #2102, opened 2026-04-13
[WIP] kms: block revisions until encryption config and kms sidecar are in sync

bertinatto:
Add a revision controller precondition that prevents rolling out a new revision when there is a mismatch between the encryption configuration and the pod spec.

CC @ardaguclu @p0lyn0mial 

## Summary by CodeRabbit

* **New Features**
  * Added a KMS revision consistency check that validates KMS encryption configuration and the presence of the KMS plugin in API server pods when KMS encryption is enabled.

* **Chores**
  * Updated dependency resolution mapping for an internal library to a specific module version.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -136,3 +136,5 @@ require (
 )
 
 replace github.com/onsi/ginkgo/v2 => github.com/openshift/onsi-ginkgo/v2 v2.6.1-0.20251001123353-fd5b1fb35db1
+
+replace github.com/openshift/library-go => github.com/bertinatto/library-go v0.0.0-20260423111950-e491ed3b41e5
--- a/go.sum
+++ b/go.sum
@@ -8,6 +8,8 @@ github.com/apparentlymart/go-cidr v1.0.1 h1:NmIwLZ/KdsjIUlhf+/Np40atNXm/+lZ5txfT
 github.com/apparentlymart/go-cidr v1.0.1/go.mod h1:EBcsNrHc3zQeuaeCeCtQruQm+n9/YjEn/vI25Lg7Gwc=
 github.com/beorn7/perks v1.0.1 h1:VlbKKnNfV8bJzeqoa4cOKqO6bYr3WgKZxO8Z16+hsOM=
 github.com/beorn7/perks v1.0.1/go.mod h1:G2ZrVWU2WbWT9wwq4/hrbKbnv/1ERSJQ0ibhJ6rlkpw=
+github.com/bertinatto/library-go v0.0.0-20260423111950-e491ed3b41e5 h1:gChhqn7yIjwKYycGEs5Uo+SgAd5EmARQFyoZsW52Ruc=
+github.com/bertinatto/library-go v0.0.0-20260423111950-e491ed3b41e5/go.mod h1:pQx73OLgJJtHR2WJVdkH0Zng/yywdL5hmPGw1QbGV/w=
 github.com/blang/semver/v4 v4.0.0 h1:1PFHFE6yCCTv8C1TeyNNarDzntLi7wMI5i/pzqYIsAM=
 github.com/blang/semver/v4 v4.0.0/go.mod h1:IbckMUScFkM3pff0VJDNKRiT6TG/YpiHIM2yvyW5YoQ=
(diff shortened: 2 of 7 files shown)


pkg/operator/kms/precondition.go
```diff
@@ -0,0 +1,110 @@
...
+			// Pod configmap not created yet: let the revision controller proceed
+			return true, nil
+		}
+		if err != nil {
+			return false, fmt.Errorf("kms precondition: failed to get kube-apiserver-pod configmap: %w", err)
+		}
+
+		kmsConfigured := hasKMSEncryptionProvider(encryptionConfig)
+		// FIXME(bertinatto): plugin name is hardcoded here
+		kmsInPod := podHasKMSSidecar(podCM, "kms-plugin")
```
bertinatto, line 67: The validation that these 2 functions do is still premature. More work needs to be done here.

pkg/operator/kms/precondition.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+		featureGates, err := featureGateAccessor.CurrentFeatureGates()
+		if err != nil {
+			return true, nil
+		}
+		if !featureGates.Enabled(features.FeatureGateKMSEncryption) {
+			return true, nil
+		}
+
+		encryptionConfig, err := kubeClient.CoreV1().Secrets("openshift-config-managed").Get(ctx, "encryption-config-openshift-kube-apiserver", metav1.GetOptions{})
```
ardaguclu, line 48: (My questions may not be accurate)

shouldn't we query from revisioned Secret in openshift-kube-apiserver-operator ns rather than openshift-config-managed (which is unrevisioned)

pkg/operator/kms/precondition.go
```diff
@@ -0,0 +1,110 @@
...
+	return func(ctx context.Context) (bool, error) {
+		// Skip the check entirely when the KMS feature gate is disabled
+		if !featureGateAccessor.AreInitialFeatureGatesObserved() {
+			return true, nil
+		}
+		featureGates, err := featureGateAccessor.CurrentFeatureGates()
+		if err != nil {
+			return true, nil
+		}
+		if !featureGates.Enabled(features.FeatureGateKMSEncryption) {
```
p0lyn0mial, line 44: Is the FG check required if the presence of a KMS configuration is also behind the same FG?

pkg/operator/kms/precondition.go
p0lyn0mial, line 48: I don’t think this is what we want. We need to compare revisioned resources. This will likely require changes in library-go. We need to validate revisioned resources to ensure they won’t change and we need to do that before the system switches to that revision.

ardaguclu: I think, there might be an issue. This function is called from targetconfigcontroller which does not know revision number. So that how can targetconfigcontroller query the revisioned encryption config Secret under openshift-kube-apiserver. Is there a missing part in our design.

p0lyn0mial: > So that how can targetconfigcontroller query the revisioned encryption config Secret under openshift-kube-apiserver

it cannot, that's why we decided to add an additional function to the revision controller (library-go).

p0lyn0mial: https://docs.google.com/document/d/1htjUQmCk60AVkcOe1QzLSYNrBhO7T8hVsFDiTbR4AJM/edit?tab=t.0#heading=h.q9rx93jt2t5r

p0lyn0mial: from that doc

>> The third solution is to add a validation function to the RevisionController. This function validated the revisioned resources. In our case the kas manifest and the encryption-configuration. If the configuration is invalid we return an error. As a result a new revision is not generated until the system converges to a good state.


> We choose option 3. This solution seems the simplest. It doesn't change the existing flow and doesn't introduce new states.

ardaguclu: Thanks for explanation, it makes sense. It seems everything is under control and there is no missing pieces for this part :)

p0lyn0mial: yw.

bertinatto:
/close 
superseded by https://github.com/openshift/cluster-kube-apiserver-operator/pull/2121

PR #2121, opened 2026-04-23
[WIP] KMS: add a post check function to revision controller

bertinatto:
The idea is to validate whether the revision is valid or not.

## Summary by CodeRabbit

* **New Features**
  * Added KMS encryption validation to the cluster-kube-apiserver-operator
  * Verifies KMS plugin sidecar containers are correctly configured and deployed when KMS encryption is enabled
  * Validates consistency between encryption configuration and actual pod specifications
  * Ensures proper initialization of KMS plugin sidecar containers during cluster API server updates

### DIFF
--- a/pkg/operator/kms/postcondition.go
+++ b/pkg/operator/kms/postcondition.go
@@ -59,7 +59,7 @@ func kmsRevisionPostcondition(ctx context.Context, kubeClient kubernetes.Interfa
 		return fmt.Errorf("kms revision post-check: failed to get kube-apiserver-pod for revision %d: %w", revision, err)
 	}
 
-	expectedSidecars, err := buildExpectedKMSSidecars(encryptionSecret)
+	expectedSidecars, err := buildExpectedKMSSidecars("encryption-config", encryptionSecret)
 	if err != nil {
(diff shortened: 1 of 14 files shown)


pkg/operator/kms/postcondition.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	if !featureGates.Enabled(features.FeatureGateKMSEncryption) {
+		return nil
+	}
+
+	revSuffix := fmt.Sprintf("-%d", revision)
+
+	encryptionSecret, err := kubeClient.CoreV1().Secrets(operatorclient.TargetNamespace).Get(ctx, "encryption-config"+revSuffix, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		return nil
```
ardaguclu, line 48: I assume that notfound revisioned Secret is totally benign?

bertinatto: I assume so, that's why I return nil here. Should I handle it differently?

ardaguclu: Returning nil should be fine here.

PR #2166, opened 2026-05-27
OCPBUGS-86057: render: fall back to Authentication CR

tchap:
When deploying STS clusters via Hive/ACM with credentialsMode: Manual, users supply the Authentication CR through manifestsConfigMapRef/ manifestsSecretRef but no cluster-authentication-02-config.yaml is generated by ccoctl or the installer. As a result the render command silently boots with the default service-account issuer, triggering a kube-apiserver rollout after bootstrap that races against token expiry and causes install timeouts.

After the existing --cluster-auth-file lookup, if ServiceAccountIssuer is still empty, scan the already-loaded --rendered-manifest-files for any config.openshift.io/v1 Authentication CR using the library-go ListManifestOfType helper (the same mechanism used for FeatureGate discovery) and extract the issuer from the first match.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a fallback to discover the service account issuer from rendered input manifests when the cluster authentication file is missing or empty, ensuring the issuer is populated reliably.

* **Tests**
  * Added tests to validate service account issuer discovery from authentication manifests included in rendered inputs.

### DIFF
--- a/pkg/cmd/render/render.go
+++ b/pkg/cmd/render/render.go
@@ -29,6 +29,7 @@ import (
 	genericrenderoptions "github.com/openshift/library-go/pkg/operator/render/options"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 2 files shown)


pkg/cmd/render/render.go
```diff
@@ -248,6 +248,11 @@ func (r *renderOpts) Run() error {
 			}
 		}
 	}
+	if len(renderConfig.ServiceAccountIssuer) == 0 {
```
everettraven: Is a length check the appropriate gate here? 

Would we want to support a case where the cluster auth file is specified and exists, it sets the service account issuer to `""`, and another manifest with the same type exists and sets the service account issuer to a different value?

pkg/cmd/render/render.go
```diff
@@ -424,6 +429,19 @@ func mustReadTemplateFile(fname string) genericrenderoptions.Template {
...
+	manifests, err := r.generic.ReadInputManifests()
+	if err != nil {
+		return err
+	}
+	authManifests := manifests.ListManifestOfType(configv1.GroupVersion.WithKind("Authentication"))
+	if len(authManifests) == 0 {
+		return nil
+	}
+	klog.V(4).Infof("--cluster-auth-file not present; using Authentication CR from rendered manifests: %q", authManifests[0].OriginalFilename)
+	return discoverServiceAccountIssuer(authManifests[0].Content, renderConfig)
```
everettraven: Is there precedence for using the first one that is listed? Is the ordering of the list deterministic?

pkg/cmd/render/render.go
tchap: Yeah, I will change this to check manifests when the original file is not available.

pkg/cmd/render/render.go
tchap: Changed this to simply get `Authentication` named `cluster`.

everettraven (approved):
/lgtm

PR #2179, opened 2026-06-04
OCPBUGS-85183: Add kind-aware staleness detection for runtime-config entries

tchap:
Add a Kinds field to groupVersionKindsByOpenshiftVersion and a
findStaleGroupVersionEntries helper that checks whether listed
API versions are stale. It detects:
- unregistered GV or kind (removed during a rebase)
- kind graduated to v1 (served by default, override is stale)
- highest pre-release version not listed (version bump missed)

When Kinds is unset, a simpler check flags any entry whose group
has a higher-priority version registered.

## Summary by CodeRabbit

* **Tests**
  * Added coverage to detect stale feature-gated API entries, including missing registrations, API graduation, and version-selection behavior.
  * Added validation to ensure default API mappings remain compatible with the Kubernetes version and registered resource types.
* **Refactor**
  * Improved API mapping support to account for resource kinds alongside group/version and version-range information.
  * Preserved existing group/version selection behavior while strengthening compatibility checks.

### DIFF
--- a/pkg/operator/configobservation/apienablement/observe_runtime_config.go
+++ b/pkg/operator/configobservation/apienablement/observe_runtime_config.go
@@ -29,7 +29,7 @@ var defaultGroupVersionsByFeatureGate = map[configv1.FeatureGateName][]groupVers
 		// admission plugin uses v1beta1 informers. The openshift-apiserver can't be rebased until the
 		// o/k 1.36 rebase lands and a 1.36 branch is created on openshift/kubernetes-apiserver.
 		{
-			KubeVersionRange: semver.MustParseRange(">=1.33.0 <1.37.0"),
+			KubeVersionRange: semver.MustParseRange(">=1.33.0 <1.34.0"),
 			GroupVersion:     schema.GroupVersion{Group: "admissionregistration.k8s.io", Version: "v1alpha1"},
 			Kinds:            []string{"MutatingAdmissionPolicy", "MutatingAdmissionPolicyBinding"},
 		},
--- a/pkg/operator/configobservation/apienablement/observe_runtime_config.go
+++ b/pkg/operator/configobservation/apienablement/observe_runtime_config.go
@@ -15,7 +15,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/events"
 )
 
-var defaultGroupVersionsByFeatureGate = map[configv1.FeatureGateName][]groupVersionByOpenshiftVersion{
+var defaultGroupVersionsByFeatureGate = map[configv1.FeatureGateName][]groupVersionKindsByOpenshiftVersion{
 	"MutatingAdmissionPolicy": {
 		// Both v1alpha1 and v1beta1 versions must be served pre-GA because e2e tests exercise both APIs.
 		// A GA OpenShift release could inadvertently serve these versions if MutatingAdmissionPolicy
@@ -28,17 +28,30 @@ var defaultGroupVersionsByFeatureGate = map[configv1.FeatureGateName][]groupVers
 		// k8s.io/apiserver at 1.34 (via openshift/kubernetes-apiserver) and its MutatingAdmissionPolicy
 		// admission plugin uses v1beta1 informers. The openshift-apiserver can't be rebased until the
 		// o/k 1.36 rebase lands and a 1.36 branch is created on openshift/kubernetes-apiserver.
-		{KubeVersionRange: semver.MustParseRange(">=1.33.0 <1.37.0"), GroupVersion: schema.GroupVersion{Group: "admissionregistration.k8s.io", Version: "v1alpha1"}},
-		{KubeVersionRange: semver.MustParseRange(">=1.34.0 <1.37.0"), GroupVersion: schema.GroupVersion{Group: "admissionregistration.k8s.io", Version: "v1beta1"}},
+		{
+			KubeVersionRange: semver.MustParseRange(">=1.33.0 <1.37.0"),
+			GroupVersion:     schema.GroupVersion{Group: "admissionregistration.k8s.io", Version: "v1alpha1"},
+			Kinds:            []string{"MutatingAdmissionPolicy", "MutatingAdmissionPolicyBinding"},
+		},
+		{
+			KubeVersionRange: semver.MustParseRange(">=1.34.0 <1.37.0"),
+			GroupVersion:     schema.GroupVersion{Group: "admissionregistration.k8s.io", Version: "v1beta1"},
+			Kinds:            []string{"MutatingAdmissionPolicy", "MutatingAdmissionPolicyBinding"},
+		},
 	},
 }
 
-type groupVersionByOpenshiftVersion struct {
+type groupVersionKindsByOpenshiftVersion struct {
 	schema.GroupVersion
 	KubeVersionRange semver.Range
+	// Kinds lists the specific resource kinds this entry is about. When set,
+	// staleness checks can verify per-kind whether the API has graduated to v1
+	// or a higher pre-release version exists. When unset, only a coarse
+	// group-level version priority check is performed.
+	Kinds []string
 }
 
-func getGroupVersionByFeatureGate(groupVersionsByFeatureGate map[configv1.FeatureGateName][]groupVersionByOpenshiftVersion, kubeVersion semver.Version) (map[configv1.FeatureGateName][]schema.GroupVersion, error) {
+func getGroupVersionByFeatureGate(groupVersionsByFeatureGate map[configv1.FeatureGateName][]groupVersionKindsByOpenshiftVersion, kubeVersion semver.Version) (map[configv1.FeatureGateName][]schema.GroupVersion, error) {
 	result := make(map[configv1.FeatureGateName][]schema.GroupVersion, len(groupVersionsByFeatureGate))
 	groupByVersions := map[string][]string{}
 	for featureGate, APIGroups := range groupVersionsByFeatureGate {
--- a/pkg/operator/configobservation/apienablement/observe_runtime_config.go
+++ b/pkg/operator/configobservation/apienablement/observe_runtime_config.go
@@ -15,14 +15,19 @@ import (
 	"github.com/openshift/library-go/pkg/operator/events"
 )
 
-var defaultGroupVersionsByFeatureGate = map[configv1.FeatureGateName][]groupVersionByOpenshiftVersion{}
+var defaultGroupVersionsByFeatureGate = map[configv1.FeatureGateName][]groupVersionKindsByOpenshiftVersion{}
 
-type groupVersionByOpenshiftVersion struct {
+type groupVersionKindsByOpenshiftVersion struct {
 	schema.GroupVersion
 	KubeVersionRange semver.Range
+	// Kinds lists the specific resource kinds this entry is about. When set,
+	// staleness checks can verify per-kind whether the API has graduated to v1
+	// or a higher pre-release version exists. When unset, only a coarse
+	// group-level version priority check is performed.
+	Kinds []string
 }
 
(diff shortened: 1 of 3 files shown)


pkg/operator/configobservation/apienablement/observe_runtime_config_test.go
```diff
@@ -246,3 +248,26 @@ func TestGroupVersionsByFeatureGate(t *testing.T) {
 		})
 	}
 }
+
+func TestDefaultGroupVersionsByFeatureGateAreRegistered(t *testing.T) {
```
everettraven: I don't think that this is actually doing what is intended by the ticket linked.

From what I recall, we've run into situations in the past where we have intentionally enabled features that are disabled-by-default in Kubernetes early and placed them as only enabled in the TechPreviewNoUpgrade feature-set via creating an equivalently named feature gate in openshift/api.

When the feature goes GA in Kubernetes and we bump OpenShift to that Kubernetes version, and promote/remove the OpenShift-specific gate override, we want to make sure that we are serving the correct versions of the APIs by default. The existing override behavior here makes it so that the incorrect versions are served (v1alpha1 and v1beta1) and never enables the correct version (v1).

To prevent a situation where we serve only the alpha/beta API versions in a GA OpenShift cluster we should make sure we aren't explicitly enabling them when the feature is GA and serving a stable (v1) API.

I think this, as-is, is doing the inverse and checks whether or not the currently enabled group-versions are registered - which I believe that even today they are still technically "registered".

I think what we actually want to check is whether or not the version we are deciding to serve for the desired kubernetes version is the highest priority version.

everettraven: Also a note, I'm not sure that we want this to just be a unit test.

An e2e test is probably more robust to catch issues with both standalone OpenShift and HyperShift stamped out kube-apiserver configurations.

tchap: Makes sense, I will rework this next week.

tchap: I've been looking into reworking the test per the review feedback. The core tension is that defaultGroupVersionsByFeatureGate maps feature gates to group versions, not individual resources — but graduation happens at the resource level. For example, admissionregistration.k8s.io/v1 already exists (for ValidatingAdmissionPolicy, webhooks), yet MutatingAdmissionPolicy is still only in v1alpha1/v1beta1 (based on the vendored code).

Two realistic options:

1. "All types" heuristic: Use Scheme.KnownTypes() to check whether all types in our pre-release GV already exist in a higher-priority version. No data model changes needed. This correctly handles the MutatingAdmissionPolicy case — today it passes (MutatingAdmissionPolicy types aren't in v1) and would fail once they graduate. The limitation is that it only fires when the entire pre-release GV has been subsumed, so it could miss partial graduations.
2. Add resource kinds to the map: Extend groupVersionByOpenshiftVersion to include the specific kinds the entry is about, then check those directly. Most precise, but requires maintainers to specify the kinds when adding entries.

Just checking a higher API version exists is obviously not enough as we can end up dropping resources.

Thoughts on which direction to go? Or any other direction entirely?

everettraven: >The limitation is that it only fires when the entire pre-release GV has been subsumed, so it could miss partial graduations.

What do you mean by "partial graduations" here? Would you mind elaborating a bit more on what that case looks like?

>Add resource kinds to the map: Extend groupVersionByOpenshiftVersion to include the specific kinds the entry is about, then check those directly. Most precise, but requires maintainers to specify the kinds when adding entries.

This doesn't seem unreasonable to me.

>Or any other direction entirely?

Thinking through this a bit, the only time we actually exercise the functionality we are trying to test here is when we have a discrepancy between gate enablement, and thus "feature maturity" in Kubernetes and OpenShift.

So, when we want to enable it in TPNU earlier than when it is enabled by default we add an entry for the gate to this list.

What if we used the OpenShift gate maturity to dictate the API group + versions we expect to be served? For example, generating this with the Default feature set, only the v1 and the latest v1betaN version (I think) should be served for the type behind that gate (i.e we never serve a v1alpha1 API version by default).

everettraven: Also going to note that thinking through this more, I'm actually OK with this being a unit test.

We are going to have to come up with a slightly different solution for HyperShift though because their feature gating is different than on standalone.

tchap: > What do you mean by "partial graduations" here? Would you mind elaborating a bit more on what that case looks like?

It was perhaps a wrong term, in that context I only meant that we can't check only for API version existing, but also that the relevant resources are available in that version, so we shouldn't switch to `v1` when `MutatingAdmissionPolicy` is not available there yet, even though that API version exists for other resources. The idea was already worded properly, so nothing new here really 🙂 

> What if we used the OpenShift gate maturity to dictate the API group + versions we expect to be served? For example, generating this with the Default feature set, only the v1 and the latest v1betaN version (I think) should be served for the type behind that gate (i.e we never serve a v1alpha1 API version by default).

Sounds pretty clear. But I am not sure it really gives us any advantage. Because even when we check that the feature gate is in default on OpenShift, we should still check that it's serving `v1` when available and not `v1betaN`, I guess.

So right now it seems to me the best way is to add the kinds we are interested in to the list, then simply check that we are enabling the highest API possible? 

Having said that, it occurs to me right now that if we have all the kinds we are interested in listed, we can simply check what highest API version contains them all and simply choose the highest possible API at runtime. Which is changing the logic, not just adding a test, but...

Thoughts?

tchap: To clarify the checks I have in mind — given the kinds in the map, two things to flag:

1. **Graduated to GA**: If a kind exists in v1 for the group, the entry is stale and should be removed — v1 is served by default.
2. **Stale pre-release version**: If a kind exists in a higher pre-release version than what the map specifies (e.g., map says v1alpha1 but v1beta1 is available), the version should be updated.

This catches both full graduation and version bumps within pre-release (alpha → beta).

So `defaultGroupVersionsByFeatureGate` would change from mapping feature gates to a list of GVs, to mapping feature gates to kinds + group. The GV can then be derived at runtime by looking up the highest available version for those kinds in the scheme. That way stale entries become impossible by construction — no test needed for that part, just the GA check (1).

Alternatively, if we keep the GV in the map and just add kinds, a test can verify both checks above — flagging entries where the kinds have graduated to v1 or exist in a higher pre-release version than what's specified.

everettraven: That sounds reasonable to me.

tchap:
/retitle OCPBUGS-85183: Add kind-aware staleness detection for runtime-config entries

pkg/operator/configobservation/apienablement/stale_gv_helpers_test.go
```diff
@@ -0,0 +1,393 @@
...
+								kind, gaGV, kubeVersion,
+							),
+						})
+						continue
+					}
+				}
+
+				// Check 2: highest pre-release version is listed. Collect all versions
+				// listed for this kind across resolved entries, then verify that the
+				// highest-priority pre-release version containing this kind is among them.
```
everettraven: I don't think we want "among them" - I think we want "is the only" for the given kind. For example, the existing `v1alpha1` entry is already stale and should be removed - or at least not apply for the current kube version (1.35).

Because the actual config isn't that granular we may still end up in a scenario where another pre-release kind is in the same group but an earlier version so that earlier version inherently gets served, but that should be OK if we are just checking for specific kinds in our test.

tchap: Aligned in 908a52ac12eace05bbfd20a453c246342691db14 .

I had to align the ranges in the mapping, because the test started to fail, which was expected, as far as I can tell.

tchap:
To squash before merging, I am going to

/hold

pkg/operator/configobservation/apienablement/observe_runtime_config.go
```diff
@@ -28,17 +28,30 @@ var defaultGroupVersionsByFeatureGate = map[configv1.FeatureGateName][]groupVers
 		// k8s.io/apiserver at 1.34 (via openshift/kubernetes-apiserver) and its MutatingAdmissionPolicy
 		// admission plugin uses v1beta1 informers. The openshift-apiserver can't be rebased until the
 		// o/k 1.36 rebase lands and a 1.36 branch is created on openshift/kubernetes-apiserver.
-		{KubeVersionRange: semver.MustParseRange(">=1.33.0 <1.37.0"), GroupVersion: schema.GroupVersion{Group: "admissionregistration.k8s.io", Version: "v1alpha1"}},
-		{KubeVersionRange: semver.MustParseRange(">=1.34.0 <1.37.0"), GroupVersion: schema.GroupVersion{Group: "admissionregistration.k8s.io", Version: "v1beta1"}},
+		{
+			KubeVersionRange: semver.MustParseRange(">=1.33.0 <1.34.0"),
+			GroupVersion:     schema.GroupVersion{Group: "admissionregistration.k8s.io", Version: "v1alpha1"},
+			Kinds:            []string{"MutatingAdmissionPolicy", "MutatingAdmissionPolicyBinding"},
+		},
```
everettraven: Probably safe for us to just remove now?

pkg/operator/configobservation/apienablement/stale_gv_helpers_test.go
```diff
@@ -0,0 +1,395 @@
...
+			}
+		}
+	}
+
+	return violations
+}
+
+func TestFindStaleGroupVersionEntries(t *testing.T) {
+	scheme := newTestScheme()
+	kubeVersion := semver.MustParse("1.35.0")
```
everettraven: More of a nit, but I'd prefer if both of these were part of the test case definitions so that we have fully declarative and self-contained test cases.

IMO it makes it significantly easier to follow the test cases if I can read all of the input at the test case level instead of having to jump between different areas of the code to correlate what the test defines as the expected state and what the global scheme configures.

everettraven (commented):
A couple of small comments. Overall, this looks really good.

pkg/operator/configobservation/apienablement/observe_runtime_config.go
tchap: Actually I am wondering whether we selected the right approach in the end, because the comment right above says

> Both v1alpha1 and v1beta1 versions must be served pre-GA because e2e tests exercise both APIs.

This is not really possible after the alignment in the last commit.

Let's see how the tests do. But overall I was thinking that we may need to have multiple non-GA versions enabled. I am not sure how our clients use this TBH, so I am not sure what the desired state is.

JoelSpeed:
This is quite relevant to some ongoing conversations we are having about enabling Kube features in tech preview prior to their GA upstream. Can we continue this work please?

pkg/operator/configobservation/apienablement/stale_gv_helpers_test.go
tchap: Amended.

pkg/operator/configobservation/apienablement/observe_runtime_config.go
tchap: Yes, this is actually removed now after I rebased this PR. I also made sure this works with the records that were there in 4ff133a2d2d9a231ffc2f983f9c3c9b5ee6302e2

JoelSpeed:
In terms of direction, I think this is a positive incremental improvement that will help us catch mistakes, IMO we should move this forward and continue the discussion about whether this is an actual need for an E2E or not

everettraven (approved):
/lgtm
/approve

tchap:
Commits squashed,

/unhold

tchap:
This will be

/verified by CI

PR #2193, opened 2026-06-11
WIP: kube-apiserver: set dnsPolicy to ClusterFirstWithHostNet

bertinatto:
According to the Kubernetes documentation [1], pods running with `hostNetwork: true` should set the DNS policy to `ClusterFirstWithHostNet`. Otherwise, pods will fall back to the `Default` DNS policy behavior, inheriting the host's name resolution configuration. This prevents containers in the pod from resolving cluster-internal addresses.

For example, when configuring KMS encryption with a Vault instance running inside the cluster, the `vaultAddress` field in the APIServer resource may reference an in-cluster Service endpoint:

```
  oc patch apiserver cluster --type=merge -p '
    {"spec":{"encryption":{"type":"KMS","kms":{"type":"Vault","vault":{
      "vaultAddress":"https://vault.vault-namespace.svc:8200",
      ...
    }}}}}'
```

Without `ClusterFirstWithHostNet`, the `kube-apiserver` pod cannot resolve `vault.vault-namespace.svc`, causing KMS plugin connectivity failures.

Co-locating the Vault KMS instance with the cluster it serves **is not recommended**, but it is common in CI environments and should not be restricted. The DNS policy should be correct regardless of where Vault is deployed.

[1] https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy

## Summary by CodeRabbit

* **Configuration Updates**
  * DNS policy for the API server pod changed to ClusterFirstWithHostNet when host networking is enabled, improving DNS resolution consistency and ensuring cluster DNS is used reliably for hostname and service lookups.

### DIFF
--- a/bindata/assets/kube-apiserver/pod.yaml
+++ b/bindata/assets/kube-apiserver/pod.yaml
@@ -290,6 +290,7 @@ spec:
       readOnlyRootFilesystem: true
   terminationGracePeriodSeconds: {{.GracefulTerminationDuration}}
   hostNetwork: true
+  dnsPolicy: ClusterFirstWithHostNet
   priority: 2000001000
   priorityClassName: system-node-critical
   tolerations:
--- a/vendor/github.com/openshift/library-go/test/library/encryption/kms/vault.go
+++ b/vendor/github.com/openshift/library-go/test/library/encryption/kms/vault.go
@@ -3,7 +3,6 @@ package kms
 import (
 	"context"
 	"fmt"
-	"net"
 	"os/exec"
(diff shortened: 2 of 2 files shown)


bertinatto:
/hold

This PR should be tested extensively to make sure it doesn't break any assumptions.

In addition, this PR contains a commit that patches the vendor directory to make sure KMS tests use the internal DNS name. As a result, `verify*` jobs are expected to fail. If we decide to merge this PR, this commit should be removed.

PR #2200, opened 2026-06-16
Enable watch termination grace period (redux).

benluddy:
This should mostly eliminate aligned informer re-lists during kube-apiserver rollout.

## Summary by CodeRabbit

* **Chores**
  * Added new default configuration setting for API server shutdown termination grace period.

### DIFF
--- a/bindata/assets/config/defaultconfig.yaml
+++ b/bindata/assets/config/defaultconfig.yaml
@@ -161,6 +161,8 @@ apiServerArguments:
     - 70s # give SDN some time to converge: 30s for iptable lock contention, 25s for the second try and some seconds for AWS to update ELBs
   shutdown-send-retry-after:
   - "true"
+  shutdown-watch-termination-grace-period:
+    - 5s # shorter than the pod graceful termination period on single-node
   storage-backend:
     - etcd3
   storage-media-type:


benluddy:
/payload-aggregate periodic-ci-openshift-release-master-ci-5.0-upgrade-from-stable-4.22-e2e-aws-ovn-upgrade 10

benluddy:
xref https://github.com/openshift/cluster-kube-apiserver-operator/pull/1862#issue-3179199655

benluddy:
From https://prow.ci.openshift.org/view/gs/test-platform-results/logs/openshift-cluster-kube-apiserver-operator-2200-ci-5.0-upgrade-from-stable-4.22-e2e-aws-ovn-upgrade/2066992765111111680, it appears the disruption samplers don't honor 429 responses with Retry-After the same way client-go's REST clients do. Since this change is allowing a few seconds to drain watches after the shutdown delay duration has elapsed, we're hitting the WithRetryAfter filter that is meant to shunt clients to other API servers.

bindata/assets/config/defaultconfig.yaml
```diff
@@ -161,6 +161,8 @@ apiServerArguments:
     - 70s # give SDN some time to converge: 30s for iptable lock contention, 25s for the second try and some seconds for AWS to update ELBs
   shutdown-send-retry-after:
   - "true"
+  shutdown-watch-termination-grace-period:
+    - 5s # shorter than the pod graceful termination period on single-node
```
sdodson, line 165: Being higher extends the overall termination period? I guess if we find that 5s is insufficient here we can  explore something like the controller that adjusts goaway-chance based on SNO vs HA.

PR #2219, opened 2026-07-03
OCPBUGS-38661: Add 10-minute degraded inertia for NodeControllerDegraded

tchap:
Prevent transient node not-ready events from immediately marking the operator as Degraded at the ClusterOperator level. Other degraded conditions keep the default 2-minute inertia.

The 10-minute inertia approach was discussed on the arch call and accepted as the way to go.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of degraded cluster conditions by applying a custom inertia window for specific controller degradation events.
  * Updated degraded-state timing behavior so the inertia duration matches expected values.
* **Tests**
  * Added unit test coverage to verify inertia durations for degraded condition types.
  * Removed an obsolete placeholder test that no longer applied.

### DIFF
--- a/pkg/operator/start_test.go
+++ b/pkg/operator/start_test.go
@@ -1,6 +1,0 @@
-package operator
-
-import "testing"
-
-func TestNothing(t *testing.T) {
-}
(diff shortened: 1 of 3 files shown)


tjungblu:
/lgtm
/approve

holding for feedback from kas folks

/hold

PR #2220, opened 2026-07-07
CNTRLPLANE-3861: Add KMS plugin sidecar revision readiness check

bertinatto:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved kube-apiserver static pod readiness checks by validating KMS plugin configuration using revision-specific configmap data.
  * Updated kube-apiserver KMS plugin integration for static pods to use the newer “ensure” behavior for consistent sidecar configuration.
  * Ensured readiness is reported only when required KMS configuration is present, valid, and matches the expected pod spec.
* **Chores**
  * Updated the Go module dependency via a `replace` directive to point `github.com/openshift/library-go` to a newer fork/version.

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/blang/semver/v4"
 
+	"github.com/google/go-cmp/cmp"
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
 	configv1client "github.com/openshift/client-go/config/clientset/versioned"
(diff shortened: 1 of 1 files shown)


pkg/operator/starter.go
```diff
@@ -626,6 +631,41 @@ func installerErrorInjector(operatorClient v1helpers.StaticPodOperatorClient) fu
...
+		podCopy := pod.DeepCopy()
+		encryptionConfigSecretName := fmt.Sprintf("encryption-config-%d", revision)
+
+		if err := kmspluginlifecycle.EnsureKMSPluginSidecarInStaticPodSpec(ctx, &podCopy.Spec, "kube-apiserver", operatorclient.TargetNamespace, encryptionConfigSecretName, "encryption-config", "cluster-kube-apiserver-operator", operatorImagePullSpec, kubeClient.CoreV1(), featureGateAccessor); err != nil {
+			return false, fmt.Errorf("failed to ensure KMS plugin sidecar in static pod spec: %w", err)
+		}
+
+		if !equality.Semantic.DeepEqual(pod, podCopy) {
+			klog.V(4).Infof("Pod diff for revision %d (-before +after):\n%s", revision, cmp.Diff(pod, podCopy))
+			return false, nil
```
ardaguclu, line 663: By returning false, what do we prevent to happen?

bertinatto: we prevent the revision to be marked as `ready` (ie, the revision ConfigMap doesn't get an annotation stating it's ready). As a result, the installer pod won't copy the revision files to the node and KAS will not use that revision. In other words, the revision will continue to exist, but it won't be used until this check passes

ardaguclu:
This PR aligns with what we designed and makes sense to me. However, I'm now well versed enough to approve it. I think, we need review from @p0lyn0mial

tjungblu:
looks good to me too! @p0lyn0mial care to take a look?

p0lyn0mial:
thanks

/lgtm
/approve

ardaguclu:
/lgtm cancel
there seems to be build failure

PR #2221, opened 2026-07-08
OCPBUGS-38661: Add 10-minute degraded inertia for StaticPodsDegraded

tchap:
Static pods may temporarily report errors during upgrades while pods restart. Use a longer inertia to avoid flapping the ClusterOperator Degraded condition.

This is an approach we agreed upon on the arch call and the affected controller was explicitly mentioned.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved degraded-status handling so an additional system condition is now tracked consistently for a 10-minute window.
  * This helps the operator more accurately recognize and retain certain transient degraded states instead of dropping them too quickly.


tchap:
/hold

will ask for additional review from @p0lyn0mial .

PR #2230, opened 2026-07-16
CNTRLPLANE-3237: intro kubeAPIServerEncryptionStatusProvider

p0lyn0mial:
## Summary by CodeRabbit

* **Improvements**
  * Enhanced KMS health monitoring by consolidating command startup and readiness handling.
  * Added encryption-status support to retrieve and update the kube-apiserver encryption status via the cluster operator API for more consistent reporting.
  * Updated KMS plugin sidecar injection parameter handling during kube-apiserver static pod spec updates.
* **Chores**
  * Updated the OpenShift library dependency to a newer version.


cmd/cluster-kube-apiserver-operator/main.go
```diff
@@ -67,8 +67,7 @@ func NewOperatorCommand(ctx context.Context) *cobra.Command {
 		}
 		return client.KubeAPIServers(), nil
 	}))
-	cmd.AddCommand(kmshealth.NewCommand(ctx, kmswriters.NewKubeAPIServerWriter))
-	cmd.AddCommand(kmspreflight.NewCommand(ctx))
```
ardaguclu, line 71: Why do you remove this?

pkg/operator/encryptionstatusprovider/provider.go
```diff
@@ -0,0 +1,47 @@
...
+	client *operatorclient.Clientset
+}
+
+func (p *kubeAPIServerEncryptionStatusProvider) GetKMSEncryptionStatus(ctx context.Context) (*operatorv1.KMSEncryptionStatus, error) {
+	obj, err := p.client.OperatorV1().KubeAPIServers().Get(ctx, "cluster", metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	s := obj.Status.EncryptionStatus
+	return &s, nil
```
ardaguclu: why not  just `return obj.Status.EncryptionStatus, nil`?

cmd/cluster-kube-apiserver-operator/main.go
```diff
@@ -67,8 +67,7 @@ func NewOperatorCommand(ctx context.Context) *cobra.Command {
 		}
 		return client.KubeAPIServers(), nil
 	}))
-	cmd.AddCommand(kmshealth.NewCommand(ctx, kmswriters.NewKubeAPIServerWriter))
```
tjungblu, line 70: what's with kmswriters.NewKubeAPIServerWriter? are we going to remove this in library-go?

ardaguclu: I think, it is already removed https://github.com/openshift/library-go/pull/2365

p0lyn0mial: yes, in general library-go is a dependency of the operators, not the other way around.

Having library-go define NewKubeAPIServerProvider, NewAuthenticationProvider means library-go needs to know about each operator's CR shape which seems as an inverted dependency.

Does it make sense ?

pkg/operator/targetconfigcontroller/targetconfigcontroller.go
```diff
@@ -330,7 +330,7 @@ func managePods(ctx context.Context, client coreclientv1.ConfigMapsGetter, secre
 		required.Spec.Containers[i].Env = append(container.Env, proxyEnvVars...)
 	}
 
-	if err := kmspluginlifecycle.AddKMSPluginSidecarToStaticPodSpec(ctx, &required.Spec, "kube-apiserver", operatorclient.TargetNamespace, "encryption-config", "cluster-kube-apiserver-operator", operatorImagePullSpec, secretClient, featureGateAccessor); err != nil {
+	if err := kmspluginlifecycle.AddKMSPluginSidecarToStaticPodSpec(ctx, &required.Spec, "kube-apiserver", operatorclient.TargetNamespace, "encryption-config", "", "cluster-kube-apiserver-operator", operatorImagePullSpec, secretClient, featureGateAccessor); err != nil {
```
p0lyn0mial, line 333: @bertinatto what should we pass for `diskSecretName` from the target controller?

bertinatto: please leave the string empty or use `encryption-config`. This new field (`diskSecretName`) is to decouple the secret name (currently `encryption-config`) from the file path where it will be available on disk as a static resource. This is necessary because the revision check uses the revisioned encryption config (e.g., `encryption-config-rev`).

p0lyn0mial: ok, thx, i will leave it empty then.

ardaguclu:
This is safe change and library-go can not be bumped without this patch
/label acknowledge-critical-fixes-only

pkg/operator/encryptionstatusprovider/provider.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+type kubeAPIServerEncryptionStatusProvider struct {
+	client *operatorclient.Clientset
+}
+
+func (p *kubeAPIServerEncryptionStatusProvider) GetKMSEncryptionStatus(ctx context.Context) (*operatorv1.KMSEncryptionStatus, error) {
+	obj, err := p.client.OperatorV1().KubeAPIServers().Get(ctx, "cluster", metav1.GetOptions{})
+	if err != nil {
+		return nil, err
```
bertinatto, line 34: it might be good for debugging to wrap this error (and the one at line 40 as well) with a descriptive msg

p0lyn0mial:
failing `ci/prow/e2e-gcp-operator-serial` will be fixed by https://github.com/openshift/cluster-kube-apiserver-operator/pull/2232/changes

PR #2236, opened 2026-07-22
OCPBUGS-86789: Enable send-retry-after-while-not-ready-once on all topologies

mkowalski:
## What

Enables the `send-retry-after-while-not-ready-once` kube-apiserver protection on all control plane topologies instead of SNO-only. The `shutdown-send-retry-after` flag is already enabled globally via `bindata/assets/config/defaultconfig.yaml`, so this PR completes the pair by making the startup-side protection topology-independent.

## Why

Investigation of OCPBUGS-86789 showed that load balancers can route requests to a not-yet-ready kube-apiserver on multi-node clusters too, even while healthy replicas are available:

- On AWS, the NLB was observed routing to a target that had been unhealthy for 50s+ with two healthy masters available (see bug comments).
- On bare metal / in-cluster traffic, the `kubernetes.default` service path can go stale: the OVN load balancer backends are only as fresh as ovn-kubernetes' EndpointSlice informers, and those informers were observed losing their API watches (via the api-int VIP) exactly during control plane rollouts — e.g. Prow run `periodic-ci-openshift-release-main-nightly-5.0-e2e-metal-ipi-ovn-dualstack/2079715174809341952`, where ovnkube on master-0 logged ~50 watch failures in the minute preceding a `LateConnections`/early-request event.

Requests accepted before readiness may be processed before RBAC and admission post-start hooks have completed, which is the security concern flagged in the bug. Since no LB implementation (cloud LB, on-prem haproxy/keepalived, or the in-cluster service path) can be perfectly synchronized with `/readyz`, the only robust protection is server-side: reject early requests with 503 + `Retry-After` until the server has been ready once. Well-behaved clients retry transparently against a healthy replica; SNO behavior is unchanged.

This follows the suggestion in https://issues.redhat.com/browse/OCPBUGS-86789 that HA clusters should also have this protection enabled (audit logs on SNO confirm premature requests are cleanly rejected with 503).

## Impact on Component Readiness

Sippy regressions 44896 and 44949 (`[sig-api-machinery][Feature:APIServer][Late] API LBs follow /readyz of kube-apiserver and stop sending requests`) are fallout of the same root cause becoming visible after the Kube 1.36 rebase started emitting the previously-silent `LateConnections`/`NonReadyRequests` events.

## Notes

- The observer no longer needs the Infrastructure lister; the value is constant.
- Unit tests updated accordingly (all topologies expect `true`).

---
🤖 *This PR was generated by AI on behalf of @mkowalski, who has reviewed it.*

## Summary by CodeRabbit

* **Configuration Updates**
  * The API server now enables `send-retry-after-while-not-ready-once` by default across all cluster topology modes.
  * Configuration updates and defaults have been aligned to consistently enable this protection.

### DIFF
--- a/pkg/operator/configobservation/apiserver/observe_send_retry_after_while_not_ready_once.go
+++ b/pkg/operator/configobservation/apiserver/observe_send_retry_after_while_not_ready_once.go
@@ -3,35 +3,34 @@ package apiserver
 import (
 	"fmt"
 	"reflect"
-	"strconv"
 
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
-	configv1 "github.com/openshift/api/config/v1"
-	"github.com/openshift/cluster-kube-apiserver-operator/pkg/operator/configobservation"
 	"github.com/openshift/library-go/pkg/operator/configobserver"
 	"github.com/openshift/library-go/pkg/operator/events"
 )
 
 var sendRetryAfterWhileNotReadyOncePath = []string{"apiServerArguments", "send-retry-after-while-not-ready-once"}
 
-// ObserveSendRetryAfterWhileNotReadyOnce ensures that send-retry-after-while-not-ready-once is set for SNO clusters.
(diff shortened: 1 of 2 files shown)


mkowalski:
/close

`send-retry-after-while-not-ready-once` is guarding the startup window and not shutdown. As my area of interest is mainly shutdown (and startup if and only if the root cause is the same), no point in pursuing this further. Unless, what's unlikely, situation changes but then we can always get back to it.

PR #2248, opened 2026-07-27
CNTRLPLANE-3237: creates the kms status provider from the operator client

p0lyn0mial:
xref: https://github.com/openshift/library-go/pull/2382

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the KMS health command’s kube-apiserver encryption status reporting for more reliable results.
* **Refactor**
  * Reworked encryption status provider initialization to use the active operator client rather than constructing from configuration, keeping the same command behavior.
* **Maintenance**
  * Updated a Go library dependency to a newer revision.

### DIFF
--- a/pkg/operator/encryptionstatusprovider/provider.go
+++ b/pkg/operator/encryptionstatusprovider/provider.go
@@ -2,10 +2,8 @@ package encryptionstatusprovider
 
 import (
 	"context"
-	"fmt"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/rest"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
 	applyoperatorv1 "github.com/openshift/client-go/operator/applyconfigurations/operator/v1"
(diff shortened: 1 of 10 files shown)


pkg/operator/encryptionstatusprovider/provider.go
```diff
@@ -15,14 +13,10 @@ import (
...
-// KubeAPIServer/cluster from a rest.Config.
-func NewKubeAPIServerEncryptionStatusProviderFromConfig(restConfig *rest.Config) (kms.EncryptionStatusProvider, error) {
-	opClient, err := operatorclient.NewForConfig(restConfig)
-	if err != nil {
-		return nil, fmt.Errorf("build operator client: %w", err)
-	}
-	return &kubeAPIServerEncryptionStatusProvider{client: opClient.OperatorV1().KubeAPIServers()}, nil
+// NewKubeAPIServerEncryptionStatusProvider builds a kms.EncryptionStatusProvider for
+// KubeAPIServer/cluster from an operator client.
+func NewKubeAPIServerEncryptionStatusProvider(client operatorclient.Interface) (kms.EncryptionStatusProvider, error) {
```
p0lyn0mial, line 18: once we wire the preflight controller we could use https://github.com/openshift/cluster-kube-apiserver-operator/blob/8dfb4f88f0bc43c4f8646bb132ae675b74e9bbe7/pkg/operator/starter.go#L187

p0lyn0mial: so there would only be one function.

PR #2263, opened 2026-08-10
vendor: update library-go to encryption-refactor-fn-injection branch

p0lyn0mial:
Pick up the encryption refactoring from library-go:
- Function injection into checkAndCreateKeys/generateKeySecret
- Promotion of checkAndCreateKeys and helpers to package-level functions
- KeyComputer and EncryptionComputer for read-only key/config computation
- KMS preflight controller wired with EncryptionConfigurationComputer
- TLS group/curve mapping support (crypto package)

Uses a local replace directive pointing to the development branch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved encryption setup with a Kubernetes KMS preflight validation step.
  * Preflight checks now use the configured deployment settings and run at regular intervals.
  * Replaced the previous validation behavior that always reported success with meaningful checks and results.

### DIFF
--- a/vendor/github.com/openshift/library-go/pkg/operator/encryption/kms/pluginlifecycle/builder.go
+++ b/vendor/github.com/openshift/library-go/pkg/operator/encryption/kms/pluginlifecycle/builder.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"path/filepath"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -25,7 +26,8 @@ type KMSPluginBuilder struct {
 	staticPod                  bool
 	errorIfNotFound            bool
 
-	healthReporter *healthReporter
+	healthReporter    *healthReporter
+	injectSocketFlags bool
 }
 
 // NewKMSPluginBuilder creates a builder that defaults to deployment mode.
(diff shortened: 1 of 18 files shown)


vendor/github.com/openshift/library-go/pkg/operator/encryption/kms/preflight/cmd.go
```diff
@@ -45,6 +45,8 @@ func NewCommand(ctx context.Context) *cobra.Command {
 }
 
 func (o *options) addFlags(fs *pflag.FlagSet) {
+	fs.StringSliceVar(&o.kmsSockets, "kms-sockets", nil, "KMS plugin endpoints in unix:// URI format (e.g. unix:///var/run/kmsplugin/kms-1.sock); each is checked for reachability (Status)")
```
p0lyn0mial, line 48: @ardaguclu here are the new two flags we will need (feel free to change the names)

vendor/github.com/openshift/library-go/pkg/operator/encryption/kms/pluginlifecycle/builder.go
```diff
@@ -195,9 +206,32 @@ func (b *KMSPluginBuilder) Apply(ctx context.Context, podSpec *corev1.PodSpec, c
...
+		}
+	}
+
 	return nil
 }
 
+// injectPreflightSocketFlags appends --kms-sockets and --kms-socket-to-verify
+// to the named container's args. sockets[0] is the highest-keyID provider
+// (the one being verified).
+func injectPreflightSocketFlags(containers []corev1.Container, containerName string, sockets []string) error {
```
p0lyn0mial, line 221: @ardaguclu this is the code in the lifecycle that inject the new flags - make sure it make sense :)

vendor/github.com/openshift/library-go/pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -89,6 +89,7 @@ func (d *PodPreflightDeployer) Deploy(ctx context.Context, configHash string, en
 
 	err = pluginlifecycle.NewKMSPluginBuilder().
 		WithSecretRequired().
+		WithPreflightChecker().
```
p0lyn0mial, line 92: @ardaguclu here is how we could let the lifecycle code we are in the "preflight" mode.

PR #2275, opened 2026-08-21
OCPBUGS-112638: apienablement: enable resource.k8s.io/v1beta2 when DeviceTaintRule is enabled

haircommander:
## Summary by CodeRabbit

* **New Features**
  * Added runtime configuration support for the `DeviceTaintRule` API through the `DRADeviceTaintRules` feature gate on Kubernetes 1.36 and compatible patch releases.
  * Enabled deployments can configure device taint rules using the `resource.k8s.io/v1beta2` API.
  * Runtime configuration is generated only when the feature gate is recognized, enabled, and applicable to the Kubernetes version.


pkg/operator/configobservation/apienablement/observe_runtime_config.go
```diff
@@ -15,7 +15,11 @@ import (
 	"github.com/openshift/library-go/pkg/operator/events"
 )
 
-var defaultGroupVersionsByFeatureGate = map[configv1.FeatureGateName][]groupVersionKindsByOpenshiftVersion{}
+var defaultGroupVersionsByFeatureGate = map[configv1.FeatureGateName][]groupVersionKindsByOpenshiftVersion{
+	"DRADeviceTaintRules": {
+		{GroupVersion: schema.GroupVersion{Group: "resource.k8s.io", Version: "v1beta2"}, Kinds: []string{"DeviceTaintRule"}},
+	},
+}
```
harche, line 22: Since we already know upstream graduates DeviceTaintRule into `resource.k8s.io/v1` in 1.37 where it is served by default, should we bound this entry the way the DynamicResourceAllocation entry did back in 4.18? That way the mapping deactivates itself on the 1.37 rebase instead of pointing at an API version we no longer need. Something like:

```suggestion
var defaultGroupVersionsByFeatureGate = map[configv1.FeatureGateName][]groupVersionKindsByOpenshiftVersion{
	// DeviceTaintRule is served from resource.k8s.io/v1beta2 on kube 1.36 and
	// graduates to resource.k8s.io/v1 in 1.37, where it is served by default,
	// so this mapping is only needed on 1.36. Note that the staleness test
	// skips range excluded entries, so nothing will flag this entry for
	// removal after the 1.37 rebase. Whoever picks it up then should delete
	// this entry and also revisit the DRADeviceTaintRules registration in
	// openshift/api, since the gate goes GA and on by default upstream in 1.37.
	"DRADeviceTaintRules": {
		{KubeVersionRange: semver.MustParseRange(">=1.36.0 <1.37.0"), GroupVersion: schema.GroupVersion{Group: "resource.k8s.io", Version: "v1beta2"}, Kinds: []string{"DeviceTaintRule"}},
	},
}
```

What do you think?

haircommander: updated!

haircommander: we already will have a blocker bug for 5.1 to remove so I don't thinki the comment is necessary

harche: Thanks,

/lgtm

haircommander:
/payload-with-prs  5.0 nightly informing openshift/cluster-kube-apiserver-operator#2275

harche:
Following the investigation in [openshift/api#3004](https://github.com/openshift/api/pull/3004#issuecomment-5539818984), this runtime-config mapping is required, but the current PR cannot safely land ahead of the API registration.

In the standalone CKASO presubmit, bootkube panics with:

```text
feature "DRADeviceTaintRules" is not registered
```

[Artifact](https://storage.googleapis.com/test-platform-results/pr-logs/pull/openshift_cluster-kube-apiserver-operator/2275/pull-ci-openshift-cluster-kube-apiserver-operator-main-e2e-aws-ovn/2090862182857183232/artifacts/e2e-aws-ovn/ipi-install-install/artifacts/expanded-log-bundle/log-bundle-20260821184545/bootstrap/journals/bootkube.log)

Conversely, landing `openshift/api#3004` first enables the gate without `resource.k8s.io/v1beta2`, which prevents the kube-scheduler `DeviceTaintRule` informer from syncing and breaks TPNU cluster initialization.

Could this PR be updated so mappings for absent/unregistered feature gates are ignored rather than calling `Enabled()` and panicking? That would allow the mapping to land dormant before the API PR activates the gate.

Suggested coverage:

- unregistered `DRADeviceTaintRules`: no panic and no runtime-config entry;
- registered but disabled: no runtime-config entry;
- enabled on Kubernetes 1.36: `resource.k8s.io/v1beta2=true`;
- Kubernetes 1.35 and 1.37: no v1beta2 mapping.

Since the mapping uses a string feature-gate name, could the temporary `github.com/openshift/api` personal-fork replacement and unrelated vendor churn also be removed? That should make it possible to land this independently of `api#3004`.

The existing [API + CKASO payload run](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/openshift-cluster-kube-apiserver-operator-2275-openshift-api-3004-nightly-5.0-e2e-aws-ovn-single-node-techpreview/2092343301330964480) confirms the mapping itself works when both changes are present: the v1beta2 API is served, the scheduler informer syncs, and installation succeeds.

pkg/operator/configobservation/apienablement/observe_runtime_config.go
```diff
@@ -15,7 +15,11 @@ import (
 	"github.com/openshift/library-go/pkg/operator/events"
 )
 
-var defaultGroupVersionsByFeatureGate = map[configv1.FeatureGateName][]groupVersionKindsByOpenshiftVersion{}
+var defaultGroupVersionsByFeatureGate = map[configv1.FeatureGateName][]groupVersionKindsByOpenshiftVersion{
+	"DRADeviceTaintRules": {
+		{KubeVersionRange: semver.MustParseRange(">=1.36.0 <1.37.0"), GroupVersion: schema.GroupVersion{Group: "resource.k8s.io", Version: "v1beta2"}, Kinds: []string{"DeviceTaintRule"}},
```
harche, line 22: The mapping itself is right for 1.36: `resource.k8s.io/v1beta2` is off by default and `DeviceTaintRule` is only served from `v1beta2` there.

https://github.com/kubernetes/kubernetes/blob/4c1f91925b8dc669539edeb56bda8b7056e5bbc8/pkg/controlplane/instance.go#L488-L492

https://github.com/kubernetes/kubernetes/blob/4c1f91925b8dc669539edeb56bda8b7056e5bbc8/pkg/registry/resource/rest/storage_resource.go#L194-L201

One thing to decide deliberately: `resource.k8s.io/v1beta2=true` enables the whole version, so `deviceclasses`, `resourceclaims`, `resourceclaimtemplates` and `resourceslices` come back at v1beta2 too, even though they are GA in v1. kube-apiserver accepts a resource-level entry, and an explicit resource takes priority over the version:

https://github.com/kubernetes/kubernetes/blob/4c1f91925b8dc669539edeb56bda8b7056e5bbc8/staging/src/k8s.io/apiserver/pkg/server/storage/resource_config.go#L161-L166

`resource.k8s.io/v1beta2/devicetaintrules=true` would serve only what the scheduler needs. That needs `RuntimeConfigFromFeatureGates` to emit resource-level entries when `Kinds` is set, which is a small change. If version-wide is preferred for consistency with the older DRA entries, that's fine too, but it'd be good to record the choice here since the origin expectation isn't settled yet.

pkg/operator/configobservation/apienablement/observe_runtime_config.go
```diff
@@ -99,8 +103,9 @@ func newFeatureGateObserverWithRuntimeConfig(featureGateObserver configobserver.
 
 func RuntimeConfigFromFeatureGates(featureGates featuregates.FeatureGate, groupVersionsByFeatureGate map[configv1.FeatureGateName][]schema.GroupVersion) []string {
 	var entries []string
+	knownFeatures := sets.New(featureGates.KnownFeatures()...)
 	for name, gvs := range groupVersionsByFeatureGate {
-		if !featureGates.Enabled(name) {
+		if !knownFeatures.Has(name) || !featureGates.Enabled(name) {
```
harche, line 110: Now that the gate is registered as disabled in every feature set, this guard is no longer needed for the merge ordering we agreed on:

https://github.com/openshift/api/blob/5b6adcbced27/features/features.go#L1071-L1076

https://github.com/openshift/api/pull/3004#issuecomment-5541265332

It also changes the failure mode. Today a mistyped or later-renamed gate name in this map panics loudly:

https://github.com/openshift/library-go/blob/d17c137b84b0/pkg/operator/configobserver/featuregates/featuregate.go#L31-L40

With this change it becomes a silent no-op and the API just never gets enabled. Two options, either is fine by me:
- drop this commit, or
- keep it and add a unit test that every key of `defaultGroupVersionsByFeatureGate` is a registered gate (iterate `features.AllFeatureSets()` from the vendored `openshift/api`), so we keep the safety net at test time.

pkg/operator/configobservation/apienablement/observe_runtime_config_test.go
```diff
@@ -175,6 +175,52 @@ func TestFeatureGateObserverWithRuntimeConfig(t *testing.T) {
...
+			mappings:     mappings,
+		},
+		{
+			name:           "enabled gate",
+			featureGates:   featuregates.NewFeatureGate([]configv1.FeatureGateName{"DRADeviceTaintRules"}, nil),
+			mappings:       mappings,
+			expectedConfig: []string{"resource.k8s.io/v1beta2=true"},
+		},
+		{
+			name:         "version-filtered gate",
```
harche: nit: could you add a `1.37.0` case alongside `1.35.0`? `DeviceTaintRule` is already in `resource/v1` on master, so the entry is meant to self-deactivate on the 1.37 rebase, and this would document that:

https://github.com/kubernetes/kubernetes/blob/d0c67b05ca01f8e298b7a74ba02cfe88bbaaaa8e/staging/src/k8s.io/api/resource/v1/types.go#L1030

Related: the staleness test filters out range-excluded entries before checking, so nothing will flag this entry for removal after the rebase:

https://github.com/openshift/cluster-kube-apiserver-operator/blob/b0783a2f85920adde7fad1915154f28dd6fe368b/pkg/operator/configobservation/apienablement/stale_gv_helpers_test.go#L38

A one-line comment on the map entry pointing at the 5.1 removal bug would help whoever picks that up.

go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	github.com/onsi/ginkgo/v2 v2.27.2
 	github.com/onsi/gomega v1.39.0
 	github.com/openshift-eng/openshift-tests-extension v0.0.0-20250804142706-7b3ab438a292
-	github.com/openshift/api v0.0.0-20260805215214-cfb63858e9d7
+	github.com/openshift/api v0.0.0-20260915185715-5b6adcbced27
```
harche: nit: the PR description still says it carries openshift/api#3004, which has merged. Worth a refresh to say this is the dormant mapping and activation follows in a later api PR.

harche (commented):
Thanks

pkg/operator/configobservation/apienablement/observe_runtime_config.go
```diff
@@ -15,7 +15,13 @@ import (
 	"github.com/openshift/library-go/pkg/operator/events"
 )
 
-var defaultGroupVersionsByFeatureGate = map[configv1.FeatureGateName][]groupVersionKindsByOpenshiftVersion{}
+var defaultGroupVersionsByFeatureGate = map[configv1.FeatureGateName][]groupVersionKindsByOpenshiftVersion{
+	// Remove this mapping on the Kubernetes 1.37 rebase, when DeviceTaintRule is served from v1 by default.
+	// RuntimeConfigFromFeatureGates maps group versions, so v1beta2 is enabled version-wide.
+	"DRADeviceTaintRules": {
+		{KubeVersionRange: semver.MustParseRange(">=1.36.0 <1.37.0"), GroupVersion: schema.GroupVersion{Group: "resource.k8s.io", Version: "v1beta2"}, Kinds: []string{"DeviceTaintRule"}},
+	},
```
bertinatto, line 23: is this intended to be backported to 5.0? If not, this `<1.37.0` is going to break the 1.37 rebase validation (CC @jubittajohn @jacobsee). Will we need to:

1. merge this PR
2. backport to 5.0
3. revert this PR (for the rebase)

?

pkg/operator/configobservation/apienablement/observe_runtime_config.go
```diff
@@ -99,8 +105,9 @@ func newFeatureGateObserverWithRuntimeConfig(featureGateObserver configobserver.
 
 func RuntimeConfigFromFeatureGates(featureGates featuregates.FeatureGate, groupVersionsByFeatureGate map[configv1.FeatureGateName][]schema.GroupVersion) []string {
 	var entries []string
+	knownFeatures := sets.New(featureGates.KnownFeatures()...)
 	for name, gvs := range groupVersionsByFeatureGate {
-		if !featureGates.Enabled(name) {
```
bertinatto, line 103: do we really need `.Has()` here? I assume it's here because `Enabled()` panics if the gate doesn't exist, right?

pkg/operator/configobservation/apienablement/observe_runtime_config.go
haircommander: yup that's the plan! I'm happy to take another approach if you'd prefer

pkg/operator/configobservation/apienablement/observe_runtime_config.go
haircommander: yup indeed!

bertinatto: do we really need it for this feature gate? Would it be OK if we keep this our this PR?

I'm asking because the panic is an useful signal when we are bumping to the next version of kube, and the feature gate that's listed here is removed from the next kube version. Right now we get a panic during the validation phase, so the person doing the rebase has an opportunity to come here and clean up the gate from `defaultGroupVersionsByFeatureGate`. If we go with this change, it'll be silently ignored.

haircommander: okay updated

bertinatto (commented):
/lgtm
/approve

harche:
Thanks for the updates.

Two small things:
- squash "tolerate absent feature gates" into "validate runtime config gates", since the guard is added and then removed within the same PR
- a line in the description noting this mapping is dormant until a follow-up `openshift/api` PR enables `DRADeviceTaintRules`

haircommander:
/verified later @haircommander
/pipieline required

done @harche

bertinatto (commented):
/lgtm

harche:
> /verified later @haircommander /pipieline required
> 
> done @harche

Thanks.

/pipeline required
/lgtm

PR #2281, opened 2026-08-25
OCPBUGS-108524: Fix infinite static pod rollouts from SA token rotation

ShazaAldawamneh:
Move localhost-recovery-client-token from RevisionSecrets to CertSecrets so that SA token
data changes no longer trigger new static pod revisions. On clusters where external tools
(ArgoCD, Kyverno, compliance scanners) interact with secrets in openshift-kube-apiserver,
the token gets regenerated every ~60 minutes, and because it was revisioned, each
regeneration triggered a full static pod rollout across all control plane nodes — inflating
revisions to 2400+, bloating etcd, and causing transient API disruptions. The token is
consumed via tokenFile in the cert-syncer kubeconfig, which client-go re-reads from disk
every ~1 minute, so no pod restart is needed; the cert-syncer sidecar already syncs
CertSecrets to disk on change via its informer watch.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved recovery token handling so externally regenerated tokens can be picked up automatically without triggering unnecessary application rollouts.
  - Recovery token updates are now applied dynamically, reducing service disruption during token rotation.

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -728,7 +728,6 @@ var RevisionSecrets = []revision.RevisionResource{
 
 	// this needs to be revisioned as certsyncer's kubeconfig isn't wired to be live reloaded, nor will be autorecovery
 	{Name: "localhost-recovery-serving-certkey"},
-	{Name: "localhost-recovery-client-token"},
 
 	{Name: "webhook-authenticator", Optional: true},
 }
@@ -750,6 +749,10 @@ var CertConfigMaps = []installer.UnrevisionedResource{
 var CertSecrets = []installer.UnrevisionedResource{
 	{Name: "aggregator-client"},
 	{Name: "localhost-serving-cert-certkey"},
+	// localhost-recovery-client-token is consumed via tokenFile in the cert-syncer kubeconfig,
+	// which client-go re-reads from disk (no restart needed). Revisioning it caused spurious
+	// rollouts whenever external tools (ArgoCD, Kyverno) triggered SA token regeneration.
+	{Name: "localhost-recovery-client-token"},
 	{Name: "service-network-serving-certkey"},
 	{Name: "external-loadbalancer-serving-certkey"},
 	{Name: "internal-loadbalancer-serving-certkey"},


pkg/operator/starter.go
```diff
@@ -750,6 +749,10 @@ var CertConfigMaps = []installer.UnrevisionedResource{
 var CertSecrets = []installer.UnrevisionedResource{
 	{Name: "aggregator-client"},
 	{Name: "localhost-serving-cert-certkey"},
+	// localhost-recovery-client-token is consumed via tokenFile in the cert-syncer kubeconfig,
+	// which client-go re-reads from disk (no restart needed). Revisioning it caused spurious
+	// rollouts whenever external tools (ArgoCD, Kyverno) triggered SA token regeneration.
+	{Name: "localhost-recovery-client-token"},
```
everettraven, line 755: I could be wrong, but I'm not sure this change is as straightforward as just moving the files here.

From what I can tell, it looks like the location in which the file would be synced to the node is different based on whether it is a revisioned resource or not.

If it is a revisioned resource, it will be synced to the directory `/etc/kubernetes/static-pod-resources/<resource>/<name>`.

If it is an unrevisioned resource, it will be synced to the directory `/etc/kubernetes/static-pod-resources/kube-apiserver-certs/<resource>/<name>`.

everettraven (commented):
Having spent some time digging into the changes here, I'm not sure this is what we actually want to do.

In the case we feel this is a change we want to make, I've provided at least some context that making this change isn't as straightforward as moving the entry from one list to another.

Now onto why I think we might not want to make this change.

The localhost-recovery-client-token secret is defined as: https://github.com/openshift/cluster-kube-apiserver-operator/blob/89ac4a58e4bc7e2242b78be85361d7562a0e6d61/bindata/assets/kube-apiserver/localhost-recovery-token.yaml#L1-L8

This leverages the "legacy" service account token behavior to create a long-lived service account token that provides the cert-syncer a token that will _always_ be valid and never expire.

Having looked further into the linked bug, it looks like we are seeing rotation issues because someone has configured some kind of external tooling to forcefully rotate the secret ~ every 60 minutes, resulting in this churn. Under normal circumstances, we would likely not see this secret ever be rotated. IMO this is a case of a customer having configured something that is interacting negatively with our system components and the customer actually needs to go back and update whatever $thing that is causing this churn to stop doing so.

I think it is a fair question to ask _why_ we are still using a legacy service account tokens if there are standards/compliance reasons as to why they have something that is causing this churn. The scope of work to solve that is significantly more intrusive to the platform from what I can tell though.

pkg/operator/starter.go
ShazaAldawamneh: Thanks for the thorough review. You're right on both points:

Moving the secret to CertSecrets would change the sync path from .../secrets/localhost-recovery-client-token/ to .../kube-apiserver-certs/secrets/localhost-recovery-client-token/, which would break the cert-syncer kubeconfig that hardcodes the original path. This isn't a simple list move.

More importantly, after digging deeper into the bug, this secret is a long-lived token that should never rotate under normal operation. The churn is caused by the customer's Kyverno policy (or similar external tooling) deleting/recreating the secret in a platform-owned namespace. The operator is behaving correctly, if tracked secret data changes, it should roll out.

The right resolution is to have the customer exclude this secret (or the openshift-kube-apiserver namespace) from their Kyverno policy. I'll update the bug accordingly and close this PR.

